PR #1, opened 2017-10-02
Add submodules

gashcrumb:
@rhuss @chirino @zregvart FYI...

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,24 @@
+[submodule "rest"]
+	path = rest
+	url = https://github.com/syndesisio/syndesis-rest.git
+[submodule "ui"]
+	path = ui
+	url = https://github.com/syndesisio/syndesis-ui.git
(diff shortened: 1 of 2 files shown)


rhuss:
@gashcrumb awesome, thanks ! Let's wait a bit with the PR, as I think we are already trying to go via direct copy. I'm going to open an issue to discuss this (containing the results from our discussion yesterday).

gashcrumb:
@rhuss This seemed like a pretty easy way to bootstrap this repo though, as it's trivial to switch the submodules to just regular directories containing files:

https://stackoverflow.com/questions/1759587/un-submodule-a-git-submodule

zregvart:
Yeah I think we should approve this in order to work the CI tasks, or at least to try to create a Jenkins or Circle CI setup. Would be one more data point for our CI discussion.

rhuss:
Let's merge ...

PR #4, opened 2017-10-12
feat: add a first draft a script that builds everything.

iocanel:


build.sh
```diff
@@ -0,0 +1,130 @@
...
+if [ -n "$SKIP_IMAGE_BUILDS" ]; then
+    echo "Skipping image builds ..."
+    MAVEN_IMAGE_BUILD_GOAL=""
+fi
+
+if [ -n "$NAMESPACE" ]; then
+    echo "Namespace: $NAMESPACE"
+    MAVEN_OPTS="$MAVEN_OPTS -Dfabric8.namespace=$NAMESPACE"
+    OC_OPTS=" -n $NAMESPACE"
+fi
```
zregvart: Does it make sense to default `$NAMESPACE` to `oc project -q`?

iocanel: Good idea!

iocanel:
@zregvart: Maybe you are right! And I think that you highlight a wider issue: 

`The scripts assume non-dev flavor of templates`.

So, they don't talk to docker at all and do everything in a more openshift purist fashion. The good thing about it, is that will work directly with our rh-idev cluster. 

Still, we need to support both styles. I'll work on that.

PR #11, opened 2017-10-17
feat: add ci modules. Pipeline now uses pipeline library from the wor…

iocanel:
…kspace.


rhuss:
merge conflicts

iocanel:
I think that for now we should better create a ci monorepo and once its in a working state we should merge. So I'm closing for now.

PR #13, opened 2017-10-24
Adding a circleci config file..

chirino:


.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
...
+      - setup_remote_docker
+
+      - checkout
+
+      - run:
+          name: Collect pom.xml
+          command: |
+            git submodule init
+            git submodule update
+            find . | grep pom.xml | grep -v src/test/resources | grep -v src/main/resources | xargs cat > all-poms
```
rhuss: `find . -name pom.xml -maxdepth 2 > all-poms`

.circleci/config.yml
```diff
@@ -0,0 +1,71 @@
...
+      - save_cache:
+          key: syndesis-all-poms-{{ checksum "all-poms" }}
+          paths:
+          - ~/.m2
+
+      - run:
+          name: Collect junit reports
+          command: |
+            mkdir ./junit/
+            find . -type f -regex ".*/target/.*-reports/.*xml" -exec cp {} ./junit/ \;
```
rhuss: One xargs is much faster than multiples cp execs:

`find . -type f -regex ".*/target/.*-reports/.*xml" -print0 | xargs -0 -J % cp % ./junit/`

rhuss:
btw, I wouldn't mind to do the reviews before merging.

PR #15, opened 2017-10-24
Adding wrapper jar file.

chirino:


rhuss:
@chirino Why do we need a top level mvnw wrapper ?

PR #16, opened 2017-10-24
Trying to use an alias to make mvn use mvnw

chirino:


.circleci/config.yml
```diff
@@ -40,7 +40,9 @@ jobs:
 
       - run:
           name: Build
-          command: ./build.sh
+          command: |
+            alias mvn=`pwd`/mvnw 
+            ./build.sh
```
rhuss: @chirino don't this works as aliases are not exported into sub processes ....

PR #17, opened 2017-10-24
Update the build.sh to use the mvnw command instead of mvn.

chirino:


chirino:
retest this please

build.sh
```diff
@@ -107,25 +109,25 @@ function init_options() {
 function connectors() {
   pushd connectors
   # We are ALWAYS clean this project, because build fails otherwise: https://github.com/syndesisio/connectors/issues/93
-  mvn clean install $MAVEN_OPTS
+  "${BASEDIR}/mvnw" clean install $MAVEN_OPTS
```
rhuss: you could just use a `${MAVEN_CMD}` here which people might even be able to override if they want a different maven. 

but ok like this, too.

PR #18, opened 2017-10-25
fix: build.sh now appends MAVEN_OPTS instead of replacing.

iocanel:


rhuss:
@zregvart I think this PR is for aligning the behaviour to the behaviour of the other script options. Since `-D` is used both by java and mvn for setting system properties, I don't see much harm to set it like this when running the command.

Using a (temporary) file to transport options given by a user feels weird.

But it also could be that I didn't really understood your comment ;-)

zregvart:
@rhuss yeah I wasn't too clear in my comment also. It seems to me that we're adding to `MAVEN_OPTS` environment variables Maven command line switches (like `--batch-mode`), these can't be a part of the `MAVEN_OPTS` as `MAVEN_OPTS` is provided to `java` command, and it will fail if it encounters unknown parameter.

We should use `MAVEN_OPTS` for JVM parameters, and use either another environment variable -- I think we could use `MAVEN_CONFIG`, I see it being used for this purpose in the `mvnw`.

As a side note, if we're setting here properties that are to be used in all use cases (JVM memory settings for instance), we should use `.mvn/jvm.config` file for JVM parameters, and `.mvn/maven.config` for Maven parameters.

Perhaps for another PR then?

PR #20, opened 2017-10-25
Support a MAVEN_CMD env to configure the maven binary used to build the project.

chirino:


rhuss:
retest this please

iocanel:
@rhuss: does `retest this please` also work for circleci?

rhuss:
@iocanel :  oops, of course not. Let me trigger it manually ...

rhuss:
ehm, the circle build is not ready yet as it looks like. We should probably make it not mandatory.

build.sh
```diff
@@ -73,9 +73,10 @@ function init_options() {
 
   # Internal variable default values
   OC_OPTS=""
-  MAVEN_OPTS=""
+  MAVEN_OPTS="--batch-mode"
```
rhuss: Can't we set batch-mode only when running in a CI ? Having it as default doesn't feel right to me (I like colors ;-)

chirino: true.

PR #22, opened 2017-10-30
refactor: Jenkinsfile now uses build.sh.

iocanel:


iocanel:
For now this can't pass until we get a hybrid image with slave maven and yarn.

PR #34, opened 2017-11-15
feat: Add version info to integration exports, and improve import error reporting.

chirino:


ui/src/app/integrations/list-page/list-page.component.html
```diff
@@ -20,10 +20,13 @@
 </ng-template>
 </syndesis-list-toolbar>
 
-<syndesis-loading [loading]="loading | async">
-  <syndesis-integrations-list [complete]="true"
-                              [integrations]="filteredIntegrations | async"></syndesis-integrations-list>
-</syndesis-loading>
+<!-- Support importing by DND right into the integraiton list area. Very google drive like -->
+<div ng2FileDrop style="height:100%" [uploader]="uploader">
```
deeleman: There is no need to inject styles right into the DOM element. We already have `list-page.component.scss` for that.

ui/src/app/integrations/list-page/list-page.component.ts
```diff
@@ -72,6 +71,16 @@ export class IntegrationsListPage implements OnInit {
           null,
           []
         );
+      } else if (status === 400) {
+        const r = JSON.parse(response);
```
deeleman: Naming things just as `r` is considered a code smell in JavaScript-land nowadays. It hinders code comprehension in long scripts. Perhaps `jsonResponse` would work better here ;-)

rest/model/src/main/java/io/syndesis/model/connection/Action.java
```diff
@@ -38,6 +38,8 @@
 @JsonIgnoreProperties(value = {"properties", "inputDataShape", "outputDataShape"}, allowGetters = true)
 public interface Action extends WithId<Action>, WithName, WithTags, WithConfigurationProperties, Serializable {
 
+    String VERSION = "1";
```
rhuss: Do you really want one version per object ? Sounds overkill to me, especially it is not guaranteed that object from different versions play well together. 

Couldn't we defined a 'model version' ?

rest/model/src/main/java/io/syndesis/model/Kind.java
```diff
@@ -24,30 +24,63 @@
...
-
-    Permission(io.syndesis.model.user.Permission.class),
-    Role(io.syndesis.model.user.Role.class),
-    User(io.syndesis.model.user.User.class),
-
-    TechExtension(io.syndesis.model.techextension.TechExtension.class),
-    TechExtensionAction(io.syndesis.model.techextension.TechExtensionAction.class),
+    Action(
+        io.syndesis.model.connection.Action.class,
+        io.syndesis.model.connection.Action.VERSION),
```
rhuss: How does this translate to deployment.json where we have

```yaml
   ...
  kind: "Action"
  ....
```

I'm really not sure whether we should put the version into the kind.

rest/model/src/main/java/io/syndesis/model/connection/Action.java
rhuss: Another question: Who is responsible for updating the version number ? How it is guaranteed that its not forgotten to increment when the model changes ?

ui/src/app/integrations/list-page/list-page.component.ts
chirino: I tend to agree when you have a large code block, but in small code blocks like this, shorter variable names seem easier to track.

chirino:
@rhuss yeah going with 1 schema version or going with a version per object is debatable.  I could go either way.  Seem the K8s style is 1 version per entity, thats why I went in this direction, but I could easily change this to be 1 version per entity schema.

chirino:
@rhuss note that this version info is only carried in ModelData objects so does not affect the DB or runtime at all.  It could show up in deployment.json, but since that's embedded into the app it's versions are expected to match the current versions.

We could build per entity version migrators. Could we use this while doing a DB version upgrade?  Not sure.  I'm sure some upgrades would require no migration DB changes if they just added optional fields.  But I think I'm getting way to ahead of what this PR was trying to tackle, which was just include a little bit of version info so that future versions of the app could identify this as an older export.

ui/src/app/integrations/list-page/list-page.component.html
chirino: fixed

ui/src/app/integrations/list-page/list-page.component.ts
chirino: but fixed

rhuss:
> But I think I'm getting way to ahead of what this PR was trying to tackle, which was just include a little bit of version info so that future versions of the app could identify this as an older export.

Agreed, therefore I am all for one single and global version. If they mismatch, then simply then no export is allowed for now. We could late add migrator which could take the update from version to version. I'm really afraid of a 'mixed' model, where objects start to evolve with different release cycles.

All or nothing is probably the way to go for now.

chirino:
I've changed this PR around so that it uses a single global version.  I refactors that dao.schema.version property from being a config item to being baked into that app at:
This version is what gets put into the export file.

https://github.com/chirino/syndesis/blob/c19a0df92b4c3a9778b68ca757dec01f98f94928/rest/model/src/main/java/io/syndesis/model/Schema.java#L23

chirino:
Fixes #126

PR #359, opened 2017-11-16
add reworked ux designs for integration details page

dongniwang:
Adding updated UX designs for integration details page. 

The designs covers the following user interactions:
- Integration Details Page basic layout
  - Expandable row for integration history (versions)
  - Kebab menu
  - Integration visualization (normal vs. many steps/connections)
- Creating/Replacing integration draft and publish a version from integration history
  - Status indicators at row level
  - Having notifications in place for publish action
  - Modal for confirming user actions

A few questions:
- How does republish integration work? 
- Can we provide users with information as to what became out of date? 

Reference: 
https://github.com/syndesisio/syndesis/issues/331
https://github.com/syndesisio/syndesis/issues/184


dongniwang:
Please review:
@sjcox-rh @seanforyou23 @amysueg @chirino @rhuss @TovaCohen @kcbabo @gashcrumb

gashcrumb:
+1 from me too

TovaCohen:
The updates look great. Here are suggestons for changes to two messages.
This image: integration_details_page_rework_11.16_publishaversion.png
Shows this message:
   "This action will stop the current published version and publish the version you've selected."
Do "stop" and "deactivate" mean the same thing? Elsewhere in the UI, the user selects "Deactivate" so I am wondering if this message should use  "deactivate" in place of "stop". Or maybe we should replace the other instances of "deactivate" with "stop". If they mean the same thing, the shorter term "stop" is preferred. Also, the convention is to not use contractions because non-native English speakers do not always understand them. The last part of the sentence should be: "... and publish the version you selected."

This image: integration_details_page_rework_11.16_replaceadraft.png
Shows this message:
   This action will replace the draft with Version 3.
   Are you sure you want to replace this draft?
I expect that the replaced draft is lost. Also, you are replacing "the" draft". Saying "this" draft implies that there are others. How about this:
   This action will replace the draft with Version 3. An integration can have exactly one draft. The draft being replaced will be lost.
    Are you sure you want to replace the draft?

gashcrumb:
@dongniwang you'll need to rebase this before it can be merged, let us know if you need a hand with that.

ux/designs/integration_details/integration_details_page.md
```diff
@@ -0,0 +1,115 @@
...
+#### Basic layout
+
+![basic layout](img/integration_details_page_rework_11.16_genericlayout.png)
+
+1. Moved status label to the top section right next to Integration Name. Using “check mark” icon to indicate this integration is active.
+
+2. Moving Draft out of the integration history. Users would be able to “Publish” and “Edit” the draft.
+
+3. Added row indicator (green bar) and checkmark icon to indicate active and running version. For active version, users would only be able to perform “Replace Draft” with the active version.
+
```
rhuss: For symmetrical reasons (e.g. to have a kebab menu on all rows), what's about to move the "Stop Integration" and the "Replace Draft" integration into the kebab menu for an active integration ? (So 'Stop integration' is the opposite of 'Publish' in the other integrations ?

ux/designs/integration_details/integration_details_page.md
```diff
@@ -0,0 +1,115 @@
...
+### Interacting with the page
+
+#### Kebab menu
+
+![kebab menu](img/integration_details_page_rework_11.16_kebab.png)
+
+1. Breadcrumb & toolbar level kebab, users are able to:
+
+  - Stop Integration
+  - Replace Draft
```
rhuss: With what would the Draft be replaced (as this is a menu without revision context) ?

Maybe a "Remove Draft" would be better at this point.

ux/designs/integration_details/integration_details_page.md
```diff
@@ -0,0 +1,115 @@
...
+
+  - Replace Draft
+  - Publish
+
+
+#### Modal for confirming actions
+
+![replace draft modal](img/integration_details_page_rework_11.16_replaceadraft.png)
+
+Once users click on “Replace Draft”, display a modal to notify users and to confirm the replace draft action.
```
rhuss: Same for "Remove Draft" (or "Clear Draft") in case the change above is ok.

ux/designs/integration_details/integration_details_page.md
```diff
@@ -0,0 +1,115 @@
...
+2. The version spinning up would have a spinner indicator.
+
+![being published](img/integration_details_page_rework_11.16_being_published.png)
+
+In this case, the published version is being “decommissioned” and the new version is being published.
+
+1. Show spinner for overall integration status.
+
+2. Show spinners for both the “decommissioning” version and the being published version.
+
```
rhuss: As an option, it would be nice to to see here would be an indicator what the progress is about and what the target stater (published / stopped) is.

E.g. the publish step actually means several 'sub-step':

* Building integration runtime
* Deploying integration runtime
* Running

Maybe we find some space to visiualize this (maybe as label below the progress meter).

Another idea: Coule we have two different progress meters: One showing the progress to starting up, and one for shutting down.

ux/designs/integration_details/integration_details_page.md
```diff
@@ -0,0 +1,115 @@
...
+
+#### Notifications (alerts)
+
+![warning message](img/integration_details_page_rework_11.16_warning.png)
+
+1. Display a warning message when a published integration is outdated.
+
```
rhuss: What actually means "outdated" ? As you can change a revision (you can only copy it over as draft), there is nothing like an revision being outdated imo.

ux/designs/integration_details/integration_details_page.md
amysueg: OK, @dongniwang and I agree; "Replace Draft" and "Stop Integration" will be kebab items for the current version.

ux/designs/integration_details/integration_details_page.md
amysueg: Replace Draft in this context means "replace the existing draft with a copy of the running integration"

ux/designs/integration_details/integration_details_page.md
amysueg: Since at this time there is no "Remove Draft", this comment is not relevant. We fear over-complicating this UI.

ux/designs/integration_details/integration_details_page.md
amysueg: Again, in order to keep things simple, we'd like to avoid additional information for this target user, at least until we hear feedback. If it's possible to add a hover tooltip, then we *might* be able to provide that kind of detail in the tooltip when the user hovers the cursor over the process indicator.

ux/designs/integration_details/integration_details_page.md
amysueg: This design is to satisfy the use case when there are updates available for an integration - 
Examples (from Hiram) - API client connector and technical extensions for example. We should perhaps change the wording of this message. @TovaCohen please advise. I think something like "Updates are available for this integration." might work? And that we should probably avoid using "outdated" as a term.

TovaCohen: How about: "Updates are available for one or more connections or steps in the active integration. Without these updates, your integration might not behave as expected."
I'm a little concerned about referring to the "published" integration because every integration listed in the history has been published. But there is only one "active" integration.

amysueg: That's great @TovaCohen and thank you for the quick turnaround.
I wonder if it's too long for the space allotted in the UI. @dongniwang could try this text to see what it looks like, but maybe we could use a slightly edited version?

TovaCohen: Is this short enough:
"An update is available for a connection or step. Republish to ensure correct behavior."
I think that saying that an update is available for the integration is just a little too vague. And I think that "An update" is fine even if there is more than one update, since it doesn't matter how many there are.

TovaCohen: Also, because of the location of this message, I think it is obvious that it applies to the active integration. So the message does not need to refer to the integration for which updates are available.

dongniwang: Thanks @TovaCohen! I'll go with "An update is available for a connection or step. Republish to ensure correct behavior." It's a little vague and generic. That's why I think we need the "more information" link if users want to find out what those updates are.

PR #362, opened 2017-11-16
API Client Connector design updates

sjcox-rh:
cc: @zregvart @kcbabo @chirino @gashcrumb @kahboom @dongniwang @amysueg @seanforyou23

Updated the PR to show a summary of swagger actions instead of a list. Also reordered the steps based on designs updates Dongni made last week.

Here is the design for the "Review Swagger Actions" step:
reviewswaggerfile


gashcrumb:
+1 from me, I'll give other folks a chance to have a look though before marking it approved.

PR #363, opened 2017-11-17
#294 Customizations Page

kahboom:
For API Connectors and Technical Extensions.

screenshot 2017-11-16 20 56 17

### DIFF
--- a/ui/src/app/customizations/api-connectors/api-connector-list.component.html
+++ b/ui/src/app/customizations/api-connectors/api-connector-list.component.html
@@ -1,0 +1,6 @@
+<div class="row api-connector-list">
+  <div class="col-xs-12">
+    <h1>API Client Connector</h1>
+    <p>Nullam quis risus eget urna mollis ornare vel eu leo. Find out more about this at <a href="https://access.redhat.com/documentation/en-us/red_hat_jboss_fuse/7.0-tp/html-single/fuse_ignite_sample_integration_tutorial/" rel="nofollow" target="_blank">{{ config.getSettings('branding', 'appName', 'Syndesis') }} Help</a>.</p>
+  </div>
+</div>
--- a/ui/src/app/customizations/api-connectors/api-connector-list.component.scss
+++ b/ui/src/app/customizations/api-connectors/api-connector-list.component.scss
@@ -1,3 +1,5 @@
+@import 'patternfly-sass/assets/stylesheets/patternfly/color-variables';
+
 :host {
   flex: 1 1 auto;
   height: 100%;
@@ -14,5 +16,11 @@
 
   .col-xs-12 {
     height: 100%;
+  }
+
+  & /deep/ .toolbar-pf {
+    background: inherit;
+    box-shadow: none;
+    border-bottom-color: transparent;
   }
 }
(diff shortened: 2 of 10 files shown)


ui/src/app/customizations/api-connectors/api-connector-list.component.html
```diff
@@ -0,0 +1,6 @@
+<div class="row api-connector-list">
+  <div class="col-xs-12">
+    <h1>API Client Connector</h1>
+    <p>Nullam quis risus eget urna mollis ornare vel eu leo. Find out more about this at <a href="https://access.redhat.com/documentation/en-us/red_hat_jboss_fuse/7.0-tp/html-single/fuse_ignite_sample_integration_tutorial/" rel="nofollow" target="_blank">{{ config.getSettings('branding', 'appName', 'Syndesis') }} Help</a>.</p>
```
deeleman: @kahboom A simple but highly effective way to optimize performance in Angular applications is by not computing values in the template itself. Please take a look to the code above. We have place in the template the following function call wrapped by an interpolation binding:

```{{ config.getSettings('branding', 'appName', 'Syndesis') }}```

This entails that Angular is re-running the `getSettings(...)` method on each change detection cycle. This might become an expensive operation if the function wraps a significant logic, not to mention that this kind of bindings are less readable.

What are the best workarounds for this? There're several approaches:

* using a getter
* binding that value just once to a class field
* using a _pure_ pipe. 

My favorite take in this case would be to use a pure pie. The pure pipe will only be re-run if the input reference type changes, which is unlikely the case, and you can also re-use it in some other instances of the application. This also allows to remove logic from the underlying component controller, since there is no need to inject the `config` object anymore:

```html
<a href="blah blah" rel="nofollow" target="_blank">
  {{ 'branding' | syndesysConfig: ['appName', 'Syndesis'] }} Help
</a>
```

A `SyndesisConfigPipe` would expect a string with the setting key we want to bring (`branding`) and the pipe should be fed with the group key that setting belongs to (`appName`). Additionally, the pipe payload can accept a second argument, which is the default value.

This approach features a minimum performance cost and makes that data binding available for the entire application, leading to less code in our components.

ui/src/app/customizations/api-connectors/api-connector-list.component.scss
```diff
@@ -0,0 +1,26 @@
...
+}
+
+.api-connector-list {
+  height: 100%;
+
+  .col-xs-12 {
+    height: 100%;
+  }
+
+  & /deep/ .toolbar-pf {
```
deeleman: uh.. .What out for this in the future ;) https://github.com/syndesisio/syndesis/issues/366

ui/src/app/customizations/api-connectors/api-connector-list.component.html
gashcrumb: @deeleman is there an issue open for this?  We could totally adopt this approach, let's create an issue to expose the config via a pipe in this manner if it's not already open.

kahboom: Sorry, I completely copy-pasta'd this, but I agree that this would be a good approach. I will adjust accordingly once we can work refactoring this throughout the app. @gashcrumb I checked very briefly but couldn't find anything so I opened one here: #392

ui/src/app/customizations/api-connectors/api-connector-list.component.scss
kahboom: Copy pasta! What can I say..

PR #365, opened 2017-11-17
cleanup: Get more dry by having some of the mvn modules share a common parent pom.

chirino:


connectors/connectors/day-trade/day-trade-get-connector/pom.xml
```diff
@@ -156,7 +156,17 @@
           </execution>
         </executions>
       </plugin>
-
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>default-jar</id>
```
zregvart: I don't understand: why are we no longer creating JARs for connectors? And if this is true for all connectors now, should we just put this in connectors parent, we'll surely create a connector without `maven-jar-plugin` disabled...

connectors/pom.xml
```diff
@@ -171,6 +171,59 @@
             <generateGitPropertiesFilename>${project.build.outputDirectory}/git.properties</generateGitPropertiesFilename>
           </configuration>
         </plugin>
+
+
+        <!-- Disable some of these checks until we get the code base in basepom shape -->
+        <plugin>
+          <groupId>com.mycila</groupId>
+          <artifactId>license-maven-plugin</artifactId>
```
zregvart: This one should be relatively easy to keep, it has a `:format` goal to add the license headers.

pom.xml
```diff
@@ -113,4 +139,57 @@
...
+              <exclude>archetypes/**/src/main/resources/archetype-resources/**</exclude>
+            </excludes>
+          </configuration>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+    <plugins>
+      <plugin>
+        <groupId>org.commonjava.maven.plugins</groupId>
+        <artifactId>directory-maven-plugin</artifactId>
```
zregvart: That's a nice plugin to find out about, thanks 🙌

rest/pom.xml
```diff
@@ -846,9 +847,9 @@
         <plugin>
           <artifactId>maven-checkstyle-plugin</artifactId>
           <configuration>
-            <configLocation>../checkstyle.xml</configLocation>
+            <configLocation>${basedir}/../checkstyle.xml</configLocation>
```
zregvart: Should we use `syndesis.basedir` here?

rest/pom.xml
```diff
@@ -846,9 +847,9 @@
...
             <propertyExpansion>
-              checkstyle.suppression.file=../checkstyle-suppressions.xml
+              checkstyle.suppression.file=${basedir}/../checkstyle-suppressions.xml
```
zregvart: And here, should we use `syndesis.basedir` here?

runtime/runtime/pom.xml
```diff
@@ -146,22 +146,42 @@
...
+            </ignoredUnusedDeclaredDependencies>
+          </configuration>
+        </plugin>
+
+        <!-- Disable some of these checks until we get the code base in basepom shape -->
+        <plugin>
+          <groupId>com.mycila</groupId>
+          <artifactId>license-maven-plugin</artifactId>
+          <configuration>
+            <skip>true</skip>
```
zregvart: I think we can have this enabled, adding license headers should be just invoking the `:format` goal.

s2i/pom.xml
```diff
@@ -88,6 +110,14 @@
           </images>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>com.mycila</groupId>
+        <artifactId>license-maven-plugin</artifactId>
+        <configuration>
+          <!-- TODO: re-enable -->
+          <skip>true</skip>
```
zregvart: Easy to keep also

ui/pom.xml
```diff
@@ -53,6 +78,26 @@
...
 
   <build>
+    <pluginManagement>
+      <plugins>
+         <!--Disable some of these checks until we get the code base in basepom shape-->
+        <plugin>
+          <groupId>com.mycila</groupId>
+          <artifactId>license-maven-plugin</artifactId>
+          <configuration>
+            <skip>true</skip>
```
zregvart: Easy to keep

connectors/connectors/day-trade/day-trade-get-connector/pom.xml
chirino: We do generate a jar.  It's the camel-connector-maven-plugin:2.20.0:jar which creates the jar.  We need to disable the default maven-jar-plugin generation since that fails when it notices the jar has already been created.

connectors/pom.xml
chirino: Lots of these files are also generated. And state don't edit me. So I wanted to skip the headache of addressing these at this time.

rest/pom.xml
chirino: Indeed.  Will do.

rest/pom.xml
chirino: yep.. I had put this in before I figured out the syndesis.basedir trick.

chirino:
Yeah lots of these pmd plugin disables are just in place to keep this initial PR small.  I forsee us doing lots a smaller subsequent PRs where we enable the check and bring in the fixes.

PR #370, opened 2017-11-17
chore: use latest 1.32.0 SNAPSHOT of datamapper

igarashitm:
So for me this change just worked to retrieve latest 1.32.0 SNAPSHOT while it was always retrieving `20171018190510` before. @syndesisio/ui-api WDYT?


gashcrumb:
@igarashitm LGTM!

PR #382, opened 2017-11-20
refactor: Move helper scripts to top-leve "tools-dir"

rhuss:


zregvart:
@rhuss seems that  #385 was merged before this so it needs a rebase

PR #383, opened 2017-11-20
feat: introduce errorprone checks

zregvart:
This adds errorprone[1] compiler extension that provides several new
checks[2] some of which were not caught by our existing checks.

Also goes one step more and declares all compiler warnings as errors.

Well, you know me, I need my checks :)

To argument this one -- actual bugs not caught by other checks:
 - Not using correct enum in comparison (https://github.com/syndesisio/syndesis-rest/compare/master...zregvart:errorprone?expand=1#diff-e9da60db20876b6c73daa9758bf6fb50L184)
 - Using platform charset (numerous places)
 - Not checking return values of methods (numerous places, most notably fluent assertThat from AssertJ)

[1] http://errorprone.info/
[2] http://errorprone.info/bugpatterns


chirino:
@zregvart This seems to need a rebase.

zregvart:
@chirino rebased, could you take another look?

chirino:
Still some conflicts.

PR #386, opened 2017-11-20
Fixes #368 by refusing to import a existing extension

### LINKED ISSUE #368: Tech extension : tech extension upload should fail if an extension with the same id already defined
When invoking the endpoint `/api/v1beta1/extensions` there is no validation about if an extension with the same `extensionId` is already defined in the system so there is no warning about potential impacts.

In such case we should reject the extension and ask the user to perform an update.

nicolaferraro:


lburgazzoli:
retest this please

PR #390, opened 2017-11-20
feat: Add technical extensions page to customizations, stub in import and detail page

gashcrumb:
Also adds some updates to the generated model file.

Changed the generator code so it no longer will overwrite model.ts since changes have been manually added in past commits.  Added `Extension` and related types to the model.

Relates to #264 #262 and #263

### DIFF
--- a/ui/src/app/customizations/customizations.component.html
+++ b/ui/src/app/customizations/customizations.component.html
@@ -1,12 +1,9 @@
 <ul class="row nav nav-tabs nav-tabs-pf toolbar-pf">
   <li routerLinkActive="active">
-    <a [routerLink]="['api-connectors']">API Client Connector</a>
+    <a [routerLink]="['api-connectors']">API Client Connectors</a>
   </li>
-  <li>
-    <a>Technical Extension</a>
-  </li>
-  <li>
-    <a>Custom Connector</a>
+  <li routerLinkActive="active">
+    <a [routerLink]="['tech-extensions']">Technical Extensions</a>
   </li>
 </ul>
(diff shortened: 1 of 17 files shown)


gashcrumb:
taking off WIP label as I think we could stand to have the code generator changes merged.

ui/src/app/customizations/customizations.component.html
```diff
@@ -1,12 +1,9 @@
 <ul class="row nav nav-tabs nav-tabs-pf toolbar-pf">
   <li routerLinkActive="active">
-    <a [routerLink]="['api-connectors']">API Client Connector</a>
+    <a [routerLink]="['api-connectors']">API Client Connectors</a>
```
kahboom: The design reads 'API Client Connector' though: https://github.com/sjcox-rh/syndesis-ux/blob/70f15abbfc56fdd11af34f7adb9cd0ccbda194c9/designs/apiconnector/apiconnector.md

I agree plural sounds better though.

gashcrumb: Funny on the tech extension design it actually is plural, just figured I'd adjust it accordingly :-)

I think we should start logging bugs for these kinds of things @kahboom WDYT?

kahboom: +1

sjcox-rh: Agreed, let's keep it consistent and make API plural, so API Client Connectors. Sorry for the inconsistency! @kahboom @gashcrumb

PR #393, opened 2017-11-21
feat: Design proposal for custom connectors

zregvart:
This is the initial design proposal for custom connectors.

### DIFF
--- a/project/proposals/119-custom-connectors.md
+++ b/project/proposals/119-custom-connectors.md
@@ -1,0 +1,229 @@
+# Custom connectors
+
+ * Issue: https://github.com/syndesisio/syndesis-project/issues/119
+ * Sprint: 20
+
+## Background
+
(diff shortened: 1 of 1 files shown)


rhuss:
@zregvart is it already ready for review (as it is still marked as wip) ?

zregvart:
> @zregvart is it already ready for review (as it is still marked as wip) ?

needs updating to the latest API in #450

rhuss:
@zregvart ok, let me know when its good to go so I (or somebody else, too) can do the review.

zregvart:
@rhuss I'm still not satisfied with the breadth of the design. I think it shows the most important aspects of that are needed for UI work and we can fill in the details when we see that this works in practice.

rhuss:
Fine with that, going to review it later today ... (but hopefully before the meeting marathon this afternoon. If not then afterwards ...)

PR #395, opened 2017-11-21
feat(connector-template): adds connector template support and Swagger Connector generator

zregvart:
Connector templates can be considered blueprints for creating
Connectors. Given a connector template a Connector generator
implementation generates a Connector.

This adds a SwaggerConnectorGenerator to sculpt the Connector with
actions, shapes and properties from Swagger specification.


zregvart:
@jimmidyson this is the same code (rebased) from https://github.com/syndesisio/syndesis-rest/pull/757 that somehow didn't make the cut to single repository, can you take a look?

PR #406, opened 2017-11-21
feat: use the syndesis/syndesis-s2i builder image for faster builds.

chirino:
Also get the s2i/pom.xml to actually build the docker image.


chirino:
Part of #185.  This change should help further reduce build times.

PR #412, opened 2017-11-21
feat: implement tech extension upload and most of the detail page

### LINKED ISSUE #264: Implement the import technical extension pages
|</img> @gashcrumb | [2017-11-09](https://github.com/syndesisio/syndesis-ui/issues/1203) | enhancement, Priority - High, sprint requirement  |
|-|-|-|

This is a number of pages in this section -> https://github.com/syndesisio/syndesis-ux/blob/master/designs/technical_extensions/tech_ext.md#import-new-technical-extension

Might want to break this up into smaller tasks.

@lburgazzoli FYI

gashcrumb:
Makes it possible to upload a tech extension via the UI and import it, fixes the list so it works properly, also adds some of the detail page.  Need to implement deleting and updating tech extensions, as well as fetch them via the step store (or somehow) in the integration editor.

Relates to #262 #263 
Fixes #264 

@lburgazzoli FYI...

### DIFF
--- a/ui/src/app/customizations/tech-extensions/detail/tech-extension-detail.component.ts
+++ b/ui/src/app/customizations/tech-extensions/detail/tech-extension-detail.component.ts
@@ -1,14 +1,32 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, OnInit, OnDestroy } from '@angular/core';
+import { ActivatedRoute, Router } from '@angular/router';
+import { Observable } from 'rxjs/Observable';
+import { Subscription } from 'rxjs/Subscription';
+import { Extension } from '../../../model';
 import { ExtensionStore } from '../../../store/extension/extension.store';
 
 @Component({
   selector: 'syndesis-tech-extension-detail',
   templateUrl: 'tech-extension-detail.component.html'
 })
-export class TechExtensionDetailComponent implements OnInit {
-  constructor(private store: ExtensionStore) { }
+export class TechExtensionDetailComponent implements OnInit, OnDestroy {
+  extension: Extension;
+  loading: Observable<boolean>;
+  subscription: Subscription;
+  constructor(private store: ExtensionStore,
+              private router: Router,
+              private route: ActivatedRoute) {
+    this.loading = store.loading;
+               }
 
   ngOnInit() {
-    // WIP
+    this.subscription = this.store.resource.subscribe(extension => {
+      this.extension = extension;
+    });
+    this.route.paramMap.first(params => params.has('id')).subscribe(params => this.store.load(params.get('id')));
    }
+
+  ngOnDestroy() {
+    this.subscription.unsubscribe();
+  }
 }
--- a/ui/src/app/customizations/tech-extensions/detail/tech-extension-detail.component.ts
+++ b/ui/src/app/customizations/tech-extensions/detail/tech-extension-detail.component.ts
@@ -1,7 +1,6 @@
-import { Component, OnInit, OnDestroy } from '@angular/core';
+import { Component, OnInit } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
 import { Observable } from 'rxjs/Observable';
-import { Subscription } from 'rxjs/Subscription';
 import { Extension } from '../../../model';
 import { ExtensionStore } from '../../../store/extension/extension.store';
 
@@ -9,24 +8,17 @@ import { ExtensionStore } from '../../../store/extension/extension.store';
   selector: 'syndesis-tech-extension-detail',
   templateUrl: 'tech-extension-detail.component.html'
 })
-export class TechExtensionDetailComponent implements OnInit, OnDestroy {
-  extension: Extension;
-  loading: Observable<boolean>;
-  subscription: Subscription;
-  constructor(private store: ExtensionStore,
+export class TechExtensionDetailComponent implements OnInit {
+  extension$: Observable<Extension>;
+  loading$: Observable<boolean>;
+  constructor(private extensionStore: ExtensionStore,
               private router: Router,
               private route: ActivatedRoute) {
-    this.loading = store.loading;
-               }
+    this.loading$ = this.extensionStore.loading;
+    this.extension$ = this.extensionStore.resource;
+  }
 
   ngOnInit() {
-    this.subscription = this.store.resource.subscribe(extension => {
-      this.extension = extension;
-    });
-    this.route.paramMap.first(params => params.has('id')).subscribe(params => this.store.load(params.get('id')));
+    this.route.paramMap.first(params => params.has('id')).subscribe(params => this.extensionStore.load(params.get('id')));
    }
-
-  ngOnDestroy() {
-    this.subscription.unsubscribe();
-  }
 }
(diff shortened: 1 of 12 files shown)


ui/src/app/customizations/tech-extensions/detail/tech-extension-detail.component.ts
```diff
@@ -1,14 +1,32 @@
...
+              private router: Router,
+              private route: ActivatedRoute) {
+    this.loading = store.loading;
+               }
 
   ngOnInit() {
-    // WIP
+    this.subscription = this.store.resource.subscribe(extension => {
+      this.extension = extension;
+    });
```
deeleman: You can streamline the component implementation by using observables as is, so you can declutter the component by removing subscriptions and stuff. First, turn the `extension`class field into this:

`extension$: Observable<Extension>;`

And then remove the subscription above and use this expression instead:

`this.extension$ = this.store.resource;`

TIP: If you name object instances by the camelCased version of its injected type - `ExtensionStore` in this case, code becomes more expressive. The above would look more self-explanatory if it was 

`this.extension$ = this.extensionStore.resource;`

Now you can completely remove the `subscription: Subscription` field and the `ngOnDestroy` blocks below. The only action required then will be to go the components template and apply this simple change:

```html
<syndesis-loading [loading]="loading | async">
   <div class="row" *ngIf="extension$ | async; let extension = extension$">
```

And that's it! ;)

deeleman: FYI, the Angular engine automatically "completes" any observable binding piped with the `async` pipe. That is why it is recommended to bind observable information directly, rather than subscribing to it (which entails tearing down the subscriptions by hand upon disposing the component and stuff)

ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
```diff
@@ -1,15 +1,111 @@
...
 
 @Component({
   selector: 'syndesis-tech-extentions-import',
   templateUrl: 'tech-extension-import.component.html'
 })
+export class TechExtensionImportComponent {
+
+  uploader: FileUploader;
+  response: Extension = undefined;
+  error: FileError = undefined;
```
deeleman: Why are these fields initialized as `undefined`? I've seen this pattern repeated all over the app and I still cannot wrap my head around on why we are doing this.

ui/src/app/integrations/list/list.component.scss
```diff
@@ -10,6 +11,7 @@
       }
     }
   }
+  */
```
deeleman: If this piece of CSS is no longer required, I'd say we should remove it instead of commenting it out.

ui/src/app/integrations/list/list.component.scss
```diff
@@ -1,6 +1,7 @@
 @import 'patternfly-sass/assets/stylesheets/patternfly/color-variables';
 .integrations.list {
   color: #656565;
```
deeleman: Harcoded hex colors in component stylesheets make application theming more complex upon time. Don't we have a centralized `_palette.scss` partial or something? I'd move this one there.

ui/src/app/store/extension/extension.service.ts
```diff
@@ -1,13 +1,23 @@
...
-  constructor(restangular: Restangular) {
+  constructor(restangular: Restangular, private http: Http) {
     super(restangular.service('../v1beta1/extensions'), 'extension');
   }
+
+  public getUploadUrl() {
+    return this.restangularService.one().getRestangularUrl();
+  }
+
+  public importExtension(id: string) {
```
deeleman: I miss the type annotation in these methods, so one can spot  at a glance what do they return instead of having to inspect them through the IDE's tschecker or looking into the method body.

ui/src/app/store/extension/extension.store.ts
```diff
@@ -17,4 +17,12 @@ export class ExtensionStore extends AbstractStore<
...
     return 'Extension';
   }
+
+  public getUploadUrl() {
+    return this.service.getUploadUrl();
+  }
+
+  public importExtension(id: string) {
+    return this.service.importExtension(id);
+  }
```
deeleman: I'd annotate the methods with their returning types.

ui/src/app/customizations/tech-extensions/tech-extensions-list.component.ts
```diff
@@ -10,19 +11,22 @@ import {
...
+import { Extensions, Extension } from '../../model';
 
 @Component({
   selector: 'syndesis-tech-extensions-list',
-  templateUrl: 'tech-extensions-list.component.html'
+  templateUrl: 'tech-extensions-list.component.html',
+  styleUrls: ['./tech-extensions-list.component.scss']
 })
 export class TechExtensionsListComponent implements OnInit {
   extensions: Observable<Extensions>;
```
deeleman: The convention for observable fields and variables is to name them with a trailing `$` (`extensions$` in this case), so one can recognize at a glance that this variable or that class property is observable and hence requires to be handled with special care.

ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
```diff
@@ -1,15 +1,111 @@
...
+      }
+      if (status === 200) {
+        this.response = <Extension> resp;
+        return;
+      }
+      this.error = {
+        level: 'alert alert-danger',
+        message: resp.userMsg || 'An unknown error has occurred'
+      };
+    };
```
deeleman: I'd say this is too much logic for a constructor. I know that constructors are meant to initialize stuff but perhaps this is too much. On the other hand most of this logic seems that might be of interest for other components that require upload functionalities.

Perhaps it would be worth to promote this to a parent class this component might extend from or even an injectable service that receives the uploader, error, response and fileSelect in its initialize payload.

ui/src/app/integrations/list/list.component.scss
gashcrumb: We do on the line above.  I don't even recall what this is for and it doesn't match any patternfly colors so I'll take it out.

ui/src/app/integrations/list/list.component.scss
gashcrumb: Yup, totally meant to do that, will take it out.

ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
gashcrumb: Habit, will take that out :-)

ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
gashcrumb: All this code is basically for initializing the file upload widget, first creating an instance and then overriding some handlers.  I could move it to `ngOnInit` I guess.

ui/src/app/customizations/tech-extensions/detail/tech-extension-detail.component.ts
gashcrumb: Will do, thanks!

ui/src/app/store/extension/extension.service.ts
gashcrumb: Sure!

ui/src/app/store/extension/extension.store.ts
gashcrumb: Yup, will do...

rhuss:
@deeleman does this look ok to you ? If so, could you please approve to the this PR merged

thanks ;-) ...

PR #420, opened 2017-11-22
Fixes #410: additional endpoints for extension details

### LINKED ISSUE #410: Can we inline the integrations used by an extension rather than requiring a separate call
Otherwise implementing this page is going to be chatty between the UI and the rest API:

https://github.com/syndesisio/syndesis/blob/master/ux/designs/technical_extensions/tech_ext.md#tech-extension-list

It also complicates implementing the detail page as we have to make a second API call to fetch what integrations are used by an extension.

nicolaferraro:
Added two more endpoints:
- `/extensions/details`
- `/extensions/{id}/details`

Returning all the data needed by ui.

### DIFF
--- a/rest/model/src/main/java/io/syndesis/model/extension/ExtensionDetail.java
+++ b/rest/model/src/main/java/io/syndesis/model/extension/ExtensionDetail.java
@@ -20,6 +20,7 @@ import io.syndesis.model.ResourceIdentifier;
 import org.immutables.value.Value;
 
 import java.io.Serializable;
+import java.util.Collections;
 import java.util.Set;
 
 @Value.Immutable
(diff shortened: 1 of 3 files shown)


rest/model/src/main/java/io/syndesis/model/extension/ExtensionDetail.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.Serializable;
+import java.util.Set;
+
+@Value.Immutable
+@JsonDeserialize(builder = ExtensionDetail.Builder.class)
+public interface ExtensionDetail extends Serializable {
+
+    Extension getExtension();
+
+    Set<ResourceIdentifier> activeIntegrations();
```
lburgazzoli: An extension may not have any active integration, should't we have a default returning an empty set ?

nicolaferraro: Yep, also the method signature should be `getActiveIntegrations()`

lburgazzoli:
retest this please

PR #427, opened 2017-11-22
Build the ui’s docker image with mvn too.

chirino:
Consolidate some of the config into the parent pom.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,10 +34,17 @@
   <description>syndesis :: ${project.artifactId}</description>
 
   <properties>
+
+    <syndesis-ui-image-name>syndesis/syndesis-ui</syndesis-ui-image-name>
+    <syndesis-s2i-image-name>syndesis/syndesis-s2i</syndesis-s2i-image-name>
+
+    <docker.maven.plugin.fabric8.version>0.22.1</docker.maven.plugin.fabric8.version>
+    <maven.exec.plugin.version>1.2.1</maven.exec.plugin.version>
     <clean-maven-plugin-version>2.4.1</clean-maven-plugin-version>
     <frontend-maven-plugin-version>1.6</frontend-maven-plugin-version>
     <node.version>v6.11.5</node.version>
     <yarn.version>v1.2.1</yarn.version>
+
   </properties>
   
   <modules>
--- a/pom.xml
+++ b/pom.xml
@@ -159,6 +159,7 @@
               <exclude>**/spring.factories</exclude>
               <exclude>**/*.adoc</exclude>
               <exclude>archetypes/**/src/main/resources/archetype-resources/**</exclude>
+              <exclude>s2i/src/main/docker/project/**</exclude>
             </excludes>
           </configuration>
         </plugin>
--- a/ui/pom.xml
+++ b/ui/pom.xml
@@ -144,49 +144,65 @@
           </filesets>
         </configuration>
       </plugin>
-      <plugin>
-        <artifactId>maven-antrun-plugin</artifactId>
-        <version>1.8</version>
-        <executions>
-          <execution>
-            <phase>package</phase>
-            <goals>
-              <goal>run</goal>
-            </goals>
-          </execution>
-        </executions>
-        <configuration>
-          <target>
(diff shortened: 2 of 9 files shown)


ui/pom.xml
```diff
@@ -152,4 +146,63 @@
...
+                <sync todir="${basedir}/docker/dist">
+                  <fileset dir="${basedir}/dist"/>
+                </sync>
+              </target>
+            </configuration>
+          </plugin>
+          <plugin>
+            <groupId>io.fabric8</groupId>
+            <artifactId>docker-maven-plugin</artifactId>
+            <version>${docker.maven.plugin.fabric8.version}</version>
```
rhuss: We could also use the fabric8-maven-plugin here for doing the build so we can still work with image-streams on minishift (and automatic redeployments), also for the ui.

The advantage using stream should be that the pods get automatically redeployed after the build without the need to explicitly kill them. 

I'm happy to merge now, and switch over later to fabric8-maven-plugin.

ui/pom.xml
```diff
@@ -152,4 +146,63 @@
...
+              <execution>
+                <phase>package</phase>
+                <goals>
+                  <goal>run</goal>
+                </goals>
+              </execution>
+            </executions>
+            <configuration>
+              <target>
+                <sync todir="${basedir}/docker/dist">
```
rhuss: Could we use `${project.build.directory}/docker/dist` instead of copying over into a top-level dir ?  This would then end up in `target/` and hence also gets cleaned up with a  `mvn clean`. I'd like to avoid mixing directories with committed (source) files and generated files (dist). 

For example, it's hard to clean up such a setup and if e.g. a files gets removed which should not appear in dist/ but you already had a run before then you end up with the old files still in `docker/dist`.  The ant sync taks takes care of that, but I'd still would prefer the 'mvn way' (i.e. working in target, with `mvn clean` giving you a fresh state).

rhuss: Of course one would need to copy over the Dockerfile + helper files then, too, to target, before calling docker.

verifier/pom.xml
```diff
@@ -281,6 +281,37 @@
...
+            <version>${fabric8-maven-plugin.version}</version>
+            <executions>
+              <execution>
+                <id>build-docker-image</id>
+                <goals>
+                  <goal>build</goal>
+                </goals>
+                <phase>package</phase>
+                <configuration>
+                  <mode>kubernetes</mode>
```
rhuss: Not sure whether we should enforce kubernetes mode, but I'm fine with that for now until I have verified that image streams are working nicely, too.

pom.xml
```diff
@@ -34,10 +34,17 @@
   <description>syndesis :: ${project.artifactId}</description>
 
   <properties>
+
+    <syndesis-ui-image-name>syndesis/syndesis-ui</syndesis-ui-image-name>
+    <syndesis-s2i-image-name>syndesis/syndesis-s2i</syndesis-s2i-image-name>
+
+    <docker.maven.plugin.fabric8.version>0.22.1</docker.maven.plugin.fabric8.version>
```
rhuss: 0.23.0 is the latest

verifier/pom.xml
chirino: Yeah.. would be nice to detect openshift env, and enable openshift style image builds.

ui/pom.xml
chirino: Yeah that sounds like a good idea.

ui/pom.xml
chirino: Yeah not a bad idea.  Isn't the a src/main/docker dir for that?

rhuss: yeah, we can put the Dockerfile into this directory. But we then still have to copy over these files + dist/ into a common directory below `target/` and then refer `<dockerFile>` to this merged directory.

I can optimize this later (we can use an assembly also together with a Dockerfile) but I would do this then when moving to fabric8-maven-plugin (not sure what version fabric8-maven-plugin uses which docker-maven-plugin).

rhuss:
Happy with this now. @chirino if you dont mind, I'm merging now and tune it later.

PR #438, opened 2017-11-23
fix: install syndesis-parent POM

zregvart:
Seems that our builds are failing as the syndesis-parent POM cannot be
resolved. This adds a stage to install just the syndesis-parent POM in
the local Maven repository.


zregvart:
Found a better solution in defining relative paths in all Maven modules.

PR #442, opened 2017-11-23
#271 #441 API Connector

### LINKED ISSUE #441: API Connector Setup
Create the module, components, templates, and routes in preparation for the API connector. Since the data structure is likely to change in the coming days, we cannot use dummy data based on a sustainable data structure. In the meantime, we can continue to scaffold out the API connector templates and adjust accordingly.

kahboom:
Adds the module, basic component, template, and routes for the API Client Connector list, wizard, and detail pages.

fix #441 

cc @zregvart


s2i/src/main/docker/project/pom.xml
```diff
@@ -1,21 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
```
kahboom: Erhm, I didn't do this. Guessing we should be adding this to `.gitignore` or something?

kahboom:
Not much to see here, but this is the empty state. The rest are pretty barren as well, but this is just for the setup. @dongniwang @sjcox-rh Maybe you can chime in on the latin text? It's in the designs here: https://github.com/sjcox-rh/syndesis-ux/blob/70f15abbfc56fdd11af34f7adb9cd0ccbda194c9/designs/apiconnector/apiconnector.md

screenshot 2017-11-23 16 34 37

PR #450, opened 2017-11-24
feat: Custom connector REST API

### LINKED ISSUE #434: Implement REST API for custom connectors
The REST API from POC (#308) only allows listing custom connector templates and creating new custom connectors. The design proposal is in PR #393 to refactor and add other missing endpoints.

zregvart:
Extends the contract of `ConnectorGenerator` to provide a basic
information about the new custom connector.

Adds REST API for Custom connectors, first set of API endpoints deals
with enumerating connector templates:

 - `GET /api/v1/custom/connectors` - lists all known connector templates
 - `GET /api/v1/custom/connectors/{id}` - fetches specific connector
   template
 - `POST /api/v1/custom/connectors/{id}/info` - receives a custom
   connector definition, combines it with the template and using a
   connector generator provides basic information about the connector
   (used for confirmation/summary step before creating a new custom
   connector)
 - `POST /api/v1/custom/connectors/{id}/validation` - validates the
   given custom connector definition
 - `POST /api/v1/custom/connectors/{id}` - creates a new custom
   connector

Fixes #434

### DIFF
--- a/rest/connector-generator/src/main/java/io/syndesis/connector/generator/ConnectorGenerator.java
+++ b/rest/connector-generator/src/main/java/io/syndesis/connector/generator/ConnectorGenerator.java
@@ -21,7 +21,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import io.syndesis.core.Names;
+import io.syndesis.core.KeyGenerator;
 import io.syndesis.model.connection.Connector;
 import io.syndesis.model.connection.ConnectorTemplate;
 import io.syndesis.model.connection.CustomConnector;
@@ -45,8 +45,7 @@ public abstract class ConnectorGenerator {
             .orElseGet(() -> determineConnectorName(connectorTemplate, customConnector));
 
         final Connector.Builder connectorBuilder = new Connector.Builder()//
-            .id(customConnector.getId()
-                .orElseGet(() -> Names.sanitize(connectorTemplate.getName()) + ":" + Names.sanitize(name)))//
+            .id(KeyGenerator.createKey())//
             .name(name)//
             .description(customConnector.getDescription())//
             .icon(customConnector.getIcon())//
--- a/rest/connector-generator/src/main/java/io/syndesis/connector/generator/ConnectorGenerator.java
+++ b/rest/connector-generator/src/main/java/io/syndesis/connector/generator/ConnectorGenerator.java
@@ -17,6 +17,7 @@ package io.syndesis.connector.generator;
 
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -25,9 +26,13 @@ import io.syndesis.model.connection.Connector;
 import io.syndesis.model.connection.ConnectorTemplate;
 import io.syndesis.model.connection.CustomConnector;
 
-public interface ConnectorGenerator {
+public abstract class ConnectorGenerator {
 
-    default Connector baseConnectorFrom(final ConnectorTemplate connectorTemplate,
+    public abstract Connector generate(ConnectorTemplate connectorTemplate, CustomConnector customConnector);
+
+    public abstract Connector info(ConnectorTemplate connectorTemplate, CustomConnector customConnector);
+
+    protected final Connector baseConnectorFrom(final ConnectorTemplate connectorTemplate,
         final CustomConnector customConnector) {
         final Set<String> properties = connectorTemplate.getProperties().keySet();
 
@@ -36,10 +41,13 @@ public interface ConnectorGenerator {
             .filter(e -> properties.contains(e.getKey()))//
             .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
 
+        final String name = Optional.ofNullable(customConnector.getName())
+            .orElseGet(() -> determineConnectorName(connectorTemplate, customConnector));
+
         final Connector.Builder connectorBuilder = new Connector.Builder()//
-            .id(customConnector.getId().orElseGet(
-                () -> Names.sanitize(connectorTemplate.getName()) + ":" + Names.sanitize(customConnector.getName())))//
-            .name(customConnector.getName())//
+            .id(customConnector.getId()
+                .orElseGet(() -> Names.sanitize(connectorTemplate.getName()) + ":" + Names.sanitize(name)))//
+            .name(name)//
             .description(customConnector.getDescription())//
             .icon(customConnector.getIcon())//
             .properties(connectorTemplate.getConnectorProperties())//
@@ -49,7 +57,14 @@ public interface ConnectorGenerator {
         return connectorBuilder.build();
     }
 
-    Connector generate(ConnectorTemplate connectorTemplate, CustomConnector customConnector);
-
-    Connector info(ConnectorTemplate connectorTemplate, CustomConnector customConnector);
+    /**
+     * Determines the newly created connector name.
+     *
+     * @param connectorTemplate connector template
+     * @param customConnector custom connector definition
+     */
+    protected String determineConnectorName(final ConnectorTemplate connectorTemplate,
+        final CustomConnector customConnector) {
+        return "unspecified";
+    }
 }
--- a/rest/connector-generator/src/main/java/io/syndesis/connector/generator/ConnectorGenerator.java
+++ b/rest/connector-generator/src/main/java/io/syndesis/connector/generator/ConnectorGenerator.java
@@ -23,23 +23,25 @@ import java.util.stream.Collectors;
 import io.syndesis.core.Names;
 import io.syndesis.model.connection.Connector;
 import io.syndesis.model.connection.ConnectorTemplate;
+import io.syndesis.model.connection.CustomConnector;
 
 public interface ConnectorGenerator {
 
-    default Connector baseConnectorFrom(final ConnectorTemplate connectorTemplate, final Connector template) {
+    default Connector baseConnectorFrom(final ConnectorTemplate connectorTemplate,
+        final CustomConnector customConnector) {
         final Set<String> properties = connectorTemplate.getProperties().keySet();
 
-        final Map<String, String> configuredProperties = template.getConfiguredProperties()//
+        final Map<String, String> configuredProperties = customConnector.getConfiguredProperties()//
             .entrySet().stream()//
             .filter(e -> properties.contains(e.getKey()))//
             .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
 
         final Connector.Builder connectorBuilder = new Connector.Builder()//
-            .id(template.getId()
-                .orElseGet(() -> Names.sanitize(connectorTemplate.getName()) + ":" + Names.sanitize(template.getName())))//
-            .name(template.getName())//
-            .description(template.getDescription())//
-            .icon(template.getIcon())//
+            .id(customConnector.getId().orElseGet(
+                () -> Names.sanitize(connectorTemplate.getName()) + ":" + Names.sanitize(customConnector.getName())))//
+            .name(customConnector.getName())//
+            .description(customConnector.getDescription())//
+            .icon(customConnector.getIcon())//
             .properties(connectorTemplate.getConnectorProperties())//
             .configuredProperties(configuredProperties)//
             .connectorGroup(connectorTemplate.getConnectorGroup());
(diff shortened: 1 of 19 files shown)


zregvart:
@kahboom this should be the API to use, I'm going to update the #393 to reflect this, I've checked the generated Swagger and it should be OK

rest/connector-generator/src/main/java/io/syndesis/connector/generator/ConnectorGenerator.java
```diff
@@ -17,35 +17,54 @@
...
 import io.syndesis.core.Names;
 import io.syndesis.model.connection.Connector;
 import io.syndesis.model.connection.ConnectorTemplate;
+import io.syndesis.model.connection.CustomConnector;
 
-public interface ConnectorGenerator {
+public abstract class ConnectorGenerator {
 
-    default Connector baseConnectorFrom(final ConnectorTemplate connectorTemplate, final Connector template) {
+    public abstract Connector generate(ConnectorTemplate connectorTemplate, CustomConnector customConnector);
```
rhuss: more a question but I always wonder when to use an abstract class vs. an interface with default methods these days. Any hint ?

rest/model/src/main/java/io/syndesis/model/connection/CustomConnector.java
```diff
@@ -0,0 +1,39 @@
...
+import io.syndesis.model.WithId;
+import io.syndesis.model.WithName;
+import io.syndesis.model.WithProperties;
+
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonDeserialize(builder = CustomConnector.Builder.class)
+@SuppressWarnings("immutables")
+public interface CustomConnector extends WithId<CustomConnector>, WithName, WithProperties {
```
rhuss: Just a question for my understanding: The "CustomConnector" is specific for a custom _API_ connector or is suitable for _any_ custom connector, which is supposed to be an uploadable artefact (not scheduled yet for GA, but discussed already in Ignite GA Backlog). This was supposed to be the second extension hook beside tech extensions.

rhuss: What is the difference to `Connector` within the domain ?

rest/connector-generator/src/main/java/io/syndesis/connector/generator/swagger/SwaggerConnectorGenerator.java
```diff
@@ -76,24 +78,49 @@
     }
 
     @Override
-    public Connector generate(final ConnectorTemplate connectorTemplate, final Connector template) {
-        final Map<String, String> configuredProperties = template.getConfiguredProperties();
+    public Connector generate(final ConnectorTemplate connectorTemplate, final CustomConnector customConnector) {
+        final Connector connector = basicConnector(connectorTemplate, customConnector);
```
rhuss: So a `Connector` is created from a `CustomConnector` ? From the naming I first implied that a CustomConnector **is** a Connector (with some kind of specialization). Isn't it more like a `ApiConnectorTemplate` or `ApiConnectorSource` ?

rhuss: Please correct me, but if I understand right, then a 'ConnectorTemplate' has the type information and the  'CustomConnector'  the runtime information ?

rest/connector-generator/src/main/java/io/syndesis/connector/generator/ConnectorGenerator.java
zregvart: Can't hide implementation details like `baseConnectorFrom` and `determineConnectorName` from clients and I was too lazy to create a base class for generators and keep the interface with only public API.

rest/connector-generator/src/main/java/io/syndesis/connector/generator/swagger/SwaggerConnectorGenerator.java
zregvart: Yeah, `ConnectorTemplate` was already taken. Here is the lowdown:
 - `ConnectorTemplate` should define all non user provided aspects of a new connector
 - `CustomConnector` defines only user provided aspects
 - `ConnectorGenerator` mixes those two to produce a `Connector`

Given that the `ConnectorTemplate` and `Connector` are taken I could not find a better name than a `CustomConnector`. Naming is hard.

You might also notice that this goes beyond API connectors, and allows creating any kind of connector based on the template and generator implementation.

rest/model/src/main/java/io/syndesis/model/connection/CustomConnector.java
zregvart: Yeah naming is hard. We had discussed this [before](https://github.com/syndesisio/syndesis-rest/pull/757#discussion_r150194277). I'm happy to rename it to something more appropriate.

The issue is that you don't give the REST API a `Connector` to create, but only parts of the data needed (`CustomConnector`), there are bits that don't really interest the end user, think Maven coordinates, or properties that can be determined from a given specification in the case of API connector.

Perhaps a name like `ConnectorSettings` or similar might be better?

Come to think of it, perhaps that `WithId` is not needed here.

rest/connector-generator/src/main/java/io/syndesis/connector/generator/swagger/SwaggerConnectorGenerator.java
rhuss: What's about 'ConnectorParameters' instead of 'CustomConnector' ? 

I mean, you have a _connector template_ fill in some _connector parameters_ and then get a _connector_ . Sounds like a common 'template pattern' ...

Even if it is not 100% fit (as a connector takes a part from the template, and another part from the customconnector) it has at least a known semantics. Does this catch your case, too ?

rest/model/src/main/java/io/syndesis/model/connection/CustomConnector.java
rhuss: `ConnectorSettings` sounds also legit to me. I really like to avoid `CustomConnector` as it implies to be a sort of `Connector` (which is not).

zregvart:
@rhuss I've removed the `id` property and renamed `CustomConnector` to `ConnectorSettings` fancy taking another look?

rhuss:
@zregvart will have a look now, sorry was busy with meeting until now ;-(

rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/CustomConnectorHandler.java
```diff
@@ -0,0 +1,92 @@
...
+import io.syndesis.connector.generator.ConnectorGenerator;
+import io.syndesis.dao.manager.DataManager;
+import io.syndesis.model.connection.Connector;
+import io.syndesis.model.connection.ConnectorTemplate;
+import io.syndesis.model.connection.ConnectorSettings;
+import io.syndesis.rest.v1.operations.Violation;
+
+import org.springframework.context.ApplicationContext;
+
+public final class CustomConnectorHandler {
```
rhuss: Coudl we rename this to `ConnectorSettingsHandler` ?

rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/CustomConnectorHandler.java
```diff
@@ -0,0 +1,92 @@
...
+        this.connectorTemplateId = connectorTemplateId;
+        this.dataManager = dataManager;
+        this.applicationContext = applicationContext;
+    }
+
+    @POST
+    @Path("/info")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    @ApiOperation("Creates a new Connector based on the ConnectorTemplate identified by the provided `connector-template-id` and the data given in `customConnector`")
```
rhuss: adapt documentation.

Best make a search through the project with 'customConnector' and change everything to `connectorSettings` (could be that I missed something, too)

rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorTemplateHandler.java
```diff
@@ -15,30 +15,87 @@
...
+    @ApiParam(name = "id", example = "swagger-connector-template")
+    @ApiResponses(@ApiResponse(code = 200, response = Connector.class, message = "Newly created Connector"))
+    public Connector create(@PathParam("id") @NotNull final String templateId, final ConnectorSettings connectorSettings) {
+        final ConnectorTemplate connectorTemplate = getDataManager().fetch(ConnectorTemplate.class, templateId);
+
+        if (connectorTemplate == null) {
+            throw new EntityNotFoundException("Connector template: " + templateId);
+        }
+
+        final ConnectorGenerator connectorGenerator = applicationContext.getBean(templateId, ConnectorGenerator.class);
```
rhuss: in https://github.com/syndesisio/syndesis/pull/450/files#diff-89486c5a63a324575071f5699d43aae1R70 you check for `null`, but not here. Not sure what is correct, maybe this duplicate code could be extracted to a single method ? (however not worth the effort if an extra class would need to be introduced only for this)

rest/rest/src/test/java/io/syndesis/rest/v1/handler/connection/CustomConnectorHandlerTest.java
```diff
@@ -0,0 +1,71 @@
...
+        when(applicationContext.getBean("connector-template", ConnectorGenerator.class)).thenReturn(connectorGenerator);
+        when(connectorGenerator.info(same(template), same(connectorSettings))).thenReturn(connector);
+
+        final Connector connectorInfo = handler.info(connectorSettings);
+
+        assertThat(connectorInfo).isSameAs(connector);
+    }
+
+    @Test
+    public void shouldVlidateCustomConnectors() {
```
rhuss: typo

rest/runtime/src/test/java/io/syndesis/runtime/connector/CustomConnectorITCase.java
```diff
@@ -0,0 +1,136 @@
...
+
+@ContextConfiguration
+public class CustomConnectorITCase extends BaseITCase {
+
+    private final ConnectorTemplate template;
+
+    public static class ConnectorTemplateResultList {
+        public List<ConnectorTemplate> items;
+
+        public int totalCount;
```
rhuss: could be later verified, too.

rest/dao/src/main/resources/io/syndesis/dao/deployment.json
```diff
@@ -860,13 +860,13 @@
...
           "required": true,
-          "type": "string",
+          "type": "file",
           "javaType": "java.lang.String",
           "tags": [ "blob" ],
           "deprecated": false,
           "secret": false,
           "componentProperty": true,
-          "description": "Swagger specification of the service"
+          "description": "Upload the swagger fore for your Custom API Client Connector. Custom API's are RESTful APIs and can be hosted anywhere, as long as a well-documented swagger is available and conforms to OpenAPI standards."
```
rhuss: typo

rest/dao/src/main/resources/io/syndesis/dao/deployment.json
```diff
@@ -860,13 +860,13 @@
           "group": "producer",
           "label": "producer",
           "required": true,
-          "type": "string",
+          "type": "file",
```
rhuss: what does this change imply ? Does trigger an upload button in the UI ?

rhuss:
Thanks for the renaming btw, it imo it really looks much better (and even better understandable for a noob like me ;-)

rest/dao/src/main/resources/io/syndesis/dao/deployment.json
zregvart: There is only one property for the `specification`, and it can accept either full specification (as String), or a URL to fetch the specification from. As per [design](https://github.com/syndesisio/syndesis/blob/master/ux/designs/apiconnector/apiconnector.md). This is an attempt to signal the UI to provide file browser for specification upload.

rest/runtime/src/test/java/io/syndesis/runtime/connector/CustomConnectorITCase.java
zregvart: I would not, as it gets data from `deployment.json`, so the number of templates will vary over time and the test would need to be changed. The test checks if the the template it sets up is present, presumably we are not testing DataManager implementation here...

PR #453, opened 2017-11-24
refactor: Introduced image streams + other build optimizations

rhuss:
+ a new profile for only creating images (without the rest)

Current workflows:

* minishift start
* apply openshift templates (should be done via a mvn profile or script, too)
* mvn -Pimage,flash

This will build & deploy automatically:

- rest
- verifier
- ui


zregvart:
I kinda liked the `flash` profile name, we can't reuse that one here?

Seems that it's not so easy to remove the ~~d-m-p~~ directory-maven-plugin:

```
[ERROR] Failed to execute goal com.mycila:license-maven-plugin:3.0:check (default) on project integration-runtime-bom: Execution default of goal com.mycila:license-maven-plugin:3.0:check failed: Cannot read header document ${syndesis.basedir}/header.txt. Cause: Resource ${syndesis.basedir}/header.txt not found in file system, classpath or URL: no protocol: ${syndesis.basedir}/header.txt -> [Help 1]
```

rhuss:
> I kinda liked the flash profile name, we can't reuse that one here?

wdym ? I just moved it up to the parent pom.xml with the same name.

rhuss:
> Seems that it's not so easy to remove the d-m-p directory-maven-plugin:

Strange, it works locally for me, and it still runs in the parent.xml (and shouldn't then the property propagated ?) --> https://github.com/syndesisio/syndesis/pull/453/files#diff-600376dffeb79835ede4a0b285078036R376

zregvart:
> > I kinda liked the flash profile name, we can't reuse that one here?
> 
> wdym ? I just moved it up to the parent pom.xml with the same name.

just looked at the description really, lazy me, there it says `fast`, all good if I bothered to look at the commit diff 😊

rhuss:
sorry, my fault. I updated the issue description.

rhuss:
@zregvart yippie, finally ;-) (at least I made runtime and tests basepom conformant, too ;-).

Could you please give a quick review and approve if ok ?

PR #459, opened 2017-11-26
feat: Update to Camel 2.20.1

rhuss:
and introduched a camel.version in top-level pom to refer to.


connectors/connectors/day-trade/day-trade-get-connector/src/main/java/io/syndesis/connector/daytrade/DayTradeGetComponent.java
```diff
@@ -1,12 +1,12 @@
...
+ * <p>
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- *
- *         http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
zregvart: these `<p>`'s look wrong

rhuss: Sorry, not sure where this comes from. I haven't tuned the license manually (as I really never care about license formatting at first priority and tbh, I would be really ok to switch off the license check and do an automatic validation / adaption / verification once in a while ... I already spent to much time on making this plugin happy (instead of vice-versa as it should be).

For real code checks I have a different opinion ....

Nevertheless, I will investigate.

rhuss:
What's about the generated code of the camle-connector-maven-plugin which gets checked in and does not add any license headers at all ? wonder why the license check plugin doesnt croak here ...

rhuss:
Nearly no class in connectors has a license header btw.

rhuss:
> license check plugin doesnt croak here

found out that its switched off in connectors as any other check from basepom, too.

rhuss:
@zregvart would be awesome if we could merge this early tomorrow morning. For sure, review is impossible for now, but we need to go on (as there is still a lot of things to clean up for the monorepo).

The most substantial things I did here:

* Add license header to all of the connectors
* Switched verifier to use all basepom checks
* Update to Camel 2.20.1 (but not sure about the connector namings, so we need to revisit this later)

The circle test pass though.

rhuss:
Sorry, I reverted the license header update in connectors, that was a stupid idea. I will create a separate PR for the license update.

rhuss:
@zregvart Is it ok to fix the license formatting in a different PR when I add license headers to all files ?

Issue #460, opened 2017-11-27
fix license headers


zregvart:
We seem to be a bit fuzzy about license headers in connectors, perhaps we should revisit license headers enforcement througout the (Java) codebase and fix existing discrepancies.

rhuss:
It's easy todo with Intellij's "Update Copyright" support for the connectors directory and the result looks ok. This has also the benefit that all license headers look the same (which the basepom license check does not enforce).

I agree, that we probably should switch off the license checks in the build, as it causes more friction than benefit IMO. As the tooling is pretty decent to update all headers in one step, I'm happy todo this from time to time (e.g. before a release) with a dedicated PR (and then switch on the licencse check for release builds only).

@syndesisio/backend wdyt ?

zregvart:
I've modified [license pull request check](https://ci.fabric8.io/view/syndesis/job/syndesis-pullreq-license-check/) to run against syndesisio/syndesis.

zregvart:
Done ✔️

PR #461, opened 2017-11-27
fixes license check and formats license headers

zregvart:
See #460


rhuss:
@zregvart there are some test failure because the license has been added to generated shell scripts (but not for the expected text).

Otherwise it looks good to me, thanks !

zregvart:
@rhuss fixed the test resource and the CircleCI gives a ✔️, fancy having another look?

PR #473, opened 2017-11-27
Lets get the circleci build generating images too. And display the yarn cache dir being used during the build.  fix: s2i-image name was not correct fix: copy in the locally built SNAPSHOT artifacts into the docker build

chirino:


s2i/pom.xml
```diff
@@ -80,21 +80,36 @@
...
+            </configuration>
+          </execution>
+          <execution>
+            <id>copy-local-repo</id>
+            <phase>package</phase>
+            <goals>
+              <goal>exec</goal>
+            </goals>
+            <configuration>
+              <executable>bash</executable>
```
rhuss: won't work on windows obviously. Maybe we should add this to the README.

s2i/pom.xml
```diff
@@ -80,21 +80,36 @@
...
+              <arguments>
+                <argument>${basedir}/copy_mvn_repo.sh </argument>
+                <argument>${syndesis.rest.version}</argument>
```
rhuss: Why is there a specific "rest" version ? Should be always 'syndesis.version' (one of the benefits of a monorepo)

s2i/copy_mvn_repo.sh
```diff
@@ -0,0 +1,23 @@
...
+      <id>local</id>
+      <url>file:/deployments/project/.m2/repository</url>
+    </snapshotRepository>
+    <repository>
+      <id>local</id>
+      <url>file:/deployments/project/.m2/repository</url>
+    </repository>
+  </distributionManagement>
+"""
+perl -p -i -e "s|</project>|${XML}</project>|" "$TARGET_DIR/pom.xml"
```
rhuss: Love it ;-) But think sed could be good  enough here, too (so that it works for systems without perl, too).

rhuss: But I'm fine with this of course (actually I would rewrite it into a complete perl script or maybe even groovy script to make it cross plattform)

s2i/pom.xml
```diff
@@ -80,21 +80,36 @@
...
+                <argument>${basedir}/src/main/docker/project</argument>
```
rhuss: Copying into a directory below `src/` is not so nice

s2i/copy_mvn_repo.sh
rhuss: However, the script has a serious flaw: You will add endlessly `<distributionManagement>` sections to `src/main/docker/pom.xml` regardless of whether this section is already included or not. Also modifying committed source files has issues (remember the connectors ?)

build.sh
```diff
@@ -74,14 +74,19 @@ function init_options() {
...
-  # If  we are running in cicleci lets configure thigs to avoid running out of memory:
+  # If  we are running in cicleci 
   if [ "$CIRCLECI" == "true" ] ; then
-    MAVEN_PARAMS="$MAVEN_PARAMS -Dbasepom.test.fork-count=2"
+    # Use batch mode so we get friendiler log files.
+    MAVEN_PARAMS="$MAVEN_PARAMS --batch-mode"
+    # lets configure thigs to avoid running out of memory:
+    MAVEN_PARAMS="$MAVEN_PARAMS -Dbasepom.test.fork-count=2 --batch-mode -P image"
+    # Enable the image profile so that they get built
+    MAVEN_PARAMS="$MAVEN_PARAMS -P image"
```
rhuss: Tomorrow I will update build.sh to be more flexible with respect to combining profiles from commandline arguments given to the script. I know that profiles dont work optimally in Maven (e.g. the default profile is only activated if no other profile is activated), but I hope I can alleviate the pain with nice script options.

s2i/copy_mvn_repo.sh
chirino: the pom.xml is generated on every mvn build, so no endless dups of '<distributionManagement>'.   Tried sed.  It fails to handle '\n''s well.  Our build.sh is based on bash already.  Seem we are already tied down to bash based systems.

s2i/pom.xml
chirino: This would be just one item in a long list no?

s2i/pom.xml
chirino: just reusing what was defined previously.

s2i/pom.xml
chirino: Sure but that's where it's been currently generating to.  Moving to target would be better tho.

PR #475, opened 2017-11-27
feat: Serve connector image icons from REST API

### LINKED ISSUE #259: We should serve icons from syndesis-rest
|</img> @zregvart | [2017-10-30](https://github.com/syndesisio/syndesis-ui/issues/1154) | enhancement, Priority - Low  |
|-|-|-|

With API Connectors we'll need to serve icons from syndesis-rest as these will be defined at runtime. First step would be to define a endpoint to fetch the icons from the backend.

jimmidyson:
Fixes #259


rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
```diff
@@ -141,4 +153,44 @@ public Kind resourceKind() {
...
+        Connector connector = get(connectorId);
+        String connectorIcon = connector.getIcon();
+
+        // If there is no specified icon, return 404
+        if (connectorIcon == null) {
+            return Response.status(Response.Status.NOT_FOUND).build();
+        }
+
+        // If the specified icon is a data URL, or a non-URL like value (e.g. font awesome class name), return 404
+        if (connectorIcon.startsWith("data:") || !connectorIcon.contains("/")) {
```
rhuss: Couldn't we serve the data URL by returning the (converted) data back to the client ?

rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
```diff
@@ -141,4 +153,44 @@ public Kind resourceKind() {
...
+            return Response.status(Response.Status.NOT_FOUND).build();
+        }
+
+        OkHttpClient httpClient = new OkHttpClient();
+        try {
+            final okhttp3.Response externalResponse = httpClient.newCall(new Request.Builder().get().url(connectorIcon).build()).execute();
```
rhuss: Who is supposed to server the icon ? Do we have an Icon service for this only or is this coming from somewhere externally (maybe the verifier service ?)

rhuss:
Sorry havent read the comments in the issue yet. Be back in an hour an revisit my questions then.

rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
jimmidyson: The URL could be an external site, it could be the verifier, it could be a separate image icon service. I think leaving it like this gives flexibility.

A separate caching image service would be great though, so perhaps this is something we can extend to later. I need to work on the connector icon generator anyway so that could be refactored in that perhaps.

rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
jimmidyson: We could, but why?

rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
rhuss: Ok, but can we harden the external access at least a bit ? E.g. like setting a (small) timeout for the request ? And returning a default broken icon if the access fails ?

rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
rhuss: So that we can store the icon directly in the DB and do not have access the external service ? (e.g. allow to upload an icon in the UI instead of refering it to an exernal service)

rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
```diff
@@ -141,4 +153,44 @@ public Kind resourceKind() {
...
+            };
+
+            Response.ResponseBuilder actualResponse = Response.ok(streamingOutput, contentType);
+            if (!Strings.isEmpty(contentLength)) {
+                actualResponse.header(CONTENT_LENGTH, contentLength);
+            }
+
+            return actualResponse.build();
+        } catch (IOException e) {
+            throw new SyndesisServerException(e);
```
rhuss: Do we really want it to be a full-blown error when the icon can't be served from an external server which can go down anytime? I think that's too much of a dependency here. Instead, I would return a default icon.

zregvart:
Just as a side note, when we define a new API Connector the user can provide a URL to the Swagger specification, in that case we will also make a request to a 3rd party service.

rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
jimmidyson: Data URLs are stored in the DB anyway in the `Connector.icon` field. We don't really need to convert them back to an image I don't think, the image source in the browser will just look like:

    embedded folder icon

rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
jimmidyson: Good idea. A default placeholder icon. @syndesisio/uxd Could you please provide/suggest a default icon we can use for connectors?

jimmidyson:
> Just as a side note, when we define a new API Connector the user can provide a URL to the Swagger specification, in that case we will also make a request to a 3rd party service.

Right, we should be caching the icon at this point, say the favicon for example. How should we handle cache headers on the retrieved file? Should we create a full on respectful cache for these images, or should we just ignore cache directives and cache in perpetuity?

zregvart:
> Right, we should be caching the icon at this point, say the favicon for example. How should we handle cache headers on the retrieved file? Should we create a full on respectful cache for these images, or should we just ignore cache directives and cache in perpetuity?

Not sure, we'll be allowing editing of custom connector's icons ([Custom connector detail page UX](https://github.com/syndesisio/syndesis/blob/master/ux/designs/apiconnector/apiconnector.md#api-connector-details)), perhaps we could generate a `Etag` based on the hash of the icon?

PR #478, opened 2017-11-28
fix(ui): Fix angular-cli definition in pom.xml

### LINKED ISSUE #477: Angular-cli warning when building UI with Maven plugin
We need to change angular-cli coordinate to new format. 
 
```
--- frontend-maven-plugin:1.6:npm (angular-cli install) @ syndesis-ui ---
Running 'npm install --no-optional -g angular-cli' in /Users/dsimansk/Git/syndesis/ui
npm WARN deprecated angular-cli@1.0.0-beta.28.3: angular-cli has been renamed to @angular/cli. Please update your dependencies.
```

dsimansk:
Fixes #477

### DIFF
--- a/ui/pom.xml
+++ b/ui/pom.xml
@@ -86,7 +86,7 @@
               <goal>npm</goal>
             </goals>
             <configuration>
-              <arguments>install --no-optional -g angular-cli</arguments>
+              <arguments>install --no-optional -g @angular/cli</arguments>
             </configuration>
           </execution>
           <execution>


ui/pom.xml
```diff
@@ -86,7 +86,7 @@
               <goal>npm</goal>
             </goals>
             <configuration>
-              <arguments>install --no-optional -g angular-cli</arguments>
+              <arguments>install --no-optional -g @angular/cli</arguments>
```
rhuss: Is the npm that we install (from Node 6) already new enough to understand scoped packages ? (afaik available since Node 2.7.0)

Actually I would like to update Node to version 8, too, but it seems that this automatic installation for node from within the build does not work on Circle (leads to an endless loop with permission errors0.

dsimansk: I has worked for me our karma-xvfb image, that has Node 6 installed. But of course it can wait until to be done with upgrade to Node 8.

rhuss: np, was just asking ;-)

PR #480, opened 2017-11-28
chore: add .mention-bot configuration file

### LINKED ISSUE #418: Re-enable mention-bot on syndesis repository
We should re-enable purebot for code review assignement on the syndesisio/syndesis repository.

zregvart:
Fixes #418

### DIFF
--- a/.mention-bot
+++ b/.mention-bot
@@ -1,0 +1,8 @@
+{
+  "message": "@pullRequester, thanks! @reviewers, please review this.",
+  "assignToReviewer": true,
+  "createReviewRequest": true,
+  "actions": ["opened", "reopened"],
+  "skipAlreadyAssignedPR": true,
+  "skipAlreadyMentionedPR": true
+}


jimmidyson:
Wonder if we should now use the [code](https://github.com/blog/2392-introducing-code-owners) [owners](https://help.github.com/articles/about-codeowners/) facility of github?

zregvart:
> Wonder if we should now use the code owners facility of github?

could be I see that it supports groups also...

zregvart:
> > Wonder if we should now use the code owners facility of github?
> 
> could be I see that it supports groups also...

Actually, I'm not sure, I would like that we don't split into groups, i.e. that everyone works on all parts of the stack (eventually). With `git blame` way of assigning we don't tie review to a particular person or group.

2c

rhuss:
My suggestion: Try this out, check whether it works, and if not, try GitHub code owners.

PR #519, opened 2017-11-29
feat: move s2i module in under app directory.

chirino:


app/syndesis-s2i/pom.xml
```diff
@@ -22,16 +22,16 @@
 
   <parent>
     <groupId>io.syndesis</groupId>
-    <artifactId>syndesis-parent</artifactId>
+    <artifactId>app</artifactId>
```
rhuss: Shouldn't we keep syndesis-parent ? or at least syndesis-app ?

chirino: syndesis-parent is the parent of app.  So I think we are fine.

chirino:
retest this please

rhuss:
@chirino instead of fixing that let me do this move on the weekend together with the other stuff.

rhuss:
Closing this now as I'm doing the move now anyway ;-)

PR #520, opened 2017-11-29
feat: remove use of connector catalog module.

### LINKED ISSUE #514: Avoid the Camel Catalog from being used by the rest module.
Using the camel catalog causes downloading online artifacts via groovy grape.  This is slow and also error prone since sometimes you need to clear that grape cache.

The catalog is only used by the project generator to encode the action properties into and endpoint URI.  We should defer that so that it's done at runtime when the route is created.  This change means that the syndesis.yml format will change a bit so that it will now carry a json property map of all the action config.

chirino:
fixes #514.


lburgazzoli:
@chirino do you mind if we keep this on hold ? I'm working on getting rid of connector artifacts & so on which as side effect removed the catalog and simplifies how the current connectors are handled so we can also get rid of some crazy naming stuffs.

chirino:
@lburgazzoli sounds fine /w me.  Could you integrate this change into your PR?

lburgazzoli:
Yes it will be included

rhuss:
Ok for me to close the PR then ? Please feel free to reopen if there is still something to apply.

PR #523, opened 2017-11-30
feat: Import integration in draft status.

### LINKED ISSUE #109: Import integration in draft status
|</img> @chirino | [2017-11-06](https://github.com/syndesisio/syndesis-rest/issues/779) |
|-|-|

chirino:
Fixes #109.

### DIFF
--- a/rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
+++ b/rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
@@ -197,6 +197,7 @@ public class IntegrationHandler extends BaseHandler
             .createdDate(rightNow)
             .currentStatus(determineCurrentStatus(encryptedIntegration))
             .userId(sec.getUserPrincipal().getName())
+            .stepsDone(new ArrayList<>())
             .build();
 
         return Creator.super.create(sec, updatedIntegration);
@@ -217,6 +218,7 @@ public class IntegrationHandler extends BaseHandler
             .lastUpdated(new Date())
             .currentStatus(currentStatus)
             .addRevision(currentRevision)
+            .stepsDone(new ArrayList<>())
             .build();
 
         Updater.super.update(id, updatedIntegration);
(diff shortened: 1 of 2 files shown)


rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -197,6 +197,7 @@ public Integration create(@Context SecurityContext sec, @ConvertGroup(from = Def
             .createdDate(rightNow)
             .currentStatus(determineCurrentStatus(encryptedIntegration))
             .userId(sec.getUserPrincipal().getName())
+            .stepsDone(new ArrayList<>())
```
rhuss: Shouldn't this be the default in the constructor of an integration ?

rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationSupportHandler.java
```diff
@@ -115,11 +120,18 @@ public int importModels(ModelExport export) throws IOException {
...
+                    integration = new Integration.Builder()
+                        .createFrom(integration)
+                        .desiredStatus(Integration.Status.Draft)
+                        .build();
+
                     // Do we need to create it?
                     if (dataManager.fetch(Integration.class, integration.getId().get()) == null) {
-                        dataManager.create(integration);
+                        LOG.info("Creating integration: "+integration.getName());
+                        integrationHandler.create(sec, integration);
```
rhuss: Architectural wise I think it's not so nice to call a Rest endpoint from another rest endpoint as this hinders the API to evolve freely and introduces an unnecessary coupling. Instead, I would extract the common used code into a lower-layer class and use this from both, here and the endpoint.

rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
chirino: Not the case when we are creating an integration from an import.

chirino: Not the case when we are creating an integration from an import.

PR #495, opened 2017-11-29
refactor(custom-connector): POC cleanup

zregvart:
This cleans up implementation and tests of the Swager custom connector
generator.

PR #534, opened 2017-11-30
Simplify /info endpoint for custom connectors

zregvart:
**Based on #495 and #500 please review that first**

This removes `/api/v1/custom/connectors/{id}/validation` endpoint and changes the response from `/api/v1/custom/connectors/{id}/info` to be a bit simpler and in the same time to contain errors and warnings.

//cc @deeleman

### DIFF
--- a/rest/model/src/main/java/io/syndesis/model/Violation.java
+++ b/rest/model/src/main/java/io/syndesis/model/Violation.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.syndesis.rest.v1.operations;
+package io.syndesis.model;
 
 import java.lang.annotation.Annotation;
 
--- a/project/proposals/119-custom-connectors.md
+++ b/project/proposals/119-custom-connectors.md
@@ -150,23 +150,99 @@ Accept: application/json
 }
 
 HTTP/1.1 200 OK
-Content-Length: 2
 
 {
-  "name": "Swagger Petstore",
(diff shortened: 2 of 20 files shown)


rhuss:
@zregvart could you please rebase ? I will review it the right after, should be the last PR to review for the connectors so far, right ?

zregvart:
@rhuss should be ok to review, thanks 💯

zregvart:
> should be the last PR to review for the connectors so far, right ?

yes, indeed

rest/model/src/main/java/io/syndesis/model/Violation.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.syndesis.rest.v1.operations;
+package io.syndesis.model;
```
rhuss: what's the difference between "message" and "error" in Violation ?

rhuss: in fromConstraintViolation() is this all null safe ? (sorry can not comment directly on the code).

rhuss:
@zregvart there are compile errors ....

zregvart:
@rhuss should be OK now, and this is why you don't have long running branches

rest/model/src/main/java/io/syndesis/model/Violation.java
zregvart: The `message` holds the text of an error (`"XYZ is required"`), and error contains the class of error (`"NotEmpty"`).

Should be null safe, it's the way bean validation works, constraints are applied via annotations, and each violation has a resolved message (or a placeholder message), if that's what you mean.

PR #535, opened 2017-11-30
Promote tech extensions to v1 API

### LINKED ISSUE #489: HTTP 404 being returned for /api/v1beta1/extensions
After cleaning up my minishift instance and reinstalling via a local build I'm getting a 404 response for `/api/v1beta1/extensions` as well as `/api/v1beta1/swagger.json`.  The `Extension` definition shows up in `/api/v1/swagger.json` as well as some of the new API connector APIs so I'm fairly confident I'm running against my local build.

No error in the rest pod log, am wondering if the oauth proxy config changed to cause this perhaps?

nicolaferraro:
This fixes #489.
Filestore has been enabled by default. Api `v1beta1` is removed and tech extensions promoted to `v1`.


gashcrumb:
Can you update [this path](https://github.com/syndesisio/syndesis/blob/master/ui/src/app/store/extension/extension.service.ts#L12) too?  No reason it needs to be done in a separate PR, just delete `../v1beta1/` from that path.

nicolaferraro:
Sure!

PR #541, opened 2017-11-30
[fix] Fixing inputShape issue due to difference in case on Postgresql

KurtStam:


connectors/connectors/sql/sql-common/src/main/java/io/syndesis/connector/sql/DatabaseMetaDataHelper.java
```diff
@@ -72,7 +72,7 @@ public static ResultSet fetchProcedures(final DatabaseMetaData meta, final Strin
         Set<String> tablesInSchema = new HashSet<>();
         ResultSet rs = meta.getTables(catalog, schemaPattern, tableNamePattern, new String[] { "TABLE" });
         while (rs.next()) {
-            tablesInSchema.add(rs.getString(3));
+            tablesInSchema.add(rs.getString(3).toUpperCase());
```
zregvart: Should we use Locale here? i.e.

    .toUpperCase(StandardCharsets.UTF_8)

connectors/connectors/sql/sql-common/src/main/java/io/syndesis/connector/sql/DatabaseMetaDataHelper.java
```diff
@@ -90,6 +90,10 @@ public static ResultSet fetchProcedures(final DatabaseMetaData meta, final Strin
             SqlParam param = params.get(i);
             String columnName = param.getColumn();
             ResultSet column = meta.getColumns(catalog, schema, tableName, columnName);
+            if (column.getFetchSize() == 0) {
+                //Postgresql does lowercase instead, so let's try that if we don't have a match
+                column = meta.getColumns(catalog, schema, tableName.toLowerCase(), columnName.toLowerCase());
```
zregvart: Locale?

PR #553, opened 2017-12-01
refactor: Naming and add -Pimages -Dfabric8.mode=openshift for "system-tests"

rhuss:


rhuss:
@iocanel fancy to have a quick look ? It's mostly a no brainer. 

I assume that for --mode system-test you *always* need to create images in openshift mode, right ?
So I added -Pimages -Dfabric8.mode=kubernetes as default params so that you dont have to provide it from the outside.

iocanel:
@rhuss: why do we need to also update the Jenkinsfile?

rhuss:
for the options ? (you don't need to add --image-mode s2i anymore when using it with --system-test). not sure though ...

PR #556, opened 2017-12-01
feat: show uses column on extension list

### LINKED ISSUE #474: Add 'uses' column to tech extension list
Currently this column can't be shown as it requires fetching the usage count for each extension as a separate call, ideally we'd inline this attribute in the extension object.

gashcrumb:
fixes #474 

@lburgazzoli FYI...

### DIFF
--- a/ui/src/app/customizations/tech-extensions/tech-extensions-list.component.html
+++ b/ui/src/app/customizations/tech-extensions/tech-extensions-list.component.html
@@ -38,7 +38,7 @@
             </div>
             <div class="list-pf-additional-content">
               <span>
-                Used by <strong>{{ item.uses || 0 }}</strong> {{ item.uses === 1 ? 'integration' : 'integrations' }}
+                Used by <span [innerHtml]="item.uses | i18nPlural: itemUseMapping"></span>
               </span>
             </div>
           </div>
--- a/ui/src/app/customizations/tech-extensions/tech-extensions-list.component.html
+++ b/ui/src/app/customizations/tech-extensions/tech-extensions-list.component.html
@@ -37,9 +37,9 @@
               <div class="list-pf-description" text-overflow-pf>{{ item.description }}</div>
             </div>
             <div class="list-pf-additional-content">
-              <div *ngIf="item.integrations">
-                <strong>Used by {{ item.integrations }}</strong>
-              </div>
+              <span>
+                Used by <strong>{{ item.uses || 0 }}</strong> {{ item.uses === 1 ? 'integration' : 'integrations' }}
+              </span>
             </div>
           </div>
         </ng-template>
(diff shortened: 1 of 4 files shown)


ui/src/app/customizations/tech-extensions/tech-extensions-list.component.html
```diff
@@ -37,9 +37,9 @@ <h1>Technical Extensions</h1>
               <div class="list-pf-description" text-overflow-pf>{{ item.description }}</div>
             </div>
             <div class="list-pf-additional-content">
-              <div *ngIf="item.integrations">
-                <strong>Used by {{ item.integrations }}</strong>
-              </div>
+              <span>
+                Used by <strong>{{ item.uses || 0 }}</strong> {{ item.uses === 1 ? 'integration' : 'integrations' }}
```
deeleman: This kind of bindings are better managed through Angular's [I18nPluralPipe](https://angular.io/api/common/I18nPluralPipe) pipe, so can automate values such as `no integrations`, `one integration`, `154,456 integrations`...

The caveat is that the controller gets more verbose, but at least the template gets cleaner and less operations are made when interpolating data. I'd add this field in your component controller

```
itemUseMapping: {[valueComparator: string]: string
} = { '=0': 'No integrations', '=1': 'one integration.', 'other': '# integrations' };
```

And then in your template:

``` 
Used by <strong>{{ (item.uses || 0) | i18nPlural: itemUseMapping }}</strong>
``` 

and if you don't feel like tweaking your PR, just go ahead! LGTM LOL

deeleman: There is also I'd like to point out: We should never "shortcircuit" bindings with fallback data in the template (I mean that `item.uses || 0` thingie). Otherwise we'll have to do this everytime we need to inspect and print the value of `item.uses` elsewhere in the code.

I believe the right place for doing this is the service handler that binds the Http response into the `item` model, whatever it is.

gashcrumb: Awesome, tx!  I'll make those changes, should have an update in a few.

ui/src/app/customizations/tech-extensions/tech-extensions-list.component.ts
```diff
@@ -26,6 +26,11 @@ export class TechExtensionsListComponent implements OnInit {
   listConfig: ListConfig;
   @ViewChild(TechExtensionDeleteModalComponent)
   deleteModal: TechExtensionDeleteModalComponent;
+  itemUseMapping: { [valueComparator: string]: string } = {
+    '=0': '<strong>0</strong> integrations',
+    '=1': '<strong>1</strong> integrations',
```
deeleman: Probs you want the output to be `<strong>1</strong> integration` and not `integrations` (please note the 's')

gashcrumb:
Have an update to the detail page I want to add to this also, marking as WIP temporarily

gashcrumb:
K, this is ready to go...

gashcrumb:
@deeleman this probably needs to be approved again since I added one more commit...

PR #568, opened 2017-12-03
refactor: Moved Maven modules to "app/"

rhuss:
* Introduced a "syndesis-policy" project in tools to hold the checkstyle
  and license config
* Consistently changed the display names of modules.


zregvart:
@rhuss did you use `git mv`, seems that we lost history for some files, i.e. [build.sh](https://github.com/rhuss/syndesis/commits/2583c7707534bc8616c975e286ba5142e7d2470f/app/build.sh)

zregvart:
> @rhuss did you use git mv, seems that we lost history for some files, i.e. build.sh

sorry my bad, the history is there if I do a `git log --follow`

rhuss:
@zregvart correct, git correctly recognized the file as to be moved (at least before I did commit them).  It's always fascinating how git manage to keep track of all of them.

If you don't mind I would merge now (when the tests are green and for sure before tomorrow).

PR #583, opened 2017-12-04
Add Copy Object Action for AWS-S3 connector to syndesis-rest

oscerd:
Actually we have only the action to copy an object to AWS S3, I'll add more in the next days.


oscerd:
updated

zregvart:
Merged this manually, the bot's have a hard time working today.

PR #584, opened 2017-12-04
feat: Allow to switch templates when installing with -i docker / openshift

rhuss:
By default the template with the openshift imagestreams (syndesis-restricted.yml) is deployed. With '-i docker' this can be switched to syndesis-dev-restricted.yml


app/build.sh
```diff
@@ -574,18 +575,27 @@ run_minishift() {
     if [ $(hasflag --full-reset) ] || [ $(hasflag --reset) ]; then
         # Only warning if minishift is not installed
         minishift delete
+        sleep 2
```
zregvart: Would it make sense to do a `minishift delete --clear-cache --force` would then this sleep be neccesary?

rhuss: Maybe not. Let me try that ...(oops forget to submit the comment ;-)

Issue #543, opened 2017-12-01
Remove examples from connectors


zregvart:
They just make the build longer (objectively not by much), and we don't use them at runtime. Not sure if there is any other use for them.

KurtStam:
It helped me understanding how the connectors work. My vote is to keep them.

oscerd:
Maybe it would be a good idea to maintain the examples somewhere else. I believe they are useful to understand how the connectors work.

zregvart:
@oscerd yeah I've created #828 and I'll move the removed examples there (at some point), also they need updating: Salesforce connector that creates/updates specifically Salesforce Contact object was removed.

PR #586, opened 2017-12-04
fix: Remove test dependencies of compile scope deps.

rhuss:
Wonder btw whether we shouldnt disable exampls by default ?


zregvart:
To me it seems that examples serve only one purpose, to test the connector implementation outside syndesis. I think we should remove all examples and refrain from creating any new ones (see #543).

PR #587, opened 2017-12-04
feature: enable debugging in -dev templates

zregvart:
This sets the environment variable `JAVA_DEBUG` to `true` for
`syndesis-rest` and `syndesis-verifier` pods in `syndesis-dev-*.yml`
templates.

With that environment variable set the Java process is launched with
remote debugging enabled and debugger interface on port `5005`.

Next one can port forward that port via

     oc port-forward syndesis-rest-... 5005:5005

and use the debugger to connect to `localhost:5005`.

### DIFF
--- a/app/build.sh
+++ b/app/build.sh
@@ -78,6 +78,11 @@ With "--minishift" Minishift can be initialized and installed with Syndesis
     --watch                   Watch startup of pods
 -i  --image-mode <mode>       Which templates to install: "docker" for plain images, "openshift" for image streams
                               (default: "openshift")
+
+With "--dev" common development tasks are simplified
+
+    --debug <name>           Setup a port forwarding to <name> pod (example: rest)
+
 Examples:
 
 * Build only backend modules, fast               build.sh --backend --flash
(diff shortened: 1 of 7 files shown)


app/build.sh
```diff
@@ -78,6 +78,11 @@ With "--minishift" Minishift can be initialized and installed with Syndesis
     --watch                   Watch startup of pods
 -i  --image-mode <mode>       Which templates to install: "docker" for plain images, "openshift" for image streams
                               (default: "openshift")
+
+With "--dev" common development tasks are simplified
+
+    --debug <name>           Setup a port forwarding to <name> pod (example: rest)
+
```
rhuss: one space to less indent ;-) (yeah, they call me the "German" for some reason :)

zregvart: Now I have to create another PR for the missing space 😱

zregvart: Now I have to create another PR for the missing space 😱

rhuss: no, its fine. you can even fix this on upstream/master via the webui.

2s2f

PR #593, opened 2017-12-04
feat: disable cache when debugging

zregvart:
Since there is little or no way to influence Infinispan cache from
outside of the running backend, when developing direct changes to
database will not be refleted until the `syndesis-rest` pod is
restarted.

With this if `JAVA_DEBUG` environment variable is set to `true` maximum
size of Infinispan will be set to `0` -- effectively disabling the
cache. Now changes direct to the database will be reflected in the
responses of the REST API.

### DIFF
--- a/app/rest/runtime/src/main/java/io/syndesis/runtime/InfinispanCacheConfiguration.java
+++ b/app/rest/runtime/src/main/java/io/syndesis/runtime/InfinispanCacheConfiguration.java
@@ -22,6 +22,7 @@ import org.infinispan.manager.DefaultCacheManager;
 import org.infinispan.manager.EmbeddedCacheManager;
 import org.infinispan.spring.provider.SpringEmbeddedCacheManager;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.cache.CacheManager;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -36,15 +37,15 @@ public class InfinispanCacheConfiguration {
     private int maxEntries;
 
     @Bean
+    @ConditionalOnProperty(name = "JAVA_DEBUG", havingValue = "false", matchIfMissing = true)
     public EmbeddedCacheManager embeddedCacheManager() {
-        if (embeddedCacheManager == null) {
-            embeddedCacheManager = new DefaultCacheManager(
-                    new GlobalConfigurationBuilder().nonClusteredDefault().globalJmxStatistics().enable()
-                            .defaultCacheName("syndesis-cache").build(),
-                    new ConfigurationBuilder().simpleCache(true).memory().evictionType(EvictionType.COUNT)
-                            .size(maxEntries).jmxStatistics().enable().build());
-        }
-        return embeddedCacheManager;
+        return cacheManager(maxEntries);
+    }
+
+    @Bean
+    @ConditionalOnProperty(name = "JAVA_DEBUG", havingValue = "true")
+    public EmbeddedCacheManager embeddedNonCachingManager() {
+        return cacheManager(0);
     }
 
     @Bean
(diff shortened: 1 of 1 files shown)


chirino:
We are caching writes?

zregvart:
> We are caching writes?

Not sure, this helps if you don't use the API to change the database, i.e. directly using SQL. I use this sometimes as it is easier than redeploying and resetting the database.

app/rest/runtime/src/main/java/io/syndesis/runtime/InfinispanCacheConfiguration.java
```diff
@@ -36,15 +37,15 @@
     private int maxEntries;
 
     @Bean
+    @ConditionalOnProperty(name = "JAVA_DEBUG", havingValue = "false", matchIfMissing = true)
```
rhuss: Couldn't find it in the docs, but does this fire also on environment variables or only on properties ?

rhuss:
btw, I'm fine for adding this for debug only, but using no cache at all when applying the dev templates feels dangerous to me (as you always develop then against no cache)

@zregvart Actually I would love to add this to the build script for `./build.sh --dev --debug`:

* Remove JAVA_DEBUG from the dev templates
* Patch rest DC (or maybe only the RC) to add JAVA_DEBUG on the already deployed templates
* Redeploy Rest
* Add a `trap` to remove this option again and do a rollback to the old version
* Start port forwarding

That way we can have reall a _debug session_ which ideally does not leave stuff behind.

app/rest/runtime/src/main/java/io/syndesis/runtime/InfinispanCacheConfiguration.java
zregvart: both, I think it goes to `Environment` that goes to `PropertySource`s, but that's from memory

zregvart:
@rhuss are we using `-dev` templates for any other reason than local development? If so then 👍, otherwise I'd like my development setup to be as ready as it can be, perhaps a `-debug` template?

rhuss:
> @rhuss are we using -dev templates for any other reason than local development?

Yes, but I'm a fan of development against a system which is as close as possible to the production system. Enabling a debug port doesn't change the functionality, disabling a cache does. No problem to _temporarily_ disables the cache for debugging, though.

> otherwise I'd like my development setup to be as ready as it can be, perhaps a -debug template?

Sorry for being so negative this morning, but I'm also against an explosion of templates. Actually, I really try to _reduce_ the number of templates (and for the reason above, I would like to use the same templates for production and development, but for this I have to fix the imagestream issue). 

Also I think no one uses the non-restricted version of the templates so I would like to get rid of them, too. And we don't  need the -ci templates (which disable health checks) anymore, I think. But I will open an other issue for this.

-----

Temporarily enabling the JAVA_DEBUG option is what the fabric8-maven-plugin does. So actually you could also use `mvn fabric8:debug` directly from the Maven project you want to debug and it does exactly the steps I described above. So maybe this is an option, too ?

zregvart:
> Yes, but I'm a fan of development against a system which is as close as possible to the production system. Enabling a debug port doesn't change the functionality, disabling a cache does. No problem to temporarily disables the cache for debugging, though.

You're right, we should develop as close to production system, and we should also try to eliminate anything that slows down the turnaround in development. I'm not that eager for this change to go in so we can close this. I would like to see a way of reducing the number of steps when changing the `deployment.json`, right now they are:
 - rsh into syndesis-db pod, and do a `psql -U syndesis` followed by `delete from jsondb;`
 - build syndesis-rest and deploy to minishift
 - delete `syndesis-rest` pod
 - reload the UI

Perhaps we can have a `flush-cache` and a `reset-db-from-json` endpoints in [TestSupportHandler](https://github.com/syndesisio/syndesis/blob/c90bb60e37c6ea0d5041a0ddab8ef09cbe2e91d1/app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/tests/TestSupportHandler.java), or can we extract deployment.json to a well known folder and use `oc rsync` to keep them synchronized?

> Temporarily enabling the JAVA_DEBUG option is what the fabric8-maven-plugin does. So actually you could also use mvn fabric8:debug directly from the Maven project you want to debug and it does exactly the steps I described above. So maybe this is an option, too ?

So perhaps a `--dev --sync` and `--dev --debug` options in `build.sh` (that we'll eventually rename)?

rhuss:
@zregvart Completely agree that your use case should be much easier. I'm perfectly fine with adding such an enpoint (but I thought we already had such an endpoint ?)

Also I wonder whether updating the schema version and then redeploying the pod doesn't work for you, too ? (or is this too involved).

PS: Openend #603 for script naming.

zregvart:
> @zregvart Completely agree that your use case should be much easier. I'm perfectly fine with adding such an enpoint (but I thought we already had such an endpoint ?)

We do, that endpoint only reloads from `deployment.json` in classpath. I feel it's too much turnaround for a simple change to `deployment.json` (like changing a property from required true to false)

> Also I wonder whether updating the schema version and then redeploying the pod doesn't work for you, too ? (or is this too involved).

Yeah with #34 the version is hardcoded
https://github.com/syndesisio/syndesis/blob/77abc8abc60713cc6fd9bde31439d05badef8954/app/rest/model/src/main/java/io/syndesis/model/Schema.java#L23

And I think it would be quicker if we could make changes to `deployment.json` without redeploying `syndesis-rest`.

PR #595, opened 2017-12-05
[bugfix] Quick fix for intermittent test failure while working on multi-record select

KurtStam:


KurtStam:
@rhuss yes, but I'm working right this area. Locally this test had been changed significantly, but I'm not ready to check that in yet.

rhuss:
@KurtStam ok, fine with that.

Issue #600, opened 2017-12-05
Integrations - unify "Add a Step"/"Choose a Step" labels


mastepan:
I would use **Choose a Step** as there already is **Choose a Connection**
This can also cause naming issues in our testsuite.
![choosestep](https://user-images.githubusercontent.com/8707251/33597830-df4d8666-d9a0-11e7-9c4e-35a4afdcc784.png)
![chooseconnection](https://user-images.githubusercontent.com/8707251/33597871-02fc16f4-d9a1-11e7-8ed0-226b3a9cb9c2.png)

dongniwang:
PR https://github.com/syndesisio/syndesis/pull/604 is merged. Closing this issue.

PR #604, opened 2017-12-05
fix: #600 Integrations - unify "Add a Step"/"Choose a Step" labels

mastepan:
[#600 Integrations - unify "Add a Step"/"Choose a Step" labels](https://github.com/syndesisio/syndesis/issues/600)

### DIFF
--- a/app/ui/src/app/integrations/edit-page/step-select/step-select.component.html
+++ b/app/ui/src/app/integrations/edit-page/step-select/step-select.component.html
@@ -20,7 +20,7 @@
                      [routerLink]="['/integrations', currentFlow.integration.id]">{{ integrationName }}</a>
                   <ng-container *ngIf="!integrationName">New Integration</ng-container>
                 </li>
-                <li class="active">Add a Step</li>
+                <li class="active">Choose a Step</li>
               </ol>
             </div>
             <div class="toolbar-pf-action-right">
@@ -32,7 +32,7 @@
     </div>
     <div class="wizard-pf-body clearfix">
       <div class="title">
-        <h1>Add a Step</h1>
+        <h1>Choose a Step</h1>
       </div>
       <p>A step specifies an operation on the data in the integration.  Click one to add it.</p>
       <syndesis-loading [loading]="loading$ | async">


rhuss:
@syndesisio/uxd Does this change looks good to you ?

TovaCohen:
I'm just seeing this change now and if there is a good internal reason for it, I won't get in the way. 
But I think there is a subtle difference between Choose and Add, that was helpful.
The user must choose the start connection and the finish connection, but adding a step is optional. With the start and finish connections in place, a prompt to Choose a Step slightly implies that it is required. Using a different verb, Add, might make it more apparent that having a step in an integration is not required. It's subtle and maybe the benefit of always using "Choose" is greater.

dongniwang:
> The user must choose the start connection and the finish connection, but adding a step is optional. With the start and finish connections in place, a prompt to Choose a Step slightly implies that it is required. 

I agreed with what Tova suggested here. Keeping the "add" would help users understand that the steps are optional. 

cc: @amysueg @sjcox-rh

amysueg:
+1 to @dongniwang @TovaCohen

rhuss:
I'm not complete sure whether the 'optional' argument applies to this screenshot as the user *already* has chosen to add a step (otherwise he wouldn't be on this screen). Now she has to *chose* the concrete step. 

I agree with @mastepan that its not much different to a connection at this point (when a user alread has selected to add a step)

It's perfectly fine when here its about "adding" a step:

image

But when the user clicked on "add step" the next screen should be about choosing which step:

image

So for consistencies sake, I'm really for changing "Add" to "Choose" on this screen.

> I agreed with what Tova suggested here. Keeping the "add" would help users understand that the steps are optional.

At this point, the step is not optional anymore. She can cancel this action but if she proceeds she has to choose a step.

TovaCohen:
It seems inconsistent to me for the user to click "Add a step" and then get a page that says "Choose a Step", but it's not a big deal. I stand by my original comment. But there are good arguments on both sides. Someone just needs to make a decision. :-)

rhuss:
> It seems inconsistent to me for the user to click "Add a step" and then get a page that says "Choose a Step", but it's not a big deal

Yeah, I see the point. But then, the "Choose a Step" on the left column in the screenshot is also inappropriate, isn't it?

Actually, I also think it's not a big deal. So let's @kcbabo decide ;-)

kcbabo:
'Choose' is fine with me.

dsimansk:
@rhuss should it have been merged eventually? I was under impression that more votes favoured current state.

PR #616, opened 2017-12-05
fix: Enable --checks on tests

rhuss:
but not on --system-test as they are always run in parallel


rhuss:
Hmm, actually it looks like that --checks is enabled by default and its --skip-checks which disable the checks, so there should be no need for this extra option.

If this is the case we should close this PR without applying.

rhuss:
Let me close this. It works already as expected.

PR #618, opened 2017-12-05
fix: use proper GAV for REST Swagger connector

zregvart:
Missed this when refactoring, the GAV for REST Swagger connector points
to a wrong artifact.


rhuss:
@zregvart there are test failure, I guess you have to adapt the tests for this change.

Issue #436, opened 2017-11-23
Implement support for HTTP Basic Authentication


zregvart:
For the API connector we need HTTP Basic Authorization support. When generating a custom connector we need to add properties for username/password that can be set when a connection is generated for the connector.

PR #619, opened 2017-12-05
chore(rest-swagger): add support for OAuth2

### LINKED ISSUE #606: Add support for new properties in REST Swagger connector
I seem to have lost an issue for this, creating a new one.

We should set HTTP headers needed for authorization based on the values given in the Custom API Connector/Connection properties. For instance if `authenticationType` is `oauth2` we need to take the value of `accessToken` and set it on the `Authorization` header.

zregvart:
Adds support for setting the `Authorization` header for OAuth2
authentication. Also adds support for `basePath` component property.

Fixes #606

### DIFF
--- a/app/connectors/connectors/rest/rest-swagger-connector/src/main/java/io/syndesis/connector/swagger/SwaggerConnectorComponent.java
+++ b/app/connectors/connectors/rest/rest-swagger-connector/src/main/java/io/syndesis/connector/swagger/SwaggerConnectorComponent.java
@@ -40,6 +40,10 @@ import org.apache.commons.io.IOUtils;
 
 public class SwaggerConnectorComponent extends DefaultConnectorComponent {
 
+    private String accessToken;
+
+    private AuthenticationType authenticationType;
+
     private String specification;
 
     public SwaggerConnectorComponent() {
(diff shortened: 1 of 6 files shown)


app/connectors/connectors/rest/rest-swagger-connector/src/main/java/io/syndesis/connector/swagger/SwaggerConnectorComponent.java
```diff
@@ -40,24 +40,73 @@
...
+    public void setAuthenticationType(final AuthenticationType authenticationType) {
+        this.authenticationType = authenticationType;
+    }
+
     public void setSpecification(final String specification) {
         this.specification = specification;
     }
 
+    /* default */ void addAuthenticationHeadersTo(final Map<String, Object> headers) {
+        if (authenticationType == AuthenticationType.oauth2) {
```
rhuss: Normally I'd prefer to add a `AuthenticationType.addToHeaders()` methods which avoid such if-else checks. But I'm fine with this (i.e. if authenticationType can  be null)

zregvart: I think that way would be nice when we add the third way of authenticating (#436)

zregvart:
I like this WIP workflow, not sure about the name, but it works nicely.

PR #620, opened 2017-12-05
feature: Add option --kill-pods (-k) for killing pods at the end of build

rhuss:
This is useful if the docker mode is used for creating images
(and the dev template)

Also: --docker and --openshift are now shortcuts for --image-mode docker
and --image-mode openshift


rhuss:
Ah, yes, and changed --namespace to --project (as we are on OpenShift) and allow a --project for --minishift, so that you don't have to reinstall whole minishift but just re-create the project.

PR #621, opened 2017-12-05
chore(ci): store test metadata in circleci

zregvart:
This adds test summary to CircleCI builds (see [build 1224](https://circleci.com/gh/syndesisio/syndesis/1224)):

screenshot-2017-12-5 circleci

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,6 +58,14 @@ save_m2: &save_m2
     paths:
       - repository/io/syndesis
 
+save_junit: &save_junit
+  run:
+    name: Collect junit reports
+    when: always
+    command: |
+      mkdir -p /workspace/junit/
+      find . -type f -regextype posix-extended -regex ".*/target/(surefire|failsafe)-reports/.*xml" | xargs -i cp --backup --suffix=.xml {} /workspace/junit/
+
 jobs:
   # UI has no dependencies, just load cache
   ui:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -58,6 +58,14 @@ save_m2: &save_m2
     paths:
       - repository/io/syndesis
 
+save_junit: &save_junit
+  run:
+    name: Collect junit reports
+    when: always
+    command: |
+      mkdir -p /workspace/junit/
+      find . -type f -regextype posix-extended -regex ".*/target/(surefire|failsafe)-reports/.*xml" -exec cp {} /workspace/junit/ \;
```
rhuss: Fine for me, could be optimized with something like `| xargs cp -t /workspace/junit` or `| xargs -i cp {} /workspace/junit` instead of `-exec` as it much more lightweight. But hey, if its only about some xml files, its not worth the effort to check whether the base image supports xargs or cp with the extension mentioned above.

However, it this is large set, calling an extra cp for every file can really be time consumer (speaking out of experience).

(I know the original version used also find -exec, but well, let's get better ;-)

zregvart: Yeah, I added a `--backup` not to overwrite existing files there (there might be some if we have two same-named tests for whatever reason), with `--suffix=.xml` if a file `test.xml` exists one that would otherwise overwrite it will be named `test.xml.xml` (and one after that `test.xml.xml.xml`, so that should be fun).

PR #630, opened 2017-12-06
chore(image-generator): add support for connector templates

zregvart:
We need to add artifacts for connectors that can be generated from
connector templates. This generates an ad hoc action and adds a step
with it to the integration being stubbed for the project generator in
order to accomplish that.


zregvart:
retest this please

PR #634, opened 2017-12-06
Fix sql connectors

lburgazzoli:
@KurtStam I noticed that the connector framework has some issues so I've amended your sql conenctors to workaround it, the issues are:
- a connector property should be marked as `componentProperty` or it is then forwarded to the underlying component which may fail if the option is unknown (in this case the offending option was `procedureName`)
- customizers are ignored by spring as the connector maven plugin does override spring.factories so I've moved datasource bits to the main connector class (and amended the related test).

Do you mind reviewing it ?


lburgazzoli:
retest this please

lburgazzoli:
retest this please

lburgazzoli:
@KurtStam logging removed

Issue #468, opened 2017-11-27
ActiveMQ Connector - UX Design


dhirajsb:
Need collateral for ActiveMQ connection images, etc. 
Also need to sort UI ordering of connection properties, filter out 'to' actions for first step.

amysueg:
@dhirajsb - need a bit more information on this one to determine what you need. thanks. 
@sjcox-rh @dongniwang fyi

amysueg:
UX PR issue #638 is merged.
Closing this issue.

PR #638, opened 2017-12-06
Add ux design for AMQ connection and icon for AMQ

dongniwang:
UX design for the AMQ connection along with the AMQ icon in .png and .svg. 

Reference: https://github.com/syndesisio/syndesis/issues/468


rhuss:
@dhirajsb could you please have a look and approve if the design is ok ?

PR #639, opened 2017-12-06
Add an additional plugin repository needed in the build and additional dependency ignores.

cunningt:
Add some ignoreUsedDependencies that are needed in the build, and add a plugin repository statement for SNAPSHOTS because the syndesis-build-helper-maven-plugin is used and I'm getting errors when running mvn clean and it can't resolve the plugin.

### DIFF
--- a/app/rest/runtime/pom.xml
+++ b/app/rest/runtime/pom.xml
@@ -60,6 +60,18 @@
       <url>https://jcenter.bintray.com</url>
     </pluginRepository>
 
+    <pluginRepository>
+      <id>sonatype-snapshot</id>
+      <name>Sonatype snapshot repository</name>
+      <url>https://oss.sonatype.org/service/local/repositories/snapshots/content</url>
+      <releases>
+        <enabled>true</enabled>
+      </releases>
+      <snapshots>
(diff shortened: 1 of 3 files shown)


app/rest/runtime/pom.xml
```diff
@@ -60,6 +60,18 @@
       <url>https://jcenter.bintray.com</url>
     </pluginRepository>
 
+    <pluginRepository>
+      <id>sonatype-snapshot</id>
+      <name>Sonatype snapshot repository</name>
+      <url>https://oss.sonatype.org/service/local/repositories/snapshots/content</url>
+      <releases>
+        <enabled>true</enabled>
```
rhuss: Shouldn't be set this to `false` ?

rhuss:
@zregvart @cunningt actually I wonder why this does not bite us upstream. happy with the additional repo.

zregvart:
Is this something we still need to address?

rhuss:
@cunningt sorry, I forgot the original rationale for this PR. Is this still required, and if so, could you please elaborate a bit on the issues ? Thanks !

PR #640, opened 2017-12-06
Describe Data Step UX Design

dongniwang:
Add UX design for Describe Data Step. 

User story: https://github.com/syndesisio/syndesis/issues/206
Epic: https://github.com/syndesisio/syndesis/issues/194


rhuss:
Short question: Ths PR also contains the changes for the AMQ connector. I guess this is an accdent, so I could rebase this PR to only include the changes to Data Step Design (if the AMQ design is already merged)

dongniwang:
@rhuss If I understand it correctly, the only conflict is in the ux design directory .md file and I resolved the conflict this morning with help from @seanforyou23. The AMQ design and Describe Data Step design are on two separate branches.

dsimansk:
@dongniwang @rhuss I think we're still missing AMQ connection icon. Should it be part of UX design?

dsimansk:
Nevermind, found it and this is probably wrong place to ask for it :)

dongniwang:
@dsimansk yeah, the .png and .svg files are included in this PR.

ux/designs/describedatastep/describedatastep.md
```diff
@@ -0,0 +1,64 @@
...
+  Upload file validation:
+    - JSON Schema - .json
+    - JSON instance - .json
+    - XML Schema - .xsd
+    - XML instance - .xml
+
+
+![Image of import via url](img/describeData_configStep_URL.png)
+
+1. User can also import schema via a valid URL. Provide “http://“ as syntax hint for users. Once users start typing, they can override the syntax hint.
```
rhuss: would use 'https://' as hint (as this should be the default)

ux/designs/describedatastep/describedatastep.md
```diff
@@ -0,0 +1,64 @@
...
+
+2. When a file is in the upload process, the drop file zone and choose file button is disabled.
+
+![Image of upload error](img/describeData_configStep_uploadError.png)
+
+1. If file uploaded isn’t a supported file type, display error message in form of inline notification.  
+
+  E.g. If user selects JSON schema in the first step, an uploaded file should be in .json format.
+
+  Upload file validation:
```
rhuss: add an empty line which helps markdown to format this as a list.

rhuss:
Sorry for the delay btw. @kcbabo @syndesisio/ui-api any comments to this PR ?

kcbabo:
LGTM.  Minor comment on the description of the step:
Description of Describe Data should not reference what you can't do.  That should be a validation that happens in the step.  Probably should be something like "Declare data type when not specified by a connection".

kcbabo:
BTW, validation of a file based on extension seems pretty shaky to me.  IMO, the only way that extension comes into the picture is as a default filter in a file chooser.  There is no rule that JSON schema files or instance documents have to end in .json, for example.

dongniwang:
> BTW, validation of a file based on extension seems pretty shaky to me. IMO, the only way that extension comes into the picture is as a default filter in a file chooser. There is no rule that JSON schema files or instance documents have to end in .json, for example.

Yeah, it's not validating based on file extension, the wording I have on the .md file is misleading. 

@dhirajsb please provide your input here. Thanks!

dongniwang:
Updated the describe data step description to 

> Declare data type when not specified by a connection

cc: @TovaCohen

ux/designs/describedatastep/describedatastep.md
dhirajsb: Are we expecting Citizen Integrators to be able to setup non-http routes, since that's the only way they will be able to access anything non-https outside OpenShift?

dhirajsb:
Are we expecting Citizen Integrators to be able to setup non-http routes, since that's the only way they will be able to access anything non-https outside OpenShift?
Also, the file extension was meant to be a simple filter (in a way a quick and dirty validation) to avoid loading non-json files. Although, there is no standard for json schema like XML is, so might be limiting for that. 
The contents will have to be fully validated (by actually parsing, and in case of jsonschema/instance docs, differentiated) using a rest endpoint in the backend.

kcbabo:
> Are we expecting Citizen Integrators to be able to setup non-http routes, since that's the only way they will be able to access anything non-https outside OpenShift?

I believe this is a response to Roland's comment, but wanted to point out that routes are not required for outbound communication from within an OpenShift cluster, right?

> Also, the file extension was meant to be a simple filter (in a way a quick and dirty validation) to avoid loading non-json files. Although, there is no standard for json schema like XML is, so might be limiting for that.

Agree that it should be used as an initial filter for the file chooser dialog.  We should just separate that from validation, since the file extension has nothing to do with validating the actual content of the file.

> The contents will have to be fully validated (by actually parsing, and in case of jsonschema/instance docs, differentiated) using a rest endpoint in the backend.

Yep, makes sense.

dongniwang:
@dhirajsb Anything needs to be updated or we're good for now?

Issue #467, opened 2017-11-27
Support OpenAPI 3.0


zregvart:
See also **https://issues.jboss.org/browse/ENTESB-11541**

We currently have support for Swagger 2.0 in the API connector, we should also support OpenAPI 3.0

zregvart:
@kcbabo I seem to recall that we needed OpenAPI 3.0 support as P1, and I think this issue came about in order to support OAuth1 which is not supported in Swagger 2.0.
Is this still required (supporting OAuth1 and OpenAPI 3.0)? I feel it might be better to wait for feedback and see if there is a need for it, I haven't seen any OpenAPI 3.0 or OAuth1 APIs while searching for demo APIs to use.

zregvart:
We'll add support for OpenAPI 3.0 and OAuth1 outside of API Client connector epic

heiko-braun:
Migrated to **https://issues.jboss.org/browse/ENTESB-11541**

PR #641, opened 2017-12-06
fix(swagger-generator): host property generation

### LINKED ISSUE #631: Swagger connector generator generates wrong host connector parameter
In Swagger connector generator we use the `host` parameter from the Swagger specification, we should also consider `schemes` property and generate the URL as the underlying REST Swagger Camel component.

zregvart:
Swagger connector generator now considers `schemes` from Swagger
specification and the Swagger specification URL in order to determine
proper `host` property for the connector being generated.

Fixes  #631


jimmidyson:
I assume we only support Swagger 2.0? Are we ignoring OpenAPI 3.0 for now?

app/rest/connector-generator/src/main/java/io/syndesis/connector/generator/swagger/PropertyGenerators.java
```diff
@@ -114,6 +131,38 @@
...
+            schemeToUse = "https";
+        } else {
+            schemeToUse = schemes.get(0).toValue();
+        }
+
+        final String host = swagger.getHost();
+        String hostToUse;
+        if (StringUtils.isEmpty(host)) {
+            hostToUse = requireNonNull(specificationUrl, "Swagger specification does not provide a `host` definition "
+                + "and the Swagger specification was uploaded so the originating URL is lost to determine the scheme to use").getHost();
```
jimmidyson: `so it is impossible to determine the originating URL` perhaps?

app/rest/connector-generator/src/main/java/io/syndesis/connector/generator/swagger/SwaggerConnectorGenerator.java
```diff
@@ -107,7 +110,11 @@ public ConnectorSummary info(final ConnectorTemplate connectorTemplate, final Co
     /* default */ Connector basicConnector(final ConnectorTemplate connectorTemplate, final ConnectorSettings connectorSettings) {
         final Swagger swagger = parseSpecification(connectorSettings);
 
-        requiredSpecification(connectorSettings);
+        final String specification = requiredSpecification(connectorSettings);
```
jimmidyson: `specificationURL` perhaps?

zregvart:
> I assume we only support Swagger 2.0? Are we ignoring OpenAPI 3.0 for now?

That is correct, I can't seem to find OpenAPI 3.0 Java implementation that we can use, and also it simplifies things a bit. Also #467

zregvart:
@jimmidyson we should be good to go here, mind having another look?

PR #642, opened 2017-12-06
Add UX design for Database connection

sjcox-rh:
UX designs for creating an integration using the Database connection. Covers the various use cases when using this connection.

Please reference UX story: [329 ](https://github.com/syndesisio/syndesis/issues/329)
Epic: [179](https://github.com/syndesisio/syndesis/issues/179)


TovaCohen:
Is there a UX design for selecting the Database connector and creating a Database connection?
What input fields are required to create a Database connection?

KurtStam:
@TovaCohen It's the same workflow as for any other Connector so I don't think we need to include it. Besides it is the same connector as the Stored Procedure connector so nothing changed there since TP2.

sjcox-rh:
Thanks @KurtStam!

PR #643, opened 2017-12-06
The return of the checks

zregvart:
Seems that we lost checks from being run, this reinstates the checks.

 - Makes the license plugin non-inheritable so its not run on modules, only on parent build.
 - We should rely on `basepom.check.skip-all` set from parent and the profile configuration there, not in individual modules
 -  fixes reported dependency checks/codestyle checks

### DIFF
--- a/app/rest/rest/pom.xml
+++ b/app/rest/rest/pom.xml
@@ -209,6 +209,16 @@
       <artifactId>kubernetes-model</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>com.squareup.okio</groupId>
+      <artifactId>okio</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>okhttp</artifactId>
+    </dependency>
+
     <!-- === Testing Dependencies ======================================================== -->
 
     <dependency>
@@ -253,6 +263,12 @@
     <dependency>
       <groupId>org.springframework.security</groupId>
       <artifactId>spring-security-web</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>com.squareup.okhttp3</groupId>
+      <artifactId>mockwebserver</artifactId>
       <scope>test</scope>
     </dependency>
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,7 +83,7 @@ jobs:
       - run:
           name: Build UI
           command: |
-            ./app/build.sh --batch-mode --module ui --image-mode=docker | tee build_log.txt | grep -v " Download"
+            ./app/build.sh --batch-mode --module ui --image-mode=docker | tee build_log.txt
       - store_artifacts:
           path: ./build_ui_log.txt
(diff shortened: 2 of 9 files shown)


zregvart:
So, [CircleCI fails](https://circleci.com/gh/syndesisio/syndesis/1379) with:
```
[INFO] --- maven-dependency-versions-check-plugin:2.0.4:check (default) @ openshift ---
[INFO] Checking dependency versions
Too long with no output (exceeded 10m0s)
```

@syndesisio/backend at what cost to build performance are we willing to sacrifice some/all of the checks?

Aside to that, [we can](https://circleci.com/docs/2.0/configuration-reference/#run) set `no_output_timeout` to a larger value than 10 min.

app/rest/rest/pom.xml
```diff
@@ -256,6 +266,12 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
```
iocanel: Why?

app/rest/rest/pom.xml
```diff
@@ -209,6 +209,16 @@
       <artifactId>kubernetes-model</artifactId>
     </dependency>
 
+    <dependency>
```
iocanel: Why?

rhuss:
> @syndesisio/backend at what cost to build performance are we willing to sacrifice some/all of the checks?

Build performance is still essential, just at the frequency as PRs are coming in. Also, unfortunately, Circle does not support cancelling running builds for new commits (for workflows as we use it), so this is even a bigger issue.

10 minutes only for this dependency check (i.e. because of dependencies don't change that often) sound over the top for me. 

I'm happy with a dedicated second build, running periodically for doing checks for things which are not supposed to change often.

And again for the license check: Why not disabling it and just run a license format before a release (or periodically)?

zregvart:
@iocanel, it's easier to reply here.

Checks performed by [dependency-versions-check](https://github.com/ning/maven-dependency-versions-check-plugin) singled out dependencies that we haven't declared according to their use. That would mean that a module depending on our module could get a dependency we depend on only for tests, or could get a dependency of a different version than the one we use. The plugin enforces rigidity with regards to dependency management.

Not 100% sure that the implemented fix is a correct one, I didn't check to see if we're actually using these dependencies at runtime or at test. I think these were marked by the plugin to be used at compile (kubernetes-model) and only to be used in tests (mockwebserver), so that's why I set the scope.

zregvart:
Let's table this for a later point in time, seems that [the build](https://circleci.com/gh/syndesisio/syndesis/1456#artifacts/containers/0) took 1h (I canceled it). I suspect network throttling, but could be also an issue in the dependency plugin. Locally it works OK with the dependencies already cached in the local Maven repository.

app/rest/rest/pom.xml
rhuss: guess because of the maven dependency check croaking ? (and someone using okio directly in the code) But @zregvart probably knows best ;-)

zregvart:
@rhuss seems like the build passes now, the 10min for rest build is due to dependency download I think that will be reduced once we cache those.

Issue #645, opened 2017-12-07
Duplicate definitions for popNotification


seanforyou23:
This same function is defined in three places

ConnectionsListComponent
IntegrationViewBase
OAuthAppModalComponent

We probably want these consolidated so we can manage adjustments at the app level.

PR #646, opened 2017-12-07
chore(notifications): consolidate duplicate definitions of popNotific…

seanforyou23:
…ation

consolidate duplicate definitions of popNotification
move duplicate definitions to centralized service
update references to popNotification

Wanted to go ahead and get this work submitted separate from the other notifications work I'm doing since it's not directly impacted by it and stands as a nice cleanup on its own.

Takes care of [645](https://github.com/syndesisio/syndesis/issues/645)

### DIFF
--- a/app/ui/src/app/common/ui-patternfly/notification-service.ts
+++ b/app/ui/src/app/common/ui-patternfly/notification-service.ts
@@ -39,4 +39,17 @@ export class NotificationService extends NS {
   getNotificationsObservable(): Observable<Notification[]> {
     return this.notificationsSubject.asObservable();
   }
+
+  // Toast notification wrapper, used for notifications
+  // that are not persistent and do not have actions associated
+  popNotification(notification) {
+    this.message(
+      notification.type,
+      notification.header,
+      notification.message,
+      false,
+      null,
+      []
+    );
+  }
 }
--- a/app/ui/src/app/common/ui-patternfly/notification-service.ts
+++ b/app/ui/src/app/common/ui-patternfly/notification-service.ts
@@ -21,9 +21,9 @@ export class NotificationService extends NS {
     type: string,
     header: string,
     message: string,
-    isPersistent: boolean,
-    primaryAction: Action,
-    moreActions: Action[]
+    isPersistent: boolean = false,
+    primaryAction?: Action,
+    moreActions: Action[] = []
   ): void {
     super.message(
       type,
@@ -42,14 +42,7 @@ export class NotificationService extends NS {
 
   // Toast notification wrapper, used for notifications
   // that are not persistent and do not have actions associated
-  popNotification(notification) {
-    this.message(
-      notification.type,
-      notification.header,
-      notification.message,
-      false,
-      null,
-      []
-    );
+  popNotification({type, header, message}): void {
+    this.message(type, header, message);
   }
 }
--- a/app/ui/src/app/common/ui-patternfly/notification-service.ts
+++ b/app/ui/src/app/common/ui-patternfly/notification-service.ts
@@ -42,7 +42,7 @@ export class NotificationService extends NS {
 
   // Toast notification wrapper, used for notifications
   // that are not persistent and do not have actions associated
-  popNotification({type, header, message}): void {
-    this.message(type, header, message);
+  popNotification({type, header, message, isPersistent = false}): void {
+    this.message(type, header, message, isPersistent);
   }
 }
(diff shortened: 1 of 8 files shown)


app/ui/src/app/common/ui-patternfly/notification-service.ts
```diff
@@ -39,4 +39,17 @@ export class NotificationService extends NS {
...
+  // that are not persistent and do not have actions associated
+  popNotification(notification) {
+    this.message(
+      notification.type,
+      notification.header,
+      notification.message,
+      false,
+      null,
+      []
+    );
```
deeleman: I've noticed that the `NotificationService.message()` is also called from other instances of the application with a similar payload in the signature. So now that we're updating the service, perhaps we can also simplify calls to the `message()` method by introducing default parameter values like this:


```typescript
message(
    type: string,
    header: string,
    message: string,
    isPersistent = false, // <--- This
    primaryAction?: Action, // <--- This
    moreActions: Action[] = [] // <--- This
  ): void {
    super.message(
      type,
      header,
      message,
      isPersistent,
      primaryAction,
      moreActions
    );
    this.notificationsSubject.next(this.getNotifications());
  }
```

This would also allow for a more leaner approach when calling the method by means of ES6 deconstruction:

```typescript
popNotification(notification): void {
  const { type, header, message } = notification;
  this.message(type, header, message);
}
```

Also please note that now we'll annotate the return type in the code example. Make a habit of annotating methods with the returned type, or 'void' if none, so we can better our code into a self-documenting experience.
Leaner, simpler, hipster code, yay

app/ui/src/app/connections/list/list.component.ts
```diff
@@ -59,25 +59,22 @@ export class ConnectionsListComponent implements OnInit {
...
-          });
-        }, 10);
+        this.notificationService.message(
+          NotificationType.DANGER,
+          'Delete Failed',
+          `Failed to delete connection: ${err}`,
+          true,
+          null,
+          []
+        );
```
deeleman: We're calling both `NotificationService.popNotification()` and `NotificationService.message()` from the same code construct in order to trigger opposite messages. Perhaps we should abstract everything from `NotificationService.popNotification()` by extending its method payload.

app/ui/src/app/customizations/tech-extensions/tech-extension-delete-modal.component.ts
```diff
@@ -34,26 +34,20 @@ export class TechExtensionDeleteModalComponent implements OnInit, OnDestroy {
...
-            'Deleted!',
-            'The technical extension "' + this.extension.name + '" has been deleted',
-            false,
-            null,
-            [],
-          );
+          this.notificationService.popNotification({
+            type: NotificationType.SUCCESS,
+            header: 'Deleted!',
+            message: 'The technical extension "' + this.extension.name + '" has been deleted'
```
deeleman: I'd use template strings instead:

message: `The technical extension ${this.extension.name} has been deleted`

app/ui/src/app/customizations/tech-extensions/tech-extension-delete-modal.component.ts
```diff
@@ -34,26 +34,20 @@ export class TechExtensionDeleteModalComponent implements OnInit, OnDestroy {
...
-            'Deleted!',
-            'The technical extension "' + this.extension.name + '" could not be deleted due to: ' + err.userMsg || 'Unknown error',
-            false,
-            null,
-            [],
-          );
+          this.notificationService.popNotification({
+            type: NotificationType.WARNING,
+            header: 'Deleted!',
+            message: 'The technical extension "' + this.extension.name + '" could not be deleted due to: ' + err.userMsg || 'Unknown error'
```
deeleman: I'd rather use template strings here too. It makes de code more readable.

app/ui/src/app/integrations/detail-page/detail.component.ts
```diff
@@ -199,24 +193,18 @@ export class IntegrationsDetailComponent extends IntegrationViewBase
...
-          'Update Failed',
-          'Failed to update ' + attr + ': ' + reason,
-          false,
-          undefined,
-          undefined
-        );
+        this.notificationService.popNotification({
+          type: NotificationType.WARNING,
+          header: 'Update Failed',
+          message: 'Failed to update ' + attr + ': ' + reason
```
deeleman: This seems another good candidate for introducing template strings instead.

app/ui/src/app/common/ui-patternfly/notification-service.ts
deeleman: Actually, the above can be simplified even more, at the cost of replacing strongly typed payloads by type inference, like this:

```typescript
popNotification({ type, header, message }): void {
  this.message(type, header, message);
}
```

PR #650, opened 2017-12-07
Tech extensions: remove feature toggle leftovers

lburgazzoli:


lburgazzoli:
@rhuss I think you need to approve it again after fixing some issues

app/runtime/runtime/src/main/java/io/syndesis/integration/runtime/stephandlers/FunctionHandler.java
```diff
@@ -58,6 +59,7 @@ public ProcessorDefinition handle(Function step, ProcessorDefinition route, Synd
         Map<String, Object> headers = step.getProperties();
         if (ObjectHelper.isNotEmpty(headers)) {
             options = headers.entrySet().stream()
+                .filter(entry -> Objects.nonNull(entry.getValue()))
```
zregvart: This seems unrelated to tech extensions feature toggle

lburgazzoli: uh yeah, that was supposed to be included in a different pr

PR #649, opened 2017-12-07
feat: Fetch of custom connector

zregvart:
Implements new REST endpoint for fetching custom connectors by id.

Fixes #648

PR #657, opened 2017-12-07
feat: Delete of custom connector

### LINKED ISSUE #652: Implement REST endpoint to delete a custom connector
We need REST operation to fetch a single custom connector by id.

UX [API Connector List](https://github.com/syndesisio/syndesis/blob/master/ux/designs/apiconnector/apiconnector.md#api-connector-list)
UI #271

zregvart:
**Based on #649 please review it first**

Implements new REST endpoint for deleting custom connectors by id.

Fixes #652

### DIFF
--- a/app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/CustomConnectorHandler.java
+++ b/app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/CustomConnectorHandler.java
@@ -41,6 +41,7 @@ import io.syndesis.model.connection.Connector;
 import io.syndesis.model.connection.ConnectorSettings;
 import io.syndesis.rest.util.PaginationFilter;
 import io.syndesis.rest.util.ReflectiveSorter;
+import io.syndesis.rest.v1.operations.Deleter;
 import io.syndesis.rest.v1.operations.Getter;
 import io.syndesis.rest.v1.operations.PaginationOptionsFromQueryParams;
 import io.syndesis.rest.v1.operations.SortOptionsFromQueryParams;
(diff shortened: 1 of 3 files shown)


app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/CustomConnectorHandler.java
```diff
@@ -35,19 +35,22 @@
...
+import io.syndesis.rest.v1.operations.Getter;
 import io.syndesis.rest.v1.operations.PaginationOptionsFromQueryParams;
 import io.syndesis.rest.v1.operations.SortOptionsFromQueryParams;
 import io.syndesis.rest.v1.util.PredicateFilter;
 
 import org.springframework.context.ApplicationContext;
 
 @Api(tags = {"custom-connector", "connector-template"})
-public final class CustomConnectorHandler extends BaseConnectorGeneratorHandler {
+public final class CustomConnectorHandler extends BaseConnectorGeneratorHandler implements Getter<Connector>, Deleter<Connector> {
```
jimmidyson: I'm still not convinced we need this at all. If we do in future, maybe we can add it in, but right now there is no difference between handling get/delete for custom connectors to normal connectors, is there? I don't like having two different endpoints for effectively same functionality.

jimmidyson:
Seeing this duplication strengthens my belief that list/get/delete should all be done via normal `/connectors` endpoint. I'd rather work on a generic filtering process in `Lister`.

zregvart:
Agree, lets base those from `/connector` endpoint

PR #658, opened 2017-12-07
#271 API Connector List

kahboom:
### Overview
This PR is for the API Client Connector list page. The search and sort work, as they are part of the list component from patternfly-ng.

###  Screenshots

screenshot 2017-12-07 11 02 53

The details page is carved out but needs to be updated, and will be in a subsequent PR today.

### Outstanding Items
- Properly display icons for connectors (separate endpoint? need to speak with @deeleman and @jimmidyson about this).
- Delete action.
- Update latin text once I hear back from UXD and Docs (@sjcox-rh @TovaCohen ).
- Remove `/deep/` from styling.
- Refrain from using Restangular.
- Properly define `ApiConnector` model (likely won't be extending existing `Connector` model).
- Update API endpoint once it is working properly locally, specific to custom connectors. Currently using `/connectors` in the meantime.
- Display column with number of uses, per design, once that is ready in the API (@zregvart actively working on this).


app/ui/src/app/customizations/api-connector/api-connector-list.component.html
```diff
@@ -2,20 +2,20 @@
   <div class="row">
     <div class="col-xs-12">
       <h2>API Client Connector</h2>
-      <p>Nullam quis risus eget urna mollis ornare vel eu leo. Find out more at <a>{{ config.getSettings('branding', 'appName', 'Syndesis') }} Help</a></p>
+      <p>Nullam quis risus eget urna mollis ornare vel eu leo. Find out more at <a>{{ config.getSettings('branding', 'appName', 'Syndesis') }} Help</a>.</p>
```
deeleman: Please do not bind a function call into an Angular interpolation. Otherwise you'll wind up calling `config.getSettings('branding', 'appName', 'Syndesis')` on each cycle of the change detector loop.

Tip: add this to a class field initialized at the `OnInit` stage.

app/ui/src/app/customizations/api-connector/api-connector-list.component.html
```diff
@@ -44,8 +46,8 @@ <h2>API Client Connector</h2>
                 </div>
               </div>
               <div class="list-pf-additional-content">
-                <div *ngIf="item.connectors">
-                  <strong>Used by {{ item.connectors }}</strong>
+                <div *ngIf="item.apiConnectors">
+                  <strong>Used by {{ item.apiConnectors }}</strong>
```
deeleman: May be you'll want to implement a 118nPluralPipe here to refine the text output, like `Used by 1 connector`, `Used by 23 connectors`, `Used by no connectors yet`... 

Here you have a [link to the API specification](https://angular.io/api/common/I18nPluralPipe) and also an actual [example on Syndesis](https://github.com/syndesisio/syndesis/pull/556/files#diff-8cb4ab233e8d35edb6d1bd79beaeb809R19).

app/ui/src/app/customizations/api-connector/api-connector-list.component.ts
```diff
@@ -24,16 +26,19 @@ const category = getCategory('ApiConnectors');
   styleUrls: ['./api-connector-list.component.scss']
 })
 export class ApiConnectorListComponent implements OnInit {
-  connectors: Observable<Connectors>;
-  filteredConnectors: Subject<Connectors> = new BehaviorSubject(<Connectors>{});
-  loading: Observable<boolean>;
+  apiConnectors$: Observable<ApiConnectors>;
+  filteredApiConnectors$: Subject<ApiConnectors> = new BehaviorSubject(<ApiConnectors>{});
```
deeleman: You do not need to annotate class members when initializing those straight away. The TypeScript compiler will do that by type inference:

```typescript
filteredApiConnectors$ = new BehaviorSubject(<ApiConnectors>{});
``` 

will suffice. Also, let's keep consistency between types. `BehaviorSubject` is an specialized class that inherits from `Subject`. If you want to use `BehaviorSubject`, you should annotate the class member accordingly (although not required in this case)

app/ui/src/app/customizations/api-connector/api-connector-list.component.ts
```diff
@@ -24,16 +26,19 @@ const category = getCategory('ApiConnectors');
...
+  loading$: Observable<boolean>;
   listConfig: ListConfig;
 
   constructor(
     public config: ConfigService,
-    private store: ConnectorStore,
+    private store: ApiConnectorStore,
```
deeleman: I'd suggest to name the class member as `apiConnectorStore` instead of just `store`. This is a common practice borrowed from ES5 and ES6, where there is no type inference and we resource to the variable/field member to inform about its type. This also ensures not name collisions will occur whenever we inject another WhateverStore in our component. Eg: A component that pulls information from both the `ApiConnectorStore` (for connector details) and a future `SyndesisI18NStore` (for localization)...

kahboom: Good point!

app/ui/src/app/customizations/api-connector/api-connector-list.component.html
kahboom: Definitely, but since we don't even have the values yet I didn't write out the complete statement (which should end with 'connection' or 'connections' obviously). I can add it in a separate PR once that's ready.

app/ui/src/app/customizations/api-connector/api-connector-list.component.html
kahboom: We discussed this before. Totally agree, and I had originally fixed this. Let me update it now. 👍

app/ui/src/app/customizations/api-connector/api-connector-list.component.html
deeleman: Why wait? Having no data is a good way to also test how the app behaves in edge cases! Go ahead and do some nice experiments with text here. Since `item.apiConnectors` will yield nothing, you can see how a `no connections` or `0 connections` would fit in the overall UI.

;)

app/ui/src/app/customizations/api-connector/api-connector.service.ts
```diff
@@ -0,0 +1,19 @@
...
+import { ApiConnector, ApiConnectors } from './api-connector.model';
+
+@Injectable()
+export class ApiConnectorService extends RESTService<ApiConnector, ApiConnectors> {
+  constructor(restangular: Restangular) {
+    super(restangular.service('connectors'), 'apiConnector');
+  }
+
+  public list(): Observable<ApiConnectors> {
+    return super.list().map( apiConnectors => {
```
deeleman: Ups! I found a naughty rogue space in the lambda

app/ui/src/app/customizations/api-connector/api-connector.model.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export interface ApiConnectorProperties extends BaseEntity {
+  specification: {
+    kind: string;
+    displayName: string;
+    required: boolean;
+    type: string;
+    javaType: string;
+    //tags: Array<Tag>;
```
deeleman: If you do not plan to use this commented property now, just remove it. Pushing commented out code is usually considered a source for technical debt for the future

app/ui/src/app/customizations/api-connector/api-connector-list.component.ts
```diff
@@ -58,7 +66,23 @@ export class ApiConnectorListComponent implements OnInit {
...
+  }
+
+  handleClick(event: any) {
+    const apiConnector = event.item;
+    this.router.navigate([apiConnector.id], { relativeTo: this.route });
+  }
+
   ngOnInit() {
-    this.store.loadAll();
+    this.getAppName();
```
deeleman: the `getAppName()` method is used only once and we do not expect to execute it outside the component domain (it is not even possible, since the component API is not exported thru an `exportAs` property in the decorator).

We also only need to initialize `this.appName` just once, so I'd remove the `getAppName()` method and insert its implementation here, like this:

```typescript
ngOnInit() {
  this.appName = this.config.getSettings('branding', 'appName', 'Syndesis');
  this.apiConnectorStore.loadAll();
}
``` 

the lesser the code, the better ;)

kahboom: Sure thing!

app/ui/src/app/customizations/api-connector/api-connector.model.ts
kahboom: We will need it, but perhaps not for this sprint. I'll just remove it.

app/ui/src/app/customizations/api-connector/api-connector.service.ts
kahboom: lol

PR #663, opened 2017-12-08
Added support for SSL connections to AMQ Connector, added ClientID option, updated text descriptions

dhirajsb:


app/connectors/connectors/jms/activemq-connectors/activemq-model/src/main/java/io/syndesis/connector/jms/ActiveMQUtil.java
```diff
@@ -0,0 +1,116 @@
...
+                    keyManagers = null;
+                } else {
+                    keyManagers = createKeyManagers(clientCertificate);
+                }
+
+                // create client trust manager
+                final TrustManager[] trustManagers;
+                if (ObjectHelper.isEmpty(brokerCertificate)) {
+                    // use a trust all TrustManager
+                    trustManagers = new TrustManager[] {new TrustAllTrustManager()};
```
zregvart: Perhaps we should go with the default TrustManager, i.e. only invoke `setKeyAndTrustManagers` if `brokerCertificate` is given?

app/connectors/connectors/jms/activemq-connectors/activemq-model/src/main/java/io/syndesis/connector/jms/TrustAllTrustManager.java
```diff
@@ -0,0 +1,58 @@
...
+import java.security.cert.X509Certificate;
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.X509ExtendedTrustManager;
+
+/**
+ * TrustManager that trusts all certificates.
+ * Use only for AMQ connections to SSL Brokers, when no Broker certificate provided.
+ * @author dhirajsb
+ */
+class TrustAllTrustManager extends X509ExtendedTrustManager {
```
zregvart: I would not do that, we're enabling MITM

app/connectors/connectors/jms/activemq-connectors/activemq-model/src/main/java/io/syndesis/connector/jms/AbstractActiveMQConnector.java
```diff
@@ -39,6 +39,15 @@
     @Metadata(label = "basic", description = "Password for authorization credential")
     private String password;
 
+    @Metadata(label = "basic", description = "Client ID for durable subscriptions")
+    private String clientID;
+
+    @Metadata(label = "basic", description = "AMQ Broker X.509 PEM Certificate")
+    private String brokerCertificate;
```
zregvart: I don't see these added to the `deployment.json` or Camel Connector JSON schema files

app/connectors/connectors/jms/activemq-connectors/activemq-model/src/main/java/io/syndesis/connector/jms/AbstractActiveMQConnector.java
```diff
@@ -39,6 +39,15 @@
...
+
+    @Metadata(label = "basic", description = "AMQ Client X.509 PEM Certificate")
+    private String clientCertificate;
```
zregvart: Needs to be added to `deployment.json` and Camel Connector JSON schema files

app/connectors/connectors/jms/activemq-connectors/activemq-model/src/main/java/io/syndesis/connector/jms/ActiveMQUtil.java
dhirajsb: no, we still need to set a special TrustAllManager, or it won't connect.

app/connectors/connectors/jms/activemq-connectors/activemq-model/src/main/java/io/syndesis/connector/jms/TrustAllTrustManager.java
dhirajsb: yes, you are right. But this is required for AMQ connections to Brokers using SSL. Without it, a certificate is always required. Which is not an easy thing to do. 
But, this was a requirement according to @chirino.

app/connectors/connectors/jms/activemq-connectors/activemq-model/src/main/java/io/syndesis/connector/jms/AbstractActiveMQConnector.java
dhirajsb: I haven't added these since we are not exposing them at the moment, just allowing the MTIM exposed SSL connections. 
Otherwise, we'd need to ask the integration developer to extract keys from Broker truststore, convert them to PEM (for String *Certificate), and then paste them in Syndesis. Might be too much to ask according to @chirino. 
It's not difficult to add these fields, so maybe we can discuss that in another issue.

app/connectors/connectors/jms/activemq-connectors/activemq-model/src/main/java/io/syndesis/connector/jms/AbstractActiveMQConnector.java
dhirajsb: Same as the comment below.

app/connectors/connectors/jms/activemq-connectors/activemq-model/src/main/java/io/syndesis/connector/jms/AbstractActiveMQConnector.java
dhirajsb: @chirino , maybe we can add a couple of large text areas in Connection page for users to paste PEM certs, if they can obtain them. If they don't provide them, we can flash a big, flashing neon warning saying it's vulnerable to MTIM attacks. What do you think?

dhirajsb:
@kcbabo care to comment on how you feel about trusting all Brokers, exposing such AMQ connections to MTIM attacks? If we don't do that, user will have to provide a PEM cert (text form) for the Broker.

chirino:
It's very common to setup developement brokers with self signed certs.  An option to be able to connect to them is needed.

app/connectors/connectors/jms/activemq-connectors/activemq-model/src/main/java/io/syndesis/connector/jms/TrustAllTrustManager.java
chirino: This is a good option to have. All AMQ brokers on openshift can only be accessed via tls. So self signed certs at common in Dev envs

chirino: This is a good option to have. All AMQ brokers on openshift can only be accessed via tls. So self signed certs at common in Dev envs

dhirajsb:
@chirino perhaps there could be an explicit option, saying "Trust This Broker" or "Skip Broker Certificate Check"? Connector will only use the TrustAllTrustManager if that option is selected. 
If it hasn't been selected, and the protocol uses ssl, then the cert fields will be enabled and mandatory. Does that make more sense?

zregvart:
I would hope that we would offer a overarching solution for certificate management in Syndesis at some point, and that we could offer functionality from connectors to report TLS issues and have one common way of managing them in. This could be a good fit for the verifier and the connection validation we do there currently.

kcbabo:
We are at an eval stage right now, therefore I think it's more important for users to be "hands on" as soon as possible to provide feedback.  I agree 100% that configuring the trust store for clients is required for GA.  If we can fit in TP3, then great.  If not, suggest we file a separate issue and file up with a fix in TP4.

dhirajsb:
Just FYI, I have adde ability to configure the trust store with a text PEM cert right now. 

I'll expose it in the next PR  for TP3, without making it mandatory. I'll also add an explicit option to allow users to bypass it for dev Brokers.

PR #665, opened 2017-12-08
chore(syndesis-policy): Remove tabs from license header to satisfy checkstyle

jimmidyson:


zregvart:
Do we need to reformat existing license headers or we're good?

jimmidyson:
@zregvart We need to do that in a subsequent PR after module is released to central or it could break local builds - I'll do that ASAP.

Issue #603, opened 2017-12-05
Rename "build.sh" to "syndesis"


rhuss:
... as it now carries much as only 'building' Syndesis (but also installing to minishift, system-test and dev parts).

Question is also if `syndesis` or `syndesis.sh` (or any other name).

I would then also put a delegate script to `tools/bin` as folks seems to like this location, too (not sure whether a pure symlink works for GitHub)

zregvart:
I think we only need to remove the existing `build.sh` I would wait just a bit for https://doc.syndesis.io/#_one_tool_to_rule_them_all to have all the details and then we can close this.

gashcrumb:
Seems like we did this in some fashion, good enough.

PR #677, opened 2017-12-09
refactor(build.sh): Moved main script to tools/bin

rhuss:
* Renamed from build.sh to syndesis   
* Replaced tools/bin/syndesis-install with a shortcut to 'syndesis   --minishift'   
* Added reference from app/build.sh to tools/bin/syndesis

Just add either tools/bin to your path or make a symbolic link from tools/bin/syndesis into a directory within your path


rhuss:
Fixes #603

PR #679, opened 2017-12-10
feature: Add generation of a "Syndesis Developer Handbook"

rhuss:
this is still WIP


rhuss:
No wip anymore. Could please someone approve it so that I can check whether the documentation is really built ? (happens only on `master`)

Jenkinsfile
```diff
@@ -12,7 +12,7 @@ pipeline {
     stages {
         stage('prepare') {
             steps {
-             withCredentials([string(credentialsId: "api.rh-idev.openshift.com", variable: 'TOKEN')]) {
+             withCredentials([string(credentialsId: "${env.IGNITE_ACCOUNT_CREDENTIAL}", variable: 'TOKEN')]) {
```
dsimansk: Is this change intentional? I guess `IGNITE_ACCOUNT_CREDENTIAL` and `--server=https://api.rh-idev.openshift.com` won't work very well.

rhuss: Yeah, not really sure why this slipped in. But this was a fix from @iocanel upstream so I wonder why it disappears here as a diff.

PR #680, opened 2017-12-11
refactor(synesis): Split up monolythic script into commands

rhuss:
Each command is now moved into an own scripts from `commands/`.
Also support for `syndesis <command>` has been added, but old syntax
still works.

Re-organized help message to be be more context sensitive and hence
shorter normally:

syndesis -h ---> global usage
syndesis build -h --> only build relevant options


oscerd:
LGTM

PR #684, opened 2017-12-11
chore(rest): adds `uses` counter to Connector

### LINKED ISSUE #666: We need `uses` property on Connector
For API connector list we need to provide UI with the `uses` property on Connector that would hold the count of Integrations that use a particular Connector. This is an optional value and it should not be persisted in jsondb but calculated dynamically.

zregvart:
This adds a uses counter to Connector that counts the number of integrations that make use of a particular connector.

This in turn means that endpoints for getting and listing connectors need to load all integrations

Fixes #666

### DIFF
--- a/app/rest/dao/src/main/java/io/syndesis/dao/manager/DataManager.java
+++ b/app/rest/dao/src/main/java/io/syndesis/dao/manager/DataManager.java
@@ -142,7 +142,7 @@ public class DataManager implements DataAccessObjectRegistry {
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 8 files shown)


zregvart:
Seems that I've miss merged this

jimmidyson:
yeah some confusing stuff in there - let me know when it's ready to review.

One quick note - I do have @rhuss's voice in my head now about obfuscated functional code... might want to change some of that to be more imperative.

zregvart:
@jimmidyson I fixed an error prone warning and rebased so it should be OK now. Not sure what functional code can be refactored as imperative -- our current field types use `Optional` a lot and it's difficult not to write in functional style.

PR #711, opened 2017-12-12
Upgraded Kubernetes client versions

paoloantinori:
Required as part of Support Page feature.

Needs more than even feedback from CI since I haven't run the complete test suite, but only `rest` and `runtime` modules.


oscerd:
LGTM, I guess we can approve.

rhuss:
Let's merge and fight dragons later.

PR #730, opened 2017-12-12
fix(swagger-generator): set connectorGroupId al...

### LINKED ISSUE #729: Set `connectorGroupId` along with `connectorGroup` in SwaggerConnectorGenerator
We're setting `connectorGroup` on the newly generated custom Connector, but not the `connectorGroupId`. We're filtering by `connectorGroupId` and it should be set also.

zregvart:
...ong with connectorGroup

We might get a connectorGroup from ConnectorTemplate, if it is set on
the template we should in addition to connectorGroup set the
connectorGroupId to allow filtering in the UI.

Fixes #729


kahboom:
Awesome! Ty @zregvart

PR #743, opened 2017-12-13
feat: SQL Connectors: adding driver for Mysql

KurtStam:


app/connectors/connectors/sql/sql-connector/pom.xml
```diff
@@ -95,8 +95,12 @@
       <groupId>org.postgresql</groupId>
       <artifactId>postgresql</artifactId>
     </dependency>
+    <dependency>
+      <groupId>mysql</groupId>
+      <artifactId>mysql-connector-java</artifactId>
```
rhuss: where does the version come from ?

KurtStam: @rhuss good question, seems to be coming from spring-boot. We should probably add it to the camel-parent? Same for the postgres driver.

PR #746, opened 2017-12-13
Updated skipCertificateCheck description

dhirajsb:


dhirajsb:
@zregvart got another simple one for you :)

app/rest/dao/src/main/resources/io/syndesis/dao/deployment.json
```diff
@@ -823,7 +823,7 @@
           "javaType" : "java.lang.Boolean",
           "deprecated" : false,
           "secret" : false,
-          "description" : "Skip Certificate check for development environment"
+          "description" : "For convenience in only development environments. Do not set for secure production environments."
```
rhuss: grammar: "For convenience in development only environments".

But maybe: "Skip certificate check. Only for development, don't use it in production as it is highly insecure" ? is a bit shorter ?

dhirajsb: I am going with what @TovaCohen suggested. I am not dogmatic about it. I could change it, if you guys can agree. :)

TovaCohen: The field name is "Skip Certificate Check" so I don't think it should be in the tooltip also. 
A Red Hat style standard is to not use contractions.
The word "only" modifies "development" not "environment" so it should be in front of "development". When we talk we might  say "development environments only" but it is not grammatically correct and should not be in technical documentation.
We could leave out "only" and use:
"For convenience in development environments. Do not set in secure production environments."

rhuss: Of course as a non-native speaker I completely believe you. Never seen that combination before and thought it was a typo. One never stops learning, thanks for clarifying  ... ;-)

Issue #669, opened 2017-12-08
integration with DB connection fails to deploy


Stefan365:
If I create integration with DB connections 
(example, according to: https://drive.google.com/file/d/1eNO13gB4Q5i13SxkC1MFM8AT9vRxLIZm/view
or
https://drive.google.com/file/d/1A6LgMd4Y8SvYw_pjvE4GqlJpGVlZTto2/view)

Integration do not reach 'Active' status (status 'In progress' persists).

oc get pods:
```
skusobnaintegra-1-build       0/1       Completed          0          3m
skusobnaintegra-2-vpt3j       0/1       CrashLoopBackOff   4          2m
```

oc logs skusobnaintegra-1-build:
```
Receiving source from STDIN as archive ...
Pulling image "172.30.1.1:5000/syndesis-project-r/skusobnaintegra:latest" ...
pulling image error : Error: image syndesis-project-r/skusobnaintegra:latest not found

Starting S2I Java Build .....
S2I source build with plain binaries detected
Copying binaries from /tmp/src to /deployments ...
Checking for SpringBoot archive...
... done
 
Pushing image 172.30.1.1:5000/syndesis-project-r/skusobnaintegra:latest ...
Pushed 0/26 layers, 0% complete
Pushed 1/26 layers, 4% complete
Push successful
```

oc logs skusobnaintegra-2-vpt3j:
```
ERROR: Neither $JAVA_MAIN_CLASS nor $JAVA_APP_JAR is set and 0 found in /deployments (1 expected)
```

dsimansk:
There one more interesting log message:
```
2017-12-08 12:39:40.617  INFO [-,,,] 1 --- [pool-3-thread-1] i.s.connector.catalog.ConnectorCatalog   : Downloading Maven GAV: io.syndesis:sql-start-connector:1.2-SNAPSHOT
2017-12-08 12:39:42.709  INFO [-,,,] 1 --- [pool-3-thread-1] i.s.c.integration.online.BaseHandler     : Integration [db-int]: Created project files and starting build
Exception in thread "pool-21-thread-1" java.lang.IllegalArgumentException: Cannot find endpoint with scheme sql-start-connector
	at org.apache.camel.catalog.AbstractCamelCatalog.doAsEndpointUri(AbstractCamelCatalog.java:879)
	at org.apache.camel.catalog.AbstractCamelCatalog.asEndpointUri(AbstractCamelCatalog.java:869)
[...]

PR #747, opened 2017-12-13
bug: integration with DB connection fails to deploy #669

KurtStam:


app/rest/runtime/pom.xml
```diff
@@ -175,11 +175,26 @@
...
+          </dependency>
           <dependency>
             <groupId>io.syndesis</groupId>
             <artifactId>sql-stored-connector</artifactId>
             <version>${syndesis-connectors.version}</version>
           </dependency>
+          <dependency>
+            <groupId>io.syndesis</groupId>
+            <artifactId>sql-stored-start-connector</artifactId>
+            <version>${syndesis-connectors.version}</version>
```
rhuss: why not `${syndesis.version}` ?

Why is ${syndesis-connectors.version} defined anyway ? Should be removed imo and point to syndesis.version

KurtStam: Y I agree, let's open a different issue for that since it's still in a lot of place and I followed suit here to be consistent. https://github.com/syndesisio/syndesis/issues/750

PR #752, opened 2017-12-14
Changed certificate properties type from boolean to string

dhirajsb:


dhirajsb:
Missed this in the last PR :) Copypasta struck again!

PR #758, opened 2017-12-14
fix(connectors): remove scheduler options from ...

zregvart:
...endpoint options

Fixes ##757


rhuss:
@lburgazzoli @zregvart Luca has some ideas around scheduling of connectors, so not sure if this commit would conflict with this ?

lburgazzoli:
@zregvart @rhuss don't think so, it is more for new connectors

PR #765, opened 2017-12-14
Simplify connectors

lburgazzoli:
This PR introduces some new options to the connector definition so we can get rid of artifact per connector and the connector catalog used at project generation time, below a sample of how a new connector definition looks like:

```json
{
  "kind": "connector",
  "data": {
    "id": "twitter",
    "name": "Twitter",
    "description": "Direct and event-driven interaction with Twitter timelines, searches, and messages",
    "icon": "fa-twitter",
    "dependencies": [
      {
        "type": "MAVEN",
        "id": "io.syndesis:camel-component-proxy:@syndesis.version@"
      },
      {
        "type": "MAVEN",
        "id": "org.apache.camel:camel-twitter:@camel.version@"
      }
    ],
    "actions": [
      {
        "name": "Mention",
        "description": "Search for tweets that mention you",
        "id": "io.syndesis:twitter-mention-connector:latest",
        "pattern": "From",
        "actionType": "connector",
        "descriptor": {
          "componentScheme": "twitter-timeline",
          "inputDataShape": {
            "kind": "none"
          },
          "outputDataShape": {
            "kind": "java",
            "type": "twitter4j.Status"
          },
          "configuredProperties": {
            "timelineType": "MENTIONS",
            "delay": "30000"
          },
          "propertyDefinitionSteps": []
        }
      }
    ]
  }
}
```

There are some new "sections":
- `dependencies`: this lists the dependencies your conenctor needs, in this case as the twitter-conenctor do not have any custom code, it is limited to `io.syndesis:camel-component-proxy` which is the underlying component does the magic and `org.apache.camel:camel-twitter` which is the component we want to use as basis for our conenctor. Note that each dependency has a type so beside maven dependencies, you can eventually require some extensions to be included in the integration so you can put some custom code there if needed. Please note that dependencies can also be defined at action level.
- `descriptor.componentScheme` represent the scheme of the component we want to wrap (note that this can be configured also at `conenctor` level), this is **optional** till we have migrated all the connectors and it is the **discriminator** between news and old style connectors.
- `descriptor.configuredProperties` sets the options we want to set by default, it is the equivalent of the default values set on camel conenctor level

If you need to customize the behavior of the conenctor you can list a number of classes that implement `io.syndesis.camel.component.proxy.ComponentProxyCustomizer` on `descriptor`:

```json
"descriptor": {
  "connectorCustomizers": {
    "my.connector.Customizer1",
    "my.connector.Customizer2"
  }
}
```

For complex customization you can implement your own `io.syndesis.camel.component.proxy.ComponentProxyFactory`:
  

```json
"descriptor": {
  "connectorFactory": "my.connector.Factory"
}
```

Of course the two options can be mixed and you can define them on `action`` or on `connector`.

An additional feature you can turn on/of is that you can automatically split the outcome of the connector by configure the `split` node on the `descriptor` level, i.e:

- to automatically split the outcome if camel knows how to split it (i.e. collections), you only need to add an empty `split` node:

```json
"descriptor": {
  "split": { }
}
```

- to split the outcome with a tokenizer:

```json
"descriptor": {
  "split": { 
    "language": "tokenize",
    "expression": ","
  }
}
```

you can use any language known by camel including bean, i.e.

```json
"descriptor": {
  "split": { 
    "language": "bean",
    "expression": "my.own.Splitter::split"
  }
}
```

Notes:
- is backward compatible - or at least it should - so we do not need to upgrade all the connectors now (I did it for twitter)
- I've also updated atlasmap to 1.32.1 in rest/pom.xml not sure if other parts need to be upgraded


lburgazzoli:
retest this please

rhuss:
Just not yet started the review (terrible train wifi), but I already have some questions:

* Would it be possible (and would it make sense) to optimize the naming (things like 'pattern' or 'configuredProperties' could deserve a facelift), or is this required for backwards compatibility ? I mean when the discrimantor is the componentSchema, then we would be free to rename the other parts, too, or ? Might make sense also for to make clear that these are different things.

* If connectors depend on runtime, then you should also adapt the build script which has some highlevel dependency description among modules to calculate the order for building.

well, only two comments so far.

But looks really good, hope I can play with this more when I'm doing my Trello X-Mas POC ;-)

lburgazzoli:
1a. pattern is not something I've introduced :)
1b. configuredProperties is an interface that is already defined as part of the model so I reused it so I can apply/reuse some helpers and logic already in place (i.e. filtering)
2. will do (btw, I hope we can get rid of the runtime bits soon)

rhuss:
Yeah, I know. I already had some hefty discussion around "pattern" (cant find the issue here), but I thought it would be a good occassion to get rid of it ;-)

lburgazzoli:
I’m for adding something like a capabilities object to list what an action can do or we can leverage data shapes, like no inbound data shape means it can only be a ‘from’

zregvart:
@lburgazzoli seems like there is a simple to fix conflict

lburgazzoli:
@chirino @zregvart I think I need one more "approve" as I did push fixes for conflicts

app/rest/dao/src/test/resources/logback-test.xml
```diff
@@ -27,7 +27,8 @@
   </appender>
 
   <root level="INFO">
-    <appender-ref ref="STDOUT" />
+    <!--<appender-ref ref="STDOUT"/>-->
```
jimmidyson: Is this a deliberate change? We're now not seeing test log in build output which is kinda a pain to debug, going to file, even when running tests in IDE.

lburgazzoli: I do not remember to have changed it deliberately but for dao, there's almost no log in any test except the for the Hibernate Validator initialization. The only test that does some logging is ReadApiClientDataTest but it does soon System.out so I guess I did it to suppress Hibernate Validator noises.

PR #769, opened 2017-12-14
Enforced ordering in pods activation

### LINKED ISSUE #767: Env provisioning fails with a rate of 25% on some systems
Some of our pods depend on other to be fully functional.
An example of this is `syndesis-rest` that depends on `syndesis-db` and `syndesis-verifier`

paoloantinori:
fixes #767

### DIFF
--- a/app/deploy/generator/04-syndesis-verifier.yml.mustache
+++ b/app/deploy/generator/04-syndesis-verifier.yml.mustache
@@ -90,6 +90,10 @@
           volumeMounts:
           - name: config-volume
             mountPath: /deployments/config
+        initContainers:
+        - name: init-syndesis-db
+          image: busybox
+          command: ['sh', '-c', 'until nslookup syndesis-db; do echo waiting for service; sleep 2; done;']
         volumes:
         - name: config-volume
           configMap:
--- a/app/deploy/generator/03-syndesis-ui.yml.mustache
+++ b/app/deploy/generator/03-syndesis-ui.yml.mustache
@@ -76,6 +76,13 @@
               memory: 255Mi
             requests:
               memory: 50Mi
+        initContainers:
+        - name: init-syndesis-rest
+          image: busybox
+          command: ['sh', '-c', 'until nslookup syndesis-rest; do echo waiting for service; sleep 2; done;']
+        - name: init-syndesis-atlasmap
+          image: busybox
+          command: ['sh', '-c', 'until nslookup syndesis-atlasmap; do echo waiting for service; sleep 2; done;']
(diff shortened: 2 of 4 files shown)


app/deploy/generator/04-syndesis-verifier.yml.mustache
```diff
@@ -90,6 +90,10 @@
           volumeMounts:
           - name: config-volume
             mountPath: /deployments/config
+        initContainers:
+        - name: init-syndesis-db
+          image: busybox
+          command: ['sh', '-c', 'until nslookup syndesis-db; do echo waiting for service; sleep 2; done;']
```
rhuss: Actually I wouldn't want the verifier to depend on the DB only because a sample connection uses this DB. I see it more likely that we forget to remove this than that someone hits the "validate" button on the UI before the DB starts. And if so, then the validation fails, which is fine (that is what validation is for after all)

Other than that the verifier doesnt use the db at all.

rhuss:
@paoloantinori @KurtStam  could we please revert that PR until we discussed the points above ?

rhuss:
@paoloantinori also, if changing the templates, don't forget to recreate them with `run.sh`

paoloantinori:
Ordering is just  an optimization. Just like we have some live probes
configured with a delay of 180s because we already know that it does not
make sense start checking them from t=0;

Same applies here. Microservices must be reactive, but there is no point in
performing some steps that we know have a specific logical order, out of
order, just to let the recovery mechanism to kick in.

The main reason for this modification is because with I have a failure with
a rate of 25% each time I reprovision the whole environment.

rhuss:
@paoloantinori actually I see to have not a fixed startup order as a way to harden resilience and force the application to be more resilitent. It's better to fix the cause than the syptoms.

Beside this we had several issues with init containers in the past, so that we removed every init container. I would not reintroduce them without a very good reason and with good testing.

'hope you dont mind that I revert the patch with #776 (where I had to regenerate the templates anyway), but I'm happy to discuss this further.

paoloantinori:
As said, this is no different than the other reliability enhancement that we have in the templates, like 
```
          livenessProbe:
            httpGet:
              port: 8080
              path: /api/v1/version
            initialDelaySeconds: 180
```

Anyhow it more about the problem than not about the solution.
On my system I have a failure rate of 25% and I have heard of other spotty failure patterns.

In my tests, startup coordination improved to reduce this error rate to 0.

I'm open to any other approach that can improve the success rate.
I have re-open the issue under a more generic description to encapsulate the problem.

PR #771, opened 2017-12-14
feat: Enable updating connector icons via the REST API

jimmidyson:
/cc @zregvart


app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorIconHandler.java
```diff
@@ -0,0 +1,157 @@
...
+        } catch (IOException e) {
+            throw new IllegalArgumentException("Error while reading multipart request", e);
+        }
+    }
+
+    @GET
+    public Response get() {
+        String connectorIcon = connector.getIcon();
+        if (connectorIcon == null) {
+            return Response.status(javax.ws.rs.core.Response.Status.NOT_FOUND).build();
```
zregvart: nit: perhaps static import NOT_FOUND so we don't have the whole package name here? Or just use `Response.Status.NOT_FOUND` like below

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorIconHandler.java
```diff
@@ -80,8 +82,17 @@ public Connector create(MultipartFormDataInput dataInput) {
                 throw new IllegalArgumentException("Can't find a valid 'file' part in the multipart request");
             }
 
+            MediaType mediaType = filePart.getMediaType();
+            if (!mediaType.getType().equals("image")) {
+                String guessedMediaType = URLConnection.guessContentTypeFromStream(new BufferedInputStream(result));
```
zregvart: Perhaps just add comment here that `URLConnection.guessContentTypeFromStream` will reset the input stream back to start, so our future selves don't trip on this?

jimmidyson: Done.

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorIconHandler.java
jimmidyson: Done. Was a mess with Response import from xml packages...

Issue #560, opened 2017-12-01
Make DeploymentDescriptorIT test great again


zregvart:
https://github.com/syndesisio/syndesis/blob/d6df69488f8e0b6885b1ea82218e3bb88dd22979/rest/dao/src/test/java/io/syndesis/dao/DeploymentDescriptorIT.java#L68

KurtStam:
The test seems to download the snapshots from maven, not the local tree. So even though I think I fixed the data to make the test pass, I won't pass till the snapshots are deployed. @zregvart does that sound right? If so that is kinda a side effect of moving to the mono repo.

zregvart:
Seemed to work OK for me when the connectors are installed in the local Maven repository (it would get them from there), could be just a build order issue? Perhaps we could also troll the source tree and take Camel schema JSONs from there? Part of them are code generated so that would only fail if the code generation was not done.

KurtStam:
@zregvart my connectors have the SQL URL param set to 'required'. My local .m2 directory has a sql-connector-1.2-SNAPSHOT.jar with the sql-connector.json reflecting this. Grabe seems to ignore this and proceeds to download the artifact from maven central (or another listed repo) and gets an older snapshot, with the param required set to false. Test fails. 

I would expect to fix the failure without having to push a release to central.

zregvart:
Perhaps we could take JSON schemas straight from filesystem, now that we're in a single repository that should work.

KurtStam:
@zregvart y, though I would expect it to work with the local .m2 as well.

zregvart:
Yeah, though looking at the JSON files directly could be also good for development experience:
 - change connector JSON
 - run the test (no need to package/install)
 - fix deployment.json...

Though the connector simplifications by @lburgazzoli will remove some of the need for this.

KurtStam:
y. though what happens if I dive directly into the `rest` directory without building the `connectors`. It would fail right? That's bad.

zregvart:
I'm thinking of going directly to filesystem for the Connector JSON files, we would have to try not to use `camel-connector-schema.json` if possible (it's generated/updated by connectors maven plugin), and go straight to `camel-connector.json`. If we end up needing (and we might) perhaps we can invoke the logic from the connector mojo to generate `camel-connector.json`. Could end up being a largish task in the end...

PR #773, opened 2017-12-14
bug: DB Connectors, setting url param to required in the sql-stored-s…

KurtStam:
…tart-connector

@zregvart I want to enable this test again, but I can't fix it without pushing a new release of the connectors. See details in
https://github.com/syndesisio/syndesis/issues/560


zregvart:
@KurtStam is this still a WIP?

KurtStam:
The PR fixes an issue with synchronizing data and can be merged. However this does will not re-enable the test, as things would work till a new snapshot of the connectors is released.

PR #775, opened 2017-12-14
refactor(tools): sprinkle `eval $(minishift oc-env)`

zregvart:
Adds `eval $(minishift oc-env)` to few scripts that are used in
development so we can have the `oc` binary at the same version as the
OpenShift running on minishfit. Another plus is that we don't need to
install `oc` binary separately to startup minishift via `syndesis
minishift --install`

### DIFF
--- a/tools/bin/commands/build
+++ b/tools/bin/commands/build
@@ -170,6 +170,7 @@ order_modules() {
 }
 
 kill_pods() {
+    eval $(minishift oc-env)
     for pod in $@; do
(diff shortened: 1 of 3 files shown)


rhuss:
looks good in general, but I would go for probing it as there are also people who develop with pure 'oc cluster up' and without minishift. So would first check for oc, if not there, then for minishift, and if this is not there as well, bail out.

rhuss:
Sorry, I'm afraid you need to rebase and resolve a conflict.

rhuss:
I'm going to close this for now and add a probing facilitiy together with #1201

PR #777, opened 2017-12-14
API Connector Details: basePath + host

kahboom:


app/ui/src/app/customizations/api-connector/api-connector.service.ts
```diff
@@ -2,7 +2,7 @@ import { Injectable } from '@angular/core';
 import { Observable } from 'rxjs/Observable';
 import { Restangular } from 'ngx-restangular';
 
-import { RESTService } from '../../store/entity/rest.service';
+import { RESTService } from '../../store/entity';
```
deeleman: Hey! Now you can use aliases here like:

`import { RESTService } from '@syndesis/ui/store/entity';`

This way, you can skip refactoring the import paths if you ever need to move `api-connector.service.ts` elsewhere.

kahboom: Cool, even better!

deeleman: In fact, the entity sub-folder features its own barrel, so you can shorten all this as:

`import { RESTService } from '@syndesis/ui/store`

PR #782, opened 2017-12-15
chore(rest): add REST Swagger connector to syndesis-build-helper

zregvart:
This adds the REST Swagger connector as a dependency of the syndesis-build-helper Maven plugin when invoked in runtime module to generate the `/META-INF/camel/camel-meta.json` that includes it.


zregvart:
_ts2f_

PR #804, opened 2017-12-15
feature: Documentation for "syndesis" started

rhuss:


doc/sdh/build/cmd_build.adoc
```diff
@@ -0,0 +1,155 @@
+[[syndesis-build]]
+## `syndesis build`
+
+The main goal of `syndesis` is to build Syndesis.
```
zregvart: Perhaps just to point that the `syndesis` script is in `tools/bin` and that it can be added to $PATH?

rhuss:
since I'd like to get the docs online asap, hope you don't mind that little self-merge. Its mostly about docs, and some minor updates to the buildscript (like a --man options for opening the online HTML docs in the browser).

Happy to discuss and revert changes later on.

🎄🎅🏽🤶🕯🎄

Issue #695, opened 2017-12-11
Syndesis s2i cannot resolve camel version that is built within pipeline


cunningt:
Syndesis s2i does a maven build within the Dockerfile.    It seems that this cannot resolve the camel version that we've built inside the pipeline that is hosted on the ea repository (in this case, 2.21.0.fuse-000007).   

I'm not sure what the correct fix here is - can we add the EA repository as a `<repository/>` here?

`[INFO] F8: Removing intermediate container 94393f01f722
[INFO] F8: Step 4/5 : RUN cd /tmp/artifacts/m2/project  && mvn --batch-mode -Dmaven.repo.local=/tmp/artifacts/m2 package -DskipTests -e -Dfabric8.skip=true  && rm -rf /tmp/artifacts/m2/project  && chgrp -R 0 /tmp/artifacts/m2  && chmod -R g=u /tmp/artifacts/m2
[INFO] F8: ---> Running in fbc828d93927
[INFO] F8: [91m[0m
[INFO] F8: [INFO] Error stacktraces are turned on.
[INFO] F8: [INFO] Scanning for projects...
<cut>
[INFO] F8: [ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] Non-resolvable import POM: Could not find artifact org.apache.camel:camel-spring-boot-dependencies:pom:2.21.0.fuse-000007 in central (https://repo.maven.apache.org/maven2) @ line 39, column 19
[ERROR] 'dependencies.dependency.version' for org.apache.camel:camel-spring-boot-starter:jar is missing. @ line 177, column 17
 @
[INFO] F8: [ERROR] The build could not read 1 project -> [Help 1]
[INFO] F8: org.apache.maven.project.ProjectBuildingException: Some problems were encountered while processing the POMs:
[ERROR] Non-resolvable import POM: Could not find artifact org.apache.camel:camel-spring-boot-dependencies:pom:2.21.0.fuse-000007 in central (https://repo.maven.apache.org/maven2) @ line 39, column 19`

zregvart:
Done with #811

PR #811, opened 2017-12-15
Need to add a custom settings.xml for the s2i image so we can resolve artifacts from SNAPSHOTs and from the ea repository.

cunningt:
This is an attempt to solve #695 - we're getting an error building s2i inside the pipeline once alignments are made because the camel artifacts needed here are on the ea repository, and in the Dockerfile mvn build, there's no repository for ea.

### DIFF
--- a/app/s2i/pom.xml
+++ b/app/s2i/pom.xml
@@ -78,6 +78,31 @@
           </artifactItems>
         </configuration>
       </plugin>
+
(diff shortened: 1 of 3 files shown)


zregvart:
@cunningt seems that `app/s2i/src/main/resources/settings.xml` is missing the license header.

PR #815, opened 2017-12-18
feat: First draft of release script.

iocanel:
This takes care of the maven release part of the release.

Pushing images to dockerhub or other artifacts to their corresponding repositories hasn't been automated yet.

### DIFF
--- a/tools/bin/commands/release
+++ b/tools/bin/commands/release
@@ -1,0 +1,149 @@
+#!/bin/bash
+
+NEXUS_URL="https://oss.sonatype.org"
+NEXUS_STAGING_PLUGIN_VERSION="1.6.7"
+NEXUS_SERVER_ID="oss-sonatype-staging"
+NEXUS_STAGING_REPO_PREFIX="iosyndesis" #This is the parent groupId with dots removed.
+NEXUS_DEFAULT_STAGING_PROFILE_ID="61015c63f807ce" #This this is the id of the staging profile. Just go to the list of staging profiles, click the one u need and grab that id from the url (its not displayed in the actual list).
+
+REPO_ANNOUNCEMENT_PREFIX="OpenedStagingProfile"
+
+release::description() {
+    echo "Perform a release"
+}
+
+release::usage() {
+    cat - <<EOT
+    --dry-run                 Dry run. This will stage and drop a release
+    --release-version         The release version
+    --dev-version             The development version
+    --profiles                Additional profiles to specify
+    --settings-xml            Path to a custom settings.xml to use for the reelase
+
+    --sonatype-username       The sonatype username
+    --sonatype-password       The sonatype password
+
+    --docker-registry         The docker registry
+    --docker-server-id        The docker server id (as defined in settings.xml)
+    --docker-username         The docker username
+    --docker-email            The docker email
+    --docker-password         The docker password
+
+
+EOT
+}
+
+release::run() {
+    source "$(basedir)/commands/build"
+    local dry_run=$(hasflag --dry-run)
+    local release_version=$(readopt --release-version)
+    local dev_version=$(readopt --dev-version)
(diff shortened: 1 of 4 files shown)


tools/bin/commands/release
```diff
@@ -0,0 +1,149 @@
+#!/bin/bash
+
+NEXUS_URL="https://oss.sonatype.org"
+NEXUS_STAGING_PLUGIN_VERSION="1.6.7"
+NEXUS_SERVER_ID="oss-sonatype-staging"
+NEXUS_STAGING_REPO_PREFIX="iosyndesis" #This is the parent groupId with dots removed.
+NEXUS_DEFAULT_STAGING_PROFILE_ID="61015c63f807ce" #This this is the id of the staging profile. Just go to the list of staging profiles, click the one u need and grab that id from the url (its not displayed in the actual list).
```
zregvart: does it make sense to make this default and allow override via command line switch?

tools/bin/commands/release
```diff
@@ -0,0 +1,149 @@
+#!/bin/bash
+
+NEXUS_URL="https://oss.sonatype.org"
+NEXUS_STAGING_PLUGIN_VERSION="1.6.7"
```
zregvart: I think 1.6.8 is latest, should we stick with latest?

tools/bin/commands/release
```diff
@@ -0,0 +1,149 @@
...
+        git push origin master
+        git push orign $tag
+    fi
+}
+
+# ======================================================
+# Staging / Release Helper Functions
+
+latest_release() {
+    git log | grep "prepare release $artifactId-" | head -n 1 | awk -F "$artifactId-" '{print \$2}'
```
zregvart: now that we have a single version is this same as `git tag|tail -1`, also I can't see this being used here, perhaps it can be removed?

tools/bin/commands/release
```diff
@@ -0,0 +1,149 @@
...
+}
+
+open_staging_repository() {
+    local maven_opts=$1
+    local maven_args="org.sonatype.plugins:nexus-staging-maven-plugin:${NEXUS_STAGING_PLUGIN_VERSION}:rc-open -DnexusUrl=${NEXUS_URL} -DserverId=${NEXUS_SERVER_ID} -DstagingProfileId=${NEXUS_DEFAULT_STAGING_PROFILE_ID} -DopenedRepositoryMessageFormat='$REPO_ANNOUNCEMENT_PREFIX:%s' $maven_opts"
+    #Call maven and just grab the last line that contains $REPO_ANNOUNCEMENT_PREFIX discard the line that displays the whole command and get the string that is between `:` and `'`.
+    call_maven "$maven_args" tail -n 1 | grep -v maven | grep $REPO_ANNOUNCEMENT_PREFIX | awk -F ":" '{print $2}' | awk -F "'" '{print $1}'
```
zregvart: Isn't there a `${profile.id}.properties` file and other goals pickup on this, [docs ](https://github.com/sonatype/nexus-maven-plugins/blob/master/staging/maven-plugin/README.md) kinda suggest that is the case, not sure if it holds true

iocanel: My understandings is that this property file is only created by the `deploy` goal and is reused if found by `close`, `release` and `drop`.

Either way, we would still need to do some `cat`, `grep` and `awk` to fech  the staging repository name, so its not something that would change, simplify anything.

tools/bin/commands/release
iocanel: Yes, this needs to go away

tools/bin/commands/release
iocanel: Not sure. I don't think this is something that will ever change.

tools/bin/commands/release
iocanel: Sure, why not.

PR #816, opened 2017-12-18
fix: extension id field replaced by extensionId instead

### LINKED ISSUE #812: Tech extensions: update page displays internal ID
Should we display ID of tech ext. in such a format?

screen shot 2017-12-18 at 08 39 15

deeleman:
Fixes #812 

Warning: For some reason my local data fixture does not yield any values for the `extensionId` or `name` properties, so I could not test this at full.

Perhaps @lburgazzoli might want to test this in his local dev env.

### DIFF
--- a/app/ui/src/app/customizations/customizations.module.ts
+++ b/app/ui/src/app/customizations/customizations.module.ts
@@ -35,7 +35,7 @@ const routes: Routes = [
     ]
   // TODO: Move to its own NgRoutingModule
   }, {
-    path: 'tech-extensions/import/:id',
+    path: 'tech-extensions/import/:id/:extensionId',
     component: TechExtensionImportComponent
   }, {
     path: 'tech-extensions/import',
--- a/app/ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
+++ b/app/ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
@@ -77,6 +77,7 @@ export class TechExtensionImportComponent implements OnInit {
     if (!this.extensionName) {
       // safety net
       this.extensionId = undefined;
+      this.id = undefined;
     }
     const uploadUrl = this.extensionStore.getUploadUrl(this.id);
     this.uploader = new FileUploader({
--- a/app/ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
+++ b/app/ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
@@ -29,6 +29,7 @@ export class TechExtensionImportComponent implements OnInit {
   response: Extension;
   error: FileError;
   importing = false;
+  id: string;
   extensionId: string;
   extensionName: string;
 
@@ -41,9 +42,9 @@ export class TechExtensionImportComponent implements OnInit {
 
   getGenericError() {
     return {
-        level: 'alert alert-danger',
-        message: '<strong>This is not a valid file type.</strong> Try again and specify a .jar file'
-      };
+      level: 'alert alert-danger',
+      message: '<strong>This is not a valid file type.</strong> Try again and specify a .jar file'
+    };
   }
 
   doImport() {
@@ -53,13 +54,13 @@ export class TechExtensionImportComponent implements OnInit {
       this.response = undefined;
       return;
     }
-    this.extensionStore.importExtension(this.response.id).toPromise().then( value => {
+    this.extensionStore.importExtension(this.response.id).toPromise().then(value => {
       this.notificationService.popNotification({
         type: NotificationType.SUCCESS,
         header: 'Imported!',
         message: 'Your technical extension has been imported'
       });
-      const id = this.extensionId || this.response.id;
+      const id = this.id || this.response.id;
       this.router.navigate(['/customizations/tech-extensions', id], { relativeTo: this.route });
     }).catch((reason: any) => {
       this.error = {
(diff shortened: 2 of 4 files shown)


app/ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
```diff
@@ -70,8 +70,8 @@ export class TechExtensionImportComponent implements OnInit {
   }
 
   ngOnInit() {
-    this.extensionId = this.route.snapshot.paramMap.get('id');
-    this.extensionName = this.route.snapshot.paramMap.get('name');
+    this.extensionId = this.route.snapshot.paramMap.get('extensionId');
+    this.extensionName = this.route.snapshot.queryParamMap.get('name');
```
dsimansk: I have a suspicion that this change will break `getUploadUrl`. Shouldn't there be both `id` and `extensionId` values?
https://github.com/syndesisio/syndesis/blob/a3b6962c89cb7c7b96597e87c5c52140591eb5fe/app/ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts#L79

lburgazzoli: I think that too as `extension.id` is what it is used for back-end operations whereas `extension.extensionId` is what UI should be shown to the end user (yeah, confusing)

dsimansk: Testing with this change produces the following error when `Update` is performed on existing Extension.
```json
{
  "errorCode": 400,
  "userMsg": "An error has occurred while trying to process the technical extension. Please, check the input file.",
  "developerMsg": "An error has occurred while trying to process the technical extension. Please, check the input file. Invalid key. Cannot contain ., %, $, #, [, ], /, or ASCII control characters 0-31 or 127. Key: :com.github.lburgazzoli:syndesis-extensions"
}
```

app/ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
```diff
@@ -70,13 +71,14 @@ export class TechExtensionImportComponent implements OnInit {
...
 
   ngOnInit() {
-    this.extensionId = this.route.snapshot.paramMap.get('id');
-    this.extensionName = this.route.snapshot.paramMap.get('name');
+    this.id = this.route.snapshot.paramMap.get('id');
+    this.extensionId = this.route.snapshot.paramMap.get('extensionId');
+    this.extensionName = this.route.snapshot.queryParamMap.get('name');
     if (!this.extensionName) {
       // safety net
       this.extensionId = undefined;
```
dsimansk: Shouldn't be "safety net" also changed to `this.id = undefined;`? Otherwise 👍

deeleman: I do not know. The thing is that I do not understand the logic behind that LOC nor its purpose. I actually do not understand anything from this file and am just following my gut since Angular apps do not operate this way, but I guess you've got a point, since we will want to keep the same previous logic prior to execute `this.extensionStore.getUploadUrl()`. I'll push a new commit...

lburgazzoli:
@dsimansk do you have any spare time to test this ?

dsimansk:
@lburgazzoli well, I've found some :) Anyway it's ok now upload and update-upload works OK. I'll approve this PR.

app/ui/src/app/customizations/tech-extensions/tech-extensions-list.component.html
```diff
@@ -48,7 +48,7 @@ <h1>Technical Extensions</h1>
                      let-index="index">
             <button type="button"
                     class="btn btn-default"
-                    [routerLink]="['import', item.id, { 'name': item.name }]">Update</button>
+                    [routerLink]="['import', item.id, item.extensionId]" [queryParams]="{ name: item.name }">Update</button>
```
dsimansk: Ok, there's still one issue left. For the current implementation `Update` button from list page works, because of this change. But there is another `Update` button on Detail page that isn't routed properly now.

https://github.com/syndesisio/syndesis/blob/98ff655cfa8e50aaa049d1f0c2d70f27d5e1ff08/app/ui/src/app/customizations/tech-extensions/detail/tech-extension-detail.component.html#L40

dsimansk:
@deeleman I think I have figured out the final one-liner. May I push it your fork? :)

deeleman:
go ahead @dsimansk and sorry for not getting back to you earlier. I reassigned @kahboom to this so she could take over, but your contribution will be very welcome.

dsimansk:
@deeleman done. @kahboom should be good to go if you got a minute.

app/ui/src/app/customizations/customizations.module.ts
```diff
@@ -35,7 +35,7 @@ const routes: Routes = [
     ]
   // TODO: Move to its own NgRoutingModule
   }, {
-    path: 'tech-extensions/import/:id',
+    path: 'tech-extensions/import/:id/:extensionId',
```
jimmidyson: I'm not keen on putting the `extensionId` field in the URL when we already have `extension.id` present. I've implemented a different option in #860 which I think does the same as this but without changing the path. Really happy to listen to you @deeleman on which way you think is best to go though, let's pick and commit :)

deeleman: Issue here is that we need both properties: `id` is required for fetching the upload URL and `extensionId` is required for display purposes. We could fetch both by using querying the backend with any of them but this would require more overhead.

I am not familiar with the internals of the associated store, but if you feel that by passing `extensionId` to `this.extensionStore.load(...)` you'll get the information you require, then go ahead.

jimmidyson: I'd personally prefer loading the details from the server if they are not already cached. Considering we also need the name, etc this seems to make sense to me rather than adding more info to the URL.

gashcrumb: Either that or introduce a little service to hold the data required and use the data that we've fetched from the list page.  

The issue is that the update/install page needs the extensionId field to determine whether to show the 'install' or 'update' version of the page, I didn't personally want to make another call just to get this data when we've already fetched it from the list page.

To be honest, I figure this code is going to be reworked once we move this section of the app away from using the restangular store and service.

gashcrumb: meh, I guess the logic would be a lot more consistent if we just fetch and only keep the ID in the URL, so go for it :-)

PR #818, opened 2017-12-19
Update API Connector Model

kahboom:


app/ui/src/app/customizations/api-connector/api-connector.models.ts
```diff
@@ -1,14 +1,11 @@
-import { BaseEntity } from '@syndesis/ui/model';
 import { StringMap, BaseReducerModel } from '@syndesis/ui/platform';
 
-export interface ApiConnector extends BaseEntity {
+export interface ApiConnector  {
```
deeleman: Actually It is okay to extend ApiConnector from BaseEntity, since it is supposed to feature the two properties enforced by BaseEntity’s interface scheme (id and kind).

kahboom: Makes sense!

app/ui/src/app/customizations/api-connector/api-connector-detail.component.ts
```diff
@@ -12,7 +12,7 @@ import { ApiConnectorStore } from './api-connector.store';
 })
 export class ApiConnectorDetailComponent implements OnInit {
   // TODO: Update Restangular/Store setup config to return a properly typed ApiConnectorData instance object
```
deeleman: Perhaps you'll want to remove this comment line, since it is no longer required.

PR #822, opened 2017-12-19
Add AMQ 6.3 template

dsimansk:
@jimmidyson 
I've slimed down AMQ 6.3 product [template](https://github.com/jboss-openshift/application-templates/blob/master/amq/amq63-basic.json) to our needs. I'd appreciate some comments especially to username/password. 
Imho the most convenient is to pre-define for the end user? Also the current limit is 512 MB, that could be decreased as low as 256 MB.


zregvart:
@dhirajsb can you review this?

app/deploy/support/amq63.yml
```diff
@@ -0,0 +1,138 @@
...
+  is ${MQ_USERNAME}/${MQ_PASSWORD}.
+metadata:
+  annotations:
+    description: Application template for JBoss A-MQ brokers. These can be deployed
+      as standalone or in a mesh. This template doesn't feature SSL support.
+    iconClass: icon-amq
+    openshift.io/display-name: Red Hat JBoss A-MQ 6.3 (Ephemeral, no SSL)
+    openshift.io/provider-display-name: Red Hat, Inc.
+    tags: messaging,amq,jboss
+  name: amq63-basic
```
jimmidyson: Should we change the name so this is syndesis specific?

app/deploy/support/amq63.yml
```diff
@@ -0,0 +1,138 @@
...
+    name: MQ_TOPICS
+  - description: User name for standard broker user. It is required for connecting to
+      the broker. If left empty, it will be generated.
+    displayName: A-MQ Username
+    value: amq
+    name: MQ_USERNAME
+  - description: Password for standard broker user. It is required for connecting to
+      the broker. If left empty, it will be generated.
+    displayName: A-MQ Password
+    value: topSecret
```
jimmidyson: Setting a value doesn't match with the description: `If left empty, it will be generated.`

I'm not keen on hard coding passwords like this, but if it's the same as for the DB connector, an example one, then i'm happy enough with it I guess.

app/deploy/support/amq63.yml
```diff
@@ -0,0 +1,138 @@
...
+    value: broker
+  - description: Queue names, separated by commas. These queues will be automatically
+      created when the broker starts. If left empty, queues will be still created dynamically.
+    displayName: Queues
+    name: MQ_QUEUES
+  - description: Topic names, separated by commas. These topics will be automatically
+      created when the broker starts. If left empty, topics will be still created dynamically.
+    displayName: Topics
+    name: MQ_TOPICS
+  - description: User name for standard broker user. It is required for connecting to
```
jimmidyson: Description doesn't match hard coded value, it will never be generated.

app/deploy/support/amq63.yml
dsimansk: I'll bring back the generators. It seems smoother and can be easily overwritten by user.

app/deploy/support/amq63.yml
dsimansk: Of course, the mustache template would be ideal to re-gen for the release, but... Lets stick with `syndesis-amq`?

PR #826, opened 2017-12-19
refactor: cleanup

zregvart:
This adds appropriate `@SuppressWarnings` annotations, removes unused
imports, declares methods that can be used as static and removes
unneeded `else` when the body of `if` returns from method.


zregvart:
Thanks @oscerd seems that I have broken the build, let me fix it first.

zregvart:
@lburgazzoli I had to remove e893bf0c68c86440b4a01ae36f4d222ed93dcd36, obviously we cannot mark `configuredProperties` as `@JsonIgnore` if we wish to load them from JSON files.

@oscerd mind taking another look seems that the build is now OK.

PR #827, opened 2017-12-19
fix(connectors): removes refactored salesforce-model dependency

zregvart:
In the Salesforce connector refactor (#817) renamed dependency `salesforce-model` (replaced by `salesforce-common`) was left behind in some of the connector POMs. This removes those.


oscerd:
+1.

PR #832, opened 2017-12-19
UX: Updated Text for API Client Connectors

sjcox-rh:
Updated designs to include text suggestions from Tova

No design changes, just text. Please reference this [Invision doc](https://redhat.invisionapp.com/share/MAEBH0WCS#/screens/262342791) to see exactly where changes were made. (Please note, this is an old design so the order of steps has changed since this)

@kahboom @TovaCohen, Please review, thanks!


kahboom:
Oh this is for the wizard, let me ping @deeleman so that he's aware as well.

zregvart:
cc @deeleman

PR #837, opened 2017-12-19
feat: New API HTTP Client for UI with composable REST interface

### LINKED ISSUE #834: Build an application-wide ApiHttpClient on top of Angular's HttpClient
With the inception of our new centralized [Store](https://github.com/syndesisio/syndesis/issues/829), and it's application to the [Api Connector Wizard](https://github.com/syndesisio/syndesis/issues/272), Http calls are meant to be decoupled from the components and handled separately. Ideally components will subscribe to store selectors (or named slices of state) and will dispatch actions, that will eventually have an impact on the store, regardless slice of persistence state is updated.

Whatever happens in the middle, it is not the component's concern, nor the store concern either. On the other hand, the interaction between the API and the UI should be constrained to just exchange serialized objects, regardless what components or structs will eventually make use of them. For the sake of consistency, all interactions (either requests and responses) should honor a previously agreed contract schema.

In short:

1. A centralized Store contains an immutable representation of state.
2. Components (which now become stateless dumb viewmodels) subscribe to that state and represent it on their template views (rendering text, displaying loading cues, etc)-
3. When users interact with the template, the component will dispatch actions.
4. An action object is spawned containing information about its type (eg *update connector*) and its payload, whatever it is.
5. Actions are intercepted by reducers, which will take a snapshot of the current state, update the snapshot according to the action requested and replace the current state instance with the new snapshot.
6. The component, as a first-class subscriber to the store wrapping our state, re-updates itself automagically (it now features a minimal implementation BTW, and does not update the state itself nor digests data at all).
7. On a side process, an Effect has listened to that same Action, and has inspected the action payload. It sends then a call to a given method of the related feature provider.
8. The feature provider contains methods that send Http requests to the API and returns the response payload in an observable stream, which are also returned by the feature provider method alike.
9. The Effect which executed that feature provider method now will push the observable response obtained down the RxJS stream and eventually dispatches an action with it.
10. Return to [4], and so on so forth. The application components, all of them subscribed to the store state, will keep updating their views accordingly...

The goal of this issue is to put together an API Http client required in step 8 that will be hooked up with the side Effect model pattern (as an Injectable class by means of Angular's built-in dependency injector) and will allow for sending HTTP calls to the API, with less boilerplate than our current implementation in Restangular.

The API Http client will need to fulfill the following requirements:

- Be able to build and return REST API endpoints by composition.
- Support the different verbs honored by a REST API
- Provide a super-simple interface to reduce boilerplate inside feature providers or the Effect stream itself.
- Provide extended support for typed responses and errors.
- Abstract internally the API config settings (not those need to be configured on a per request basis)

As a bonus, the following would be nice to have:

- Provide extended support for json/multipart upload functionality
- Expose observable properties tracking the progress of Http request/upload

### Proposed implementation

The new `ApiHttpClient` wraps both Angular's native `HttpClient` and our current `ConfigService` and allows to build any kind of API endpoint by composing either object entries or rest parameters. Eg:

```typescript
// For a configured endpoint named 'apiConnectors' as connectors/{connectorId}
this.apiHttpClient.getEndpoint('apiConnectors', { connectorId: 441 });

// or for a configured endpoint named 'apiConnectorbyId' as connectors/{0}
this.apiHttpClient.getEndpoint('apiConnectorbyId', 441);

// will yield /api/v2/connectors/441
```

Getting REST API endpoints is cool but obviously we need to actually submit calls, containing specific payloads, to the API, so we also expose a chainable method to pick a given API endpoint by key and send a specific payload (or not):

```typescript
// For an endpoint named 'apiConnectors' as connectors/{connectorId}/{action} we can
// post a given payload { file: 'foo.png' } and expect an observable Http response 
// typed as ApiHttpResponse
this.apiHttpClient
  .setEndpoint('apiConnectors', { connectorId: 441, action: 'icon' })
  .post<ApiHttpResponse>({ file: 'foo.png' });

// or in case we wanted to GET data for a given API endpoint
// as /api/v2/connectors/{0}/details/{1}
this.apiHttpClient
  .setEndpoint('apiConnectors', 441, 'actions')
  .get<ApiHttpResponse>();

// Both method calls will yield an Observable<ApiHttpResponse> object instance 
// (or whatever type we define in the method generic type signature)
```

Please note that we leverage the same function signature as for `apiHttpClient.getEndpoint()`, where we enter a mandatory string pointing to a API endpoint key available in our settings and then the required parameters to compound up that URL, either as rest arguments or as an object. The execution of `.setEndpoint()` will return an instance object featuring `get()`, `post()`, `put()` and `delete()` methods, plus a `url` string property where we can inspect the target URL, if needed.

Last but not least, the `ApiHttpClient` class also exposes longer versions of the methods above, as shortcuts with the signature `([key, ...params], payload)`. Eg:

```
this.apiHttpClient
  .post<ApiHttpResponse>(['apiConnectors', { connectorId: 441, action: 'icon' }], { file: 'foo.png' });

this.apiHttpClient
  .get<ApiHttpResponse>(['apiConnectors', 441, 'actions']);
``` 

### Multipart upload functionality

Folllowing up with the above, and bearing in mind Angular functionalities, we can easily implement multipart uploads through this interface in our own class:

```typescript
// Will write down in the browser console 1%, 2%, etc once the upload begins
this.apiHttpClient.progressEvent$.subscribe(progress => {
  console.log(`${progress.percentage}%`);
});

// Will trigger an upload of a 'file' object along with { name: 'foo.png' }
// payload in a multipart message posted to /api/v2/connectors/441/icon
this.apiHttpClient
  .setEndpoint('apiConnectors', { connectorId: 441, action: 'icon' })
  .upload<ApiHttpResponse>(file, { name: 'foo.png' });
```

### Colophon

This will not only allow to provide a terse Http interaction with the API for creating new connectors and upload files, but also can be reused for all our other feature modules, current and future, allowing to phase out Restangular once fully replaced and then commit to upgrade to Angular 5.

deeleman:
Fixes #834 

Please refer to the linked issue for a comprehensive explanation on the rationale for this new class. This new internal service is designed for sending Http calls straight to our REST API through a more intuitive and flexible interface, allowing for replacing Restangular in the Api Connectors implementation or even on any other feature module in the future.

Please pay special attention to the new config nodes created at the config files. Ideally, we should go porting all URLs hardcoded in services and components to a centralized location and those files seem the right place until we come up with a more approppiate solution.

This PR welcomes `CoreModule` into our project, which is the module of choice for service providers, according to the official Angular recommendations for avoiding duplicated dependency instances in our injector.

[EDIT]; The Http service now implements multipart upload functionalities and upload progress events.

### DIFF
--- a/app/ui/src/app/app.module.ts
+++ b/app/ui/src/app/app.module.ts
@@ -21,6 +21,7 @@ import { TourNgxBootstrapModule } from 'ngx-tour-ngx-bootstrap';
 import { NotificationModule } from 'patternfly-ng';
 import { DataMapperModule } from '@atlasmap/atlasmap.data.mapper';
 
+import { CoreModule } from './core';
 import { environment } from '../environments/environment';
 import { AppComponent } from './app.component';
 import { AppRoutingModule } from './app.routing';
@@ -82,6 +83,7 @@ export function mapperRestangularProvider(
     FormsModule,
     ReactiveFormsModule,
     HttpModule,
+    CoreModule.forRoot(),
     DynamicFormsCoreModule.forRoot(),
     RestangularModule.forRoot([ConfigService], restangularProviderConfigurer),
     TabsModule.forRoot(),
--- a/app/ui/src/app/core/core.module.ts
+++ b/app/ui/src/app/core/core.module.ts
@@ -20,10 +20,10 @@ export class CoreModule {
       ngModule: CoreModule,
       providers: [
         SYNDESIS_PROVIDERS.ApiHttpProviderService,
-
-        { provide: ApiHttpService, useClass: SYNDESIS_PROVIDERS.ApiHttpProviderService },
-
-        { provide: Window, useValue: window }
+        {
+          provide: ApiHttpService,
+          useClass: SYNDESIS_PROVIDERS.ApiHttpProviderService
+        },
       ]},
     ];
   }
--- a/app/ui/src/app/core/core.module.ts
+++ b/app/ui/src/app/core/core.module.ts
@@ -1,0 +1,30 @@
+import { ModuleWithProviders, NgModule, Optional, SkipSelf } from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { HttpClientModule } from '@angular/common/http';
+
+import { ApiHttpService } from '@syndesis/ui/platform';
+import * as SYNDESIS_PROVIDERS from './providers';
+
+@NgModule({
+  imports: [CommonModule, HttpClientModule],
+})
+export class CoreModule {
+  constructor(@Optional() @SkipSelf() parentModule: CoreModule) {
+    if (parentModule) {
+      throw new Error('Syndesis CoreModule is already loaded. Import it in the AppModule only');
+    }
+  }
+
+  static forRoot(): Array<ModuleWithProviders> {
+    return [{
+      ngModule: CoreModule,
+      providers: [
+        SYNDESIS_PROVIDERS.ApiHttpProviderService,
+
+        { provide: ApiHttpService, useClass: SYNDESIS_PROVIDERS.ApiHttpProviderService },
+
+        { provide: Window, useValue: window }
+      ]},
+    ];
+  }
+}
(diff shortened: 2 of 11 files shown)


deeleman:
The Http service now implements multipart upload functionalities and upload progress events.

PR #838, opened 2017-12-19
feat: Create multipart handler to handle uploading icon at the same time as creating custom connector

### LINKED ISSUE #702: Multipart support for the `POST /api/v1/connectors/custom/{templateId}` endpoint
We need to support upload of Swagger specification, currently set as a `specification` a String `configuredProperty` of ConnectorSettings, and of a custom icon. This is for new API Connector wizard.

See [UX design](https://github.com/syndesisio/syndesis/blob/master/ux/designs/apiconnector/apiconnector.md#upload-swagger).

jimmidyson:
Fixes #702

### DIFF
--- a/app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
+++ b/app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
@@ -71,7 +71,8 @@ public class ConnectorHandler extends BaseHandler implements Lister<Connector>, 
     private final Verifier verifier;
 
     public ConnectorHandler(final DataManager dataMgr, final Verifier verifier, final Credentials credentials, final Inspectors inspectors,
-                            final ClientSideState state, final EncryptionComponent encryptionComponent, final ApplicationContext applicationContext, final IconDataAccessObject iconDao) {
+                            final ClientSideState state, final EncryptionComponent encryptionComponent, final ApplicationContext applicationContext,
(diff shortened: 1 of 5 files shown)


zregvart:
Would this change necessitate the use of multipart or would it scale down and accept _plain_ `application/json`? The reason I'm asking is, if I understood @deeleman correctly, the UI will try to handle interaction with the API in a generic way and send multipart payloads only if there is a file included in the payload. So when the user opts not to include an icon the API should accept the plain version also.

Did I understand this correctly @deeleman?

jimmidyson:
This change doesn't change the existing handler that consumes `application/json`. There is even an integration test for that, so it will handle both plain `application/json` and `multipart/form-data` request, with or without an `icon` form param (so you can send multipart request with just single form param containing connector settings if you want).

deeleman:
We can proceed as you wish:

1. We can send different headers depending on wether there is a file or not
2. We can send the same header (which would be `multipart/form-data`) all the time. This is how [it is configured ATM](https://github.com/syndesisio/syndesis/pull/837/files#diff-4ba4ed964768bcb2a871b86075766461R80).

Just let me know what do you need and I'll update our setup accordingly. 

BTW, we're sending the files inside the form data as properties mapped to the filename. Pls let me know if this is correct (check the file implementation - it is very straightforward and we can tweak it as you wish)

deeleman:
> we're sending the files inside the form data as properties mapped to the filename

Disregard this comment. We'll use whatever name (eg `icon`) you guys need.

PR #840, opened 2017-12-20
feat: Support HTTP Basic Authentication in REST Swagger connector generator

### LINKED ISSUE #436: Implement support for HTTP Basic Authentication

zregvart:
Adds support for HTTP Basic Authentication (not digest!) in REST Swagger connector generator.

Fixes #436

### DIFF
--- a/app/connectors/connectors/rest/rest-swagger-connector/src/main/java/io/syndesis/connector/swagger/SwaggerConnectorComponent.java
+++ b/app/connectors/connectors/rest/rest-swagger-connector/src/main/java/io/syndesis/connector/swagger/SwaggerConnectorComponent.java
@@ -31,6 +31,7 @@ import java.io.OutputStream;
 import java.net.URI;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
@@ -46,7 +47,11 @@ public final class SwaggerConnectorComponent extends DefaultConnectorComponent {
 
     private AuthenticationType authenticationType;
 
+    private String password;
+
     private String specification;
+
+    private String username;
 
     public SwaggerConnectorComponent() {
         this(null);
@@ -64,8 +69,16 @@ public final class SwaggerConnectorComponent extends DefaultConnectorComponent {
         return authenticationType;
     }
 
+    public String getPassword() {
+        return password;
+    }
+
     public String getSpecification() {
         return specification;
+    }
+
+    public String getUsername() {
+        return username;
     }
 
     public void setAccessToken(final String accessToken) {
@@ -76,13 +89,24 @@ public final class SwaggerConnectorComponent extends DefaultConnectorComponent {
         this.authenticationType = authenticationType;
     }
 
+    public void setPassword(final String password) {
+        this.password = password;
+    }
+
     public void setSpecification(final String specification) {
         this.specification = specification;
+    }
+
+    public void setUsername(final String username) {
+        this.username = username;
     }
 
     /* default */ void addAuthenticationHeadersTo(final Map<String, Object> headers) {
         if (authenticationType == AuthenticationType.oauth2) {
             headers.put("Authorization", "Bearer " + accessToken);
+        } else if (authenticationType == AuthenticationType.basic) {
+            final String usernameAndPassword = username + ":" + password;
+            headers.put("Authorization", "Basic " + Base64.getEncoder().encode(usernameAndPassword.getBytes(StandardCharsets.UTF_8)));
         }
     }
 
--- a/app/rest/connector-generator/src/test/java/io/syndesis/connector/generator/swagger/BaseSwaggerGeneratorExampleTest.java
+++ b/app/rest/connector-generator/src/test/java/io/syndesis/connector/generator/swagger/BaseSwaggerGeneratorExampleTest.java
@@ -39,7 +39,7 @@ abstract class BaseSwaggerGeneratorExampleTest extends AbstractSwaggerConnectorT
     private final String specification;
 
     public BaseSwaggerGeneratorExampleTest(final String connectorQualifier, final String name) throws IOException {
-        specification = resource("/swagger/" + name + ".swagger.json");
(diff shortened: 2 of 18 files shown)


app/connectors/connectors/rest/rest-swagger-connector/src/main/java/io/syndesis/connector/swagger/SwaggerConnectorComponent.java
```diff
@@ -76,13 +89,24 @@ public void setAuthenticationType(final AuthenticationType authenticationType) {
...
 
+    public void setUsername(final String username) {
+        this.username = username;
+    }
+
     /* default */ void addAuthenticationHeadersTo(final Map<String, Object> headers) {
         if (authenticationType == AuthenticationType.oauth2) {
             headers.put("Authorization", "Bearer " + accessToken);
+        } else if (authenticationType == AuthenticationType.basic) {
+            final String usernameAndPassword = username + ":" + password;
```
jimmidyson: Do we care if username has `:` in it?

app/rest/connector-generator/src/test/java/io/syndesis/connector/generator/swagger/BaseSwaggerGeneratorExampleTest.java
```diff
@@ -39,7 +39,7 @@
     private final String specification;
 
     public BaseSwaggerGeneratorExampleTest(final String connectorQualifier, final String name) throws IOException {
-        specification = resource("/swagger/" + name + ".swagger.json");
+        specification = resource("/swagger/" + name + ".swagger.json", "/swagger/" + name + ".swagger.yaml");
```
jimmidyson: Supporting YAML definition is separate to this PR, but I'm OK to include it here or have it in separate PR - me likey granular PRs :)

app/rest/connector-generator/src/test/java/io/syndesis/connector/generator/swagger/TestHelper.java
```diff
@@ -56,4 +56,12 @@ private TestHelper() {
         return resource;
     }
 
+    /* default */ static String resource(final String path, final String alternative) throws IOException {
```
jimmidyson: I like this but how about expanding it even more, varargs loop through until find first resource? Slightly more flexible and might help us in future.

app/rest/connector-generator/src/test/java/io/syndesis/connector/generator/swagger/BaseSwaggerGeneratorExampleTest.java
jimmidyson: At least in a separate commit anyway please.

app/rest/connector-generator/src/test/resources/swagger/petstore.param_connector.json
```diff
@@ -172,11 +172,11 @@
   },
   "actions": [
     {
-      "id": "io.syndesis:rest-swagger-connector:1.2-SNAPSHOT:_id_:updatePet",
+      "id": "io.syndesis:rest-swagger-connector:@syndesis-connectors.version@:_id_:updatePet",
```
jimmidyson: Would be good to do this is separate commit please - needed change but easier to track in separate commit as it's unrelated to core change of this PR.

app/rest/connector-generator/src/test/resources/swagger/petstore.unified_connector.json
```diff
@@ -185,11 +185,11 @@
   },
   "actions": [
     {
-      "id": "io.syndesis:rest-swagger-connector:1.2-SNAPSHOT:_id_:updatePet",
+      "id": "io.syndesis:rest-swagger-connector:@syndesis-connectors.version@:_id_:updatePet",
```
jimmidyson: Same comment - separate commit please, well same commit as other similar changes.

app/connectors/connectors/rest/rest-swagger-connector/src/main/java/io/syndesis/connector/swagger/SwaggerConnectorComponent.java
zregvart: Probably, I don't think connector is proper place to make such decisions, it should take the property on face value. Not sure if there is a way to enforce property syntax for configured properties. Created #841

app/rest/connector-generator/src/test/java/io/syndesis/connector/generator/swagger/BaseSwaggerGeneratorExampleTest.java
zregvart: We already parse Swagger files using the Swagger API that supports both JSON and YAML, so this doesn't provide additional functionality. It was just somewhat easier to craft a YAML Swagger file for this test.

app/rest/connector-generator/src/test/java/io/syndesis/connector/generator/swagger/TestHelper.java
zregvart: Let's deal deal with that in the future, right now we need only one alternative (we check for JSON or YAML file present).

app/connectors/connectors/rest/rest-swagger-connector/src/main/java/io/syndesis/connector/swagger/SwaggerConnectorComponent.java
jimmidyson: This code path could be reached from multiple places so being defensive here makes sense to me as this is creating the basic auth header.

However... from https://tools.ietf.org/html/rfc2617#section-2 colon isn't valid in username so i think we can assume here that it is safe.

app/rest/connector-generator/src/test/java/io/syndesis/connector/generator/swagger/BaseSwaggerGeneratorExampleTest.java
jimmidyson: Ah OK.

PR #842, opened 2017-12-20
feat: Enable updating connector and icon via multipart request

jimmidyson:


app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectorHandler.java
```diff
@@ -179,4 +185,66 @@ public Kind resourceKind() {
...
+                String guessedMediaType = URLConnection.guessContentTypeFromStream(new BufferedInputStream(connectorFormData.getIconInputStream()));
+                if (!guessedMediaType.startsWith("image/")) {
+                    throw new IllegalArgumentException("Invalid file contents for an image");
+                }
+                MediaType mediaType = MediaType.valueOf(guessedMediaType);
+                Icon.Builder iconBuilder = new Icon.Builder().mediaType(mediaType.toString());
+
+                Icon icon = getDataManager().create(iconBuilder.build());
+                iconDao.write(icon.getId().get(), connectorFormData.getIconInputStream());
+                connectorToUpdate = new Connector.Builder().createFrom(connectorToUpdate).icon("db:" + icon.getId().get()).build();
```
zregvart: We have an almost identical code block CustomConnectorHandler: https://github.com/syndesisio/syndesis/blob/b2104477226ad73ebe1507ecbcdf106d3e603b94/app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/CustomConnectorHandler.java#L81-L96

would it make sense to pull this in a helper/base class?

zregvart:
@jimmidyson we decided on the standup today it's better to have this merged, and we can always refactor later

PR #857, opened 2017-12-20
fix(ui): display integration name and description

### LINKED ISSUE #430: Integration detail: name is missing/empty
```html
<h1 _ngcontent-c3="">
    <syndesis-editable-text _ngcontent-c3="" _nghost-c5="">
    <!----><!---->

    <!---->
  </syndesis-editable-text>
</h1>
```

TP2
screen shot 2017-11-23 at 08 05 50

Master
screen shot 2017-11-23 at 08 06 05

zregvart:
Seems that changing `*ngIf` to `[ngIf]` helped. Wonder what would be idiomatic Angular solution to this issue?

Fixes #430

### DIFF
--- a/app/ui/src/app/common/editable/editable-tags.component.ts
+++ b/app/ui/src/app/common/editable/editable-tags.component.ts
@@ -4,7 +4,7 @@ import { EditableComponent } from './editable.component';
 @Component({
   selector: 'syndesis-editable-tags',
   template: `
-    <ng-template *ngIf="!editing; else editingTemplate">
+    <ng-template [ngIf]="!editing">
       <em class="text-muted" *ngIf="!value || value.length === 0">
         {{ placeholder }}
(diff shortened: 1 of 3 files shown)


dsimansk:
@zregvart ez pz? I couldn't figure it out, looking in the wild for solution, thanks 👍

deeleman:
@kahboom The *ngIf directive is ALWAYS a structural directive. Property binding is only used for interpolating/render data in the view. Whatever depends on toggling content in and out from the DOM must be handled through structural directives and hence use an asterisk. The original code posted by @zregvart  was right, so here we're basically patching an unexpected behavior that we'll keep dragging around unless we find the source of the issue.

Now this is merged it might not be relevant anymore but let's keep an eye on this kinda behavior in the future so we can improve the use we make of Angular ;)

Issue #855, opened 2017-12-20
Create user doc for Amazon S3 connector


TovaCohen:
The Jira doc issue is here:
https://issues.jboss.org/browse/FUSEDOC-2301

TovaCohen:
Andrea reviewed this doc. It's ready for TP3. 
Can be viewed here: 
https://access.qa.redhat.com/documentation/en-us/red_hat_jboss_fuse/7.0-tp/html-single/integrating_applications_with_ignite/#connecting-to-s3

Issue #854, opened 2017-12-20
Create user doc for uploading an extension


TovaCohen:
The Jira doc issue is here: https://issues.jboss.org/browse/FUSEDOC-2286

TovaCohen:
Doc for extensions for TP3 is in place. 
Preview available here: 
https://access.qa.redhat.com/documentation/en-us/red_hat_jboss_fuse/7.0-tp/html-single/integrating_applications_with_ignite/

Issue #853, opened 2017-12-20
Create user doc for AMQ connector


TovaCohen:
Creating this issue so I can refer to it in a pull request. 
The Jira doc issue is here: https://issues.jboss.org/browse/FUSEDOC-2221 
Not sure I need this GitHub issue plus the Jira issue. We'll see.

TovaCohen:
TP3 doc for AMQ connector is done.

PR #858, opened 2017-12-20
Draft doc for #853, #854 #855

TovaCohen:
This is preliminary doc that needs to be reviewed. But it would be good to merge it so it can be published on the Pantheon site and available to all for review as a whole.


fbolton:
@zregvart I'm also happy for this doc to be merged. It'll be easier to preview that way.

PR #862, opened 2017-12-21
fix: Disable API connector delete button if API connector is in use by any integrations

### LINKED ISSUE #847: Disable Delete button on API Connector list page if uses != 0
This is the simplest way of implementing #712.

cc @kahboom

jimmidyson:
Fixes #847

### DIFF
--- a/app/ui/src/app/customizations/api-connector/api-connector-list.component.html
+++ b/app/ui/src/app/customizations/api-connector/api-connector-list.component.html
@@ -50,7 +50,7 @@
           <ng-template #actionTemplate
                        let-item="item"
                        let-index="index">
-            <button type="button" class="btn btn-default">Delete</button>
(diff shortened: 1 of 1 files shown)


zregvart:
Just wondering if we need to declare `uses` first in the `Connector`:

https://github.com/syndesisio/syndesis/blob/9bd33dd4df1ec420891c4b19423a28c6ca94dd24/app/ui/src/app/model.ts#L132-L142

jimmidyson:
We probably should but it's used already in the HTML all over the place -
 https://github.com/jimmidyson/syndesis/blob/fe9a178aeee8ce542bc8833bb708601d3e38cf1f/app/ui/src/app/customizations/api-connector/api-connector-list.component.html#L44-L53 - so I think it's safe enough.

Issue #637, opened 2017-12-06
UI Changes for ActiveMQ connector per UX design


dhirajsb:
UX design has brought up some changes that need to be made to the UI based. The AMQ icon needs to be added, some action options must be enabled/disabled based on other options.

dhirajsb:
@seanforyou23 there is also a problem with action filtering:

- the `Start Step` in an Integration should only show `Subscribe for messages` action
- the `Finish Step` in an integration should only show `Publish messages` action, or at the very least filter out `Subscribe for messages`

seanforyou23:
Thanks for pointing this out @dhirajsb. I'll use the pattern property on the action to make this determination. Does this only apply for activemq connectors?

zregvart:
Every action regardless of connector type should have a `pattern` property with `"From"` or `"To"` values.

See #253

seanforyou23:
Perfect, thank helps. Thanks @zregvart!

jimmidyson:
@seanforyou23 Any news on this? Is this going to be done for TP3 this week or defer to TP4?

seanforyou23:
@jimmidyson progress has been slower than expected. I believe I'll still be only partially complete by the end of day today. If it's ok, I'd like to defer to TP4. A chance I'll still have it out in time, but it's down to the wire at this point and I just can't say for sure that I'll have it wrapped up in time. Apologies, the dynamic forms plugin we use has gotten the best of me this week.

zregvart:
I'd rather see smaller tasks that we can plan and deliver so let's split this into individual issues for all tasks that we need and close this one. Having this as a open-ended issue doesn't really help us plan or discuss or surface any tech debt issues that need to be addressed for all connectors/actions.

Some of these issues can be addressed by bigger work addressing our tech debt around dynamic forms (in #843).

@dhirajsb @seanforyou23 what do you think?

gashcrumb:
One the PR that fixes #886 lands the UI will support specifying control orders.  Ideally we should drive this from the backend, for the moment though I've added a client-side tweak:

https://github.com/syndesisio/syndesis/pull/919/files#diff-28b4a582e947fd283b192ca194ed84b5R45

We could drive this from the backend by adding a `controls` property alongside the `properties` descriptor, just needs to be an array of strings and sets the order of form controls via property IDs.  Also there's wildcard support, so any form left out just gets appended without a problem.  I think probably for tp4 it'd be good to start adopting this in the descriptors.

[...]

PR #871, opened 2017-12-21
fix(actions-filter): only display relevant actions...

seanforyou23:
... for start and subsequent steps in create integration flow.

Add hooks for filtering actions list in the integrations action select

This helps the following issues:
https://github.com/syndesisio/syndesis/issues/637
https://github.com/syndesisio/syndesis/issues/253

### DIFF
--- a/app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
+++ b/app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
@@ -31,6 +31,7 @@ export class IntegrationsSelectActionComponent extends FlowPage
   actionsSubscription: Subscription;
   position: number;
   step: Step;
+  currentStep: number;
 
   constructor(
     public connectorStore: ConnectorStore,
@@ -101,16 +102,48 @@ export class IntegrationsSelectActionComponent extends FlowPage
     }
   }
 
+  // filters out actions that do not use the 'From' pattern
+  connectorStartActionsFilter(connector: Connector): Connector {
+    const filteredConnectorActions = connector.actions.filter(action => action.pattern === 'From');
+    connector.actions = filteredConnectorActions;
+    return connector;
+  }
+
+  // filters out actions that do not use the 'To' pattern
+  connectorFinishActionsFilter(connector: Connector): Connector {
+    const filteredConnectorActions = connector.actions.filter(action => action.pattern === 'To');
+    connector.actions = filteredConnectorActions;
+    return connector;
+  }
+
   ngOnInit() {
-    this.actions = this.connector
-      .filter(connector => connector !== undefined)
-      .map(connector => connector.actions);
+    // set a local var for current step
+    this.route.params.pluck<Params, string>('position').subscribe(value => this.currentStep = parseInt(value));
+
+    // if it's a start step
+    if (this.currentStep === this.currentFlow.getFirstPosition()) {
+      this.actions = this.connector
+        .filter(connector => connector !== undefined)
+        .map(this.connectorStartActionsFilter)
+        .map(connector => connector.actions);
+    }
+
+    // if it's anything other than a start step
+    if (this.currentStep > this.currentFlow.getFirstPosition()
+      && this.currentStep <= this.currentFlow.getLastPosition()) {
+      this.actions = this.connector
+        .filter(connector => connector !== undefined)
+        .map(this.connectorFinishActionsFilter)
+        .map(connector => connector.actions);
+    }
+
     this.actionsSubscription = this.actions.subscribe(_ =>
       this.currentFlow.events.emit({
         kind: 'integration-action-select',
         position: this.position
       })
     );
+
     this.routeSubscription = this.route.params
       .pluck<Params, string>('position')
       .map((position: string) => {
--- a/app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
+++ b/app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
@@ -102,30 +102,15 @@ export class IntegrationsSelectActionComponent extends FlowPage
     }
   }
 
-  // filters out actions that do not use the 'From' pattern
-  connectorStartActionsFilter(connector: Connector): Connector {
-    const filteredConnectorActions = connector.actions.filter(action => action.pattern === 'From');
-    connector.actions = filteredConnectorActions;
-    return connector;
-  }
-
-  // filters out actions that do not use the 'To' pattern
-  connectorFinishActionsFilter(connector: Connector): Connector {
-    const filteredConnectorActions = connector.actions.filter(action => action.pattern === 'To');
-    connector.actions = filteredConnectorActions;
-    return connector;
-  }
-
   ngOnInit() {
     // set a local var for current step
-    this.route.params.pluck<Params, string>('position').subscribe(value => this.currentStep = parseInt(value));
+    this.currentStep = +this.route.snapshot.paramMap.get('position')
 
     // if it's a start step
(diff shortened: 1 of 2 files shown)


app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
```diff
@@ -101,16 +102,48 @@ export class IntegrationsSelectActionComponent extends FlowPage
     }
   }
 
+  // filters out actions that do not use the 'From' pattern
+  connectorStartActionsFilter(connector: Connector): Connector {
+    const filteredConnectorActions = connector.actions.filter(action => action.pattern === 'From');
+    connector.actions = filteredConnectorActions;
```
zregvart: Would this overwrite the `connector.actions` property so when the connector is used for the second time at a different place in the integration then those actions might be missing?

I might be talking nonsense, I'm TypeScript/Angular newb.

seanforyou23: It's a good question, I'm not 100% sure. What's a good way to test this?

seanforyou23: I don't think it would, because if it did we'd be getting a twice filtered list in the "Finish" step, but we don't. I think that filter is only applied within the context of a single view. Would be good to have someone verify this, or let me know how I can just to be sure.

zregvart:
@deeleman @kahboom @jimmidyson can you help @seanforyou23 here, is that `connector.actions` mutation safe in this context?

app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
deeleman: It would actually overwrite it, I'm afraid. Remember that JavaScript operates by reference, and JavaScript objects are mutable (this is by design on JavaScript: TypeScript and Angular are unrelated to this design principle), so referencing that object to a variable and changing its property will change that property elsewhere in other reference application-wide. 

Correct me if I'm wrong, but apparently what you pursue here is to fetch all connectors that are undefined (including null connectors would be useful too) and then, in the context of `OnInit`, assign the actions from the `connector`object whose `pattern` property is set to `From` to the `actions` property in the component, right?

After quickly assessing the code and borrowing some help from the RxJS operands,  I'd suggest to apply this other approach straight on LOC 124 (hence removing the `connectorStartActionsFilter()` method, which should have been flagged as `private` anyways):

```typescript
if (this.currentStep === this.currentFlow.getFirstPosition()) {
  this.actions = this.connector
    .filter(connector => connector)
    .switchMap(connector => connector.actions.filter(action => action.pattern === 'From'));
}
```

app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
```diff
@@ -101,16 +102,48 @@ export class IntegrationsSelectActionComponent extends FlowPage
...
+        .map(this.connectorStartActionsFilter)
+        .map(connector => connector.actions);
+    }
+
+    // if it's anything other than a start step
+    if (this.currentStep > this.currentFlow.getFirstPosition()
+      && this.currentStep <= this.currentFlow.getLastPosition()) {
+      this.actions = this.connector
+        .filter(connector => connector !== undefined)
+        .map(this.connectorFinishActionsFilter)
```
deeleman: Please refer to my previous comment.

app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
```diff
@@ -101,16 +102,48 @@ export class IntegrationsSelectActionComponent extends FlowPage
...
+    connector.actions = filteredConnectorActions;
+    return connector;
+  }
+
   ngOnInit() {
-    this.actions = this.connector
-      .filter(connector => connector !== undefined)
-      .map(connector => connector.actions);
+    // set a local var for current step
+    this.route.params.pluck<Params, string>('position').subscribe(value => this.currentStep = parseInt(value));
```
deeleman: This is not the Angular way to fetch params from URLs. In first place please remember that route information (URL path fragments, querystring params) is meant to be aynchronous. Everything's a stream in Angular, including routes.

If you feel that users will always load this page or whatever it is from another unrelated component/page, then perhaps you can get away with inspecting the route snapshot in a synchronous fashion, like this:

```typescript
this.currentStep = +this.route.snapshot.paramMap.get('position')
```

This is important since the interface type of the `params` property can change anytime from a StringMap to whatever the Angular team comes up with and your code will break. The `+` will turn the returned string into the number you need. And please, never use `parseInt(..., 10)` without the base index (10 for integers). Otherwise you might confront unexpected results depending on the structure of the parsed string.

If you prefer to stay async for whatever reason, then try this:

```typescript
this.route.paramMap.subscribe(params => this.currentStep = +params.get('position'))
```

app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
```diff
@@ -101,16 +102,48 @@ export class IntegrationsSelectActionComponent extends FlowPage
...
+
     this.actionsSubscription = this.actions.subscribe(_ =>
       this.currentFlow.events.emit({
         kind: 'integration-action-select',
         position: this.position
       })
     );
+
     this.routeSubscription = this.route.params
       .pluck<Params, string>('position')
```
deeleman: Please refer to my previous comment RE route params and the proper syntax to achieve this the Angular way.

app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
deeleman: AS I mentioned elsewhere, you'll probably want to import the `switchMap` operand into the component in order to be able to use it. Try to implicitly use it first, and should the compiler complains, then add this line at the top, right after the angular imports:

`import 'rxjs/add/operator/switchMap';`

Ideally we should have been adding RxJS operands from a centralized vendor file generated as a bundle by either the Angular CLI or Webpack since the project's kickstart, so this should not be necessary.

app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
```diff
@@ -102,22 +103,35 @@ export class IntegrationsSelectActionComponent extends FlowPage
...
+        .filter(connector => connector !== undefined)
+        .switchMap(connector => [connector.actions.filter(action => action.pattern === 'From')]);
+    }
+
+    // if it's anything other than a start step
+    if (this.currentStep > this.currentFlow.getFirstPosition()
+      && this.currentStep <= this.currentFlow.getLastPosition()) {
+      this.actions = this.connector
+        .filter(connector => connector !== undefined)
+        .switchMap(connector => [connector.actions.filter(action => action.pattern === 'To')]);
```
deeleman: Here we're replicating code in two blocks just to check 2 two different types of value for the same variable. This is not DRY. Please take a look into RxJS *lettable operators* as a way to remove code duplication.

Also, you can replace `.filter(connector => connector !== undefined)` by the broader `.filter(connector => !!connector)`

app/ui/src/app/integrations/edit-page/action-select/action-select.component.ts
```diff
@@ -102,22 +103,35 @@ export class IntegrationsSelectActionComponent extends FlowPage
   }
 
   ngOnInit() {
-    this.actions = this.connector
-      .filter(connector => connector !== undefined)
-      .map(connector => connector.actions);
+    // set a local var for current step
```
deeleman: Do we really need to add a comment like this? Code is (and always should be) self-explanatory ;)

PR #873, opened 2017-12-21
fix(ui): stop ng-chart from initiating change storm

### LINKED ISSUE #733: Dashboard: kebab menu in Top 5 Integration table is not working
When clicking on any action nothing happens. It seems that actions aren't initialized as links in HTML code properly.

screen shot 2017-12-13 at 09 01 18

zregvart:
I traced the ever-repainting kebab menu to chart component triggering
change event, seems that ng-chart is adding functions to array and that
in change detection triggers repaint of other directives.

I know I explain this wrong, I'm a Angular newb, though it seems to
work with this change. Perhaps there is a better way of doing this, not
sure.

Fixes #733


jimmidyson:
I have no idea about this... @deeleman help please!

zregvart:
@deeleman can you have a quick look and approve this if it's OK?

PR #877, opened 2017-12-21
feature: fully reactive Swagger fetch component step

deeleman:
This completes STEP ONE of the API Connector Create wizard, including a lot of stuff required by the entire application as a whole:

- Reactive Store, with actions, reducers and side model pattern effects based on the Redux architecture
- New ApiHttpService with multipart upload functionalities (we still do not upload anything tho)
- Forms validated clientside (we should implement URL format validation as well)
- Components are no stateless and any business logic has been removed.
- Lotsa new SASS code as the foundation of our new SASS architecture
- Some new global components aiming to ensure we DRY for once and for all

Still to do: 

- Hook up file upload in the first step with our new ApiHttpService.
- Check how the injected data is finally displayed on the review step (could not manage to go past the errors)
- Implement the dynamic form for updating Auth settings
- Hook up the last step with the Redux container, not that it finally works.


jimmidyson:
Build error:

    ERROR in ng:///workspace/app/ui/src/app/common/button.component.ts.ButtonComponent.html (3,32): Property 'submit' does not exist on type 'ButtonComponent'.

PR #878, opened 2017-12-22
273 API Connector Details Data Fetch

kahboom:
Fetch data using new Http Client.


app/ui/src/app/customizations/api-connector/api-connector-detail.component.ts
```diff
@@ -1,30 +1,33 @@
 import { Component, OnInit } from '@angular/core';
-import { ActivatedRoute, Router } from '@angular/router';
-import { Observable } from 'rxjs/Observable';
+import { ActivatedRoute } from '@angular/router';
 
-import { ApiConnector, ApiConnectorData } from './api-connector.models';
+import { ApiConnectorData } from './api-connector.models';
 import { ApiConnectorStore } from './api-connector.store';
+import { ApiConnectorService } from '@syndesis/ui/customizations/api-connector/api-connector.service';
```
deeleman: Referencing paths like this does not benefit future refactorings. The point for using aliased barrels is precisely not to define paths to the end file. Here you should:

1. Reference the file by its relative path (ultimately both files belong in the same domain): `'./api-connector.service'`
2. Add `api-connector.service` to the index barrel and reference it like `'@syndesis/ui/customizations/api-connector'`. This will work fine **as long as** you don't have to expose (this is, export) this same file outside the `customizations/api-connector`. Otherwise you might wind up creating an undesired circular reference.

app/ui/src/app/customizations/api-connector/api-connector-detail.component.ts
```diff
@@ -1,30 +1,33 @@
...
-      .subscribe(id => this.apiConnectorStore.load(id));
+    const id = this.route.snapshot.paramMap.get('id');
+
+    if (id) {
+      this.apiConnectorService.getApiConnector(id).subscribe(apiConnectorData => {
+        this.apiConnector = apiConnectorData;
+        this.loading = false;
+      });
+    }
   }
```
deeleman: You can even take a more formal _Reactive_ approach for this, by removing the conditional and relying on RxJS operators:

```typescript
  ngOnInit() {
    this.route.paramMap
      .first(params => params.has('id'))
      .map(params => params.get('id'))
      .switchMap(id => this.apiConnectorService.getApiConnector(id))
      .subscribe(apiConnectorData => {
        this.apiConnector = apiConnectorData;
        this.loading = false;
      });
  }
```

app/ui/src/app/customizations/api-connector/api-connector.service.ts
```diff
@@ -4,13 +4,18 @@ import { Restangular } from 'ngx-restangular';
...
+import { ApiHttpService } from '@syndesis/ui/platform';
 
 @Injectable()
 export class ApiConnectorService extends RESTService<ApiConnectorData, ApiConnectors> {
-  constructor(restangular: Restangular) {
+  constructor(restangular: Restangular, private apiHttpService: ApiHttpService) {
     super(restangular.service('connectors?query=connectorGroupId%3Dswagger-connector-template'), 'apiConnector');
   }
 
+  public getApiConnector(id): Observable<ApiConnectorData> {
```
deeleman: This is nice. Two suggestions:

- Class methods and properties are implicitly `public` by default in TypeScript, so you'll want to remove the `public` accessor (and never use it again except for some edge case scenarios).
- Although it doesn't fall in the scope of this task, you might want to replace the Restangular implementation in the `list()` method below by a more lean ApiHttpService implementation like:

```typescript
list(): Observable<ApiConnectors> {
  return this.apiHttpService
    .setEndpointUrl('connectors?query=connectorGroupId%3Dswagger-connector-template')
    .get< ApiConnectors >();
}
``` 

this obviously is not part of this issue but would allow to remove all the Restangular boilerplate from this service file at once. It's up to you ;)

jimmidyson:
Build error:

    ERROR in ng:///workspace/app/ui/src/app/customizations/api-connector/api-connector-detail.component.html (28,3): Property 'apiConnectorData$' does not exist on type 'ApiConnectorDetailComponent'.

deeleman:
@kahboom I submitted a PR to your branch with several enhancements and fixes. Please review it before proceeding to merge this.

PR #885, opened 2018-01-02
fix: don't proceed to the next page until the verifier returns no errors

### LINKED ISSUE #718: Action Configure: Exception Handling
Exception handling in the UI for the DB connector on save.

Exception Handling

Full designs [here](https://github.com/syndesisio/syndesis/blob/cac32465b92d780a3f0440c5b542b1ece489ae99/ux/designs/databaseconnection/databaseconnection.md#create-an-integration-with-database-as-a-start-connection).

cc @KurtStam

gashcrumb:
fixes #718

@KurtStam FYI...

### DIFF
--- a/app/ui/src/app/integrations/edit-page/action-configure/action-configure.component.html
+++ b/app/ui/src/app/integrations/edit-page/action-configure/action-configure.component.html
@@ -19,6 +19,7 @@
     <button type="button"
             class="btn btn-primary"
             (click)="continue()"
+            [disabled]="!formGroup?.valid"
             tourAnchor="integrations.done">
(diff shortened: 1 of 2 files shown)


gashcrumb:
Note we can't actually disable the `Done` button as that's the only way the user can initiate the UI posting the data to the verifier.  We can look at doing async validation on the configuration form in the future but this at least prevents the user from proceeding if there's verifier errors.

KurtStam:
@gashcrumb thanks :). Want me to approve or do you wait for Rachel to take a look?

gashcrumb:
@KurtStam think it's a trivial enough change to be honest...

PR #895, opened 2018-01-03
[chore] adding readme on development tips

KurtStam:
fixes: https://github.com/syndesisio/syndesis/issues/88

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -1,0 +1,62 @@
+# Development Tips
+
+## Start the Rest API w/o security
+
+Steps:
+ 1. Start PostgreSQL from docker
+
+        $ docker run -d --rm -p 5432:5432 -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password -e POSTGRES_DB=syndesis postgres
+
+ 2. Start Syndesis backend (I run io.syndesis.runtime.Application from IDE), with JVM options:
+ 
+         -Dendpoints.test_support.enabled=true -Ddeployment.load-demo-data=false -Dencrypt.key=hello
+
+ 3. Install tinyproxy
+        
+        $ brew install tinyproxy --with-reverse
+
+    Start [tinyproxy](https://tinyproxy.github.io/) from a directory that has `tinyproxy.conf` (below)
+
+        $ tinyproxy -d -c tinyproxy.conf
+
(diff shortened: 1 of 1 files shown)


app/README.md
```diff
@@ -0,0 +1,62 @@
+# Development Tips
+
+## Start the Rest API w/o security
```
gashcrumb: Why is this needed?  Like what reason would I need to go through these steps?  How is this different than the normal development workflow?

app/README.md
```diff
@@ -0,0 +1,62 @@
...
+    the `apiEndpoint` needs to point to tinyproxy (`http://localhost:9090`):
+    
+        "apiBase": "http://localhost:9090"
+     
+     or
+
+        "apiEndpoint": "http://localhost:9090/api/v1",
+        "mapperEndpoint": "http://localhost:9090/mapper/v1",
+
+  5. Start the UI with from `$SYNDESIS/app/ui` (You may need to `npm rebuild node-sass`)
```
gashcrumb: If this is strictly related to the UI should this information maybe be added to `app/ui/README.md` instead?

app/README.md
KurtStam: @gashcrumb It allows for quick turn around times when working on an issue that spans REST and UI. That said we should add the instructions for the `normal` workflow too if you can.

app/README.md
KurtStam: @gashcrumb seems easier to me to keep it all in one place.

app/README.md
gashcrumb: Yep, it's here -> https://github.com/syndesisio/syndesis/blob/master/app/ui/README.md#day-to-day-workflow

app/README.md
gashcrumb: Shouldn't that be the 'developer handbook' then?  Not entirely sure where that gets it's content from...

gashcrumb:
Aha, according to [this PR](https://github.com/syndesisio/syndesis/pull/679) we might want to add this info here somewhere -> https://github.com/syndesisio/syndesis/tree/master/doc/sdh

I'll create an issue to do the same with the readme under UI, that file needs to get cleaned up anyways as there's loads of stale information in there...

gashcrumb:
Let's just merge this for now and I'll add a todo item on my issue regarding integrating this information...

PR #897, opened 2018-01-03
feat(forms): allowing conditional display of model driven form fields

seanforyou23:
A work in progress for adding the ability to conditionally display form fields depending on values of another. Some questions;

1. The order of params passed to super() in syndesis-form-control.component.ts have changed (to match latest ng-dynamic-forms), which causes runtime errors when loading a form. How can I track down what's reliant on the order of these params?
2. The relation property for the ConfiguredConfigurationProperty interface, is there a better type to give this than "any"?

- Updated to latest ng-dynamic-forms
- Added new services for syndesis-form-control.component
- Applied api changes to syndesis-form-control.component
- Added a property to describe conditional display of form controls from a configuration file
- Applied form service tweaks to utilize new relation property
- Updated tests to reflect new api parts


gashcrumb:
> The order of params passed to super() in syndesis-form-control.component.ts have changed (to match latest ng-dynamic-forms), which causes runtime errors when loading a form. How can I track down what's reliant on the order of these params?

At the point when you pass along the injected services to the base class's constructor the order of the parameters are enforced at compile time.  I wonder if the runtime error is more due to other differences in the base class, like this block:

https://github.com/udos86/ng-dynamic-forms/blob/master/packages/ui-bootstrap/src/dynamic-bootstrap-form-control.component.ts#L65-L68

We may need to adjust our version to follow suit which will then also have an affect on the usage of this component (like changing `focus` to `dfFocus` I think?) mostly in this file in spots like:

https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/common/ui-patternfly/syndesis-form-control.component.html#L42-L44

Basically our copy of the original bootstrap form control component is just there to swap out the HTML template so we can adjust a bit here and there to better suit the patternfly CSS, so the component code itself should kinda follow the original file it was based on fairly closely.

gashcrumb:
>The relation property for the ConfiguredConfigurationProperty interface, is there a better type to give this than "any"?

Yeah, looks like it should be a `DynamicFormControlRelationGroup` -> https://github.com/udos86/ng-dynamic-forms/blob/master/packages/core/src/model/misc/dynamic-form-control-relation.model.ts

gashcrumb:
Currently blocked by https://github.com/syndesisio/syndesis/issues/938

gashcrumb:
@seanforyou23 I think we should park this for now and close this PR until we complete the related upgrades to angular and ng-dynsmic-forms.

Issue #898, opened 2018-01-03
Custom API Connector details endpoint must return an 'actionsSummary' property


deeleman:
This is required for #272 and it currently a blocker that prevents us from finalizing this epic.

The UI application performs calls to `/connectors/{id}` in order to fetch the data graph required to render the custom API connector details page. The expected returning message should feature the same response model we get when validating swagger files when creating a new api connector, in order to preserve data consistency with the create wizard calls so the application can reuse the connector summary components.

More specifically, any custom connector based on a swagger file will feature an `actionsSummary` property with a `actionCountByTags` and `totalActions` nested properties. The closest we get from the aforementioned API call is a property named `summary` that does not feature such information either way and we will want to observe the same data model across the board anyways.

Definition of done: Refactor the `/connectors/{id}` endpoint to return a revamped response model where the `summary` property is renamed to `actionsSummary` (or vice versa) and contains the required `actionCountByTags` and `totalActions` properties.

[EDIT]: I just checked that the `/api/{version}/connectors?query=connectorGroupId%3D**{templateId}**` endpoint returns a cursor of API connectors featuring each one the same data scheme as the detail endpoint above. therefore we would need the aforementioned changes to be featured in the output of this endpoint as well.

deeleman:
Apologies in advance but we need to bring this one back from the grave I'm afraid.

Although we get the expected `actionsSummary` property in the response, it comes nested inside the `summary` property (which was meant to be removed) along with other properties (name, description, etc), instead of being a top level property.

So, this is what we currently get when fetching custom API connectors:

```
"description": "blah blah blah",
    "summary": {
        "actionsSummary": {
            "actionCountByTags": {
                ...
```

And this is what we need (`summary` is gone as is replaced by `actionSummary` as is, along with `errors` and `warnings` - fields such as `name`, `description` and `properties` were already present at the top level of the object):

```
"description": "blah blah blah",
    "actionsSummary": {
            "actionCountByTags": {
                ...
        
```

PR #902, opened 2018-01-03
272 - Create wizard complete with multipart upload functionality

deeleman:
This PR finalizes the implementation of the create api connector wizard, plus introduces the lazy loading pattern we should follow from now on to reduce http calls to the backend by leveraging the NGRX store when rendering data on components.

Although not an intrinsic part of the create wizard, the list and detail pages have also been refactored to leverage the Store functionality when rendering data.

The detail page is not complete since we will require #898 to be delivered first, in order to have a consistent data graph for the summary component in the detail view.

I noticed the list page still does not implement delete functionality.


app/ui/src/app/customizations/api-connector/api-connector.models.ts
```diff
@@ -1,98 +1,64 @@
...
-  specification: {
-    kind: string;
-    displayName: string;
-    required: boolean;
-    type: string;
-    javaType: string;
-    description: string
+export interface ApiConnectorData {
+  id?: string;
+  actionsSummary?: {
```
kahboom: Wow this changed quite a bit.

app/ui/src/app/core/providers/api-http-provider.service.ts
```diff
@@ -45,59 +47,74 @@ export class ApiHttpProviderService extends ApiHttpService {
...
+      .catch(error => Observable.throw(this.catchError(error)));
   }
 
   delete<T>(endpoint: string | any[]): Observable<T> {
     const { endpointKey, endpointParams } = this.deconstructEndpointParams(endpoint);
     const url = this.getEndpointUrl(endpointKey, ...endpointParams);
-    return this.httpClient.delete<T>(url);
+    return this.httpClient
+      .delete<T>(url)
+      .catch(error => Observable.throw(this.catchError(error)));
```
kahboom: Interesting that we weren't catching those before..

app/ui/src/app/customizations/api-connector/api-connector.reducer.ts
```diff
@@ -48,10 +61,69 @@ export function apiConnectorReducer(state = initialState, action: any): ApiConne
...
+        hasErrors: false,
+        errors: []
+      };
+    }
+
     case ApiConnectorActions.FETCH: {
       return {
         ...state,
         loading: true,
+        loaded: false,
```
kahboom: We are not returning a default state in case there are no changes?

kahboom: Ah nevermind, I see it here: https://github.com/deeleman/syndesis/blob/0c76af5ce20716b0253338c02615f194a77f34fd/app/ui/src/app/customizations/api-connector/api-connector.reducer.ts#L155 👍

app/ui/src/app/customizations/api-connector/api-connector.service.ts
```diff
@@ -1,23 +1,43 @@
...
       .setEndpointUrl('getApiConnectorDetails', { id })
       .get<ApiConnectorData>();
   }
 
-  list(): Observable<ApiConnectors> {
+  getApiConnectorList(): Observable<ApiConnectors> {
     return this.apiHttpService
       .setEndpointUrl('getApiConnectorList', { template: 'swagger-connector-template'})
-      .get<{ items: ApiConnectors }>() // TODO: Provide a better typing T here
+      .get<{ items: ApiConnectors }>() // TODO: Provide a better <T> typing here
```
kahboom: TODO: No more TODOS! 😭

app/ui/src/app/customizations/customizations.module.ts
```diff
@@ -25,7 +25,8 @@ const routes: Routes = [
       },
       {
         path: 'api-connector',
-        component: ApiConnectorListComponent
+        component: ApiConnectorListComponent,
+        canActivate: [ApiConnectorLazyLoaderGuard]
```
kahboom: Wow we even have our own guard now!

app/ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
```diff
@@ -9,10 +10,10 @@ import {
   ParsedResponseHeaders
 } from 'ng2-file-upload';
 import { NotificationType } from 'patternfly-ng';
-import { Extension } from '../../../model';
-import { ExtensionStore } from '../../../store/extension/extension.store';
-import { NotificationService } from 'app/common/ui-patternfly/notification-service';
-import {Observable} from "rxjs/Observable";
+
+import { Extension } from '@syndesis/ui/model';
```
kahboom: 👍

kahboom:
LGTM, surely we need to resolve the Delete for List Page, but that's a separate PR altogether. I really like where this is headed.

app/ui/src/app/customizations/customizations.module.ts
deeleman: Yep, that guard takes care of checking if the ApiConenctors have been fetched from the API already and will trigger the fetch action if not, which will set the overall state in "loading", so when the user makes it to ANY page requiring that data, the loading state will be already on and the connectors will be rendered once loaded. Next time you return to that same page, no fetch (hence no http call) action will be triggered.

Try it out yourself: Go to the API Connectors page with the Networks tab open. You'll see a http request being made. Go back, then return to the page: No more http requests are triggered, since the information is now fetched from the store.

Try this out now: Go straight to the any connector detail view. Hard reload the page and see how you're fetching ALL the connectors, including the one you need. Now click on the lsit link and see that you're no longer requesting anything form the backend.

Last test: Go to the wizard and create a new connector (use this URL for fetching a awagger file: `http://petstore.swagger.io/v2/swagger.json`). Complete the steps and see how the connectors are re-fetched once the create process is done. Next time you access the list page, you're newly created connector will be already there.

Bottom line is that, by doing this, we:

* Reduce the backend overhead by minimizing HTTP calls
* Pages speed becomes blazing fast since all the information is reactively returned from the store.

Probs some people will thing the above is trivial, but think for a minute this site had millions of uniques a day hammering the API. This kinda architecture ensures site resilience and stability.

app/ui/src/app/customizations/api-connector/api-connector.service.ts
deeleman: TODO: No more technical debt. I agree that this is not ideal, but tracking stuff that needs to be catered (and cannot be fixed at this stage) is a good way to embrace the [The boy Scout Rule](http://programmer.97things.oreilly.com/wiki/index.php/The_Boy_Scout_Rule) and prevent the introduction of more technical debt in the future.

Also, doing this is a good way to prevent other programmers like: "Hey, this needs to be fixed. Can you do it now?"

app/ui/src/app/customizations/api-connector/api-connector.reducer.ts
deeleman: > We are not returning a default state in case there are no changes?

Certainly we do! But changes is not the right term here, Rachel. Our reducer handles ACTIONS. Therefore, we do return a default state (which is the current state snapshot) when the action that executes this reducer is not intercepted by any action handler in the switch/case block.

app/ui/src/app/core/providers/api-http-provider.service.ts
deeleman: Actually the application is not very resilient to backend exceptions, so it breaks half of the times when something goes wrong upon requesting data.

By combining catch operators propagated from the ApiHttpService with catch operators in RxJS streams (usually located at the related Effects middleware class) which funnel the caught error to a a particular XXXX_FAIL action, we ensure the application is resilient and never gets in a bad state that prevents the user from keep browsing.

This also allows to leverage generic error components and display proper error messages in the components with minimum boilerplate, since all instances of state feature a common `hasErrors::Boolean` and `errors::Array<Error>` properties.

app/ui/src/app/customizations/api-connector/api-connector.models.ts
deeleman: These models are far from being convenient and Frontend/Angular-friendly, but this is the best I could get to map the information we need from the backend in a simple and lean fashion.

Backend and Frontend will have to compromise in a better way. Otherwise this app will not scale.

app/ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
deeleman: This was unrelated to this issue, but as I pointed out earlier, any software engineer that worths this title should observe the Boy Scout Rule with no exception, and I found this when linting the PR.

app/ui/src/app/customizations/customizations.module.ts
deeleman: Also, it's worth mentioned the change applied in the overall component controller implementation. Relying on a Store reduces the boilerplate to the max, since components become mere consumers of the store and action dispatchers, not requiring any further logic in their implementation.

app/ui/src/app/customizations/api-connector/api-connector.reducer.ts
kahboom: Correct, I'm terrible with terminology.

app/ui/src/app/customizations/tech-extensions/import/tech-extension-import.component.ts
kahboom: I don't believe in separate PRs for cleanups like this, or they'll never get done. 👍

PR #913, opened 2018-01-04
fix: Add in connector summary details to connector rest api

### LINKED ISSUE #898: Custom API Connector details endpoint must return an 'actionsSummary' property

jimmidyson:
Fixes #898

### DIFF
--- a/app/rest/connector-generator/src/main/java/io/syndesis/connector/generator/swagger/BaseSwaggerConnectorGenerator.java
+++ b/app/rest/connector-generator/src/main/java/io/syndesis/connector/generator/swagger/BaseSwaggerConnectorGenerator.java
@@ -24,7 +24,9 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static java.util.Optional.ofNullable;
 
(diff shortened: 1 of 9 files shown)


jimmidyson:
@syndesisio/backend Could someone please review ASAP	to unblock @deeleman? Thanks!

KurtStam:
@jimmidyson Looking at it now.

jimmidyson:
Thanks!

PR #940, opened 2018-01-05
fix: Properly handle uploaded swagger specification

jimmidyson:
@deeleman This should sort you out.


jimmidyson:
Self-merging to unblock @deeleman

PR #960, opened 2018-01-08
JsonDB: +secondary indexes, +query depth option, +query limit option, +query after option

chirino:
They must be configured up front when you start the jsondb instance but you should get good performance when you look for records in a collection that have a specific field value.


chirino:
@rhuss do you know master will be open for tp4 PRs to make it in?

rhuss:
@chirino Not sure yet about the state of TP3 (at least we do not have a tag yet). Let me ask on the mailing list ...

chirino:
Will manual merge as this has not received a review for a week.  Passes all tests.

PR #974, opened 2018-01-09
chore(docs): update Twitter to Salesforce mapping screenshot

zregvart:
Updates Twitter to Salesforce sample integration documentation with
screenshot of a mapping table with custom Salesforce field
(TwitterScreenName) shown.


TovaCohen:
Hi Zoran, your change to this screen image implies that we are changing the sample integration to have the Twitter/ScreenName_c custom field in Salesforce. Is that correct? This would require the addition of instructions in the tutorial for how to add that custom field. Right? And then I would also need to change all instances of "Title" to ScreenName_c. Please let me know if I correctly understand what I need to do for TP3. Thanks.

zregvart:
> Hi Zoran, your change to this screen image implies that we are changing the sample integration to have the Twitter/ScreenName_c custom field in Salesforce. Is that correct? This would require the addition of instructions in the tutorial for how to add that custom field. Right? And then I would also need to change all instances of "Title" to ScreenName_c. Please let me know if I correctly understand what I need to do for TP3. Thanks.

Oh dear, I totally missed this, thanks for noticing it @TovaCohen. I was still in the mindset of doing a Create or update Salesforce action, and we're not doing that any more. Let me revert this.

TovaCohen:
Oh, okay.  Thanks!

Issue #884, opened 2018-01-02
SF connector issues, Failed to resolve endpoint


tplevko:
I created the Sf->DB integration following the steps described in the [fuse ignite sample integration tutorial](https://access.redhat.com/documentation/en-us/red_hat_jboss_fuse/7.0-tp/html-single/fuse_ignite_sample_integration_tutorial/).
The integration was created successfully, but when I try to create a new lead in salesforce, the integration hits following exception:

```
org.apache.camel.component.salesforce.api.SalesforceException: Unhandled exception: Failed to resolve endpoint: salesforce-on-create-component://getSObject?rawPayload=true&sObjectName=Lead due to: Cannot auto create component: salesforce-on-create-component
```

The whole log file is attached.
[test-2.log](https://github.com/syndesisio/syndesis/files/1597681/test-2.log)

jimmidyson:
@tplevko I don't see that first error in the first log, but I do see:

    java.lang.IllegalArgumentException: clientId must be specified

Could you please create the second issue as a separate issue so we can discuss there rather than mixing up the two different issues? Thanks!

jimmidyson:
Could you possibly export the integration and attach it so I can import it and inspect what's been created?

tplevko:
@jimmidyson, I created separate issue for the second problem: https://github.com/syndesisio/syndesis/issues/888
The exported integration:
[sfdbtest-export.zip](https://github.com/syndesisio/syndesis/files/1599779/sfdbtest-export.zip)

jimmidyson:
@dsimansk @tplevko Could you please run through this again to see if it still an issue? Thanks!

iocanel:
@jimmidyson: FWIW, I don't seem to be able to reproduce.

iocanel:
@jimmidyson: Wrong! I just did!

jimmidyson:
Damnit!

zregvart:
Fixed by #975

PR #963, opened 2018-01-09
fix(salesforce): Use connector component scheme for enrich

zregvart:
Changes the enrich endpoint URI to
`DefaultConnectorComponent::getComponentScheme()`  from
`DefaultConnectorComponent::getComponentName() + "-component"`.

With this the enrich step is using the correct scheme in the
endpoint URI, to fetch the affected Salesforce object the
notification was received for.

Fixes #884

PR #975, opened 2018-01-09
fix (#884): Use the correct component name in the salesforce s…

iocanel:
…treaming connector enricher.


dsimansk:
There's #963 from @zregvart already

PR #978, opened 2018-01-09
fix(project-generator): Filter step skips an index

### LINKED ISSUE #888: Sf connector issue, no endpoint could be found
There is an issue when creating the Tw->Sf integration using the [fuse ignite sample integration tutorial.](https://access.redhat.com/documentation/en-us/red_hat_jboss_fuse/7.0-tp/html-single/fuse_ignite_sample_integration_tutorial/) The error I hit in this case is:

```
org.apache.camel.RuntimeCamelException: org.apache.camel.FailedToCreateRouteException: Failed to create route flow1 at: >>> Filter[{io.syndesis.integration.runtime.util.JsonSimplePredicate@6fff46bf} -> [To[atlas:mapping-step-3.json], To[salesforce-create-sobject-3?sObjectName=Contact]]] <<< in route: Route(flow1)[[From[twitter-timeline-1]] -> [Filter[{io.synde... because of No endpoint could be found for: salesforce-create-sobject-3?sObjectName=Contact, please check your classpath contains the needed Camel component jar.
```

The whole log file is also attached.
[twsftest-2.log](https://github.com/syndesisio/syndesis/files/1599749/twsftest-2.log)

The exported integration is also attached.
[twsftest-export.zip](https://github.com/syndesisio/syndesis/files/1599755/twsftest-export.zip)

zregvart:
When generating `syndesis.yaml` it's important that the index of the
step is in line with the generated `application.properties` in
`/controllers/src/main/java/io/syndesis/controllers/integration/IntegrationSupport.java`

In the case when the filter step is followed by two or more steps the
nested steps are created with the same context, and thus share the same
index.

This tracks the context that needs to be used to maintain the index.

Fixes #888

### DIFF
--- a/app/rest/project-generator/src/main/java/io/syndesis/project/converter/visitor/FilterStepVisitor.java
+++ b/app/rest/project-generator/src/main/java/io/syndesis/project/converter/visitor/FilterStepVisitor.java
@@ -29,11 +29,13 @@ public abstract class FilterStepVisitor implements StepVisitor {
     @Override
     public Collection<io.syndesis.integration.model.steps.Step> visit(StepVisitorContext stepContext) {
         Step step = stepContext.getStep();
+        StepVisitorContext currentContext = stepContext;
         if (step instanceof FilterStep && step.getStepKind().equals(getStepKind())) {
             Filter filter = createFilter((FilterStep) step);
             List<io.syndesis.integration.model.steps.Step> steps = new ArrayList<>();
             while (stepContext.hasNext()) {
-                steps.addAll(visit(stepContext.next()));
+                currentContext = currentContext.next();
+                steps.addAll(visit(currentContext));
             }
             filter.setSteps(steps);
             return Collections.singletonList(filter);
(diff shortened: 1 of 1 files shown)


app/rest/project-generator/src/main/java/io/syndesis/project/converter/visitor/FilterStepVisitor.java
```diff
@@ -29,11 +29,13 @@
     @Override
     public Collection<io.syndesis.integration.model.steps.Step> visit(StepVisitorContext stepContext) {
         Step step = stepContext.getStep();
+        StepVisitorContext currentContext = stepContext;
         if (step instanceof FilterStep && step.getStepKind().equals(getStepKind())) {
             Filter filter = createFilter((FilterStep) step);
             List<io.syndesis.integration.model.steps.Step> steps = new ArrayList<>();
             while (stepContext.hasNext()) {
-                steps.addAll(visit(stepContext.next()));
+                currentContext = currentContext.next();
```
iocanel: How is that any different than the line above?

app/rest/project-generator/src/main/java/io/syndesis/project/converter/visitor/FilterStepVisitor.java
```diff
@@ -43,7 +45,7 @@
         StepVisitorFactory<?> factory = generatorContext.getVisitorFactoryRegistry().get(stepContext.getStep().getStepKind());
         StepVisitor visitor = factory.create();
 
-        return visitor.visit(stepContext);
+        return visitor.visit(currentContext);
```
iocanel: Same.

PR #981, opened 2018-01-09
fix: StepVisitorContext.next() shouldn't mutate the object.

iocanel:


app/rest/project-generator/src/main/java/io/syndesis/project/converter/visitor/StepVisitorContext.java
```diff
@@ -42,7 +43,7 @@ default boolean hasNext() {
     @Override
     default StepVisitorContext next() {
         final int index = getIndex();
-        final Queue<Step> remaining = getRemaining();
+        final Queue<Step> remaining = new ArrayDeque<>(getRemaining());
```
lburgazzoli: Just asking, can we instead use a read-only list and use the index ? Wants to avoid to copy objects at each iteration if possible.

iocanel: Good idea!

lburgazzoli: Allocation then happens only on getRemaining() as we need a sublist but that won't happen all the time

iocanel:
This change is creating a series of issues, that do need time to deal with. So, we better defer it to the future.

Issue #540, opened 2017-11-30
Improve Decoupling of API & UI


kahboom:
## tl;dr
Following from our discussion in the retrospective, there is an inherent problem in the application development lifecycle that I think is directly impeding our ability to keep up with feature requests in a timely fashion. I know this is a LONG description, but please take a moment to read, as I think _everyone_ is affected by this. It's an opportunity to improve the development process. If we don't solve this soon, technical debt will continue to increase and the overall product roadmap will be jeopardized.

The two items to address are 1) decoupling the API & UI; 2) making a more parallel work flow rather than linear and blocking. Solutions proposed are at the bottom. We should collaborate and discuss any potential tradeoffs or alternatives.

## Introduction
At the beginning of the project we made a huge effort to decouple the UI from the backend by creating a Node.js API (now in Java). As features continue to be added to the iPaaS roadmap, we are finding it more and more challenging to develop these features without getting blocked by one or more areas.

##  Theory
When you are spending as much time trying to get the local build of the REST API working properly (to be able to work with real data) as you do actually implementing a feature, it's a problem. It almost feels like we need yet another layer of abstraction to be able to work completely decoupled from the backend. From the UI perspective, which sits kind of right in the middle, it boils down to two factors:

1) separation of concerns and true decoupling
2) timing per sprint to develop these features (linear vs parallel work flow)

## Current Work Process
Here is a rough outline of our work process:

1. Discuss requirements for the sprint.
2. Assign feature owners. Feature owners are usually backend folk (not sure if this is intentional, but seems like a very important consideration we should be consistent with and have a philosophy it supports).
3. Feature owners write a design proposal, which includes the data structure the UI should adhere to.
4. UXD collaborates with feature owners to create designs.
5. UI receives designs, looks at design proposal for the REST API, and begins work.
6. Confusion about whether to mock out data or use existing REST API work ensues. Neither option is smooth, as using existing resources is risky for reasons stated below, and mocked data is great, but when it comes time to wire up the API, there are often huge discrepencies.
7. Scrambling to get things working properly together, whether using maintainable approaches or not, in time for the demo and sprint end.

## Testing the Theory
The problems described earlier manifest in a few different ways that I've noticed impede development:

### Separation of Concerns
[...]

PR #993, opened 2018-01-09
feat: Add a /graphql and /graphiql endpoints that exposes an integrations list.

chirino:


chirino:
@zregvart @deeleman @rhuss here's an initial spike of what it would look like to add a GraphQL interface in front of our normal APIs.

Using this we basically define the view model using GraphQL schema and then the UI defines which fields from the objects they are interested in when doing requests via their queries.

This is here basically just to explore the idea.  WDYT?

app/rest/runtime/src/main/java/io/syndesis/runtime/GraphQLConfiguration.java
```diff
@@ -0,0 +1,82 @@
...
+public class GraphQLConfiguration {
+
+    @Autowired
+    IntegrationHandler integrationHandler;
+
+    @Bean
+    GraphQLSchema schema() throws IOException {
+
+        String schema;
+        try (InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream("schema.graphqls") )  {
```
iocanel: Why do we rely on TCCL when we can just do 

GraphQLConfiguration.class.getClassLoader().getResourceAsStream("shema.graphqls")?

chirino:
Works both ways.

iocanel:
Sure, but isn't it a little bit safer if we didn't rely on TCCL? Frameworks screw with TCCL all the time.

chirino:
We know exactly what framework we are using.  Don't think it's an issue.

deeleman:
I see pros and cons on using GraphQL. Basically it gives the UI a god tool to query the application data graph and get taylored responses that match the UI needs more closely. However its use is justified if the UI wants to perform queries that encompass different related entities and get an unified response model schema. This becomes  specially useful only if the UI is aware of the application data graph tree and is willing to conduct advanced queries amongst different related entities.

If we only plan to use GraphQL to submit Http request to fetch data which is already formalized as an existing model in the backend, and leverage GraphQL to filter out unwanted properties, then we do not need since we're already doing that by means of the client models.

So question here is: Will we want the UI to be aware of the backend application data graph in the future, so it can query the API as if it was querying a database? If so (that would probably save a lot of time to the backend folks by delegating that responsibility on the UI devs), then this is the way to go.

All in all, exploring this idea and implementing this quick POC would be a great way to see its real value. The only caveat is that leveraging GraphQL will require us to implement the ApolloQL library on the frontend and configure it, and we'll need to allocate some time for it.

app/rest/runtime/src/main/java/io/syndesis/runtime/GraphQLConfiguration.java
kahboom: Have to agree with @iocanel here, why risk the headache if we don't need to, and it's easy to change?

jimmidyson:
Just a quick pedantic aside:

> Add a /graphql and /graphiql rest endpoints

GraphQL != REST

kahboom:
Can someone help me understand this a bit better? I can't provide an educated opinion considering I'm not educated on GraphQL and ApolloQL. I was aware of it and what it intends to solve, but judging based on what I'm seeing, it just doesn't seem like something we need. It appears that it could replace our implementation of Web Sockets, and give the UI added flexibility, but it's a big change that seems wouldn't be easy to undo. I'm interested in hearing more about the benefits.

I think this doesn't resolve it in a way I thought it would. What we need is true decoupling (#540 😂 ), and us working with the DB directly (not directly, but querying the API almost as if we were) is the complete opposite of decoupling 😂-- it's just passing through one abstraction layer as if it didn't exist, putting more responsibility on the UI.

I'm all for added responsibility in the UI if it solves an existing problem. We need something that will allow a consistent interface (can our new models currently do that @deeleman ?), an available and common data structure we can work with if the REST API is borked or still undergoing changes we need to continue work in the UI.

Maybe I misunderstood something though, and am happy to be enlightened. :)

chirino:
@deeleman @kahboom so gaphql solves the issue of getting the UI exactly the right amount of data that it need to render the current page without needing to do multiple http requests. 

It also helps formalize the shape of the data models passed back and forth between UI and backend since you can define this with the schemea. 

Keep in mind that queries are just a simple http request, you don’t need extra libs to use it.  Just peek at the test that this PR has in it. 

I think we will always have full rest resource apis, and the UI can choose which access method is better for the job. 

An interesting thing this POC highlights is that we have the ability to transform the DB model to a slightly different shape if needed. For example you can’t use objects as map data container in graphql so we convert configuredProperties to an array.

chirino:
@jimmidyson true.. true

rhuss:
My feelings are mixed as opposed to a decoupled approach with dedicated view objects and a translation layer in between.

Advantages of GraphQL:

* Flexible querying on an existing domain object, giving the UI more freedom to pick only what is required.
* No translation in between, so no extra layer.

Disadvantages: 

* Couples directly to the domain model, so every change in the domain model needs to be communicated and potentially leads to changes in the UI code. That is not necessarily the case with dedicated, use case specific, view objects.
* Potential performance issues as iiur all objects for a query are loaded into memory and then filtered 
(iiur the current implementation right)
* More burden put onto the UI as the ui team needs to know in detail the domain model.
* Does 'API by contract' can work with this as its just a generic API on the domain models? Is a 'contract' which can be verified possible that way?
* It's late in the release cycle, and we, unfortunately, don't have the luxury to do much experimentation and exploration (looking at all the other POCs we need to do).
* Nobody of us has any real word experience with GraphQL (maybe wrong here, though)

What would be the alternatives to improve the API situation? @zregvart you mentioned that contract based, 'pure' REST API development would be an option, too ? Is this something we can introduce incrementally ?

as said at the beginning, i have mixed feelings, so not sure whether we should go down this road. However IMO its mostly the @syndesisio/ui-api which should drive the API decision.

chirino:
@rhuss I disagree it "Couples directly to the domain model".  The view models are defined in the [schema.graphqls](https://github.com/chirino/syndesis/blob/8f68464241072446ea95795033d4e5bf5dff0f97/app/rest/runtime/src/main/resources/schema.graphqls#L30).  Notice that while similar, it's not the exact model we use in the DB.  I probably should have made an even more drastic modification of the model to prove the decoupling point.  There is a translation layer happening, but it's mostly automatic if the UI model is matching the DB model.

The UI folks could be the ones that define the models in the schema.graphqls, and the back end folks figure out how to map to it via per field data fetchers in the [GraphQLConfiguration.java](https://github.com/chirino/syndesis/blob/8f68464241072446ea95795033d4e5bf5dff0f97/app/rest/runtime/src/main/java/io/syndesis/runtime/GraphQLConfiguration.java#L66) file.  Naturally, if parts of the UI module and the DB model match up, then there is less transformation work needed to fulfill the query request.

It's up to the data fetchers to figure out how much to fetch into memory.  They could become more DB aware to filter down what's loaded from the DB.  This is no different from our current rest APIs.  Note that the query operations at the top level can also take parameters to control things like paging.  So I don't think that's  an issue.

rhuss:
@zregvart @chirino @deeleman Is it ok to close this PR for the moment (or park it on a branch in the upstream repo ?) I think we are heading to another direction (swagger based), so I wonder whether we should put more efforts here for the momentn.

chirino:
fine with me.

Issue #380, opened 2017-11-20
Upgrade Angular dependency graph


deeleman:
Our current Angular version is locked down to v4.4.5, while our Angular CLI version is currently on 1.2.8 (which is a bit too outdated).

As part of our effort to keep the dependency graph for Syndesis UI up to date and also leverage the latest industry tools to improve the quality of tis code, time has come to upgrade the Angular libs and its CLI runtime, before it is too late and the cost of upgrading (Angular 6 will see the light on late March 2018) becomes too high.

Accomplishing this EPIC will also set the ground for implementing a solid Redux architecture in the near future, as described in #371

PR #1008, opened 2018-01-10
chore(NG5 upgrade): AtlasMapper dependency upgraded - #380

deeleman:
@kahboom @gashcrumb @seanforyou23 I'd need some help from you guys. Could you pull these changes and help me test that the AtlasMapper instance works as expected? I do not seem to be able to create connections in my dev env.

Creating a data mapper step when creating a new connection or editing an existing connection will bring up the AtlasMapper widget. Checking that the AtlasMapper component can be mounted is enough.


gashcrumb:
@deeleman checking now...

gashcrumb:
@deeleman worked for me without a problem, lemme review now :-)

PR #1015, opened 2018-01-11
Link form labels to form controls

seanforyou23:
This PR adds a critical accessibility enhancement to our model driven forms.

Remove duplicate label elements
Styling enhancement for label hover state

An unrelated issue surfaced during my testing. Navigate to the settings section and in any expand the first two sections in the collapsable list container (twitter/salesforce). Now, click on either label "Client ID" from twitter and salesforce, notice the focus is always placed in the first input. This is because the id's are the same, which they should not be.

I made the keys include a unique number (see changes to oauth-app-form.component.ts below) but I see now that these values are only generated once so that doesn't quite cut it.

What's the appropriate way to make these values unique from one another?

I think the semantics would be better if we used a naming convention that includes the connector type as part of the property name, for example - "twitterClientId"..

### DIFF
--- a/app/ui/src/app/common/ui-patternfly/syndesis-form-control.scss
+++ b/app/ui/src/app/common/ui-patternfly/syndesis-form-control.scss
@@ -9,9 +9,7 @@ textarea:read-only {
   border: none;
 }
 .control-label {
-  &:hover {
-    cursor: pointer;
-  }
+  cursor: pointer;
   > .hint-icon {
     position: absolute;
     top: 3px;
--- a/app/ui/src/app/common/ui-patternfly/syndesis-form-control.scss
+++ b/app/ui/src/app/common/ui-patternfly/syndesis-form-control.scss
@@ -8,6 +8,13 @@ textarea:read-only {
   background-color: transparent;
   border: none;
 }
-.control-label > label {
-  margin-bottom: 0;
+.control-label {
+  &:hover {
+    cursor: pointer;
+  }
+  > .hint-icon {
+    position: absolute;
+    top: 3px;
+    right: 3px;
+  }
 }
--- a/app/ui/src/app/settings/oauth-apps/oauth-app-form.component.ts
+++ b/app/ui/src/app/settings/oauth-apps/oauth-app-form.component.ts
@@ -15,18 +15,27 @@ import {
   DynamicInputModel
 } from '@ng-dynamic-forms/core';
 
-// Default form config object for OAuth client settings
-const OAUTH_APP_FORM_CONFIG = {
-  clientId: {
-    displayName: 'Client ID',
-    type: 'string',
-    description: 'The OAuth client ID setting for the target application'
-  },
-  clientSecret: {
-    displayName: 'Client Secret',
-    type: 'password',
-    description: 'The OAuth client secret value for the target application'
-  }
+const uniqueNum = () => {
+  console.log('uniqueNum called');
+  return Math.random().toString(36).substr(2, 16);
+}
+
+const uniqueIdProp = () => 'clientId-' + uniqueNum();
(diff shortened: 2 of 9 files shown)


app/ui/src/app/settings/oauth-apps/oauth-app-form.component.ts
```diff
@@ -15,18 +15,27 @@ import {
...
-    type: 'string',
-    description: 'The OAuth client ID setting for the target application'
-  },
-  clientSecret: {
-    displayName: 'Client Secret',
-    type: 'password',
-    description: 'The OAuth client secret value for the target application'
-  }
+const uniqueNum = () => {
+  console.log('uniqueNum called');
```
deeleman: we should not push console loggers to production.

app/ui/src/app/settings/oauth-apps/oauth-app-form.component.ts
```diff
@@ -15,18 +15,27 @@ import {
...
+  return Math.random().toString(36).substr(2, 16);
+}
+
+const uniqueIdProp = () => 'clientId-' + uniqueNum();
+const uniqueSecretProp = () => 'clientSecret-' + uniqueNum();
+const idName = uniqueIdProp();
+const secretName = uniqueSecretProp();
```
deeleman: IdName and secretName are assigned just once, so there is no point on leveraging lambda functions (uniqueIdProp and uniqueSecretPop) to generate its unique key. You can assign these values directly or even execute and assign uniqueNum just once and then append its value through a E6 template string to these two (and any future) key on OAUTH_APP-FORM_CONFIG

app/ui/src/app/common/ui-patternfly/syndesis-form-control.scss
```diff
@@ -8,6 +8,13 @@ textarea:read-only {
   background-color: transparent;
   border: none;
 }
-.control-label > label {
-  margin-bottom: 0;
+.control-label {
+  &:hover {
```
deeleman: You can save the hover combinator here, since you’ll want to display a cursor pointer all the time.

app/ui/src/app/settings/oauth-apps/oauth-app-form.component.ts
seanforyou23: Good call on template strings being a better approach. My question, about these values being assigned just a single time, was kind of the main point though. It's problematic because we end up with duplicate ID's for the form fields.

We need a way to make these ID's unique for each app in the list, until then there will be a hiccup in the UI when a user expands multiple of these containers and selects one of the labels - focus will likely be placed into the wrong input.

Maybe that should be addressed as a separate issue? For now I just put it back like it was before.

gashcrumb: Also note for this particular set of forms this may eventually be driven by the backend -> https://github.com/syndesisio/syndesis/issues/240

Also it might be best to handle the actual element IDs from within the form component itself rather than modifying the input?  I think code using the form service shouldn't need to worry about making unique element IDs, it's more a concern for the component that renders the form to the page.

seanforyou23: I added a new property ns "namespace" to the forms service createFormModel. This is used when present to prefix the id of the form controls. I wasn't sure how to handle this directly within the component view template, but as a service param it becomes sort of configurable which is nice. What do you think?

seanforyou23: Updated, changes to the form service reverted and now we're only applying the unique id's at the DOM level. Looking much cleaner, let me know what you guys think!

PR #1022, opened 2018-01-11
Support exporting multiple integrations at a time and exporting the related extensions along with them.

chirino:


app/ui/src/app/store/integration-support.service.ts
```diff
@@ -73,13 +73,17 @@ export class IntegrationSupportService {
...
+    const params = new URLSearchParams();
+    for ( const id of ids ) {
+      params.append('id', id);
+    }
+
+    const url = this.service.one('export.zip').getRestangularUrl();
+    return this.http.get(url, {
+      responseType: ResponseContentType.Blob,
+      params: params
+    });
```
deeleman: The `Http` API was deprecated back in Angular 4.2, and got replaced by the new `HttpClient` class, available from `'@angular/common/http';`. You should use that one instead. The API is almost identical.

All in all, we the UI team are now using the new `ApiHttpService` class, which internally abstracts the API endpoints setup and also provides a lean API to handle multipart requests for uploading files. You can also take advantage of its API here to reduce the overall code footprint for this method.

app/ui/src/app/integrations/list-page/list-page.component.ts
```diff
@@ -62,17 +63,22 @@ export class IntegrationsListPage implements OnInit {
...
     ) => {
-      if (status === 204) {
-        this.notificationService.popNotification({
-          type: NotificationType.SUCCESS,
-          header: 'Imported!',
-          message: 'Your integration has been imported'
-        });
+      if (status === 200) {
+        const changeEvents = JSON.parse(response) as ChangeEvent[];
+        for ( const x of changeEvents ) {
```
deeleman: Unwanted space inside parens. We definitely need to fix the TypeScript Linter. Sorry for that.

app/ui/src/app/store/integration-support.service.ts
```diff
@@ -73,13 +73,17 @@ export class IntegrationSupportService {
...
 
-  exportIntegration(id: string): Observable<Response> {
-    const url =
-      this.config.getSettings().apiEndpoint +
-      '/integrations/' +
-      id +
-      '/export.zip';
-    return this.http.get(url, { responseType: ResponseContentType.Blob });
+  exportIntegration(...ids: string[]): Observable<Response> {
+    const params = new URLSearchParams();
```
deeleman: Same as the `Http` class, the `URLSearchParams` is deprecated. You should use the `HttpParamsOptions` class instead and inject it from the constructor rather than instantiating it directly.

app/ui/src/app/integrations/list-page/list-page.component.ts
chirino: Please fix linter if these types of things bother you that much.  Just commenting on this is consuming time on things that do not add value to the finished product.

app/ui/src/app/store/integration-support.service.ts
chirino: Please submit a PR on to of this one to Convert to newer API.  This PR was not about upgrading usage of older APIs after all.

app/ui/src/app/store/integration-support.service.ts
chirino: Please submit a PR on to of this one to Convert to newer API.  This PR was not about upgrading usage of older APIs after all.

kahboom:
@chirino - Seems there are some merge conflicts before this can be merged in.

chirino:
rebased.

PR #1044, opened 2018-01-12
chore(ui): display API connector icons

### LINKED ISSUE #1029: API connector: uploaded image (icon) doesn't display anywhere
apiconnectoriconupload

zregvart:
If the icon is persisted in the database as opposed to statically served
from assets, point the icon url to the API endpoint for fetching the
icon.

Fixes #1029

### DIFF
--- a/app/ui/src/app/connections/detail-page/info.component.ts
+++ b/app/ui/src/app/connections/detail-page/info.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, Output, EventEmitter } from '@angular/core';
+import { Component, Input, Output, EventEmitter, SimpleChanges } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { Observable } from 'rxjs/Observable';
 import { Subscription } from 'rxjs/Subscription';
@@ -6,6 +6,8 @@ import { Subscription } from 'rxjs/Subscription';
 import { Connection } from '../../model';
 import { ConnectionService } from '../../store/connection/connection.service';
 import { ConnectionConfigurationService } from '../common/configuration/configuration.service';
+import { ConfigService } from '@syndesis/ui/config.service';
+import { OnChanges } from '@angular/core/src/metadata/lifecycle_hooks';
 
 @Component({
   selector: 'syndesis-connection-detail-info',
@@ -13,7 +15,7 @@ import { ConnectionConfigurationService } from '../common/configuration/configur
     <h1>
       <dl class="dl-horizontal">
         <dt>
-          <img src="../../../assets/icons/{{ connection.connectorId }}.connection.png" height="46" width="46">
+          <img src="{{ icon }}" height="46" width="46">
         </dt>
         <dd>
           <syndesis-editable-text [value]="connection.name"
@@ -51,18 +53,31 @@ import { ConnectionConfigurationService } from '../common/configuration/configur
   `
   ]
 })
-export class ConnectionDetailInfoComponent {
+export class ConnectionDetailInfoComponent implements OnChanges {
+  readonly apiEndpoint: any;
   @Input() connection: Connection;
   @Output() updated = new EventEmitter<Connection>();
+  icon: String;
 
   constructor(
     private connectionService: ConnectionService,
-    private configurationService: ConnectionConfigurationService
-  ) {}
+    private configurationService: ConnectionConfigurationService,
+    private config: ConfigService
+  ) {
+    this.apiEndpoint = config.getSettings().apiEndpoint;
+  }
 
   onAttributeUpdated(attr: string, value) {
     this.connection[attr] = value;
     this.updated.emit(this.connection);
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    if (this.connection.icon.startsWith('db:')) {
+      this.icon = `${this.apiEndpoint}/connectors/${this.connection.connectorId || this.connection.id}/icon`;
+    } else {
+      this.icon = `../../../assets/icons/${this.connection.connectorId || this.connection.id}.connection.png`;
+    }
   }
 
   /* tslint:disable semicolon */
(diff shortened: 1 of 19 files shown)


zregvart:
I'm sure that this can be done more elegantly, I just hammered at it until I got both the static and uploaded icons to show up. If no serious issues arise from this can we merge this for TP3 and refactor in TP4?

zregvart:
Seems that it might have broken the new integration flow.

zregvart:
False alarm, didn't have the Verifier running, so integration editor broke.

gashcrumb:
> I'm sure that this can be done more elegantly, I just hammered at it until I got both the static and uploaded icons to show up. If no serious issues arise from this can we merge this for TP3 and refactor in TP4?

I dunno, there's some copy/paste in there and everything, are you gonna do this refactoring?

app/ui/src/app/connections/detail-page/info.component.ts
```diff
@@ -1,19 +1,21 @@
...
+import { Component, Input, Output, EventEmitter, SimpleChanges } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { Observable } from 'rxjs/Observable';
 import { Subscription } from 'rxjs/Subscription';
 
 import { Connection } from '../../model';
 import { ConnectionService } from '../../store/connection/connection.service';
 import { ConnectionConfigurationService } from '../common/configuration/configuration.service';
+import { ConfigService } from '@syndesis/ui/config.service';
+import { OnChanges } from '@angular/core/src/metadata/lifecycle_hooks';
```
gashcrumb: You can just import this from `@angular/core`, no?

app/ui/src/app/connections/detail-page/info.component.ts
```diff
@@ -1,19 +1,21 @@
...
 
 @Component({
   selector: 'syndesis-connection-detail-info',
   template: `
     <h1>
       <dl class="dl-horizontal">
         <dt>
-          <img src="../../../assets/icons/{{ connection.connectorId }}.connection.png" height="46" width="46">
+          <img src="{{ icon }}" height="46" width="46">
```
gashcrumb: Shouldn't this be `[src]="icon"`?

Should really change that height/width to be set via stylesheet too, though I see the original code didn't bother to either.

app/ui/src/app/connections/detail-page/info.component.ts
```diff
@@ -51,20 +53,33 @@ import { ConnectionConfigurationService } from '../common/configuration/configur
...
+  ) {
+    this.apiEndpoint = config.getSettings().apiEndpoint;
+  }
 
   onAttributeUpdated(attr: string, value) {
     this.connection[attr] = value;
     this.updated.emit(this.connection);
   }
 
+  ngOnChanges(changes: SimpleChanges) {
```
gashcrumb: This executes lots of times, and I suspect we don't need to evaluate this code that often.  This should probably go in `ngOnInit` where the `connection` object should also be available and set.

app/ui/src/app/connections/list/list.component.ts
```diff
@@ -102,4 +107,10 @@ export class ConnectionsListComponent implements OnInit {
     return connection.id === this.selectedId;
   }
 
+  connectionIcon(connection: Connection) {
+    if (connection.icon.startsWith('db:')) {
+      return Observable.of(`${this.apiEndpoint}/connectors/${connection.connectorId || connection.id}/icon`);
```
gashcrumb: Why is this async?  :-S

app/ui/src/app/customizations/api-connector/api-connector-detail/api-connector-detail.component.ts
```diff
@@ -34,9 +40,17 @@ export class ApiConnectorDetailComponent implements OnInit {
       .switchMap(([id, apiConnectors]: [string, ApiConnectors]) =>
         apiConnectors.filter(apiConnector => apiConnector.id == id)
       );
+
+    this.icon$ = this.apiConnectorData$.mergeMap(connector => {
+      if (connector.icon.startsWith('db:')) {
```
gashcrumb: Have seen this `if` statement a lot.  Maybe there's a better place to check this.  My suggestion would be to enhance the service and maybe just reset/add the icon property in there, as with this current code the person refactoring it has to go through all these files and discover/update these if statements.

app/ui/src/app/dashboard/connections.component.html
```diff
@@ -32,7 +32,7 @@ <h1>Create a Connection</h1>
 
               <div class="card-pf-top-element">
                 <span class="card-pf-icon-large image-icon">
-                <img src="../../../assets/icons/{{ connection.connectorId }}.connection.png">
+                <img src="{{ connectionIcon(connection) }}">
```
gashcrumb: Yeah, definitely work looking into updating the store.  Like we could add an override to this class:

https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/store/connection/connection.service.ts

That basically sets the icon, you'd override this function:

https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/store/entity/entity.store.ts#L60-L75

and probably modify this too:

https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/store/entity/entity.store.ts#L77-L86

to update/enhance the `icon` property on the connection/connector.  Probably need to do it for both the connection and the connector stores.

app/ui/src/app/dashboard/connections.component.ts
```diff
@@ -28,6 +30,12 @@ export class DashboardConnectionsComponent implements OnInit {
 
   truncateTrail = '…';
 
+  constructor(
+    private config: ConfigService
```
gashcrumb: Yeah, if we put this in the store then you don't need all these `ConfigService` injections etc.

zregvart:
Impressive work @deeleman 🥇

PR #1048, opened 2018-01-12
Updates to correct links and number of sublevels

TovaCohen:
This is ready to merge. It fixes some broken links and limits the number of sublevels of topics. Thanks in advance!


TovaCohen:
Thanks Stan!

Issue #852, opened 2017-12-20
Shared modules


lburgazzoli:
For new connectors we should have some modules available to everyone:

- rest/core
- rest/model 
- verifier interfaces

As example, having access to verifier interfaces, would allow to provide connector specific validations along with the connectors so a connector do not need to touch verifier code.

Issue #786, opened 2017-12-15
Self containing connectors


lburgazzoli:
Beside creating the connector, adding it to syndesis requires some manual activity on different modules:

- dao
- runtime

We should try to reduce this by:

- include connector definition in connector's artifact
- include mapper inspections in connector's definition, i.e.:
```json
"outputDataShape": {
  "kind": "java",
  "type": "twitter4j.Status",
  "specification": "<put the precomputed shape here>"
}
```

Then when building the runtime we can process connectors bom and extract all the information needed by the runtime from connector's artifact.

igarashitm:
Unless syndesis run `mvn install` for the custom connector, we would need external jar support for the Java inspection - atlasmap/atlasmap#29

lburgazzoli:
The initial plan is not to change the behavior of the runtime, so the runtime will include an additional jar with connectors metadata (including inspections) but yeah, supporting custom jar is something we should have,

PR #1054, opened 2018-01-14
Implements Logging POC

chirino:


lburgazzoli:
I think there is a small issue in this PR about module dependencies as rest/core is now a dependency of runtime which is a dependency of rest/project-generator so there is a deadlock while building top level modules. I've partially addressed this in #852 as part of #786 but I think we should have a better distributions of modules, rest is too big IMHO (and runtime won't be needed anymore by project-generation when we'll get rid of yaml)

app/rest/controllers/src/main/java/io/syndesis/controllers/logs/LogsController.java
```diff
@@ -0,0 +1,379 @@
...
+ * This class tracks pod logs and ingests them into our DB.
+ */
+@Service
+@ConditionalOnProperty(value = "controllers.logs.enabled", havingValue = "true", matchIfMissing = false)
+public class LogsController implements Closeable {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LogsController.class);
+
+    private final DBI dbi;
+    private final NamespacedOpenShiftClient client;
```
iocanel: We could live here with just an OpenshiftClient here (as in 99% of the cases).

The NamespacedClient variation only provides the ability to obtain a new instance of the client that defaults to an other namespace and I don't think anyone is really using that.

app/rest/controllers/src/main/java/io/syndesis/controllers/logs/LogsController.java
```diff
@@ -0,0 +1,379 @@
...
+     *
+     * We do this instead of using the watchLog() feature of the k8s client lib because it really sucks due to:
+     *  1. You can't configure the timestamps option or the sinceTime option.  Need to resume log downloads.
+     *  2. It seems to need extra threads..
+     *  3. It might be hiding some of the http failure conditions.
+     *
+     */
+    protected void getLogs(String podName, Callback handler, String sinceTime) throws IOException {
+        try {
+            PodOperationsImpl pod = (PodOperationsImpl) client.pods().withName(podName);
```
iocanel: This goes a long way, while it could just use something like:

    WatchLog w = client.pods().withName(podName).withTimestamps().watchLog();
    processLogStream(w.getOutput);

The part where you go down to the okhttp client, implement a custom callback etc (seems to me uneeded).

app/rest/controllers/src/main/java/io/syndesis/controllers/logs/LogsController.java
chirino: Might as well go down to KubernetesClient actually.

chirino:
@lburgazzoli just need the KeyGenerator in the runtime.  Perhaps we just move it to a runtime module.

app/runtime/runtime/src/main/java/io/syndesis/integration/runtime/LoggingInterceptStrategy.java
```diff
@@ -0,0 +1,134 @@
...
+            return target;
+        }
+        return exchange -> {
+
+            String tx = exchange.getExchangeId();
+            String step = definition.getId();
+
+            System.out.println(new LogLine(tx, step, "begin", null));
+            target.process(exchange);
+            System.out.println(new LogLine(tx, step, "end", null));
```
lburgazzoli: Wondering if we could avoid creating this LogLine object and instead have a static utility method that prints directly to thread local writer (or somethign similar)

lburgazzoli:
@chirino that's fine with me

app/runtime/runtime/src/main/java/io/syndesis/integration/runtime/LoggingInterceptStrategy.java
chirino: Is this for performance reasons?

app/rest/controllers/src/main/java/io/syndesis/controllers/logs/LogsController.java
chirino: See method comment:
```java
    /*
     * Feeds the logs of the given podName to the callback handler for processing.
     *
     * We do this instead of using the watchLog() feature of the k8s client lib because it really sucks due to:
     *  1. You can't configure the timestamps option or the sinceTime option.  Need to resume log downloads.
     *  2. It seems to need extra threads..
     *  3. It might be hiding some of the http failure conditions.
     *
     */
```

Has watchLog() resolved all of those?

app/runtime/runtime/src/main/java/io/syndesis/integration/runtime/LoggingInterceptStrategy.java
lburgazzoli: Just to avoid useless temporary object to be allocated and destroyed

chirino:
rebased PR

app/runtime/runtime/src/main/java/io/syndesis/integration/runtime/LoggingInterceptStrategy.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+public class LoggingInterceptStrategy implements InterceptStrategy {
+
+
+    public UuidGenerator getUuidGenerator() {
+        return new UuidGenerator() {
+            @Override
+            public String generateUuid() {
+                // Generates always incrementing lexically sortable unique keys
+                return KeyGenerator.createKey();
```
rhuss: why is 'lexicable sortable keys' required ?  Why not a plain java.util.UUID ?

app/runtime/runtime/src/main/java/io/syndesis/integration/runtime/LoggingInterceptStrategy.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public static class LogLine {
+        public final String id;
+        public final String tx;
+        public final String step;
+        public final String action;
+        public final String message;
+
+        public LogLine(String tx, String step, String action, String message) {
```
rhuss: Shouldn't a log line alos contain a date ?

app/runtime/pom.xml
```diff
@@ -91,6 +91,12 @@
   <dependencyManagement>
     <dependencies>
       <!-- Syndesis Runtime Modules -->
+      <dependency>
+        <groupId>io.syndesis</groupId>
+        <artifactId>core</artifactId>
```
rhuss: Could we avoid this dependency ? Seems to be a bit much a dependency only for the KeyGenerator. Would be ok for me even to copy the class instead (doesn't seem to be changed often).

app/runtime/runtime/src/main/java/io/syndesis/integration/runtime/LoggingInterceptStrategy.java
chirino: this key is used to insert and order transactions in the db without needing a 2nd index based on time.

app/runtime/runtime/src/main/java/io/syndesis/integration/runtime/LoggingInterceptStrategy.java
chirino: generated keys have an embedded timestamp in them.   See:
https://github.com/chirino/syndesis/blob/52137b104e872ee3ee30a116b703362fb68b7188/app/core/core/src/main/java/io/syndesis/core/KeyGenerator.java#L88

app/runtime/pom.xml
chirino: What's in core that seems so risky to share?

chirino:
rebased off master

app/runtime/pom.xml
rhuss: Not really risky, but a question how to interprete "common": Common to the backend infra (rest, verifier, ...) or common to the integration runtime, or both ? Architectural-wise I really would like to keep that separated so that it can be evolve differently. Remember the runtime and rest will always share the same latest version number, so when you change KeyGenerator because you need it differently in the backend you have to change the integration runtime, too (and have to test whether it works for both use cases). Also, it would then require that your infra changes potentially lead to an integration upgrade maybe only because of this change (which would not be needed otherwise), if one want to keep the runtime on the latest versions (which you want, because you do not want to track different versions for your runtime).

At the end, its the same argument as they have for Microservices. The less dependencies the better. My POV is, that dependencies are inevitable if required. But they are always heavyweight (see the use case above), so we should not introduce dependencies lightly. Especially not for accessing a single small sized class.

app/ui/src/app/integrations/components/logs.component.scss
```diff
@@ -0,0 +1,13 @@
+@import 'patternfly-sass/assets/stylesheets/patternfly/color-variables';
```
rhuss: This PR contains UI releated code, so it would be awesome if @syndesisio/ui-api could have a look, too, whether we all obey the coding rules for the UI // @chirino @deeleman

chirino: Is there a feature flag framework for the UI yet?  Would I could enable that section added just for POC.

chirino:
Codacy clean up, rebased, and squashed

app/ui/src/app/integrations/components/logs.component.scss
chirino: log component only shown now if feature flag is enabled.

app/ui/src/app/integrations/detail-page/detail.component.html
```diff
@@ -180,7 +180,7 @@ <h1>
                   (click)="requestAction('delete', i)">Delete Integration</button>
         </dd>
       </dl>
-      <dl class="dl-horizontal">
+      <dl class="dl-horizontal" *ngIf="config.getSettings('features', 'logging', false)">
```
deeleman: Method calls should never be injected in HTML templates. Otherwise the Angular detection system will execute that method call every few milliseconds until the component is unmounted. You'd better execute this just once at the `OnInit` stage and make it accessible from a property in the component class.

app/ui/src/app/integrations/detail-page/detail.component.ts
```diff
@@ -135,7 +136,8 @@ export class IntegrationsDetailComponent extends IntegrationViewBase
     public notificationService: NotificationService,
     public modalService: ModalService,
     public application: ApplicationRef,
-    integrationSupportService: IntegrationSupportService
+    integrationSupportService: IntegrationSupportService,
+    public config: ConfigService
```
deeleman: Hint for injecting dependencies: In order to ensure immutability, Angular prefers injection tokens to be flagged as `private`. The `public` accessor is only used when the component inherits from another component, which we should rarely do.

app/ui/src/app/integrations/components/logs.component.scss
deeleman: All vendor libraries and related imports are now abstracted in an aliased token for our convenience and ease future refactors, so you can use `@import 'syndesis-sass';` instead.

app/runtime/runtime/src/main/java/io/syndesis/integration/runtime/LoggingInterceptStrategy.java
chirino: Lets optimize once we get a performance testing framework.

app/runtime/pom.xml
chirino: I don't mind making copies of stuff it it will make it easier to reason about, but this does not seem like we are running into these issues yet.  Could we fork the copy later if we need to?

app/ui/src/app/integrations/components/logs.component.scss
chirino: actually this file should be empty.  I don't actually plan to implement any styling for the POC.

app/ui/src/app/integrations/detail-page/detail.component.ts
chirino: thx. will do.

chirino: actually.. I just tied to remove the public accessor, fails due the html template not being able to access the config variable.

app/ui/src/app/integrations/components/logs.component.scss
deeleman: Then I'd suggest to remove the SASS file and remove the `styleUrls: [...]` line from the @Component decorator settings as well. The lesser files we have, the better.

app/runtime/pom.xml
rhuss: ok

rhuss:
@deeleman ok to merge from an UI POV ?

app/ui/src/app/integrations/detail-page/detail.component.ts
deeleman: Yep, private members cannot be accessed from the template by Webpack (which fuels the Angular CLI), but this is connected with a comment I posted aside regarding injecting object instances and executing methods from within the template. 

If you believe this is just a temporary POC, then leave it as it is and we'll refactor this when the time is right. Otherwise I'd advocate to inject this as a private class member and assign the output of the `config.getSettings(...)` method to a class property injected as a binding in the template, either at the constructor or either at the `onInit`lifecycle hook. that is the Angular way or approaching this.

app/rest/controllers/pom.xml
```diff
@@ -42,6 +42,11 @@
       <artifactId>dao</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>io.syndesis</groupId>
+      <artifactId>jsondb</artifactId>
+    </dependency>
+
```
rhuss: Why is this dependency required here ? jsondb must only be accessed from the DAO package, never directly.

app/core/core/src/main/java/io/syndesis/core/KeyGenerator.java
```diff
@@ -76,4 +79,39 @@ public static String createKey() {
             return randomVal;
         });
     }
+
+    /**
+     * Used to extract the time information that is encoded to each
+     * generated key.
+     */
+    public static long getKeyTimeMillis(String key) throws IOException {
```
rhuss: a feaure only needed for the logging, not for creating DB keys.(yet another smell that these should be two classes).

chirino: "two classes' of KeyGenerator?

chirino: All these generated keys have had an embedded timestamp component.  The keys have not changed. We are just adding methods that further expose USEFUL information.  When statements like these are made could you please also state scenarios where doing these things harms something plz.

app/rest/controllers/pom.xml
chirino: I'm not sure we need a DAO right now.  I don't plan on implementing any alternative access methods that would require a common abstraction.

app/ui/src/app/integrations/detail-page/detail.component.ts
chirino: Got it. Will fix up in a bit.

app/rest/controllers/pom.xml
rhuss: Sorry, disagree here. If we couple jsondb like this directly as dependency the DAO abstraction is useless anyway. 

If you find an other way to avoid the dependency here, that's probably fine for me, too. Only 'rest/runtime' is allowed to depend on jsondb.

app/core/core/src/main/java/io/syndesis/core/KeyGenerator.java
rhuss: I said it's a _smell_, not an issue. You didn't need the timestamp for the original usage, but you do need it now. Maybe because its so useful to embedd semantic information into a key, maybe you drink the Kool-Aid and want to add more information like this to the key later. But changing this class on behalf of one of the dependants, requires to update the other dependent, too. Or to be extra careful to not break stuff (if we still remember that there are other dependents).

Not an issue now, just a smell. But why adding this smell when it is so easy to avoid it in the first place ? That's my opinion (i.e. that runtime code should not share code with backend code, as long as we don't hit a hard wall). Separation of concerns is a commonly accepted principle. 

Don't want to argue further, as said previously, I'm somewhat ok with that.

rhuss:
jsondb works quite well for us for our backend with actually very little data (~ some 10 MBs or so). But when we talk about logging, we are talking about a huge amount of data, which is constantly written.

* How does this affect the usual operation which reads from the same table at the same time ? (e.g. how is locking handled. Is the whole table locked ? row based locking ?)
* Shouldn't it be in a separate store from our application state to be resistant against logs running amok ?
* Do we have some performance numbers to ensure that this is not the case ?
* Is Postgres good enough to hold so much data in a single table (without the need to e.g. repartition) ?

I'm by no means a database expert (as I assume no one of us is), and I really wonder whether this should be our core business or whether we should not delegate this to one of the (os) products of a billion dollar industry. It always better to finger point them than us ;-)

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/logs/LogsHandler.java
```diff
@@ -0,0 +1,142 @@
...
+        public String id;
+        public long at;
+        public String action;
+        public String messages;
+    }
+
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path(value = "/{integration}")
```
KurtStam: I prefer integrationId

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/logs/LogsHandler.java
```diff
@@ -0,0 +1,142 @@
...
+    public List<APITxLogEntry> getLogs(@PathParam("integration") String integrationId) throws IOException {
```
KurtStam: integrationId

chirino:
Squashed and rebased.  Refactor modules to so that controller + jaxrs resources associated with this db logging impl, is in a single module which are enabled by a feature flag.  Rest API is now better defined.

chirino:
@rhuss tests are green now.  Could we get this merged in so that we can provide builds to QE that they can pound and measure actual performance impacts?

chirino:
Opened up issue https://github.com/syndesisio/syndesis-qe/issues/276 to measure performance impacts.

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/LogsResource.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+        if( limit > 1000 ) {
+            limit = 1000; // max out to 1000 per request.
+        }
+
+        GetOptions options = new GetOptions()
+            .order(GetOptions.Order.DESC) // reverse the order since we want most recent exchanges first.
+            .after(from).limit(limit); // allow paging
+
+        byte[] data = jsondb.getAsByteArray(path, options);
```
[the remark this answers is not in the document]
chirino: So create 2 modules?

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/LogsResource.java
```diff
@@ -0,0 +1,187 @@
...
+import static io.syndesis.rest.dblogging.jaxrs.JsonNodeSupport.getString;
+import static io.syndesis.rest.dblogging.jaxrs.JsonNodeSupport.removeBoolean;
+import static io.syndesis.rest.dblogging.jaxrs.JsonNodeSupport.removeLong;
+import static io.syndesis.rest.dblogging.jaxrs.JsonNodeSupport.removeString;
+
+/**
+ * Provides a JAXR interface to query stored integration logging from the DB.
+ */
+@Component
+public class LogsResource {
```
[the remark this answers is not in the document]
chirino: Can do.

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/controller/KubernetesSupport.java
```diff
@@ -0,0 +1,100 @@
...
+        try {
+            PodOperationsImpl pod = (PodOperationsImpl) client.pods().withName(podName);
+            String url = pod.getResourceUrl().toString() + "/log?pretty=false&follow=true&timestamps=true";
+            if (sinceTime != null) {
+                url += "&sinceTime=" + URLEncoder.encode(sinceTime, "UTF-8");
+            }
+            // URL url = new URL(URLUtils.join());
+            Request request = new Request.Builder().url(new URL(url)).get().build();
+            OkHttpClient clone = okHttpClient.newBuilder().readTimeout(0, TimeUnit.MILLISECONDS).build();
+            clone.newCall(request).enqueue(new Callback() {
```
rhuss: How is a restart handled when a connection is dropped ?

chirino: That should be handled by this catch:
https://github.com/chirino/syndesis/blob/028131b4abf79f7d30e000f8bd26969ebc1db1c9/app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/controller/PodLogHandler.java#L101

Basically the pod log handler handles any errors by re-scheduling a run of the handler until the handler is shutdown.

rhuss:
There's a test error:

```
INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.7.0:testCompile (default-testCompile) on project db-logging: Compilation failure
[ERROR] /workspace/app/rest/db-logging/src/test/java/io/syndesis/rest/dblogging/controller/LogsControllerTest.java:[76,50] cannot find symbol
[ERROR]   symbol:   variable REVISION_ID_ANNOTATION
[ERROR]   location: interface io.syndesis.openshift.OpenShiftService
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
```

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/LogResource.java
```diff
@@ -0,0 +1,187 @@
...
+    Set<String> EVENT_FIELDS_SKIP_LIST = new HashSet<>(Arrays.asList("id", "at", "logts"));
+
+    public LogResource(final JsonDB jsondb) {
+        this.jsondb = jsondb;
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path(value = "/{integrationId}")
+    public List<Exchange> getLogs(
```
rhuss: Sorry, not sure whether I already asked: Why is this Jax RS endpoint when it is delegate to from the LogHandler in the rest api ?

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ */
+public class Exchange {
+
+    private String id;
+    private String logts;
+    private Long at;
+    private String pod;
+    private String ver;
```
rhuss: version ?

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
```diff
@@ -0,0 +1,107 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+
+import java.util.List;
+
+/**
+ */
+public class Exchange {
+
+    private String id;
+    private String logts;
```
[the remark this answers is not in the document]
rhuss: ANd if it is a timestamp, why it is not a long ?

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+package io.syndesis.rest.dblogging.jaxrs.model;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+import java.util.List;
+
+/**
+ */
+public class Exchange {
```
rhuss: Exchange sound a bit very generic. Isn't it a 'ExchangeLog' ? (just to differentiate it from a Camel exchange maybe ...)

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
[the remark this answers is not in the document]
rhuss: Just saw that it is a date in universal format. Would't call it timestamp as this is usually associated with epoch time. What's about 'logDate' ?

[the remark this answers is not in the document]
rhuss: or just 'date'

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/LogsResource.java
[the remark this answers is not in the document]
chirino: To open up the possibility of unit testing the via jaxrs in the the db-logging module.

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
chirino: Was trying to keep the amount of bits stored lower using abbreviations :)

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private String id;
+    private String logts;
+    private Long at;
```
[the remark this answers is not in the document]
chirino: at is the time exchange first started.

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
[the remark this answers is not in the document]
chirino: So this maps to the log time stamp in the K8S API. Using the pod id + logts, you can ask the kubernetes api for logs at that point in time were this exchange started.  So while it is a timestamp, it's only intended to be used as foreign key into K8s API.

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
chirino: I don't think we have any danger of both this class and a camel Exchange class being both used from the same java class :)

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/controller/DurationConverter.java
```diff
@@ -0,0 +1,79 @@
...
+            Duration d = check.apply(source);
+            if( d!=null ) {
+                return d;
+            }
+        }
+        return Duration.parse(source);
+    }
+
+    private static Pattern patternFor(String units) {
+        return Pattern.compile("^\\s*(\\d+)\\s*("+units+")?\\s*$", Pattern.CASE_INSENSITIVE);
```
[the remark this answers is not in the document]
chirino: When no unit is given, we fallback to parsing ISO-8601 format which is awful:  see:
https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/controller/PodLogState.java
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.rest.dblogging.controller;
+
+import io.syndesis.core.Json;
+
+public class PodLogState {
+    public String time;
```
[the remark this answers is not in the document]
chirino: This is just the simplest class that we can use Json.mapper() against. :) I can likely turn this into an inner class impl detail.

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/controller/PodLogHandler.java
```diff
@@ -0,0 +1,235 @@
...
+                    processLogStream(is);
+                } finally {
+                    is.close();
+                }
+            } catch (InterruptedException | IOException e) {
+                LOG.info("Failure occurred while processing controller for pod: {}", podName, e);
+                logsController.schedule(this::run, 5, TimeUnit.SECONDS);
+            }
+        } else {
+            logsController.schedule(this::run, 5, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
chirino: So we resume reading logs from where we last processed them by using the sinceTime query option in the K8s API.  See: https://github.com/chirino/syndesis/blob/c48cc1a06c7de5c3f6f7a6de3b1c32f45906ff4c/app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/controller/KubernetesSupport.java#L67

In theory we should process very few dup log entries.
Even if we did re-process log entries, the way log entries are structured in jsondb, makes those log updates idempotent.

chirino:
I'm should probably wait for  @lburgazzoli PR to get in around removing the flow model, since it likely impacts the Logging bits that were added to the runtime in this PR.  @lburgazzoli is that going in soon?

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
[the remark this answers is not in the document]
rhuss: Since K8s API refers to this at 'time' why not use time here, too ? It's even one char less. Or use 'timestamp'. 'log' is redundant.

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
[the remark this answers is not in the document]
rhuss: 'start' ?

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
rhuss: come on, you are optimizing at that level really ;-) ?

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
rhuss: Still I don't think it  _is) an 'Exchange' but a log about an exchange, so 'ExchangeLog' would fit much better IMO.

rhuss: But feel free to reject this comment like the others, too.

chirino: How about ExchangeActivity??

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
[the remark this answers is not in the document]
chirino: time sounds good.

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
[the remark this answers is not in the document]
chirino: The exchange occurred 'at' :)  plus its fewer chars.

app/rest/db-logging/src/main/java/io/syndesis/rest/dblogging/jaxrs/model/Exchange.java
chirino: yeah :)

PR #1060, opened 2018-01-15
Labels, consistant icon sizing and OAuth settings icon for API connectors

zregvart:
Sets labes for OAuth as in the documentation, defines three new CSS clases (syn-icon-[large|medium|small]) and uses those instead of width/height img attributes, and adds API connector icon to OAuth settings list.


zregvart:
thanks @gashcrumb !

PR #1049, opened 2018-01-12
fix: best to check the values and make sure there's some

gashcrumb:
fixes #1047 

Also gets rid of `Object.keys()` to make @kahboom happy :-)

@dsimansk FYI

PR #1063, opened 2018-01-15
UI, fix step page for tech ext

### LINKED ISSUE #1062: UI - problem with tech ext custom properties
There is a NPE if an extension defines a custom property but doesn't provide any default value for it.

Fix ready

paoloantinori:
fixes #1062


gashcrumb:
There's a PR open for this already -> https://github.com/syndesisio/syndesis/pull/1049

PR #1070, opened 2018-01-15
Remove runtime dependencies on core

lburgazzoli:


.circleci/config.yml
```diff
@@ -347,13 +351,13 @@ workflows:
               only: master
       - license-check
       - ui
-      - runtime
       - core:
+      - runtime
           requires:
-            - runtime
+            - core
```
rhuss: don't think that runtime requires core now (it was just @chirino's PR, but I hope that we can avoid tha)

.circleci/config.yml
```diff
@@ -347,13 +351,13 @@ workflows:
               only: master
       - license-check
       - ui
-      - runtime
       - core:
```
rhuss: colon ?

.circleci/config.yml
```diff
@@ -347,13 +351,13 @@ workflows:
...
+      - runtime
```
rhuss: colon !

tools/bin/commands/build
```diff
@@ -1,17 +1,17 @@
...
   "ui"
-  "runtime"
-  "core:runtime"
-  "connectors:runtime core"
-  "verifier:runtime core connectors"
-  "rest:runtime core connectors"
-  "s2i:runtime core connectors rest"
-  "tests:runtime core connectors rest s2i"
+  "core"
+  "runtime:core"
```
rhuss: would remove core for now here

.circleci/config.yml
```diff
@@ -347,13 +351,13 @@ workflows:
               only: master
       - license-check
       - ui
-      - runtime
-      - core:
+      - core
+      - runtime:
           requires:
-            - runtime
+            - core
```
rhuss: runtime does not require core (introduced only in hirams PR)

lburgazzoli: check now

PR #1071, opened 2018-01-15
Adding a syndesis git sub command that gives us some interesting git info about what’s changed in the project.

chirino:
Adding some handy commands that may becom more useful.. for example you could do something like:

`mvn clean install -Dfabric8.mode=kubernetes -P image -P flash -pl $(syndesis git touched_pl),runtime $*`

to basically have maven rebuild changed module and deploy them for quick manual testing.

Was also thinking of seeing if we can skip building of modules in circleci if they have not been modified.  thats what `syndesis git has_app_module_changed runtime` would be for.

### DIFF
--- a/tools/bin/commands/git
+++ b/tools/bin/commands/git
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+git::description() {
+    echo "Git tools"
+}
+
+git::usage() {
+    cat <<EOT
+    has_app_module_changed <module>     prints true if the module has changed relative to master
+    touched_pl                          prints the list of /app modules uncommitted changes
+EOT
+}
+
+git::has_app_module_changed() {
+    modules="$@"
+
+    local changed=`\
+        git diff origin/master --numstat | cut -f 3 \
+        | grep "app/" | cut -c 5- \
+        | cut -d / -f 1 \
+        | awk 'NF' | sort | uniq \
+    `
+    for var in $changed; do
+        for m in $modules; do
+          if [ "$var" = "$m" ]; then
+              echo 'true'
+              return
+          fi
+        done
+    done
+    echo 'false'
+}
+
+git::touched_pl() {
+    cd `appdir` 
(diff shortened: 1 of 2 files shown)


tools/bin/commands/git
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+git::description() {
+    echo "Git tools"
+}
+
+git::usage() {
+    cat <<EOT
+    has_app_module_changed <module>     prints true if the module has changed relative to master
+    touched_pl                          prints the list of /app modules uncommitted changes
```
zregvart: perhaps we can make this command look like other commands (use `-` instead of `_`), add shorthand (one letter) parameters? this could be as simple as `-c/--has-changed` and `-s/--show-changed`

tools/bin/commands/git
```diff
@@ -0,0 +1,55 @@
...
+    git status -s | cut -c 4- | grep / | grep -v '\.\.' | cut -d / -f 1 | awk 'NF' | sort | uniq | paste -s -d"," -
+}
+
+git::run() {
+
+    local command="$1"
+    shift
+
+    case $command in
+        has_app_module_changed)
```
rhuss: Actually I would prefer the $(hasflag) and $(readopt) pattern as we use in the other commands. That would also easily allow for shortcuts.

tools/bin/commands/git
rhuss: Could we also add a '-q' or '--quiet' which instead of printing 'true' or 'false' just exit with 0 or 1 ? We could put this kind of function in `util/github_funcs` and then e.g. reuse it for "build" to offer an option only to rebuild those modules that have changed (+ the one that depend on those modules).

Also, instead of just putting out a change it would might make sense also to tell *what* has changed (by whom).

chirino:
@rhuss please modify as needed.

chirino:
@rhuss should I close this PR out or will you push into it?

rhuss:
I'm going to create a new one. thanks for the idea, really like it !

Issue #1056, opened 2018-01-14
[Custom connectors] Create a new top level app module extension


lburgazzoli:
We should create a new top level module with some sub modules containing all the bits extension related.

i.e. we can have the following sub modules:

- **extension-bom** the bom used by extension developers, should include camel, spring-boot and syndesis dependencies
- **extension-maven-plugin** should be the new place for syndesis-maven-plugin bits related to extensions 
- **extension-api** for contains any developer facing API and annotations
- **extension-converter** for meta data converter

lburgazzoli:
@nicolaferraro can you have a look at this one while working on #1037

nicolaferraro:
:+1:

apupier:
For information, I created https://issues.jboss.org/browse/FUSETOOLS-2757 to track actions required on Fuse Tooling side after this update.

Issue #1037, opened 2018-01-12
[Custom connectors] Define dedicated metadata format for extension definition


lburgazzoli:
As today extension definition is based on syndesis model but in order to support syndesis internals upgrades we should define a dedicated and versioned metadata definition that should be translated to the syndesis model.

nicolaferraro:
I started to do #1056 and was thinking to this issue. Conceptually it is a good thing to go in this direction and create a metadata definition for users and another one for our rest services and mapping between the two.

But if we use java beans for the user-facing model, it will result in a duplicate of the core/model project, since the objects used in an extension are really complex, and they'll be more complex now that we want to introduce custom connectors.

I was thinking to an alternative approach:
- Add a schema version information in the generated extension object
- Create a converter project that:
  - takes the generated json (as raw jackson maps) 
  - applies a transformation if necessary and convert the content to core model

We can also add tests to validate generated schema of some sample projects against a json schema (we have generated it) to ensure that we handle model version changing correctly.

This way, we should not need a duplicate structure.

Wdyt @lburgazzoli ?

lburgazzoli:
That is ok to me however it may complicate the maven plugin as we need to enrich / generate the extension/connector model so we would need to use raw json stuffs.

nicolaferraro:
So the approach used in the PR is to convert from the public schema to the internal schema, do changes in the maven-plugin and convert back to the public schema when saving. The rest analyzer part has been adapted to convert from the public schema to the internal schema before saving the extension in the jsondb.

We can add a `schemaVersion` field (integer) but it may be missing in metadata fragments written by users. The only place where it can be used is when loading the extension from file in the rest part.

The field can be also stored in jsondb in the internal model. We can track this way if the uploaded extension was targeting a old schema. Also, the `syndesisVersion` used to create the extension seems important in this context. We can add both, or just the `syndesisVersion` to the metadata in both schemas. Wdyt @lburgazzoli ?

nicolaferraro:
It is also true that adding a `syndesisVersion` field is not useful, as syndesis version can be inferred from dependencies (`extension-api` is a mandatory dependency, right?).

nicolaferraro:
Let's go for adding a mandatory `schemaVersion` that is added automatically by the plugin.

Issue #1078, opened 2018-01-16
Cleanup Maven build


rhuss:
* [ ] Fix parent to include all referenced version in `<dependencyManagement>`
* Introduce a concise Maven naming for directories, artifactIds and top-level package names.
  - [x] Every directory is reflected in the artifactid (e.g. `rest/dao` --> `rest-dao` artifactId)
  - [ ] Every module has a single top-level package (e.g. for `rest/dao` --> `io.syndesis.rest.dao`)
  - [x] The are `-` allowed in directory names but should be avoided (because of the point above).
  - [x] Directory structure should not be deeper than 2
  - [x] Avoid intermediated pom.xml which do not add functionality. Reference modules directly in the top-level parent pom
* Check the dependency graph:
  - [ ] Avoid circular dependencies at all cost
  - [ ] No dependencies for the runtimes to backend modules. We might weaken this restriction later on, but for now it helps in separating those two big blocks.
  - [x] Should we add a different Maven group for runtime jars ? 
* [x] Group Maven module in top-level "modules" which are reflected via GitHub labels `modules/...`

paoloantinori:
this might have implications on other systems slow to pick up the change, so better doing earlier than later

Issue #1075, opened 2018-01-16
[extensions] not working after db optimization


paoloantinori:
```
syndesis-rest-2-jfwd2 syndesis-rest 2018-01-16 09:42:31.815  WARN [-,,,] 1 --- [ool-12-thread-1] io.syndesis.jsondb.impl.SqlJsonDB        : fetchIdsByPropertyValue failed: Index not defined for:  collectionPath: /extensions, property: status
syndesis-rest-2-jfwd2 syndesis-rest Exception in thread "pool-12-thread-1" java.lang.UnsupportedOperationException: Index not defined for:  collectionPath: /extensions, property: status
syndesis-rest-2-jfwd2 syndesis-rest 	at io.syndesis.jsondb.impl.SqlJsonDB.fetchIdsByPropertyValue(SqlJsonDB.java:247)
syndesis-rest-2-jfwd2 syndesis-rest 	at io.syndesis.jsondb.dao.JsonDbDao.fetchIdsByPropertyValue(JsonDbDao.java:88)
syndesis-rest-2-jfwd2 syndesis-rest 	at io.syndesis.dao.manager.DataManager.lambda$fetchIdsByPropertyValue$2(DataManager.java:256)
syndesis-rest-2-jfwd2 syndesis-rest 	at io.syndesis.dao.manager.DataManager.doWithDataAccessObject(DataManager.java:372)
syndesis-rest-2-jfwd2 syndesis-rest 	at io.syndesis.dao.manager.DataManager.fetchIdsByPropertyValue(DataManager.java:246)
syndesis-rest-2-jfwd2 syndesis-rest 	at io.syndesis.dao.extension.ExtensionDataManager.getInstalledPhysicalId(ExtensionDataManager.java:48)
syndesis-rest-2-jfwd2 syndesis-rest 	at io.syndesis.dao.extension.ExtensionDataManager.getExtensionBinaryFile(ExtensionDataManager.java:40)
syndesis-rest-2-jfwd2 syndesis-rest 	at io.syndesis.project.converter.DefaultProjectGenerator.addExtensions(DefaultProjectGenerator.java:221)
syndesis-rest-2-jfwd2 syndesis-rest 	at io.syndesis.project.converter.DefaultProjectGenerator.lambda$generateAddProjectTarEntries$0(DefaultProjectGenerator.java:180)
[...]

PR #1079, opened 2018-01-16
Adding the extension top-level module

nicolaferraro:
This fixes both #1056 and #1037, by introducing a new top-level model for extensions and a converter sub-module that will allow us to evolve the internal API independently from the user public API for extensions.

I've added the modules:
- extension-annotation-processor: was syndesis-annotation-processor in rest
- extension-api: was runtime-api in runtime
- extension-bom: new for extension developers
- extension-maven-plugin: was syndesis-maven-plugin in rest
- extension-examples: new extension samples taken from @lburgazzoli github projects
- extension-converter: conversion tools for maintaining the public API separate

I've moved the extension json schema to the extension-api module, so tooling can pick it up from classpath as usual, since the core/model syndesis module is no longer required to be imported by users (there should be no changes for @lhein).

I've adapted the plugin to read the correct versions from the BOM instead of relying on pom properties.


nicolaferraro:
Need to clean the repo and fix some errors...

Some projects are parentless:
- bom
- examples/log-step-example

Is this going to create problems to the CI release process?

app/extension/maven-plugin/pom.xml
```diff
@@ -17,19 +17,18 @@
...
-    <artifactId>syndesis-rest-parent</artifactId>
+    <groupId>io.syndesis.extension</groupId>
+    <artifactId>extension-parent</artifactId>
     <version>1.2-SNAPSHOT</version>
     <relativePath>../pom.xml</relativePath>
   </parent>
 
-  <artifactId>syndesis-maven-plugin</artifactId>
-  <name>REST :: Maven Plugin</name>
+  <artifactId>extension-maven-plugin</artifactId>
```
lhein: I wouldn't say this doesn't affect tooling...it effectively breaks tooling ;)

lhein: Our code freeze is tomorrow and you introduce such a disruptive change at this point? Why was it required to change the Maven Coordinates of the plugin?

apupier:
> I've moved the extension json schema to the extension-api module

fine for that. We are not using it at all for now due to https://github.com/lburgazzoli/syndesis-extension-example-xml/issues/7

app/extension/maven-plugin/pom.xml
nicolaferraro: Yes, that renaming may cause issues :sweat_smile:. Are projects generated via archetypes by tooling? do you also invoke the plugin directly?

apupier: the example is completely duplicated in Fuse tooling codebase and embedded in built product.

we don't invoke the plugin directly

lhein: How to proceed:

I need to know if these are now the final coordinates for the plugin 
and 
I need to know if there is already a staging repo hosting this plugin so I can test the tooling with the new plugin.

As already mentioned the code freeze is tomorrow. Every such kind of change u do after this code freeze has the potential of breaking the tooling's Syndesis Extension functionality. This will be lasting until we do a 10.3 release mid of April leaving possible users with a broken tooling in the time between. 

I just want to raise the level of awareness that changes in this project can have effects on the tooling. If you are unsure about side effects then please feel free to put tooling team as reviewers or as mentions. We will then help to figure out the risk.

lhein: > we don't invoke the plugin directly

We invoke this plugin during the build of the project. So if the coords change then the plugin is not found. (or if we are lucky some old version is found instead)

apupier: >> we don't invoke the plugin directly
> 
> We invoke this plugin during the build of the project. So if the coords change then the plugin is not found. (or if we are lucky some old version is found instead)

we don't rely on normal mvn build integration of m2e? We have a specific menu?

lhein: We invoke the m2e build via Eclipse to create the deployable artifact.

lburgazzoli: Extensions have now a broader scope so it is required to have such changes to keep the code-base clean and have better module naming. This change is also required to have a dedicated bom for extension so you do not need to specify spring-boot, camel and syndesis version.

As interim solution we can wrap the new plugin with the old one so the coordinates are still valid but I think the issue here is that tooling release cycle is not in sync with syndesis - or vice versa :) - so I think PM should discuss this, @rhuss what do you think ?

apupier: > I think the issue here is that tooling release cycle is not in sync with syndesis - or vice versa :) - so I think PM should discuss this, @rhuss what do you think ?

it was decided by PM a year ago to decouple the release cycles of Fuse Tooling to be part of Jboss Tools.
For this specific release, we also took care that the Tech preview 3 is matching  the Fuse Tooling 10.2.0. Seems Tech preview 3 has shifted.

apupier: > Extensions have now a broader scope so it is required to have such changes to keep the code-base clean and have better module naming. This change is also required to have a dedicated bom for extension so you do not need to specify spring-boot, camel and syndesis version.

Required for Tech preview 3 or for Tech preview 4?
Using a bom instead of the version requires to change also the Wizard UI on Fuse Tooling side.

> As interim solution we can wrap the new plugin with the old one so the coordinates are still valid

If for a tech preview 3, it might be a solution or to shift this modification to Tech Preview 4

lburgazzoli: Well, this is part of the epic "Custom connectors" (#803) which targets TP4 so yeah, this targets TP4. TP3 branch has been cut yesterday afaik.

rhuss: Yes, 1.2.x is the TP3 branch.

lburgazzoli: @lhein @apupier as this targets TP4 and TP3 has branched is this PR still a blocker ?

apupier: The next stable Fuse Tooling release (10.3.0) - in normal process - is in 3 months (I mean after the 10.2.0 which has code freeze tomorrow). But there are Milestones publicly available every 3 weeks.
The Fuse Tooling will still work with Tech Preview 3 even after this modification.
With the PR as-is, it means that a dedicated release of Fuse Tooling would need to be planned (never happened yet) to have support of new plugin for Tech Preview 4 Technical Extension in a "stable" Fuse Tooling release.
As it is support for some Tech previews, I think it is fine to have it only in milestones but PM can decide otherwise.

nicolaferraro:
Thanks @rhuss, I'll take care of changing the remaining stuff, including the archetypes that still reference the old artifacts.

PM issues raised are important and we should work to solve them (maybe outside this PR).

app/extension/maven-plugin/src/main/java/io/syndesis/maven/RepackageExtensionMojo.java
```diff
@@ -156,16 +157,35 @@ protected void addCustomBoms(Collection<MavenDependency> dependencies) {
...
     protected void addDefaultBOMs(Collection<MavenDependency> dependencies) {
         String[] defaultBoms = new String[]{
-            resolveBomVersion(SPRING_BOOT_BOM, "spring-boot.version"),
-            resolveBomVersion(CAMEL_BOM, "camel.version"),
-            resolveBomVersion(SYNDESIS_BOM, "syndesis.version"),
-            resolveBomVersion(SYNDESIS_INTEGRATION_RUNTIME_BOM, "syndesis.version")
+            resolveBomVersion(SPRING_BOOT_BOM, "org.springframework.boot:spring-boot", "spring-boot.version"),
+            resolveBomVersion(CAMEL_BOM, "org.apache.camel:camel-core", "camel.version"),
+            resolveBomVersion(SYNDESIS_BOM, "io.syndesis.extension:extension-api", "syndesis.version")
+//            resolveBomVersion(SYNDESIS_INTEGRATION_RUNTIME_BOM, "io.syndesis:integration-runtime", "syndesis.version")
```
apupier: commented out code?

nicolaferraro:
@lburgazzoli, @rhuss: issues should have been addressed. I'll remove the wip after running a full e2e test, but the code can be reviewed.

Tooling issues will be tracked in a separate Jira.

nicolaferraro:
Now that #1075 is fixed, I've done a full test and it works.

rhuss:
@nicolaferraro let me review this tomorrow morning. we should merge asap to avoid too many conflicts in the course of time.

app/extension/pom.xml
```diff
@@ -0,0 +1,139 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.syndesis</groupId>
+    <artifactId>syndesis-parent</artifactId>
+    <version>1.2-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <groupId>io.syndesis.extension</groupId>
```
rhuss: I wonder whether we should introduce a new group. I know we already have a kind of mixture, but I think having one single group `io.syndesis` has some benefits, as now the artifactId need to be unique (a counter example is "model" which exists for 'io.syndesis' and 'io.syndesis.runtime'.

Alternatively we could decide to use a specific group for every top-level module.

@syndesisio/backend wdyt ?

nicolaferraro: No preference about that. Probably `io.syndesis.extension` may be redundant since all artifacts in that module start with the prefix `extension-`.. Changing it should be a simple find/replace..

rhuss: Ok, then let's go with a top-level 'io.syndesis'. This forces us to make the artifactId unique by adding the top-level module as prefix. 

The next step is then get the naming right with #1078 , where we still could reconsider the group naming.

@nicolaferraro are you fine with replacing the group ?

nicolaferraro: Ok, I will replace the groupId with `io.syndesis` soon.

PR #1081, opened 2018-01-16
Add Prometheus deployment to gather metrics

jimmidyson:


KurtStam:
@jimmidyson Can you add a description of recipe of what's required to add this deployment? For instance I feel unfamiliar with changes to the syndesis-template.go file.

app/deploy/generator/06-syndesis-prometheus.yml.mustache
```diff
@@ -0,0 +1,165 @@
...
+          kubernetes_sd_configs:
+          - role: pod
+            namespaces:
+              names:
+                - ${OPENSHIFT_PROJECT}
+
+          relabel_configs:
+          - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_scrape]
+            action: keep
+            regex: true
```
rhuss: Not sure whether this is an issue for Prometheus, but using `true` instead of `"true"` for an otherwise string value (see the other regex) causes some trouble usually.

jimmidyson: It's using a different YAML parser and works fine. This is lifted from the [Prometheus example config](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml#L150).

jimmidyson:
> Can you add a description of recipe of what's required to add this deployment? For instance I feel unfamiliar with changes to the `syndesis-template.go` file.

What do you mean a recipe? The deployment is added to the generated OpenShift template and is deployed along with the rest of the Syndesis components. Then the Syndesis metrics API can be used to query Prometheus and return the required metrics.

jimmidyson:
> Just a general question to the scope of this PR: Does this PR already includes the connection to the runtime pods. I.e. does prometheus start to scrape when we deploy a new integration ?

Yes. The Prometheus configuration in https://github.com/syndesisio/syndesis/pull/1081/files#diff-1bd4878dec4ef0fbc5511bebfc533430R30 section means that all pods annotated with `prometheus.io/scrape=true` will be scraped. That is configured in the Syndesis OpenShift service - https://github.com/syndesisio/syndesis/pull/1081/files#diff-50f77eb5d13a610813c15d9549d92be7R195 -
 when the integration pod is deployed.

KurtStam:
@jimmidyson So what's left to do for the prometheus POC is querying prometheus when a metrics API request comes it right?

jimmidyson:
That's right. Was waiting on the metrics API being defined then could quickly whip up a Prometheus query implementation.

KurtStam:
@jimmidyson super. I'm working on a first draft of that right now.

chirino:
rebased and resolved conflicts.

PR #1098, opened 2018-01-17
chore(build): add code coverage to Codacy

zregvart:


.circleci/config.yml
```diff
@@ -65,6 +65,19 @@ save_junit: &save_junit
     command: |
       mkdir -p /workspace/junit/
       find . -type f -regextype posix-extended -regex ".*/target/(surefire|failsafe)-reports/.*xml" | xargs -i cp --backup --suffix=.xml {} /workspace/junit/
+  run:
+    name: Collect coverage and class files
+    command: |
+      cd app
+      ./mvnw -N -B build-helper:timestamp-property@timestamp jacoco:merge@collect
```
rhuss: Ah, one can specify an execution id at the goal level. was new to me ...

zregvart:
@rhuss I'm not sure, I need to look into that, the crux of the issue is that Codacy requires one Jacoco XML report, and we have several Jacoco binary reports scattered in different containers that need to be merged and along with classes supplied to Jacoco to produce this single XML.
Let me thinker about it for a bit I think code coverage is a good tool to see what we can (more) safely refactor :)

zregvart:
This is proving to be a difficult one to implement. I'm putting this on hold for the moment.

rhuss:
@syndesisio/all If we want to go for the commit-first/no-review approach as suggested in the retrospective, then we have to enhance our safeguards. I still believe the refactoring after the fact usually does not happen (but people will work around stumbling blocks), but it seems I'm alone with the opinion (i made my points that letting others doing the clean up in some distant future is not necessarily fair). 

To allow refactoring, we should care to enhance our unit test coverage or at least do not decrease it (or am I alone with this opinion, too ?). So I opt for mandatory coverage checks which guarantee that test coverage will not drop with a given PR. This check might be overruled for refactoring PRs (should be labelled for this), as it's not fair to allow refactorings only when tests have to be created for already existing code.

So can we get this PR running and make it a mandatory check for GitHub ?

deeleman:
> I still believe the refactoring after the fact usually does not happen (but people will work around stumbling blocks), but it seems I'm alone with the opinion (i made my points that letting others doing the clean up in some distant future is not necessarily fair).

You're not alone on this. Common experience proves that such cleanup will never happen and pushing code that requires refactoring right off the bat in the hopes that someone will come later to clean that mess is certainly unfair IMO. This circumstance becomes particularly relevant in the case of code smells or architectural anti-patterns, since no unit or integration test can address those, and if this principle prioritizes rapid feature development above code quality, then the sad truth is that the latter will never find its space in the product agenda because there will be always new features to implement, so this will only snowball more and more.

My 0.02$

zregvart:
The main problem with this approach for reporting code coverage to Codacy is that Codacy supports single XML Jacoco report, whereas we generate multiple. The PR contains code/configuration to generate the single XML Jacoco report but this fails on CircleCI as we generate same files from multiple steps and store them in shared workspace. Getting non-overlapping shared workspace layers is quite tricky to do. Happy for any suggestions on how to get this done.

chirino:
@zregvart should we close this out?

zregvart:
I'll reopen, when I get this working.

PR #1113, opened 2018-01-17
[WIP] UI dependency graph update

deeleman:
This is a partial update of some core dependencies, leveraging some of the work already did for #380 

This PR is a WIP


gashcrumb:
What's left to do on this one?

deeleman:
> What's left to do on this one? This all LGTM, does it just need a test locally?

@gashcrumb I'm still fixing incompatibilities in some implementations that break the build

deeleman:
Closing, as it seems not possible to put together a proper set of upgrades that will make the CI server happy and this is has consumed too much time already. We face the following issues:

1. Certain package upgrades require Angular 5, which is not compatible with some other core dependencies.
2. Some newer package versions implement a new API that is not compatible with our current implementation of them.
3. Some newer package versions share different versions of a same peer dependency.

The amount of work required to fix all the above exceeds the scope of the original PR (which was intended to take advantage of the work done for #380). I'd suggest to go upgrading and submitting individual PRs on a per dependency base, when the time is right.

PR #1122, opened 2018-01-18
chore(checks): Let's exclude UI from Codacy

### LINKED ISSUE #1112: Fix codacy quote rule for typescript files
Noticed on some of the open PRs the codacy check is failing with this:

```
' should be " 
```

Our tslint in the UI is set up the opposite, i.e. we want `'`.

zregvart:
In order for Codacy to pickup them the configuration files need to be in the root of the repository.

I'm not 100% sure about `syndesis-policy` dependency and what happens if it goes out of sync with `ruleset.xml`.

Would be cool to get feedback if this breaks anyone's workflow.

Fixes #1112

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,0 +1,4 @@
+---
+exclude_paths:
+  - app/ui/**
+


zregvart:
According to [Supported Configuration files](https://support.codacy.com/hc/en-us/articles/207994335-Code-Patterns#5-existing-tools) the `tslint.json` should be picked up from the root of the repository. According to the [PR analysis](https://www.codacy.com/app/zregvart/syndesis/pullRequest?prid=1245752) it seems that it considers the intentional violation in 288dea19ef9bd6601e95a2d6de8f0353f821172e to be a fix.
Not sure if this is a bug in Codacy or it only considers `tslint.json` from `master`, or something else...

gashcrumb:
So, maybe we should just mask typescript files from the Codacy config?  We do have a lint step in the UI build which will fail that particular build already.

zregvart:
> So, maybe we should just mask typescript files from the Codacy config? We do have a lint step in the UI build which will fail that particular build already.

Yeah let's do that, it would be nice to have UI show up Codacy so we can identify hot spots and pick stuff to refactor...

gashcrumb:
Agreed, it would be nice!.

Poking around codacy I see this quote thing being the main complaint across a number of UI typescript files.  We can hide all this by excluding the UI code but of course then we don't see other problems.  Though I haven't yet seen codacy reporting problems other than `'`.

zregvart:
Let's close this and I'll create a new PR for ignoring UI in Codacy.

PR #1123, opened 2018-01-18
Backport of release releated fixes for TP3

rhuss:
These fixes are currently needed for doing a TP3 release. This is still WIP, as the release script needs still to be updated for the template and Docker image generation / push.

* Refactored `syndesis release` and used this to make the releases
* Configured release plugin to run an intermediate build in the connectors to create the schema json's with the proper versions before the get checked in. 
* Remove dependencies on test classes among the SQL connectors (that broke the build). Shouldn't be harmful in the long run, as we should move the whole SQL connector to a single module anyway. @KurtStam is this ok for you ?


rhuss:
@syndesisio/backend anyone fancy for a review ? It's mostly about the release process in TP3 (boring stuff, I know). If you don't mind, I wouldn't, too, so lets merge it before it starts roting.

Issue #1115, opened 2018-01-17
Create runtime side of Data Buckets POC


igarashitm:
This issue is to create Data Buckets POC focused on runtime. Put the associated document into Camel Exchange Property with a Document ID as a key, also keeping primary Document as a IN Message body.

Key point is to build a mapping between Syndesis Steps and corresponding DataShapes via Document ID. The idea came from initial discussion was to do that in the Interceptor, for each Step the Inteceptor would capture the output data and push it into Exchange Property with Document ID as a key.

Assuming UI would put corresponding Document ID into datamapper mapping definition ahead, then camel-atlasmap only needs to retrieve it from Exchange Property with Document ID in the mapping as a key.

chirino:
Added PR  #1143 which should allow atlas map to do something like:

```java
Map<String, Message> messages = exchange.getProperty("CAPTURED_OUT_MESSAGES_MAP", Map.class);
Message msg = messages.get("<<Step UUID>>")
String inputData = msg.getBody(String.class);
```
To get the out documents associated with each step id.

The nice thing about this is that you don't need to add any dependencies to atlasmap to pick this up.

lburgazzoli:
@igarashitm @chirino I've added WIP label to the PR as I'm almost ready to open a PR for the yaml removal stuffs.

chirino:
@lburgazzoli. ok.  My PR for monitoring also depends on Id's being set on the processors.  The sooner you get it in the better.

lburgazzoli:
@chirino setting the processor id has been added to #1144

igarashitm:
FYI - PR atlasmap/atlasmap#246 for camel-atlasmap bit

igarashitm:
@chirino w/r/t #1191 - do we want to make a deep copy of the *org.apache.camel.Message* object in the interceptor, if the "preserve output message" option is specified for the step/connection?

igarashitm:
As per today's call we don't implement @1191 at first place.

A memo for the future work:
How about holding a `java.util.Map<String, Object>` as a Data Buckets and push the body instead of Message? The Interceptor then could *clone* it if the *preserve output message* option is enabled && the body is `Clonable`.

chirino:
Even cloning is not guaranteed to be a deep copy.  Best solution would be to serialize/marshal to store the body and then deserialize/unmarshal to get it back.  Not all bodys can be serialized. Perhaps we should create explicit steps that can do the marshal/unmarshal.

PR #1143, opened 2018-01-18
feat: Adding an out message capture intercept strategy.

chirino:
Part of #1115

### DIFF
--- a/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/IntegrationRuntimeAutoConfiguration.java
+++ b/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/IntegrationRuntimeAutoConfiguration.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.ServiceLoader;
 
 import org.apache.camel.builder.RouteBuilder;
+import org.apache.camel.spi.InterceptStrategy;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
@@ -60,5 +61,11 @@ public class IntegrationRuntimeAutoConfiguration {
         // IntegrationRouteBuilder automatically add known handlers to the list
         // of provided ones, know handlers have priority
         return new IntegrationRouteBuilder(location, handlers);
+    }
+
+    @Bean
+    @ConditionalOnProperty(prefix = "syndesis.integration.runtime", name = "capture", matchIfMissing = true)
+    public InterceptStrategy createOutMessageCaptureInterceptStrategy() {
+        return new OutMessageCaptureInterceptStrategy();
     }
 }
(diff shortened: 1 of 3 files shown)


app/runtime/runtime/src/main/java/io/syndesis/integration/runtime/SyndesisRouteBuilder.java
```diff
@@ -56,6 +56,8 @@ protected SyndesisModel loadModel() throws Exception {
     @SuppressWarnings("unchecked")
     @Override
     public void configure() throws Exception {
+        getContext().addInterceptStrategy(new OutMessageCaptureInterceptStrategy());
+
```
lburgazzoli: [CamelAutoConfiguration](https://github.com/apache/camel/blob/master/components/camel-spring-boot/src/main/java/org/apache/camel/spring/boot/CamelAutoConfiguration.java#L438-L448) should automatically register beans of type `InterceptStrategy` so it may be better to register `OutMessageCaptureInterceptStrategy` using spring-boot's `@Configuration` so you can conditionally enable it with a simple property (feature toggle)

chirino: Will do

rhuss:
@lburgazzoli could you please have short recheck of the review if the changes are fine ?

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/IntegrationRuntimeAutoConfiguration.java
```diff
@@ -61,4 +62,10 @@ public RouteBuilder integrationRouteBuilder() {
         // of provided ones, know handlers have priority
         return new IntegrationRouteBuilder(location, handlers);
     }
+
+    @Bean
+    @ConditionalOnProperty(prefix = "syndesis.integration.runtime", name = "capture", matchIfMissing = true)
```
lburgazzoli: This is usually in the form:

    @ConditionalOnProperty(prefix = "syndesis.integration.runtime.capture", name = "enabled", matchIfMissing = true)

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/OutMessageCaptureInterceptStrategy.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Used to capture the out messages of processors with configured ids.  The messages are placed into
+ * a map stored in the camel exchange property using the processor id as the map key.
+ */
+public class OutMessageCaptureInterceptStrategy implements InterceptStrategy {
+
+    public static final String CAPTURED_OUT_MESSAGES_MAP = "CAPTURED_OUT_MESSAGES_MAP";
```
lburgazzoli: I would prefix the property with `Syndesis.`

lburgazzoli:
@chirino I've added 2 small comments, then it lgtm

PR #1144, opened 2018-01-19
Use integration to build camel routes

### LINKED ISSUE #447: Use integration to build camel routes
As today the process to convert an integration to camel routes is:

    Integration Step --> Integration Step Handler (Visitor) --> Runtime Step --> Runtime Step Handler --> Camel Route

This is not needed anymore as the generated yaml file is not more directly accessible to the end user so we can avoid most of the intermediate steps

    Integration Step --> Integration Step Handler (Visitor) --> Camel Route

So we can get rid of integration-runtime/model and lot of code from integration-runtime/runtime

### LINKED ISSUE #1147: [TechExt] newline char in properties value, breaks TechExt
`\n` symbol is illegal in URIs. It needs to be escaped to `%0A`.

Camel Core doesn't do that on its own, so we have added a sanitization step specific for Tech Extensions, that are the most likely case where we can have one of them.

Typical usecase is if you feed in a template or a script.

lburgazzoli:
1. The intermediate yaml flow is finnaly over so we have more freedom to define our steps
2. There is a new top level module **integration** that contains all the bits related to the integration and replaces **runtime**
3. **integration** depends on:
    * **core**
    * **exension**

Fixes #447 
Fixes #1147

Include some bits from #1148 and #1143


lburgazzoli:
@syndesisio/backend anyone has some spare time to review this ?

lburgazzoli:
retest this please

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/ConnectorStepHandler.java
```diff
@@ -0,0 +1,209 @@
...
+                route = builder.from(componentId);
+            } else {
+                route = route.to(componentId);
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+
+        // Handle split
+        if (descriptor.getSplit().isPresent()) {
```
rhuss: Looks similar to the code in SimpleEndpointStepHandler. Would it make sense to move this out to a common place ?

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/SimpleEndpointStepHandler.java
```diff
@@ -0,0 +1,112 @@
...
+                route = builder.from(uri);
+            } else {
+                route = route.to(uri);
+            }
+        } catch (URISyntaxException e) {
+            throw ObjectHelper.wrapRuntimeCamelException(e);
+        }
+
+        // Handle split
+        if (descriptor.getSplit().isPresent()) {
```
rhuss: see ConnectorStepHandler for the same code.

app/integration/project-generator/src/main/resources/io/syndesis/integration/project/generator/templates/application.properties.mustache
```diff
@@ -0,0 +1,6 @@
+# Logging
+logging.level.io.syndesis = DEBUG
```
rhuss: Should this be hardcoded or would it make more sense to make it configurable ? // @paoloantinori this migh affect the support page, too.

app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGenerator.java
```diff
@@ -0,0 +1,249 @@
...
+    @SuppressWarnings("PMD.DoNotUseThreads")
+    private Runnable generateAddProjectTarEntries(IntegrationDeployment deployment, OutputStream os) {
+        return () -> {
+            try (
+                TarArchiveOutputStream tos = new TarArchiveOutputStream(os)) {
+                tos.setLongFileMode(TarArchiveOutputStream.LONGFILE_POSIX);
+
+                ProjectGeneratorHelper.addTarEntry(tos, "src/main/java/io/syndesis/example/Application.java", ProjectGeneratorHelper.generate(deployment, applicationJavaMustache));
+                ProjectGeneratorHelper.addTarEntry(tos, "src/main/resources/application.properties", ProjectGeneratorHelper.generate(deployment, applicationPropertiesMustache));
+                ProjectGeneratorHelper.addTarEntry(tos, "src/main/resources/syndesis/integration/deployment.json", Json.mapper().writerWithDefaultPrettyPrinter().writeValueAsBytes(deployment));
```
rhuss: is the deployment.json for reference only or how it is used ? Does is has the same format as the `deployment.json` from rest ? (if not, maybe its better to rename it to avoid confusion)

lburgazzoli: no it has different format, will rename it

app/integration/project-generator/src/main/resources/io/syndesis/integration/project/generator/templates/application.properties.mustache
lburgazzoli: At the moment it is hardcoded but yeas it make sense to make it configurable. My idea is to add some logging option to the integration so that this can be controlled (i.e. you may ahve a flag on the integration page to turn on/of debugging). Or we can use an environment variable but that would mean to change the build config.

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/SimpleEndpointStepHandler.java
lburgazzoli: done

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/ConnectorStepHandler.java
lburgazzoli: done

app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGenerator.java
lburgazzoli: done

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/ConnectorStepHandler.java
```diff
@@ -0,0 +1,209 @@
...
+        final ConnectorAction action = step.getAction().filter(ConnectorAction.class::isInstance).map(ConnectorAction.class::cast).get();
+        final ConnectorDescriptor descriptor = action.getDescriptor();
+
+        // Camel
+        final String index = step.getMetadata(Step.STEP_INDEX).orElseThrow(() -> new IllegalArgumentException("Missing index for step:" + step));
+        final String scheme = Optionals.first(descriptor.getComponentScheme(), connector.getComponentScheme()).get();
+        final CamelContext context = builder.getContext();
+        final TypeConverter converter = context.getTypeConverter();
+        final String componentId = scheme + "-" + index;
+        final ComponentProxyComponent component = resolveComponent(componentId, scheme, context, descriptor);
```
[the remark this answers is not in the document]
lburgazzoli: yes no limitation of the type of component

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/ConnectorStepHandler.java
```diff
@@ -0,0 +1,209 @@
...
+            // Set connector options with remaining properties and original values
+            // to keep placeholders and delegate resolution to the final endpoint.
+            component.setOptions(
+                properties.entrySet().stream()
+                    .filter(e -> customizersOptions.containsKey(e.getKey()))
+                    .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue))
+            );
+
+            // Remove component
+            context.removeComponent(component.getComponentId());
```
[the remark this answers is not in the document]
lburgazzoli: This component id is generated so it is specific for the connector and step

PR #1148, opened 2018-01-19
[TechExt] Added sanitization logic for new_lines in params

### LINKED ISSUE #1147: [TechExt] newline char in properties value, breaks TechExt

paoloantinori:
Fixes #1147


lburgazzoli:
Labelled as WIP as there is #1144 coming which has impact on this one

paoloantinori:
fixed with https://github.com/syndesisio/syndesis/pull/1144

Issue #1065, opened 2018-01-15
Monitoring API Design


chirino:
We should get UI feedback on the API design used for monitoring APIs since they need to render the data them those provide.  We should discuss those APIs in this issue.

Currently we identified the following Integration Metrics:
Summary : # msgs, # errors, last processed timestamp, uptime

Review Camel metrics available to see if any other metrics make sense
Per integration revision?

Also relates to: https://github.com/syndesisio/syndesis/issues/1025

chirino:
The logging poc #1054 defines the loging API to look like:

`GET /api/v1/logs/{integration}` ->
```
[
  "id": string // tx id for each integration transaction (aka integration run)
  "pod": string // pod where the tx was run in.
  "at":long // - time that the tx started
  "steps": [{ 
      "key": string // id of the step
      "value": [ {
         "id": string // id of the log entry
         "at": long // long - time that the tx started
         "action": string // can be "begin" or end "used" to signal step start/end
         "message": string // user supplied log message
      }]
  }]
]
```

@gashcrumb what do you think about that shape. I think the UI is likely more interested in stuff like duration of steps.  Do you need that from the rest API?  I guess we also need to track transaction/step overal state like in-progress, sucess, failed. etc.

chirino:
Also need to design how to page and tail the logging api.

chirino:
@KurtStam @dhirajsb have you guys thought about what the API would be for metrics?  Would it be something like:

`GET /api/v1/metrics/{integration}` ->
```
{
 TBD...
}
```

gashcrumb:
How do we indicate if an integration flow hit an error at a given step?

Would it make sense to rename `key` to `id` since it's the id of a step?  Also is this `id` field going to be reflected in the actual integration model?

Hard for me to say if anything's missing at this stage, I wonder if the return value for the API should maybe be an object instead and have this array of transactions be a property on that object, that way we can also have properties like number of exchanges processed overall etc, this will kinda depend on the UXD designs though.

chirino:
Good points @gashcrumb created #1067 so that UXD can get us some quick wireframes.

KurtStam:
@chirino I was thinking the following:

someUniqueId is to be determined; I want something unique for the process. Maybe podId but that may not be good enough. Maybe some sort of Id obtainable from the MBeans.

GET /api/v1/metrics/integration/{integrationId}/version/{versionId}/someUniqueId/{someUniqueId}
{
[...]

PR #1156, opened 2018-01-19
[feature] Adding the metrics API #1065 

KurtStam:


rhuss:
Do wee need so verbose keys ? `numberOfProcessedMessages`, could it be short `processedMessages` as the value is a number it should be clear what this is about.

 I suggest:

* messages
* errors
* lastProcessed
* uptime

I think this should be clear enough (and reduced the chance of typos).

rhuss:
For the rollup keys, I would choose different names, though (so that you can interprete the response without having to know from which endpoint it was fetched):

* totalMessages
* totalErrors
* lastProcessed
* uptime

(i assumes the lastProcessed and uptime are the same as for the other endpoint)

app/core/model/src/main/java/io/syndesis/model/integration/Integration.java
```diff
@@ -117,8 +116,6 @@ default Kind getKind() {
 
     Optional<Date> getCreatedDate();
 
-    Optional<BigInteger> getTimesUsed();
```
rhuss: Shouldn't this go to an IntegrationRevision (not sure whether the corresponding PR has been already merged) ? Why does it need to be a BigInteger ? Isn't a long good enough ?

rhuss: Wouldn't just a `count` be better than `timesUsed` ?

jimmidyson: This is being removed so sholdn't be a problem. Are the @syndesisio/ui-api folks aware of this change?

app/core/model/src/main/java/io/syndesis/model/metrics/MetricsSummary.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+import io.syndesis.model.Kind;
+import io.syndesis.model.WithId;
+
+@Value.Immutable
+@JsonDeserialize(builder = MetricsSummary.Builder.class)
+@SuppressWarnings("immutables")
+public interface MetricsSummary extends WithId<MetricsSummary>, Serializable{
```
jimmidyson: If this is just a response DTO, why does it have an ID? Doing so means that non-jsondb implementations, i.e. Prometheus, will have to insert a dummy ID. I would prefer to have a response type MetricsSummary that is expose from the API and the ID be only stored in jsondb, not returned via the API. This is tying the API to the implementation which is not a good idea.

app/rest/jsondb/src/main/java/io/syndesis/jsondb/dao/IntegrationSummaryJsonDbDao.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.stereotype.Service;
+
+@Service
+@Configuration
+@ConditionalOnProperty(value = "dao.kind", havingValue = "jsondb")
+public class IntegrationSummaryJsonDbDao extends JsonDbDao<MetricsSummary> implements MetricsSummaryDao {
```
jimmidyson: `IntegrationSummaryJsonDbDao` -> `MetricsSummaryJsonDbDao`?

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/metrics/IntegrationMetricsHandler.java
```diff
@@ -0,0 +1,95 @@
...
+import io.syndesis.model.Kind;
+import io.syndesis.model.ListResult;
+import io.syndesis.model.metrics.MetricsSummary;
+import io.syndesis.rest.v1.handler.BaseHandler;
+import io.syndesis.rest.v1.operations.Getter;
+
+@Path("/metrics/integrations")
+@Api(value = "integration-metrics")
+@Component
+public class IntegrationMetricsHandler extends BaseHandler
```
jimmidyson: This shouldn't extend `BaseHandler` as it ties the source of metrics to jsondb via the `DataManager`. Please separate this out and create an interface that can be implemented differently depending on the respective backend source. Then plug in the `DataManager` into your jsondb implementation. I can then work on the Prometheus implementation.

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/metrics/IntegrationMetricsHandler.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public MetricsSummary get(String id) {
+        return Getter.super.get(id);
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @ApiOperation("Retrieves a rolled up metrics summary for all integrations over their lifetime")
+    public MetricsSummary get() {
+        ListResult<MetricsSummary> integrationMetricsList = getDataManager().fetchAll(
```
jimmidyson: Again, ties to jsondb implementation.

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/metrics/IntegrationMetricsHandler.java
```diff
@@ -0,0 +1,95 @@
...
+                MetricsSummary.class);
+        return rollup(integrationMetricsList.getItems());
+    }
+
+    private MetricsSummary rollup(List<MetricsSummary> metricsSummaryList) {
```
jimmidyson: Should only be part of the jsondb implemenation.

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/metrics/IntegrationMetricsHandler.java
jimmidyson: Should delegate to the injected implementation, either jsondb or Prometheus, once the interface has been defined.

app/core/model/src/main/java/io/syndesis/model/metrics/MetricsSummary.java
jimmidyson: I'm not sure what the point of the ID is in the returned integration specified metrics is? If we dropped that, we could just use the same response type for aggregated metrics.

app/core/model/src/main/java/io/syndesis/model/integration/Integration.java
rhuss: Ah yes. Guess I'm red/green blind ;-)

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/metrics/IntegrationMetricsHandler.java
rhuss: @chirino but couldn't we keep the endpoint stable and delegate _after_ the endpoint ? That way we *guarantee* that the API remains the same regardless of the implementation (which would not be the case if we have to (supposed to be) identical endpoints.

app/core/model/src/main/java/io/syndesis/model/metrics/MetricsSummary.java
KurtStam: @jimmidyson, The id is the integrationId, you really want to remove it?

app/core/model/src/main/java/io/syndesis/model/metrics/IntegrationMetricsSummary.java
```diff
@@ -0,0 +1,54 @@
...
+@JsonDeserialize(builder = IntegrationMetricsSummary.Builder.class)
+@SuppressWarnings("immutables")
+public interface IntegrationMetricsSummary extends WithId<IntegrationMetricsSummary>, Serializable{
+
+    @Override
+    default Kind getKind() {
+        return Kind.IntegrationMetricsSummary;
+    }
+
+    Optional<Long> getMessages();
```
jimmidyson: What's the difference between `messages` and `totalMessages`?

app/core/model/src/main/java/io/syndesis/model/metrics/IntegrationMetricsSummary.java
```diff
@@ -0,0 +1,54 @@
...
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+import io.syndesis.model.Kind;
+import io.syndesis.model.WithId;
+
+@Value.Immutable
+@JsonDeserialize(builder = IntegrationMetricsSummary.Builder.class)
+@SuppressWarnings("immutables")
+public interface IntegrationMetricsSummary extends WithId<IntegrationMetricsSummary>, Serializable{
```
jimmidyson: Still unsure what the purpose of the `id` field is? Plus if it's the ID of the integration then it should be called `integrationId` to be explicit. But I would prefer not to have that field here at all.

app/rest/rest/src/main/java/io/syndesis/rest/metrics/SQLMetricsProviderImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    private IntegrationMetricsSummary rollup(List<IntegrationMetricsSummary> metricsSummaryList) {
+        Long totalMessages = 0L;
+        Long totalErrors = 0L;
+        Optional<Date> totalLastProcessed = Optional.empty();
+        Long totalUptime = 0L;
+        for (IntegrationMetricsSummary integrationSummary : metricsSummaryList) {
+            if (integrationSummary.getMessages().isPresent()) {
+                totalMessages += integrationSummary.getMessages().get();
+            }
+            if (integrationSummary.getErrors().isPresent()) {
```
jimmidyson: This can be simplified to:

```java
totalErrors += integrationSummary.getErrors().orElse(0);
```

app/rest/rest/src/main/java/io/syndesis/rest/metrics/SQLMetricsProviderImpl.java
```diff
@@ -0,0 +1,79 @@
...
+        return rollup(integrationMetricsList.getItems());
+    }
+
+    private IntegrationMetricsSummary rollup(List<IntegrationMetricsSummary> metricsSummaryList) {
+        Long totalMessages = 0L;
+        Long totalErrors = 0L;
+        Optional<Date> totalLastProcessed = Optional.empty();
+        Long totalUptime = 0L;
+        for (IntegrationMetricsSummary integrationSummary : metricsSummaryList) {
+            if (integrationSummary.getMessages().isPresent()) {
```
jimmidyson: This can be simplified to:

```java
totalMessages += integrationSummary.getMessages().orElse(0);
```

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/metrics/IntegrationMetricsHandler.java
```diff
@@ -0,0 +1,72 @@
...
+@Path("/metrics/integrations")
+@Api(value = "integration-metrics")
+@ComponentScan("io.syndesis.rest.metrics")
+@Component
+public class IntegrationMetricsHandler implements Resource {
+
+    private final MetricsProvider metricsProvider ;
+
+    public IntegrationMetricsHandler(MetricsProvider metricsProvider) {
+        super();
```
jimmidyson: Redundant.

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/metrics/IntegrationMetricsHandler.java
```diff
@@ -0,0 +1,72 @@
...
+        this.metricsProvider = metricsProvider;
+    }
+
+    @Override
+    public Kind resourceKind() {
+        return Kind.IntegrationMetricsSummary;
+    }
+
+    /**
+     * The id should be integrations.{integrationId}
```
jimmidyson: So not just the integration ID but prefixed with `integrations.`? Is that just a typo? Seems a bit odd?

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/metrics/IntegrationMetricsHandler.java
```diff
@@ -0,0 +1,72 @@
...
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiParam;
+import io.syndesis.model.Kind;
+import io.syndesis.model.metrics.IntegrationMetricsSummary;
+import io.syndesis.rest.metrics.MetricsProvider;
+import io.syndesis.rest.v1.operations.Resource;
+
+@Path("/metrics/integrations")
+@Api(value = "integration-metrics")
+@ComponentScan("io.syndesis.rest.metrics")
```
jimmidyson: Do we prefer to use auto-configuration in the metrics module or leave ComponentScan here? @zregvart any preference?

app/core/model/src/main/java/io/syndesis/model/metrics/IntegrationMetricsSummary.java
KurtStam: id is the integrationId. I labeled it as such in the API. For it to be persisted in the db I need an 'id' field though so this is the path of least residence if you don't mind.

app/core/model/src/main/java/io/syndesis/model/metrics/IntegrationMetricsSummary.java
KurtStam: In the second call we roll up the metrics over all integrations. Roland requested there to be keys with different fields. I don't really have a strong opinion about it. I'm happy to drop the 'total' prefixes.

app/rest/rest/src/main/java/io/syndesis/rest/metrics/SQLMetricsProviderImpl.java
KurtStam: ah right, that's much cleaner, forgot about that.

app/rest/rest/src/main/java/io/syndesis/rest/metrics/SQLMetricsProviderImpl.java
KurtStam: +1

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/metrics/IntegrationMetricsHandler.java
KurtStam: yes forgot to update the javadoc, it should simply be the `integrationId`

KurtStam:
@jimmidyson should be good to go now?

deeleman:
From a UI POV, I cannot provide any relevant feedback in regards of the suitability of the API responses until I have a final-final UI at a hand. Right now, and taking the screenshots submitted originally by @sjcox-rh (which are currently being repurposed AFAIK), I can see that `/api/v1/metrics/integrations` lacks the following:

- **Total amount of active integrations**, (informing about how many are green and how many are red, whatever that means), unless we want to infer it from the list retrieved form the "Integration Store" (which will require fetching all integrations every time we want to refresh data).
- **Active connections** (although I guess we want to infer it from the list retrieved from the "Connections Store", following the above principle)
- **Breakdown on message types** (green|red, whatever that means - we already have the total amount of messages in the API response).

We could have all the integrations and connections data already available by the time the user reaches this page if such data was persisted somewhere, but unfortunately it doesn't, so unless I query the API to retrieve ALL the integrations and connections, this information will have to become part of the API response (provided we want to move forward with this layout for the metrics dashboard).

KurtStam:
@deeleman I think the total number of active integrations is not `metrics` and can come from obtaining the integration status, as is done already on that page to create little circular diagram. Same for connections. With regard to messages count (green/red) I'm assuming red shows the number of requests that resulted in error (which is the `errors` field. 

That said, the current API is a first Draft and we can modify as we see fit.

KurtStam:
squashed, @chirino can you approve one more time?

PR #1157, opened 2018-01-19
(chore) Move integration support and adjust some filenames and paths.

gashcrumb:
This moves the integration support service into the integration domain.  I probably went a bit overboard; reworked the file/directory names a bit too and fixed up the imports a bit.  This also moves some of the integration related models out of `model.ts` and into the integration module.

Did hit an issue where some import changes caused runtime issues so had to adjust the import used in the integration store/service.

Apologizing to @kahboom now ahead of time :-)


app/ui/src/app/dashboard/dashboard.component.ts
```diff
@@ -1,7 +1,8 @@
 import { Component, OnInit } from '@angular/core';
 import { Observable } from 'rxjs/Observable';
 
-import { Connections, Integrations } from '../model';
+import { Connections } from '../model';
+import { Integrations } from '@syndesis/ui/integration';
 import { ConnectionStore } from '../store/connection/connection.store';
 import { IntegrationStore } from '../store/integration/integration.store';
```
deeleman: I do believe that you can combine these two into a single import statement, like:

```
import { ConnectionStore, IntegrationStore } from '@syndesis/ui/store';
``` 

I just doubled checked those folders and the barrels were already in place.

app/ui/src/app/customizations/tech-extensions/detail/tech-extension-detail.component.ts
```diff
@@ -1,7 +1,8 @@
 import { Component, OnInit, ViewChild } from '@angular/core';
 import { ActivatedRoute, Router } from '@angular/router';
 import { Observable } from 'rxjs/Observable';
-import { Extension, Integrations } from '../../../model';
+import { Extension } from '../../../model';
```
deeleman: Since `model.ts` is located at the very top of the `app` folder, it an be safely reached through the `@syndesis/ui/model` with no fear to circular dependencies.

app/ui/src/app/integration/integration.model.ts
```diff
@@ -0,0 +1,125 @@
...
+  };
+}
+
+export function createIntegrationRevisionSpec() {
+  return <IntegrationRevisionSpec>{
+    configuration: undefined,
+    steps: undefined,
+    connections: undefined
+  };
+}
```
deeleman: This is the only section of the entire PR that I believe needs to be addressed in a different way: Having all the functions returning object instances where all properties are set to undefined are simply unnecessary and ad a lot of boilerplate to the codebase. If we eventually move these functions elsewhere or either any of the types returned or its ancestors change its definition, the function will make the compiler complain. 

**Solution**: If a type (any type) is meant to support nullable/non-nullable properties mapped as `undefined`, then the original type property should be defined as optional by adding a `?` to the property definition. 

Example - Refactor the `Step` interface like this (please notice that its ancestor has been removed):

```typescript
export interface Step {
   action?: Action;
   connection?: Connection;
   name?: string;
   configuredProperties?: any;
   stepKind?: string;
   id?: string;
 }
``` 

Then you can refactor the `createStep` function like this, in a type safe mode (unless you further update `Step` to include a mandatory, non-nullable typed property):

```typescript
export const createStep = () => ({} as Step);
```

You can even go one step further and do this on a more generic way:

```typescript
export function integrationTypeFactory<T>(): T {
   return {} as T;
}
```

This would allow to replace all the 3 functions above at once.

app/ui/src/app/dashboard/dashboard.component.ts
gashcrumb: Yeah, both of these are two cases where I kinda debated if I should go and touch the other imports or kinda try and keep the changes a bit more focused.  I think since we'll eventually also be moving or replacing these stores I think we could leave this for now and fix this line when the stores get dealt with.

gashcrumb:
Applied that fix to `tech-extension-detail.component.ts`, squashed and rebased, ready to go pending CI

gashcrumb:
@deeleman Rebased and squashed again, ready for merging...

Issue #1095, opened 2018-01-17
As a citizen integrator, add additional information to an integration log


kcbabo:
As a citizen integrator, I want the ability to add a log step to an integration flow that allows me to record additional information in the integration log.  Additional log information will include one or more of the following:
* Message context (e.g. message headers and exchange properties)
* Message content (current message body)
* Custom information (Simple language expression)
The log step configuration page should allow me to select any/all of the above to be reported in the integration log.  When viewing a log entry produced by the log step, the details portion of the log entry should contain all of the log data produced by the step.

If a custom log message is provided, the UI should perform a basic syntax validation of the text prior to accepting the step configuration.

iocanel:
From the server side, there is nothing else to be done.
The part that I am not sure if exists, is the validation, which is ui.

I am going to throw the issue at @gashcrumb  to close, reassign or do whatever it needs.

gashcrumb:
I guess the remaining this is a syntax validator, that's kinda useful for a couple spots.  Created https://github.com/syndesisio/syndesis/issues/2189 I think we can close this user story issue though.

PR #1159, opened 2018-01-22
feat (#1095): First draft of LogStep.

iocanel:
This is a working draft of the LogStep.

It currently provides the ability to pass a simple expression for logging or to enable boy logging.


app/ui/src/app/store/step/step.store.ts
```diff
@@ -94,10 +93,15 @@ export class StepStore {
...
       properties: {
-        message: {
+        expression: {
           type: 'string',
           displayName: 'Log Message',
-          required: true,
+          required: false,
+        },
+        bodyLoggingEnabled: {
+          type: 'checkbox',
```
iocanel: gashcrumb: Do you have any idea why this appears as a text box?

lburgazzoli:
@iocanel be aware of #1144 which has huge impacts on this PR :)

app/ui/src/app/store/step/step.store.ts
gashcrumb: whoops, yeah do `type` boolean for now, need to make the form service also accept `checkbox` properly :-)

app/ui/src/app/store/step/step.store.ts
```diff
@@ -94,10 +93,15 @@ export class StepStore {
       description: "Sends a message to the integration's log",
       configuredProperties: undefined,
       properties: {
-        message: {
+        expression: {
           type: 'string',
```
gashcrumb: Might be worth changing this to `textarea`?  Then it'll be a multiline input.  Have a look at how the advanced filter step is set up for details.

app/ui/src/app/store/step/step.store.ts
```diff
@@ -94,10 +93,15 @@ export class StepStore {
...
           displayName: 'Log Message',
-          required: true,
+          required: false,
```
gashcrumb: Is a log step useful without a user-supplied message?

iocanel: Yes, in theory the user could just log the body of the message by clicking the checkbox. Does it make sense?

app/ui/src/app/store/step/step.store.ts
iocanel: Thanks! I will.

app/ui/src/app/store/step/step.store.ts
gashcrumb: Ah, of course, yep!

app/ui/src/app/store/step/step.store.ts
```diff
@@ -92,19 +91,52 @@ export class StepStore {
...
+        },
+        bodyLoggingEnabled: {
```
iocanel: @gashcrumb: This does work fine, but when I edit the checkbox always comes unchecked. Any clues?

gashcrumb: I would verify via the network tab in the browser what the step configuration looks like in the integration when it's fetched first and just make sure that the propery is set in `configuredProperties`

gashcrumb: From the [look of this](https://github.com/udos86/ng-dynamic-forms/blob/master/packages/core/src/model/dynamic-check-control.model.ts#L19) also worth verifying that the value is actually a boolean and not a string.

iocanel: I think that last time I used the network tab to inspect the value was string. Need to check again.

app/rest/project-generator/src/main/java/io/syndesis/project/converter/visitor/LogStepVisitor.java
```diff
@@ -0,0 +1,110 @@
...
+import io.syndesis.model.integration.Step;
+import io.syndesis.model.log.LogStep;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Locale;
+import java.util.stream.Collectors;
+
+
+public class LogStepVisitor implements StepVisitor {
```
chirino: I think this class might be obsolete now.  You should rebase off master.

app/ui/src/app/integration/edit-page/current-flow.service.ts
```diff
@@ -335,6 +335,11 @@ export class CurrentFlow {
         const action = event['action'];
         const properties = this.stringifyValues(event['properties']);
         const step = this.steps[position] || createStep();
+        for (const key of Object.keys(properties)) {
+          if (properties[key] === 'null') {
```
chirino: @gashcrumb this seems kinda smelly.  What if a user actually want to use the value 'null' for a property?

app/core/model/src/main/java/io/syndesis/model/log/LogStep.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+@Value.Immutable
+@JsonDeserialize(builder = LogStep.Builder.class)
+@SuppressWarnings("immutables")
+public interface LogStep extends Step {
```
chirino: So how come we need to add a new model object for log?  End users seem to be able to add steps to syndesis via extensions without changing the model.  Why can we do something similar?  @lburgazzoli  what do you think?

app/ui/src/app/integration/edit-page/current-flow.service.ts
gashcrumb: Yeah, it's a workaround really.  For some reason the forms now produce `"null"` instead of an empty string.  I've created https://github.com/syndesisio/syndesis/issues/1250 to track this down, not sure when this crept in.

app/core/model/src/main/java/io/syndesis/model/log/LogStep.java
lburgazzoli: I'd avid adding a specific class for such things as we are now in a situation were we have some untyped steps (endpoint) and some typed ones (*Filters) which make the model inconsistent so we should decide to go with all untyped or all typed.

To me this class looks like an helper class to extract things from getConfiguredProperties and the helpers can be added to LogStepHandler.

chirino: I vote to go as untyped as possible to keep it inline with what extensions are allowed to do.
@iocanel ??

chirino:
@iocanel this has failing tests.

chirino:
@iocanel, @gashcrumb looks like this just needs a few tweaks in the ui module.

gashcrumb:
lemme take a looksie...

gashcrumb:
hmm, lemme redo that rebase :-S

gashcrumb:
@iocanel weird, this is just what I get when I rebase, can you rebase and force-push your local copy?  I can always re-apply that last commit.

iocanel:
let me try to rebase

iocanel:
When I try to rebase this pull request I get a lot more conflicts than what's shown here. And they are all ui related files.

iocanel:
So, I'll resolve the conflicts in the UI and merge asap.

app/rest/project-generator/src/main/java/io/syndesis/project/converter/visitor/LogStepVisitor.java
lburgazzoli: @iocanel this class is not more needed, can we remove it ?

iocanel:
On 02.15.18, Luca Burgazzoli wrote:
>lburgazzoli commented on this pull request.
>
>
>
>> + * limitations under the License.
>+ */
>+package io.syndesis.project.converter.visitor;
>+
>+import io.syndesis.integration.model.steps.Log;
>+import io.syndesis.integration.support.Strings;
>+import io.syndesis.model.integration.Step;
>+import io.syndesis.model.log.LogStep;
>+
>+import java.util.Arrays;
>+import java.util.Collection;
>+import java.util.Locale;
>+import java.util.stream.Collectors;
>+
>+
>+public class LogStepVisitor implements StepVisitor {
>
>@iocanel this class is not more needed, can we remove it ?

We can. In fact I've removed it, at least twice so far. Not sure how it
keeps coming back.

iocanel:
Cherry picked commits into a new pull request. I am closing this one.

PR #1160, opened 2018-01-22
chore(rest): implement default icon generator

### LINKED ISSUE #705: Icon generator for custom connectors
When creating a new custom connector, for instance via [New API connector wizard](https://github.com/syndesisio/syndesis/blob/master/ux/designs/apiconnector/apiconnector.md#general-connector-info) we should generate a placeholder icon for that the user can use instead of uploading a icon.

We have implemented serving of icons in #475, now we need a initial icon generator. This could be a random generated icon, or it could be a favicon fetched from the remote API/website.

zregvart:
This adds a icon generator that generates a base64-encoded data uri for
the `icon` property of the `Connector` for the
`POST /connectors/custom/info` and `POST /connectors/custom` endpoints
if no `icon` property is given in the request body.

Icons are generated based on a SVG template and rendered to PNG.

Fixes #705

cc @deeleman


jimmidyson:
Why can't we just display them as SVG instead of PNG?

zregvart:
> Why can't we just display them as SVG instead of PNG?

Good question, and I would like to do that, this was more complex thing to do. How do the @syndesisio/ui-api feel about using SVG for icons?

jimmidyson:
Let's wait for an answer before I review.

gashcrumb:
if it's more complex let's stick with png for now and add support for svg at a later time.

jimmidyson:
> if it's more complex let's stick with png for now and add support for svg at a later time

It's more complex to do the png conversion - so if it's easy enough to display SVGs in the UI then we should stick with that and simplify this code.

zregvart:
@jimmidyson can you have another look I made the icons SVG-only and polished a bit

app/rest/connector-generator/pom.xml
```diff
@@ -99,6 +99,16 @@
...
     </dependency>
 
+    <dependency>
+      <groupId>com.github.spullara.mustache.java</groupId>
+      <artifactId>compiler</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
```
jimmidyson: Wonder can we do it without this extra dependency? Seems overkill for just escaping the SVG XML?

jimmidyson: A simple JS version that could be copied over perhaps? https://github.com/tigt/mini-svg-data-uri/blob/master/index.js

zregvart: I would not mind replacing that with something (that I did not code) but that won't remove the guava dependency from the backend, we have it in `connector-generator` and Swagger brings it in also

deeleman: Using SVG for icons is perfectly fine and we can embed those as regular `` elements and even apply transformations thru custom pipes if necessary.

If file format is an issue (which it shouldn't), a workaround is to grab whatever the backend returns (either a SVG, a PNG, a TIFF, a GIF, whatever), open a blank canvas object, embed the bitmap there (resize it if necessary), and then use the Canvas Web API to translate the canvas representation into binary data (Base64 encoded) and inject it as a PNG generated in memory into any `` tag. We can implement a custom a NG pipe so we can reuse this same logic globally, same as we did with that path from the connector icons.

Doing the above is trivial and just takes 30 minutes of fiddling with TypeScript code, but if the backend can return a path to a SVG file, we can get away with that.

zregvart:
@deeleman with this change if no `icon` was passed in `ConnectorSettings` to `POST /connectors/custom/info` and `POST /connectors/custom` endpoints, the `icon` property will be generated by the backend to contain a data URI with a SVG data.

Does that make sense for the UI?

app/rest/connector-generator/pom.xml
zregvart: @jimmidyson given that we already have guava dependency in our fat jar (brought in by Swagger) can we move forward with this without removing guava?

PR #1164, opened 2018-01-22
refactor: fixes #1163 Move the S3 connector to the new connector mode…

### LINKED ISSUE #1163: Move the S3 connector to the new connector model, like twitter-connector

oscerd:
…l, like twitter-connector


app/connectors/connectors/aws-s3-connector/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      
+      <plugin>
+        <groupId>pl.project13.maven</groupId>
+        <artifactId>git-commit-id-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
zregvart: Don't we want the git-id in the META-INF/MANIFEST.MF, is it breaking the build?

oscerd: No, I ran a full build.

oscerd: The pom, is exactly the same as the twitter-connector

PR #1171, opened 2018-01-22
Update Extension UX design to include Connector Extension, update pag…

dongniwang:
…e content

Updated UX design for Extensions. Updates include:
- Add a type column in list view to differtiate Step Extension and Connector Extension, allow users to filter based on type 
- Update Extension Details view to include type information and supported steps/actions
- Update import review page to include type information and supported steps/actions

Please review: @lburgazzoli @amysueg 

cc: @sjcox-rh @gashcrumb @kahboom


dongniwang:
@lburgazzoli thank you for the comments - I just updated the md file, there should be no more "tech extension", and I also added "or actions" in the review step. 

Please let me know if there's anything else missing. Thanks!

PR #1174, opened 2018-01-23
fix(rest): AWS connector dependency

zregvart:
We seem to be referencing AWS connector GAVs before refactor in #1163.


oscerd:
I don't think the dependency is needed in runtime, because twitter-connector is not there as dependency. @lburgazzoli is it right?

lburgazzoli:
Yeah it is not more needed. There is now a connectors-catalog that includes the bits needed by runtime.

oscerd:
Lets wait for the build to merge and then I'll open a PR to remove the depedency

oscerd:
I'm gonna open a PR to remove the dependency

PR #1176, opened 2018-01-23
fix(rest): removed aws-s3 connector dependency, since it is already i…

oscerd:
…n connectors-catalog


oscerd:
I think circleci is hanging.

zregvart:
> I think circleci is hanging.

There was an [outage](https://status.circleci.com/incidents/h3r64c7q74dt), I guess we could be at the back of the queue.

oscerd:
retest this please

PR #1184, opened 2018-01-23
chore: Handle errors that occur if we can’t decrypt stored values.

chirino:
This should help with this error I encountered the other day:
Caused by: javax.crypto.BadPaddingException: Given final block not properly padded

### DIFF
--- a/app/rest/dao/src/main/java/io/syndesis/dao/manager/EncryptionComponent.java
+++ b/app/rest/dao/src/main/java/io/syndesis/dao/manager/EncryptionComponent.java
@@ -24,6 +24,7 @@ import io.syndesis.model.integration.Step;
 import org.springframework.security.crypto.encrypt.TextEncryptor;
 import org.springframework.stereotype.Component;
 
+import java.security.GeneralSecurityException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


app/rest/dao/src/main/java/io/syndesis/dao/manager/EncryptionComponent.java
```diff
@@ -108,7 +109,12 @@ public String decrypt(final String value) {
         }
         String result = value;
         if( result.startsWith(ENCRYPTED_PREFIX) ) {
-            result = textEncryptor.decrypt(stripPrefix(result, ENCRYPTED_PREFIX));
+            try {
+                result = textEncryptor.decrypt(stripPrefix(result, ENCRYPTED_PREFIX));
+            } catch (RuntimeException e) {
```
zregvart: Could we catch GeneralSecurityException here instead?

chirino: TextEncryptor.decrypt() is not specific about what types of runtime execeptions it throws.

https://docs.spring.io/spring-security/site/docs/4.1.2.RELEASE/apidocs/org/springframework/security/crypto/encrypt/TextEncryptor.html#decrypt-java.lang.String-

zregvart: Perhaps we could do a `@SuppressWarnings` then instead, it's sad that it's not documented seems like it throws `IllegalArgumentException` and `IllegalStateException` from [CipherUtils.java]( https://github.com/spring-projects/spring-security/blob/1517e9b2221d8b3f4a86bd71db77cb10fadfaaf0/crypto/src/main/java/org/springframework/security/crypto/encrypt/CipherUtils.java)

What would be even better is to have a unit test confirming our exception handling in this case.

PR #1194, opened 2018-01-23
Add UX design for describe data as part of the action configuration

dongniwang:
Adding UX design based on the latest requirement of specifying data type as part of the action configuration. 

Also add the capability of letting users copy and paste when importing schema. 

Please review: @dhirajsb @kcbabo Thank you! 

cc: @amysueg @sjcox-rh @seanforyou23 @TovaCohen


kcbabo:
LGTM!

PR #1200, opened 2018-01-24
(chore) refactor restangular out of integration support service, pull…

gashcrumb:
… some cruft out of integration detail component

### DIFF
--- a/app/ui/src/app/integration/detail-page/detail.component.html
+++ b/app/ui/src/app/integration/detail-page/detail.component.html
@@ -1,8 +1,8 @@
 <syndesis-modal [title]="'Confirm ' + getActionTitle()"
                 [message]="'Are you sure you would like to ' + getAction() + ' the \'' + selectedIntegration?.name + '\' integration?'">
 </syndesis-modal>
-<syndesis-loading [loading]="i && loading | async">
-  <div *ngIf="i"
+<syndesis-loading [loading]="loading$ | async">
+  <div *ngIf="integration$ | async; let integration"
        class="row integration-detail">
     <div class=" toolbar-pf toolbar">
       <div class="col-sm-12">
@@ -32,9 +32,9 @@
     </div>
     <div class="col-md-12 details">
       <h1>
-        <syndesis-editable-text [value]="i.name"
+        <syndesis-editable-text [value]="integration.name"
                                 [validationFn]="validateName"
-                                (onSave)="onAttributeUpdated('name', $event)"></syndesis-editable-text>
+                                (onSave)="attributeUpdated('name', $event, integration)"></syndesis-editable-text>
       </h1>
       <dl class="dl-horizontal">
         <dt>
@@ -44,7 +44,7 @@
           <p>
             <ng-template ngFor
                          let-step
-                         [ngForOf]="i.steps"
+                         [ngForOf]="integration.steps"
                          let-index="index">
               <div *ngIf="step"
                    class="step-block inline-block text-center">
@@ -76,9 +76,9 @@
             </ng-template>
           </p>
           <p>
-            <syndesis-editable-textarea [value]="i.description"
+            <syndesis-editable-textarea [value]="integration.description"
                                         placeholder="No description set..."
-                                        (onSave)="onAttributeUpdated('description', $event)"></syndesis-editable-textarea>
+                                        (onSave)="attributeUpdated('description', $event, integration)"></syndesis-editable-textarea>
           </p>
         </dd>
       </dl>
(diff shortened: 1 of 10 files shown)


app/ui/src/app/integration/detail-page/detail.component.html
```diff
@@ -32,9 +32,9 @@
     </div>
     <div class="col-md-12 details">
       <h1>
-        <syndesis-editable-text [value]="i.name"
+        <syndesis-editable-text [value]="integration.name"
                                 [validationFn]="validateName"
-                                (onSave)="onAttributeUpdated('name', $event)"></syndesis-editable-text>
+                                (onSave)="onAttributeUpdated('name', $event, integration)"></syndesis-editable-text>
```
deeleman: I know this falls outside the scope of this refactor, but just a note for the future: Angular enforces a non `onXXXX` naming convention for output emitter properties. The reason is because Angular transpiles these properties by appending an `on-` to the generated method handler, so this would eventually become `on-onSave`.

Not relevant for this PR but just a word of caution for the future ;)

app/ui/src/app/integration/integration-support.api.ts
```diff
@@ -0,0 +1,16 @@
...
+  // TODO should this go into the integration service
+  filterOptions: '/integrations/filters/options',
+  history: '/integrations/{id}/history',
+  deployments: '/integrations/{id}/deployments',
+  deployment: '/integrations/{id}/deployments/{version}',
+  pom: '/integration-support/generate/pom.xml',
+  export: '/integration-support/export.zip',
+  import: '/integration-support/import',
+  // TODO this path should be driven by config.json also does this belong here
+  javaInspection: '../../mapper/v1/java-inspections/{connectorId}/{type}.json',
```
deeleman: Have you double checked wether this path works. With our current setup, the `ApiHttpService` will prepend the `apiHost` and `apiEndpoint` details and `javaInspection` will eventually turn into a call to:

`https://syndesis.whatever.com/api/v1../../mapper/v1/java-inspections/{connectorId}/{type}.json`

Please note the lack of slash between `v1` and `../..`. Also, this is a bit ugly either way. Perhaps we should update the `setEndpointUrl` method to allow for custom `apiEndpoint` values (an options parameter perhaps?) so we can bind `mapperEndpoint` by hand.

For the time being let's focus on checking if this endpoint works like this (with no leading slash) and then refactor on a later stage.

app/ui/src/app/integration/integration-support.service.ts
```diff
@@ -1,93 +1,67 @@
...
+  getHistory(id: string): Observable<any> {
+    return this.apiHttpService.setEndpointUrl(integrationSupportEndpoints.history, { id: id }).get();
+  }
+
+  getDeployments(id: string): Observable<any> {
+    return this.apiHttpService.setEndpointUrl(integrationSupportEndpoints.deployments, { id: id }).get();
+  }
+
+  getDeployment(id: string, version: string): Observable<any> {
+    return this.apiHttpService.setEndpointUrl(integrationSupportEndpoints.deployment, { id: id, version: version }).get();
```
deeleman: Just a quick note, for future refactors like this: In object literal parameters whose properties and mapped variables do match in their naming, we can shorten things up a bit with this syntax:

`{ id: id, version: version }` can become `{ id, version }`

app/ui/src/app/integration/detail-page/detail.component.html
gashcrumb: I might as well fix this one now then, tx!

app/ui/src/app/integration/integration-support.api.ts
gashcrumb: Yeah, I think that's a good idea, there's always a potential we'll have calls that don't fall into `/api/v1` or even have `/api/v1beta1` etc.  I'll add the `/` and actually have to adjust the code calling this anyways, so thanks for the heads up!

I kinda think BTW we might consolidate calls to the data mapper stuff into it's own service in a future refactor, WDYT?

app/ui/src/app/integration/integration-support.service.ts
gashcrumb: I'll do that now too, thanks!

app/ui/src/app/integration/integration-support.api.ts
gashcrumb: Just so we don't forget (I know I will) created https://github.com/syndesisio/syndesis/issues/1211

deeleman: I added my comments on #1211 and I also think we should update `setEndpointUrl` to automatically append that slash if missing, so the concatenated URL is legal

PR #1202, opened 2018-01-24
refactor(ui): Ordered connection options in aws s3 connector

oscerd:


app/ui/src/app/connections/common/configuration/configuration.service.ts
```diff
@@ -44,6 +44,9 @@ export class ConnectionConfigurationService {
       case 'activemq':
         controls = ['brokerUrl', 'username', 'password', 'clientId', 'skipCertificateCheck', 'brokerCertificate', 'clientCertificate'];
         break;
+      case 'aws-s3':
+        controls = ['accessKey', 'secretKey', 'region', 'bucketNameOrArn'];
```
gashcrumb: I think this is fine for now, but we should get this field set at the backend, created https://github.com/syndesisio/syndesis/issues/1204

oscerd: Yeah, absolutely. It's the only way I know/found for the moment :-)

gashcrumb: No worries, should have raised that issue earlier :-)

PR #1222, opened 2018-01-24
fix: Issues with deactivating and renaming an integration etc.

iocanel:


app/core/core/src/main/java/io/syndesis/core/KeyGenerator.java
```diff
@@ -59,7 +59,7 @@ public static String createKey() {
         buffer.putLong(getRandomPart(now));
 
         try {
-            return Base64.encodeBytes(buffer.array(), 2, 15, Base64.ORDERED);
+            return Names.sanitize(Base64.encodeBytes(buffer.array(), 2, 15, Base64.ORDERED));
```
chirino: This breaks the KeyGeneratorTest.  Generated keys are supposed to be lexically incrementing always.

chirino: I think the only thing you need to do to keys to make them valid labels in k8s is prefix them with a letter.

gashcrumb:
@iocanel wanna maybe mark this wip and I can push some UI work to this branch?

gashcrumb:
took the liberty, have the deployments listed, WIP.

chirino:
rebased off master to see if that helps CI.

app/core/core/src/main/java/io/syndesis/core/Labels.java
```diff
@@ -0,0 +1,82 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.core;
+
+import java.util.Locale;
+
+public final class Labels {
```
zregvart: I think we could unit test this quite easily. Any reason on not using/extending the existing class? https://github.com/syndesisio/syndesis/blob/814d0cdf3b443adf843f0fe1e0941331ae6a1068/app/core/core/src/main/java/io/syndesis/core/Names.java#L20

app/core/model/src/main/java/io/syndesis/model/integration/Integration.java
```diff
@@ -51,6 +51,7 @@ default Kind getKind() {
         return Kind.Integration;
     }
 
+    //TODO: This is actually the deployment version and not the deployment. At some point we need to rename it.
     Optional<Integer> getDeploymentId();
```
zregvart: Why not rename it right away?

app/rest/controllers/src/main/java/io/syndesis/controllers/integration/IntegrationController.java
```diff
@@ -175,25 +175,26 @@ private String getLabel(IntegrationDeployment integrationDeployment) {
...
-                    dataManager.update(updated);
+                    if (update.getState() == IntegrationDeploymentState.Undeployed) {
+                        dataManager.delete(IntegrationDeployment.class, integrationDeploymentId);
+                    } else {
+                        // TODO: do this in a single TX.
+                        Date now = new Date();
+                        IntegrationDeployment current = dataManager.fetch(IntegrationDeployment.class, integrationDeploymentId);
+                        IntegrationDeployment updated = new IntegrationDeployment.Builder()
+                            .createFrom(current)
+                            //.statusMessage(Optional.ofNullable(update.getStatusMessage()))
```
zregvart: Let's remove commented out code

app/rest/controllers/src/main/java/io/syndesis/controllers/integration/online/ActivateHandler.java
```diff
@@ -104,7 +104,7 @@ public StateUpdate execute(IntegrationDeployment integrationDeploymentDefinition
             stepPerformer.perform("build", this::build, deploymentData);
             stepPerformer.perform("deploy", this::deploy, deploymentData);
         } catch (@SuppressWarnings("PMD.AvoidCatchingGenericException") Exception e) {
-            logError(integrationDeployment,"[ERROR] Activation failure");
+            logError(integrationDeployment, "[ERROR] Activation failure");
```
zregvart: Since we're catching `Exception` would it be better to pass the `Exception e` to `logError` to get the whole exception with stack trace logged?

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -130,12 +131,13 @@ public static void addEntry(ZipOutputStream os, String path, byte[] content) thr
...
-        return getDataManager().fetchAll(
-            Integration.class,
+        List<Integration> integrations =  getDataManager().fetchAll(Integration.class,
             new DeletedFilter(),
             new ReflectiveSorter<>(clazz, new SortOptionsFromQueryParams(uriInfo)),
             new PaginationFilter<>(new PaginationOptionsFromQueryParams(uriInfo))
-        );
+        ).getItems().stream().map(updateCurrentState).collect(Collectors.toList());
+
+        return new ListResult.Builder<Integration>().addAllItems(integrations).totalCount(integrations.size()).build();
```
zregvart: `ListResult.of(integrations)` would be simpler

gashcrumb:
rebased...

PR #1224, opened 2018-01-24
Initial version of AMQP Connector

dhirajsb:


dhirajsb:
@zregvart is there a profile for running codacy tests locally? it's annoying to have to push changes to try them out. Also, is there a magic comment workaround for a genuinely empty catch block?

app/connectors/connectors/jms/amqp-connector/src/main/java/io/syndesis/connector/amqp/AMQPVerifier.java
```diff
@@ -0,0 +1,42 @@
...
+
+    @Override
+    public List<VerifierResponse> verify(CamelContext context, String connectorId, Map<String, Object> params) {
+        // create and register amqp-publish component, if needed
+        if (context.getComponent("amqp-publish") == null) {
+            context.addComponent("amqp-publish",
+                    new AMQPConnectorFactory().newInstance(connectorId, "amqp"));
+        }
+        return super.verify(context, connectorId, params);
+    }
```
lburgazzoli: @dhirajsb I've relaxed access to ComponentVerifier's methods so you do not need this "hack" anymore but you can just implmented `resolveComponentVerifierExtension`, see  https://github.com/syndesisio/syndesis/blob/master/app/connectors/connector-verifier/src/main/java/io/syndesis/verifier/api/ComponentVerifier.java#L93

zregvart:
> @zregvart is there a profile for running codacy tests locally? it's annoying to have to push changes to try them out. Also, is there a magic comment workaround for a genuinely empty catch block?

Codacy checks work by either taking the default configuration it provides or fetching well known files from the root of the repository, with the move to monorepo we lost that as the configuration files ended up in `app/rest`. Until that's fixed the only alternative is to enable Codacy on your own fork.

dhirajsb:
@lburgazzoli changed the verifier based on your changes and rebased. @zregvart have you added a comment pattern for empty catch blocks?

PR #1227, opened 2018-01-24
Assign ids to steps when they are created.

chirino:


deeleman:
In order to [create a simple unique hash](https://github.com/cartant/firebase-key/blob/master/source/key-key.ts), we're bringing ANOTHER dependency to the dependency graph, which we will force us to maintain, keep an eye on and ensure it does not collide with any other dependency in the future, not to mention that bundling this new library (which encompass other 15 function modules we won't be using and is mostly geared to Firebase-based products) with our other libraries will increase the overall application filesize footprint until we implement a proper tree-shaking filter (which is not in the agenda for the time being).

I'd recommend to strip out the code from that function so we can better keep the codebase in check it in the future and, if we're just after a unique numeric ID (this is, not an alphanumeric one like the one provided by the function above), use something so simple as:

```javascript
const key = () => new Date().getUTCMilliseconds();
```

Do you want it to be alphanumeric? No worries...

```javascript
function key(): string {
  const seedKey = Math.random().toString(36).split('');
  const uniqueKey = [...new Set(seedKey)].join('').substr(2, seedKey.length);
  return uniqueKey;
}

```

We need to come to terms with the idea that JavaScript and TypeScript (plus Angular) give us pretty much all what we need to build this product and that we need to keep the dependency graph in check from now on. Bringing new dependencies to the game just to save us from a couple of lines of code is not a good idea. ;-)

chirino:
We could use something as simple as a timestamp if we don't think our UI will be getting exercised by bots.  Otherwise it is a possibility to generate two steps with the same id.  Randoms are not guaranteed to be unique.  

I picked that one because the rest of syndesis uses firebase style keys.  Since it has no dependencies I figured it's low risk.  It's a nice safe, simple to use guid.  But perhaps I'm over engineering.

chirino:
You guys can feel free to replace the key() function with anything you guys feel is more suitable.   But plz merge this in soon as I need it for demos Friday.

chirino:
@gashcrumb isn't a service overkill? We just need a function that generates keys.  It's stateless and I doubt we need to inject different impls.

deeleman:
Services are not meant to be stateful classes, just singletons that are available through Angular's DI mechanism. With this said, services can be just wrappers around helper functions, proxies over setting objects, etc. And we do not even need to rely on Angular for this: We can just create a standalone function and export it from `@sydnesis/ui/platform`.

Angular itself is a good example of a framework filled up with stateless services and exportable standalone functions that expose convenience methods to do things everywhere.

chirino:
Ok. Did a roll your own impl of the key() function at: 

https://github.com/chirino/syndesis/blob/fb7662cdfc03d692cb47e6ef3d65fa119810506e/app/ui/src/app/core/key.ts

gashcrumb:
Tx!

PR #1233, opened 2018-01-25
Allow OAuth bearer token authentication against the backend API

zregvart:
We need to allow non UI clients to access our backend APIs as for some of them it is cumbersome to use OAuth flow and Cookies. This is difficult for our IDE integration to do, for example.

oauth-proxy can be configured to also check if the `Authorization` header of the incoming request contains an OAuth bearer token and check if that token is valid to pass on the request to the API.

This configures oauth-proxy when running a _restricted_ variant of the templates to that end.

Caveat is that in this configuration oauth-proxy is not pasing the `X-Forwarded-Access-Token` header, and we will loose OpenShift OAuth token that we require (for some endpoints).

When `Authentication` with `null` credentials is passed on to the `PreAuthenticatedAuthenticationProvider` it fails to authenticate the request.

This adds a fixed credentials `"<missing>"` so that the authentication provider can authenticate and pass the request through.

The drawback of this is that the `Authentication` stored in the `SecurityContext` will not have access to OpenShift OAuth token (it will be `"<missing>"`), and there is no way of fixing that unless we modify oauth-proxy to include it for requests with Bearer token.

### DIFF
--- a/app/deploy/generator/04-syndesis-oauth-proxy.yml.mustache
+++ b/app/deploy/generator/04-syndesis-oauth-proxy.yml.mustache
@@ -141,15 +141,3 @@
       type: ImageChange
 {{/WithDockerImages}}
     - type: ConfigChange
-{{#Restricted}}
-- apiVersion: v1
-  kind: ClusterRoleBinding
-  roleRef:
-    name: system:auth-delegator
-  subjects:
-  - kind: ServiceAccount
-    name: syndesis-oauth-client
-    namespace: ${OPENSHIFT_PROJECT}
-  userNames:
-  - system:serviceaccount:${OPENSHIFT_PROJECT}:syndesis-oauth-client
-{{/Restricted}}
--- a/app/deploy/generator/04-syndesis-oauth-proxy.yml.mustache
+++ b/app/deploy/generator/04-syndesis-oauth-proxy.yml.mustache
@@ -121,7 +121,6 @@
             requests:
               memory: 20Mi
 {{#Restricted}}
-        serviceAccount: syndesis-oauth-client
         serviceAccountName: syndesis-oauth-client
 {{/Restricted}}
         volumes:
--- a/app/deploy/generator/04-syndesis-oauth-proxy.yml.mustache
+++ b/app/deploy/generator/04-syndesis-oauth-proxy.yml.mustache
@@ -82,7 +82,10 @@
             - --skip-auth-preflight
             - --openshift-ca=/etc/pki/tls/certs/ca-bundle.crt
             - --openshift-ca=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
-            {{#Restricted}}- --openshift-sar={"namespace":"${OPENSHIFT_PROJECT}","resource":"pods","verb":"get"}{{/Restricted}}
+{{#Restricted}}
+            - --openshift-sar={"namespace":"${OPENSHIFT_PROJECT}","resource":"pods","verb":"get"}
+            - --openshift-delegate-urls={"/api":{"resource":"projects","verb":"get","resourceName":"${OPENSHIFT_PROJECT}","namespace":"${OPENSHIFT_PROJECT}"}}
+{{/Restricted}}
           env:
           - name: OAUTH_COOKIE_SECRET
             valueFrom:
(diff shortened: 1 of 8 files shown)


zregvart:
So I did the changes locally manually, I'm not 100% sure that that ClusterRoleBinding will be applied (I had to use `oc adm policy add-cluster-role-to-user` locally.

dsimansk:
@zregvart without `cluster-admin` rights you can't create `ClusterRoleBinding`. I wonder if we should limit this change only to `syndesis-dev-restricted` flavour. Because we reference `syndesis-restricted` in quickstart guide for example, without explicit requirement for `cluster-admin`.

zregvart:
> @zregvart imho without cluster-admin rights you can't create ClusterRoleBinding. I wonder if we should limit this change only to syndesis-dev-restricted flavour. Because we reference syndesis-restricted in quickstart guide for example, without explicit requirement for cluster-admin.

yeah that's what I was afraid of, not sure how to do this otherwise, `oauth-proxy` requires this permission, we could just add a note saying that the user needs to:

    oc login -u system:admin
    oc project $SYNDESIS_PROJECT
    oc adm policy add-cluster-role-to-user system:auth-delegator -z syndesis-oauth-client

Would that be better?

dsimansk:
Hmm you could add the set of cmds to syndesis script and avoid templates completely, right?

paoloantinori:
Hi @zregvart , looking at my old worfklow here: https://github.com/syndesisio/syndesis/issues/99

```
curlsyn() {
	OS_TOKEN=$(oc whoami -t);
    curl -L --insecure -N -H "Authorization: Bearer $OS_TOKEN" -H "X-Forwarded-Access-Token: $OS_TOKEN"  2> /dev/null "$@"
}
```

Does an approach like that cover your user scenario or is there still something that's not there?
Are you looking for a way to set just a single header instead of two?

zregvart:
@paoloantinori totally does, but it binds the client to abide with the way oauth-proxy sends the request to the backed (by adding `X-Forwarded-Access-Token`).
I would not want to put that burden on the client as if we change this behavior, i.e. by re-configuring or replacing oauth-proxy the client would need to change also.

This whole issue is really about other clients accessing our APIs and we should have as _standard_ or _expected_ experience for them.

paoloantinori:
I see. Have you tried to check in debug which headers reach the java application? I did last time but I can't remember if there was still reference to anything useful.
I was thinking if that's the case, we can write a filter that takes care to bridge tokens values from one key to another

zregvart:
Yeah we end up just with `Authorization` header and `developer:` (i.e. no password) there. I've created a PR https://github.com/openshift/oauth-proxy/pull/53 so we can receive `X-Forwarded-Access-Token` from the client's bearer token, until we can do that I'm fine with some endpoints not working as the token is unavailable.

The first focus of this is for the extension endpoints that the Fuse IDE team needs to access, later it might a nicer way of excising the backend in end to end tests.

chirino:
@zregvart are they modification to oauth-proxy that would help keep it more consistent?

chirino:
@zregvart nevermind, I just read up on your PR over at oauth-proxy.

app/deploy/generator/04-syndesis-oauth-proxy.yml.mustache
```diff
@@ -117,6 +120,10 @@
               memory: 200Mi
             requests:
               memory: 20Mi
+{{#Restricted}}
+        serviceAccount: syndesis-oauth-client
+        serviceAccountName: syndesis-oauth-client
```
dsimansk: A small nitpicking: `serviceAccount` has been deprecated in favour of `serviceAccountName`, see [the spec](https://docs.openshift.com/container-platform/3.6/rest_api/kubernetes_v1.html#v1-podspec). Imho we don't need both.

zregvart:
With the change to add `-pass-user-bearer-token` to oauth-proxy we can now remove the addition of `<missing>` as the OAuth token. The oauth-proxy will now pass the received bearer token from the client.

chirino:
Very nice.

zregvart:
@lhein this would enable the API use in a stateless way, you would only need the OpenShift OAuth token for the user. So something like this is possible:

    curl -k -H "Authorization: Bearer $(oc whoami -t)" https://syndesis.$(minishift ip).nip.io/api/v1/version

Not sure if you can use an endpoint on the OpenShift side to fetch the token by authenticating using username/password first. I think that JBoss Tools does this somehow (from screenshots I've seen), so you might want to check there.

lhein:
@apupier any clue?

apupier:
yes there are doing like that.
When creating/editing an OpenShift Connection:
image
Click on the retrieve link
provide username/password and click "Log In"
image
then click "Close"

PR #1237, opened 2018-01-25
UI, fixed api-connector refactoring

### LINKED ISSUE #1236: [UI] Issue in API Client Connectors page
Opening `/customizations/api-connector` throws an exception in browser console.

```
 https://syndesis.192.168.42.160.xip.io/api/v1getApiConnectorList 404 (Not Found)
```

It seems to miss a `/` after `v1` token

paoloantinori:
fixes #1236

### DIFF
--- a/app/ui/src/app/customizations/api-connector/api-connector.service.ts
+++ b/app/ui/src/app/customizations/api-connector/api-connector.service.ts
@@ -4,25 +4,28 @@ import { Observable } from 'rxjs/Observable';
 import { ApiHttpService } from '@syndesis/ui/platform';
 import { ApiConnectors, ApiConnectorData, CustomConnectorRequest } from './api-connector.models';
 
+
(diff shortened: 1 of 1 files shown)


app/ui/src/app/customizations/api-connector/api-connector.service.ts
```diff
@@ -4,25 +4,28 @@ import { Observable } from 'rxjs/Observable';
 import { ApiHttpService } from '@syndesis/ui/platform';
 import { ApiConnectors, ApiConnectorData, CustomConnectorRequest } from './api-connector.models';
 
+
+import { apiConnectorEndpoints } from '../../customizations/api-connector/api-connector.api';
```
deeleman: This is fine, but you better use `import { apiConnectorEndpoints } from  './api-connector.api';` instead.

This way the service will keep pointing to the right file location even if we move the entire directory or change its name.

paoloantinori:
done. it's still not clear to me which paths are relatives and which not

PR #1238, opened 2018-01-25
chore: remove runtime module remnants

zregvart:
This removes the orphan `runtime` module.


lburgazzoli:
@chirino I guess we need to port this test to the new style, correct ?

PR #1240, opened 2018-01-25
Add icon to extension

### LINKED ISSUE #1177: [Custom connectors] Allow extension developers to provide a icon for connectors
A icon can be embedded in the jar file and then used by syndesis when displaying the connector.

nicolaferraro:
Fixes #1177.

Extension developers can put a `icon.png` or `icon.svg` file in `META-INF/syndesis` and it will be picked up as icon for the connector/connection.

I've used the existing connector icon service to lookup the icon directly from the jar in the filestore when the icon string starts with `extension:`.

I've not seen a big performance improvement in extracting it in advance (on extension import) and putting it somewhere for faster access, so I've left the icon inside the jar.

I'll publish an example in https://github.com/syndesisio/syndesis-extensions shortly.


lburgazzoli:
@nicolaferraro can you squash the commits ?

PR #1252, opened 2018-01-26
UX for system metrics

sjcox-rh:
Please review designs for system wide metrics.


ux/designs/monitormetrics/monitormetrics.md
```diff
@@ -0,0 +1,26 @@
...
+## System Metrics
+
+![Image of choosing connection](img/dashboard.png)
+
+1. System wide metrics will be shown on the Home page of the application.
+1. Status aggregate cards will be used to show high level metrics.
+  - **Integrations**: Card shows number of total active integrations. Option to break down integrations via success vs. error.
+  - **Connections**: Shows total number of available connections. Optional - click on text to drill down to connection page.
+  - **Messages**: Show total number of messages. Option to break down via success vs error.
+  - **Uptime**: Shows time that current pod is running since last restart
```
rhuss: I think in the 'uptime' card there should be no time series graph as we do not record actually time series data (this could be added later if we go with the prometheus approach)

deeleman:
@KurtStam @chirino Probs you'll want to review this

ux/designs/monitormetrics/monitormetrics.md
sjcox-rh: @rhuss Makes sense. We can remove the time series graph. Keep the rest of the content? Such as the date/time started and length of time.

KurtStam:
Looks good to me.

deeleman:
@KurtStam Since this UX is now approved let's have a chat as for Monday about how can we accommodate all those UI widgets with the information provided by the API, thx.

KurtStam:
@deeleman Sounds good. I've send out an invite.

PR #1254, opened 2018-01-26
WIP!! Import integrations initial setup, first pass

kahboom:
Not that you would, but please don't merge this in yet. Quite a bit of outstanding work; leaving open for collaboration and just to show what I have so far.

Some issues:
- Resolve merge conflicts.
- Type errors resulting from incorrect, well, typings and issues naming things. And also it's 3:30am.
- Still getting used to effects, redux containers, and deciding whether or not to have a store for certain things, so these areas need to be worked on a bit.
- Naming! Importing and uploading seemed sufficient, but I wasn't sure whether or not to include something as detailed as ImportIntegrationImport or ImportIntegrationUpload.
- Templating! I'm sure the app won't build properly because of the aforementioned reasons.
- Some confusion about where there should be overlap between import and exporting, and well as multi versus single import.
- No proper tests.
- Among other things.

I've taken some parts of the API Custom Connector and adjusted them for this particular feature, and that is still undergoing work. Reducer and actions have been written accordingly, but likely contain issues. Tried to check the Swagger JSON on staging with not much luck for data structure, and I don't think we have design proposals anymore, so I guess I'll see about that tomorrow.

@deeleman - Please take a look, but also don't kill me. This has a long way to go.


app/ui/src/app/import-export/import-export.component.scss
```diff
@@ -0,0 +1,24 @@
+:host {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  overflow: visible;
+}
+
+ul {
```
deeleman: You do not need this `ul` styling anymore. Jas add the `syn-tabs` class to the navbar and some additional classnames to the listitems and you're all sel. 

Double check the example in the Monitor metrics for reference :)

app/ui/src/app/import-export/import-export.module.ts
```diff
@@ -0,0 +1,41 @@
...
+        path: 'import',
+        component: ImportIntegrationComponent
+      },
+      {
+        path: '',
+        redirectTo: 'import-export'
+      }
+    ]
+  }
+];
```
deeleman: It' usually a good practice to promote routing to its own routing module, and import it inot the gneeral domain module. Check the Metrics module for an example and download the John Papa's Angular Snippets for autogenerating that code in your IDE

app/ui/src/app/import-export/import/import.actions.ts
```diff
@@ -0,0 +1,117 @@
...
+  static IMPORT_FAIL = '[Import Integration] Import integration failed';
+  static UPLOAD = '[Import Integration] Upload integration for import request';
+  static UPLOAD_CANCEL = '[Import Integration] Upload integration for import cancelled';
+  static UPLOAD_COMPLETE = '[Import Integration] Upload integration for import complete';
+  static UPLOAD_FAIL = '[Import Integration] Upload integration for import failed';
+
+
+  /**
+   * Not actually sure that we would need fetching here.
+   */
```
deeleman: It's always better to start with what you really need and then go adding stuff as you need it. Otherwise reducers can become massive and kinda confusing ;)

So, if you see there're actions you're not fully confident that you'll want to use, remove those

app/ui/src/app/import-export/import/import.actions.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  static uploadComplete(payload: any): IntegrationUploadComplete {
+    return new IntegrationUploadComplete(payload);
+  }
+
+  static uploadCancel(): IntegrationUploadCancel {
+    return new IntegrationUploadCancel();
+  }
+}
```
deeleman: The idea for creating static methods in a Actions class to dispatch actions more easily eventually proved to generate too much boilerplate. Perhaps is better to use Action classes as is. Again, you can see the leaner approach I took for the metrics domain asa reference.

app/ui/src/app/import-export/import/import.component.scss
```diff
@@ -0,0 +1,25 @@
...
+  flex-direction: column;
+  height: 100%;
+  overflow: visible;
+}
+
+syndesis-list-toolbar {
+  flex: 0 0 auto;
+}
+
+syndesis-loading {
```
deeleman: Isn't this piece of SASS used elsewhere? If so, perhaps it would be a good idea to promote it to `src/scss/components/_loading.scss`

app/ui/src/app/import-export/import/import.component.ts
```diff
@@ -0,0 +1,68 @@
+import { Component, OnInit } from '@angular/core';
+import { Observable } from 'rxjs/Observable';
+import { NotificationType } from 'patternfly-ng';
+import { IntegrationStore, IntegrationSupportService } from '@syndesis/ui/store';
+import { Integrations } from '@syndesis/ui/model';
+import { NotificationService } from '@syndesis/ui/common';
+import {
+  FileUploader,
+  FileItem
+} from 'ng2-file-upload';
```
deeleman: No need for a breakline, it's just two classes ;P

app/ui/src/app/import-export/import/import.component.ts
```diff
@@ -0,0 +1,68 @@
...
+
+@Component({
+  selector: 'syndesis-import-integration-component',
+  templateUrl: './import-integration.component.html',
+  styleUrls: ['./import-integration.component.scss']
+})
+export class ImportIntegrationComponent implements OnInit {
+  public uploader: FileUploader;
```
deeleman: In TypeScript all members are public by default, so this accessor is redundant. not wrong, but redundant.

app/ui/src/app/import-export/import/import.component.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  loading: Observable<boolean>;
+  integrations: Observable<Integrations>;
```
deeleman: I'd suggest to conform to the `loading$` or `integrations$` naming convention.

app/ui/src/app/import-export/import/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './import.component';
+export * from './import.actions';
+export * from './import.reducer';
+export * from './import.models';
+export * from './import.module';
```
deeleman: models, actions and reducer are most likely unnecessary outside the scope of this module, so there is no need to expose those.

app/ui/src/app/import-export/import/import.reducer.ts
```diff
@@ -0,0 +1,142 @@
...
+  IntegrationFetchComplete,
+  IntegrationFetchFail,
+  IntegrationUpload,
+  IntegrationUploadComplete,
+  IntegrationUploadCancel,
+  IntegrationImport,
+  IntegrationImportComplete,
+  IntegrationImportFail,
+} from './import.actions';
+
```
deeleman: `import * as ImportActions from './import.actions';`is simpler and more scalable, so you won't have to import each new action yo create in the future.

gashcrumb:
@kahboom run `yarn build:ci` locally to catch any AOT issues, speeds up fixing them a bit.

app/ui/src/app/integration/import-export/integration-import-export.component.html
```diff
@@ -0,0 +1,7 @@
+<ul class="row nav nav-tabs nav-tabs-pf toolbar-pf">
+  <li routerLinkActive="active">
+    <a [routerLink]="['api-connector']">Import Integration</a>
```
chirino: 'api-connector' seems like an odd link for import integration.

app/ui/src/app/integration/import-export/import/integration-import.service.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Observable } from 'rxjs/Observable';
+
+import { ApiHttpService } from '@syndesis/ui/platform';
+import { IntegrationImportData, IntegrationImportRequest } from './integration-import.models';
+
+@Injectable()
+export class IntegrationImportService {
+  constructor(private apiHttpService: ApiHttpService) { }
+
+  getImport(id: string): Observable<IntegrationImportData> {
```
chirino: Don't think we need ti s class, we can't upload then review for imports.  Once the file is summitted to the importer, your done.  It's imported.  You just need to display the import results.

app/ui/src/app/integration/import-export/import/integration-import.component.ts
```diff
@@ -0,0 +1,65 @@
...
+      url: integrationSupportService.importIntegrationURL(),
+      disableMultipart: true,
+      autoUpload: true
+    });
+    this.uploader$.onCompleteItem = (
+      item: FileItem,
+      response: string,
+      status: number
+    ) => {
+      if (status === 204) {
```
chirino: Don't you have to take this result and put it in the result page of the wizard instead of poping up notifications?

app/ui/src/app/integration/import-export/integration-import-export.component.html
kahboom: This is copy-pasta.

app/ui/src/app/integration/import-export/import/integration-import.service.ts
kahboom: I am removing the `fetch`-related Redux actions, hence this being a WIP.

app/ui/src/app/integration/import-export/import/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './integration-import.component';
+export * from './integration-import.module';
+export * from './integration-import.actions';
+export * from './integration-import.models';
+export * from './integration-import.reducer';
```
deeleman: You're already importing and bootstrapping the Reducer at `integration-import.module`, hence there's no need to export it from here. all the way around, this might end up in a circular reference. Same applies to `integration-import.component` above.

app/ui/src/app/integration/import-export/import/integration-import.actions.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  constructor(public payload: IntegrationImportUploadState) { }
+}
+
+/**
+ * Step 2: Integration Edit
+ * Once the integration has been uploaded, it is available for review.
+ * Once the user has reviewed the payload, they can then press 'Done' to
+ * proceed to the Edit Integration view, which will be populated.
+ */
```
deeleman: You're describing a UI interaction here. Actions are meant to be UI-agnostic. They just deal with data/state operations.

app/ui/src/app/integration/import-export/import/integration-import.component.html
```diff
@@ -0,0 +1,21 @@
...
+        </button>
+    </div>
+  </ng-template>
+</syndesis-list-toolbar>
+
+<ng-template>
+  Select the integration file to import:
+  <form class="syn-form syn-form--integration-upload"
+        (ngSubmit)="onSubmit(integrationImportForm, uploadFileOption.checked)"
+        #swaggerForm="ngForm"
```
deeleman: Is the integration import file coming as a swagger file?

app/ui/src/app/integration/import-export/import/integration-import.component.scss
```diff
@@ -0,0 +1,25 @@
...
+  padding-top: 20px;
+  padding-bottom: 20px;
+  // Dock the scrollbar to the right
+  margin-right: -20px;
+  padding-right: 20px;
+}
+
+.max-height {
+  height:100%
+}
```
deeleman: This chunk of SASS basically duplicates to the point what we already had at `app/integration/list-page/list-page.component.scss`. This kind of code duplication must be avoided by all means, since it only increases the overall footprint of the application bundle and prevents CSS reusability and maintainability. You have several workarounds for this:

- Move this chunk of code to a shared location and import it from both components (remember that the `styleUrls` property expects an array with not one but many references).
- Turn this piece of code into a SASS partial and import it from each component's own stylesheet, should you feel that each component will eventually feature its own CSS nuances.
- Move parts of the code to SASS `%` extension classnames and use them internally within these selectors.
- Use a global mixin and import it directly.

In addition to the approaches above, you should strive to stick to the app variables we defined at `/scss/base/_variables.scss` in order to keep consistency in the the UI's look and feel. So whenever you're dealing with component CSS variables, you should strive to avoid hardcoded values as much as you can.

app/ui/src/app/integration/import-export/import/integration-import.models.ts
```diff
@@ -0,0 +1,36 @@
...
+  file?: File;
+  importResults: {
+    integrations?: string[];
+    connections?: string[];
+  };
+  list: IntegrationImportsEditState;
+}
+
+export type IntegrationImportsEditState = Array<IntegrationImportEditState>;
+
```
deeleman: What is the point of defining two different state interfaces for the same schema? And two types for an interface array with different names?

app/ui/src/app/integration/import-export/import/integration-import.effects.ts
```diff
@@ -0,0 +1,34 @@
...
+  IntegrationUploadComplete,
+  IntegrationImport,
+  IntegrationImportComplete
+} from './integration-import.actions';
+
+@Injectable()
+export class IntegrationImportEffects {
+  @Effect()
+  importIntegration$: Observable<Action> = this.actions$
+    .ofType<IntegrationUpload>(IntegrationImportActions.IMPORT)
```
deeleman: You can spare importing `IntegrationUpload ` and use `IntegrationImportActions .IntegrationUpload ` here instead. It might be more verbose but will allow you to not to keep track and import each and every action created in the future.

app/ui/src/app/integration/import-export/import/integration-import.effects.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Actions, Effect } from '@ngrx/effects';
+import { Observable } from 'rxjs/Observable';
+
+import { IntegrationImportService } from './integration-import.service';
+import {
+  IntegrationImportActions,
+  IntegrationUpload,
+  IntegrationUploadComplete,
+  IntegrationImport,
+  IntegrationImportComplete
```
deeleman: Half of these tokens are not in use

app/ui/src/app/integration/import-export/import/integration-import.models.ts
deeleman: If you want to define an interface that models the import operation (regardless it models an upload or edit operation) just do it once! And strive for interface reusability whenever you can (as in this place).

app/ui/src/app/integration/import-export/import/integration-import.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  templateUrl: './integration-import.component.html',
+  styleUrls: ['./integration-import.component.scss']
+})
+export class IntegrationImportComponent implements OnInit {
+  constructor(private store: Store<PlatformStore>) { }
+
+  ngOnInit() {
+    /**
+     * Payload for IntegrationImportUpload action
+     */
```
deeleman: Why do we need a comment for this? Isn't the variable names or the method naming self-descriptive enough?

app/ui/src/app/integration/import-export/import/integration-import.component.ts
```diff
@@ -0,0 +1,35 @@
...
+      file: null,
+      importResults: {
+        integrations: [],
+        connections: []
+      }
+    };
+
+
+    // We dispatch the desired action with the given payload
+    this.store.dispatch(new IntegrationImportUpload(payload));
```
deeleman: Do you believe you're gonna be needing to dispatch the `IntegrationImportUpload` with different payloads in the future. Otherwise there is no point to dispatch an action with an object payload that reflects and empty initial state. You already have `initialState` at the reducer for this, and you can even initialize empty instances of state within the reducer handlers if needed.

Here we're only duplicating code and action payloads should be as minimal as possible or even non-existing whereas possible.

app/ui/src/app/integration/import-export/import/integration-import.service.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  importIntegration(importRequest: IntegrationImportRequest): Observable<any> {
+    const apiHttpService = this.apiHttpService.setEndpointUrl('submitImport');
+
+    const [importIntegration, icon, specification] = [
+      importRequest,
+      importRequest.iconFile,
+      importRequest.specificationFile
+    ];
```
deeleman: Beware with copy+pasted code, since that's the easiest way to introduce technical debt into a project: `iconfile` or `specificationFile` do not exist in the `IntegrationImportRequest` interface.

app/ui/src/app/integration/import-export/integration-import-export.component.scss
```diff
@@ -0,0 +1,24 @@
...
+
+    li {
+      &.active {
+        cursor: default;
+      }
+
+      cursor: pointer;
+    }
+  }
+}
```
deeleman: All the `<ul>` styling here has been thoroughly duplicated all over the project and in order to prevent more duplication we now have a global classname named `.syn-tabs` that takes care of that for the entire project. FYI it is available at `/scss/components/_tabbed-nav.scss` and you do not need to import `syndesis-sass` to use it in your HTML because it is already part of the CSS classnames bndle.

Please avoid SASS duplication.

app/ui/src/app/integration/import-export/import/integration-import.actions.ts
kahboom: It was for reference so that developers know what this action is typically used for, but I do see what you mean. I'll remove it.

app/ui/src/app/integration/import-export/import/integration-import.service.ts
kahboom: I was trying to use the ApiConnector stuff as an example for following the Redux architecture, but ended up copying (and subsequently removing) a lot of the code. I think it would have been best to apply what I'm actively learning outside of a feature requirement.

kahboom:
Squashed a bunch of commits, fixed the lint errors and failing build, made changes according to feedback. I think I set up the Redux stuff as best as I could--only concern is the integration routing seems wonky, so I'm sorting that out. Adding actions, effects, etc. as necessary for the editor (step 2) will come later on.

A separate PR it will be heavily focused on API wiring and component/template (actually implementing the feature) using this architecture will be opened once this one looks okay.

kahboom:
This is ready for review again. Lint and tests pass, AOT build is fine, fixed the issue with the routing. This is for the architectural setup, I'm already working on opening the feature implementation PR. For more information, please see the primary issue.

kahboom:
There is a `<h1></h1>` and `<h2></h2>` tag visible on the page now from when I was testing the templating and routes, but since it's not a central page on the app I'd prefer to leave it there for this PR--it will be removed shortly with the next PR. See screenshot below:

screenshot 2018-02-14 13 53 51

kahboom:
It looks like the REST build failed for some reason; there is notice on circleCI for partially degraded service that suggests potential intermittent failures. I don't have permission to trigger a build manually, but will try something.

PR #1268, opened 2018-01-27
feature: Reworked release command

rhuss:
This is a backport from the 1.2.x branch

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -186,10 +186,26 @@
         <plugins>
 
 					<plugin>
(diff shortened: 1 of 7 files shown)


rhuss:
Fix is a backport from 1.2.3 and adds a 'release' command to syndesis. Since I need some of the docs and backports on master (from my three overlapping PRs) until tomorrow and although I'm against the no-review approach, let me be the first to start it.

They are all about syndesis scripts. some minor changes in the s2i generation and update of the templates.

PR #1286, opened 2018-01-29
fix: extension-plugin ignores fields annotated with SyndesisActionProperty for action of type STEP

lburgazzoli:


app/extension/annotation-processor/src/main/java/io/syndesis/maven/annotation/processing/SyndesisExtensionActionProcessor.java
```diff
@@ -129,6 +129,15 @@ protected boolean augmentProperties(TypeElement element, Properties props) throw
...
             props.put("entrypoint", element.getQualifiedName().toString());
+
+            // Let's search for fields annotated with SyndesisActionProperty
+            for (Element field: element.getEnclosedElements()) {
+                if (field.getKind() == ElementKind.FIELD) {
+                    addActionProperties(field, props);
+                }
+            }
+
+            return true;
```
zregvart: There is also a loop in the `else` branch that returns true on processing just the first element of enclosedElements. Smells a bit as unintentional, if it's intentional perhaps comment would help.

lburgazzoli: The else branch returns true if the method being evaluated is annotated with a specific annotation, will document it

PR #1287, opened 2018-01-29
refactor(codestyle): move policy checks to root

zregvart:
This moves content from `tools/policy` to the root of the repository
which should enable configuring Condacy checks via our own
`checkstyle.xml` (for Checkstlyle) along with `ruleset.xml` (for PMD).


rhuss:
@zregvart is this the only way to convince codacy to use custom rules ? I ask because obviously its not nice to clutter the top-level dir, but if this is the only way, I'm ok.

rhuss:
btw, I think we can remove the top-leve Jenkinsfile, can't we ? // @iocanel

zregvart:
> btw, I think we can remove the top-leve Jenkinsfile, can't we ? // @iocanel

I don't think it's used.

zregvart:
> @zregvart is this the only way to convince codacy to use custom rules ?

I'm [afraid so](https://support.codacy.com/hc/en-us/articles/207994335-Code-Patterns#5-existing-tools).

zregvart:
@rhuss @iocanel are we good to go here?

gashcrumb:
Wonder, can we re-enable the UI and fix the quote check so codacy points out `"` instead of `'`?

rhuss:
> @zregvart is this the only way to convince codacy to use custom rules ?
> 
> I'm afraid so.

@zregvart wouldn't it help if we set the projectdir to `app` (from the same docs) ? -->

```
6. Configuring the project root directory for analysis

By default, Codacy starts the analysis on the repository's root. However, you can set up a different project folder on which to start the analysis using a Codacy configuration file. This file needs to be named ".codacy.yaml" or ".codacy.yml" and must be placed in the project's root.

See example below:

---
engines:
 rubocop:
 enabled: true
 exclude_paths:
 - config/engines.yml
 base_sub_dir: test/baseDir
```

PR #1293, opened 2018-01-29
Add some ignoreUsedDependencies that are needed in the build

cunningt:
Add some ignoreUsedDependencies that are needed in the build, and add a plugin repository statement for SNAPSHOTS because the syndesis-build-helper-maven-plugin is used and I'm getting errors when running mvn clean and it can't resolve the plugin.

### DIFF
--- a/app/core/pom.xml
+++ b/app/core/pom.xml
@@ -413,6 +413,7 @@
         <artifactId>duplicate-finder-maven-plugin</artifactId>
         <configuration>
           <ignoredClassPatterns>
+            <ignoredUnusedDeclaredDependency>com.fasterxml.jackson.datatype:jackson-datatype-jdk8</ignoredUnusedDeclaredDependency>
             <ignoredClassPattern>io.fabric8.kubernetes.client.*</ignoredClassPattern>
             <ignoredClassPattern>org.infinispan.*</ignoredClassPattern>
             <ignoredClassPattern>org.springframework.security.crypto.*</ignoredClassPattern>


cunningt:
@KurtStam I redid the request from https://github.com/syndesisio/syndesis/pull/639 here - I think we do need both the pluginRepo here and the ignoreUnused statements.

KurtStam:
@cunningt thx Tom, can you rebase once more to resolve the conflict?

app/rest/runtime/pom.xml
```diff
@@ -48,6 +48,18 @@
...
+      <id>sonatype-snapshot</id>
+      <name>Sonatype snapshot repository</name>
+      <url>https://oss.sonatype.org/service/local/repositories/snapshots/content</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </pluginRepository>
```
rhuss: Where do we need snapshot plugin form Maven central ?

app/verifier/pom.xml
```diff
@@ -365,6 +365,10 @@
         <artifactId>maven-dependency-plugin</artifactId>
         <configuration>
           <ignoredUnusedDeclaredDependencies>
+            <ignoredUnusedDeclaredDependency>io.syndesis:sql-common</ignoredUnusedDeclaredDependency>
+            <ignoredUnusedDeclaredDependency>junit:junit</ignoredUnusedDeclaredDependency>
+            <ignoredUnusedDeclaredDependency>commons-io:commons-io</ignoredUnusedDeclaredDependency>
+            <ignoredUnusedDeclaredDependency>org.assertj:assertj-core</ignoredUnusedDeclaredDependency>
```
rhuss: Sure that this needs to be added here and not by adding the proper scope to dependencies ? The maven dependency plugin does not count "test" and "runtime" scoped dependencies as unused, and e.g. "junit" should be clearly a test scoped dependency.

KurtStam:
@cunningt I have approved but it needs another rebase to pick up my fix for the broken build.

PR #1299, opened 2018-01-30
fix: show validate button according to connector's tags

lburgazzoli:


zregvart:
@lburgazzoli why not add the `verifier` tag to the rest of the connector implementations that support verifier?

lburgazzoli:
@zregvart done

app/ui/src/app/connections/common/configuration/validation.component.html
```diff
@@ -1,6 +1,6 @@
 <div [ngClass]="{'text-right': placement === 'right'}">
   <!-- validate button -->
-  <button *ngIf="showValidateButton(connection?.connector?.id)"
+  <button *ngIf="showValidateButton(connection?.connector)"
```
deeleman: Just wanted to point out something for the future: DO NOT introduce function calls in template evaluation expressions such as [ngIf], [hidden] and the like. Otherwise those functions will be evaluated on every pass of the change detection loop. If the function features a heavy execution, well wind up consuming lotsa resources unnecessarily.

In this case the check required is a one time check so it should have been handled within the OnInit() lifecycle handler.

Issue #1245, opened 2018-01-25
ActionMetadata should provide human readable schema label


igarashitm:
Today we are using `ActionMetadata/outputSchema/title` as a Document label eventually. This is copied into `DataShape/type` here:
https://github.com/syndesisio/syndesis/blob/e488dffd620c30a45204f18786a1f0e6545b8383/app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectionActionHandler.java#L143
https://github.com/syndesisio/syndesis/blob/e488dffd620c30a45204f18786a1f0e6545b8383/app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/connection/ConnectionActionHandler.java#L149

We need to do either
* Put the human readable short label in `ActionMetadata/outputSchema/title`
* Or introduce `ActionMetadata/outputSchema/name`

And also `ActionMetadata/outputSchema/description` for a bit longer description. datamapper UI is showing it as a tooltip when hover on a short label.

Then we might need to introduce those into `DataShape` as well. Note that `DataShape/type` is already used for class name, so it might be better to add `name` and `description` so that we can put the readable name and description on Java type as well.

#### Screenshots for Today's label and tooltip
![test-flatprimitiveclass](https://user-images.githubusercontent.com/265462/35456023-25afe5ca-02a2-11e8-827b-60e25282f006.png)
![twitter-status](https://user-images.githubusercontent.com/265462/35456024-25bf0762-02a2-11e8-95ac-bef5cb0c8791.png)

igarashitm:
BTW, #1136 would need to allow users to specify DataShape label in the action configuration page.

igarashitm:
@KurtStam suggested to introduce new properties like `name` and `description`, as `title` is already in use for internal use in DB connector.

lburgazzoli:
@igarashitm so something like:

```java
@Value.Immutable
@JsonDeserialize(builder = DataShape.Builder.class)
public interface DataShape extends Serializable {

    // New fields
    String getName();
    String getDescription()
    String[] getTags();

    String getKind();
    String getType();
    String getSpecification();

    class Builder extends ImmutableDataShape.Builder {
    }
}
```

zregvart:
I'd add just the `implements WithName` on the `DataShape`, is there a use case for `tags` and `description`, let's not overthink this 😈

igarashitm:
[...]

PR #1303, opened 2018-01-30
ActionMetadata should provide human readable schema label

lburgazzoli:


lburgazzoli:
@igarashitm I've implemented the backend part of #1245 so you can now set an optional name and a description on datashapes, don't know what has to be done on the UI side.

zregvart:
@lburgazzoli perhaps it would be nice to add these new properties to existing connector definitions in `deployment.json` or connector definitions.

lburgazzoli:
@zregvart I've added them to the twitter connector but I do not know what is the right value for other connectors so I leave this task to the connectors developers.

PR #1316, opened 2018-01-31
chore: Replace the StringTrimmingJsonDeserializer which was applied to all String values received in json documents with StringTrimmingConverters which can be targeted at specific fields in our data models.

chirino:


app/core/model/pom.xml
```diff
@@ -38,6 +38,7 @@
     <dependency>
       <groupId>org.immutables</groupId>
       <artifactId>value</artifactId>
+      <scope>compile</scope>
```
lburgazzoli: Immutables is not required at runtime so shouldn't it have `provided` scope ?

zregvart:
@chirino there is a merge conflict now. Let's merge this and revisit that dependency issue (#1322) in another PR.

chirino:
rebased.

PR #1326, opened 2018-01-31
(feat) add support for a duration form control

### LINKED ISSUE #717: [Periodic Invocation of Start Action] Entry widget for milliseconds field.
Create a widget/component/service, or similar solution, that converts milliseconds to seconds, minutes, hours, and days.

Unit Conversion

Full design [here](https://github.com/syndesisio/syndesis/blob/cac32465b92d780a3f0440c5b542b1ece489ae99/ux/designs/databaseconnection/databaseconnection.md#create-an-integration-with-database-as-a-start-connection).

@gashcrumb 
> For some of the upcoming designs we have time periods in the generated form. We'll need to add support for a numeric type that has a unit field along with it. Ideally we should support the possibility to have upper and lower bounds per unit. So for example a field that could be in milliseconds, seconds, minutes, hours, the input should be type number with a select next to it to specify the unit.Changing the unit should update any potential min/max bounds on the input field.

NOTE: We will likely be doing away with ng-dynamic-forms at some point (or at least this is still being discussed here #843 ), so we've decided to put this on hold for now.

cc @KurtStam

gashcrumb:
download

@KurtStam FYI, note you'll want to update the `description` attribute on that field now that the user can select a time unit, currently it refers to milliseconds.

fixes #717 

Marked WIP as there's currently an override that forces this particular field to be a duration, can take that out when the connector can set `duration` for relevant fields that require this control.

### DIFF
--- a/app/ui/src/app/common/ui-patternfly/duration-form-control.component.html
+++ b/app/ui/src/app/common/ui-patternfly/duration-form-control.component.html
@@ -1,0 +1,41 @@
+<form class="form-inline duration-input">
+  <div class="input-group">
+    <input class="form-control duration-input__input"
+           [attr.accept]="model.accept"
+           [attr.autoComplete]="model.autoComplete"
+           [attr.list]="model.listId"
+           [attr.max]="model.max"
+           [attr.min]="model.min"
+           [attr.multiple]="model.multiple"
+           [attr.step]="model.step"
+           [attr.tabindex]="model.tabIndex"
+           [attr.id]="fieldId"
+           [autofocus]="model.autoFocus"
+           [dynamicId]="bindId && model.id"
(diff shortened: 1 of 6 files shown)


gashcrumb:
Took out the override, let's get this merged and it'll show up when the connector can use the right form input type.

KurtStam:
@gashcrumb yes I will update both the type and the description.

Relates to
https://github.com/syndesisio/syndesis/issues/220

KurtStam:
@gashcrumb to be sure: the value will still be send over in milliseconds?

gashcrumb:
Yup, the control produces milliseconds and calculates it's state from milliseconds.

We could in a future enhancement also let the possible durations come from the server I think but let's keep it simple for now :-)

PR #1330, opened 2018-01-31
feat(apidoc): split internal/supported API documentation

### LINKED ISSUE #1325: External REST API Docs
We need to publish API Docs for published REST API that we endorse and make available to end users.  These published APIs will need to remain available and backward compatible in future versions of syndesis so we should limit what's documented.

This could just be as simple as an open api document that's publish via mvn repo.

zregvart:
We need to document what APIs are supported and what APIs are internal.

The easiest way to do that is to do two Swagger executions configured
against supported handlers and internal handlers.

Also switches Maven plugins to a plugin linked from swagger.io.

Now there are two sets of Swagger specifications:
 - internal at /api/v1/internal/swagger.[json|yaml]
 - supported at /api/v1/swagger.[json|yaml]

And two sets of API documents:
 - internal at /api/v1/internal/index.html
 - supported at /api/v1/index.html

To simplify generation IntegrationSupportHandler was moved to it's own
package.

Fixes #1325


chirino:
@rhuss: I think we should do that for all syndesis packages that a developer could import into his IDEA.  So basically all the APIs exposed via a tech extension.  Are tech extensions the only users of our java APIs?

zregvart:
I've modified the serving of documentation and Swagger files and added an integration test to check.

Let's do `internal` packages in a different PR 🙏 

@chirino @rhuss are we good to merge this in?

PR #1335, opened 2018-01-31
[feature] Adding the metrics-collector #1065

KurtStam:
This is the implementation that reads metrics from the integration pods and stores it in the database.


app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
```diff
@@ -0,0 +1,318 @@
...
+        Long messages = 0L;
+        Long errors = 0L;
+        Optional<Date> lastProcessed = Optional.empty();
+        Optional<Date> startDate = Optional.empty(); //we may have no more live pods for this integration
+
+        for (RawMetrics raw:metrics.values()) {
+            messages += raw.getMessages();
+            errors += raw.getErrors();
+            //Let's simply grab the oldest living pod, we will need to revisit when doing rolling upgrades etc
+            if (livePodIds.contains(raw.getPod())) {
```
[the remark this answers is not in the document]
KurtStam: it's easier to debug that way as you'd know were to look. We'd have a RawMetrics entry for each pod. The lifecycle of the integration is tied to the lifecycle of the pod. RawMetrics is then rolled up into the IntegrationMetricsSummary which is then exposed to over REST to the UI. 

I just found out that at the moment we only run 1 pod per integration but if we start supporting more then we'd simply grab the oldest living pod.

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/RawMetrics.java
```diff
@@ -0,0 +1,44 @@
...
+
+@Value.Immutable
+@JsonDeserialize(builder = RawMetrics.Builder.class)
+@SuppressWarnings("immutables")
+public interface RawMetrics {
+
+    String getIntegration();
+    String getIntegrationId();
+    Optional<String> getVersion();
+    String getPod();
```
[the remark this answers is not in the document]
KurtStam: See above - Current metrics = History + Pods, Dead pod metrics gets rolled into the History.

app/pom.xml
```diff
@@ -752,6 +752,18 @@
         <version>2.5</version>
       </dependency>
 
+      <dependency>
+        <groupId>org.jolokia</groupId>
+        <artifactId>jolokia-client-java</artifactId>
+        <version>1.3.7</version>
```
rhuss: 1.4.0 is the latest version currently (though there is  not much change to 1.3.7)

app/rest/metrics-collector/pom.xml
```diff
@@ -0,0 +1,226 @@
...
+    </dependency>
+
+    <dependency>
+      <groupId>io.syndesis</groupId>
+      <artifactId>controllers</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>io.syndesis</groupId>
+      <artifactId>jsondb</artifactId>
```
rhuss: Do you require a direct DB dependency on jsondb here ? I really would like to go for *every DB access through the DAOs*.  Why else would we have then a DAO layer to abstract the DB implementation ? I have a lenghty discussion around this in #1054 around this and we ended up that we at least indicated clearly that the implementation is tight to JSONDB. 

I won't repeat myself here, all arguments are on the table (see #1054)

app/rest/metrics-collector/pom.xml
```diff
@@ -0,0 +1,226 @@
...
+    </dependency>
+
+
+    <dependency>
+      <groupId>io.syndesis</groupId>
+      <artifactId>dao</artifactId>
```
rhuss: Why bothering with the DAO if you can access jsonddb yourself ?

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/JolokiaKubernetesAuthenticator.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private final KubernetesClient kubernetes;
+
+    public JolokiaKubernetesAuthenticator(KubernetesClient kubernetes) {
+        this.kubernetes = kubernetes;
+    }
+
+    @Override
+    public void authenticate(HttpClientBuilder httpClientBuilder, String s, String s1) {
+        httpClientBuilder.addInterceptorFirst(new BearerTokenInterceptor(kubernetes.getConfiguration().getRequestConfig().getOauthToken()));
```
rhuss: Nice ;)

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
```diff
@@ -0,0 +1,318 @@
...
+                            p.getMetadata().getLabels().get("syndesis.io/deployment-id"),
+                            rawMetricsHandler))
+            );
+
+            ListResult<Integration> integrationList = dataManager.fetchAll(Integration.class);
+            Set<String> activeIntegrationIds = new HashSet<>();
+            for (Integration integration : integrationList.getItems()) {
+                String name = integration.getName();
+                String integrationId = integration.getId().get();
+                LOGGER.info("IntegrationId=" + integrationId);
```
rhuss: debug ?

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
```diff
@@ -0,0 +1,318 @@
...
+                    executor.execute(new PodMetricsReader(
+                            kubernetes,
+                            p.getMetadata().getName(),
+                            p.getMetadata().getLabels().get("integration"),
+                            p.getMetadata().getLabels().get("syndesis.io/integration-id"),
+                            p.getMetadata().getLabels().get("syndesis.io/deployment-id"),
+                            rawMetricsHandler))
+            );
+
+            ListResult<Integration> integrationList = dataManager.fetchAll(Integration.class);
```
rhuss: Isn't there an easier way to get all integration ids and names without deserializing all integrations each time the metrics are queried ? Looks super expensive to me. // @chirino

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
```diff
@@ -0,0 +1,318 @@
...
+                activeIntegrationIds.add(name);
+
+                Map<String,RawMetrics> rawMetrics = getRawMetrics(name);
+                IntegrationMetricsSummary currentSummary =
+                        computeIntegrationSummary(integrationId, rawMetrics, livePods);
+                IntegrationMetricsSummary existingSummary =
+                        dataManager.fetch(IntegrationMetricsSummary.class, integrationId);
+                if (existingSummary == null) {
+                    dataManager.create(currentSummary);
+                } else if (existingSummary.hashCode() != currentSummary.hashCode()) {
```
rhuss: Isn't equals() more appropriate here ?

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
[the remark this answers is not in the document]
rhuss: It's not true that the lifecycle of an integration == lifecycle of an pod. In fact an integration run can span multiple pods if e.g. OpenShift decides to restart the integrations pod. 

Tracking pods doesn't seem right to me, we should count on the integration level.

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
```diff
@@ -0,0 +1,318 @@
...
+        close(scheduler);
+        close(executor);
+    }
+
+
+    @Override
+    public void run() {
+        LOGGER.info("Collecting metrics for active integration pods.");
+        try {
+            List<Pod> integrationPodList = kubernetes.pods().withLabel("integration").list().getItems();
```
rhuss: Not sure whether this is relevant to this PR but shouldn't we used namespaced labels ?

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
```diff
@@ -0,0 +1,318 @@
...
+            }
+
+            integrationPodList
+                .stream()
+                .filter(p -> Readiness.isReady(p))
+                .forEach(p ->
+                    executor.execute(new PodMetricsReader(
+                            kubernetes,
+                            p.getMetadata().getName(),
+                            p.getMetadata().getLabels().get("integration"),
```
rhuss: either all prefixed with "syndesis.io" or none imo

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
rhuss: If you are looking for live pods maybe a filter on the status would be useful here.

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
```diff
@@ -0,0 +1,318 @@
...
+     * the key is either HISTORY or the podName.
+     * @throws IOException
+     */
+    /*default*/ Map<String,RawMetrics> getRawMetrics(String integrationId) throws IOException {
+        //try to obtain metrics in this integration
+        Map<String,RawMetrics> metrics = new HashMap<>();
+        String json = jsonDB.getAsString(path(integrationId));
+        LOGGER.debug("JSON: " + json);
+        if (json != null) {
+            metrics = Json.mapper().readValue(json, new TypeReference<Map<String,RawMetrics>>() {});
```
rhuss: A DAO method with proper serialization would be useful here, as we do this for any other objects., too.

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
```diff
@@ -0,0 +1,318 @@
...
+                    RawMetrics history = metrics.get(HISTORY);
+                    RawMetrics dead = entry.getValue();
+                    Date lastProcessed = history.getLastProcessed().orElse(new Date(0)).after(dead.getLastProcessed().orElse(new Date(0)))
+                            ? history.getLastProcessed().orElse(null) : dead.getLastProcessed().orElse(null);
+                    RawMetrics updatedHistoryMetrics = new RawMetrics.Builder()
+                            .integrationId(integrationId)
+                            .integration(integration)
+                            .pod(history.getIntegrationId() + ":" + dead.getPod())
+                            .messages(history.getMessages() + dead.getMessages())
+                            .errors(history.getErrors() + dead.getErrors())
```
rhuss: If you would not collect metrics per pod but per integration you would never have to do such 'curation', as the metrics 'per integration' is just incremented. After all that's what the user wants to see. I don't understand why you need to complicate this (or otherwise, I don't see the benefit on doing it that fine granuarly).

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/PodMetricsReader.java
```diff
@@ -0,0 +1,274 @@
...
+            return null;
+        }
+        try {
+            return DATE_FORMAT.parse(s);
+        } catch (ParseException e) {
+            return null;
+        }
+    }
+
+    public static void main(String args[]) {
```
rhuss: would remove the main method and is specific to your setup.

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/PodMetricsReader.java
```diff
@@ -0,0 +1,274 @@
...
+     * @param kubernetes The {@link KubernetesClient} instance.
+     * @param pod        The name of the pod.
+     * @return An instance of the {@link J4pClient}.
+     */
+    private static J4pClient forPod(KubernetesClient kubernetes, String pod) {
+        String jolokiaUrl = String.format(JOLOKIA_URL_FORMAT, kubernetes.getMasterUrl(), kubernetes.getNamespace(), pod);
+        try {
+            return new J4pClientBuilder()
+                .url(jolokiaUrl)
+                .user("user")
```
rhuss: I also would set "POST" as the default method, as its much more robust then "GET" (because of escaping issuees). Actually this should be a default for Jolokia but unfortunately it is not.

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/RawMetrics.java
```diff
@@ -0,0 +1,44 @@
...
+
+    Long getMessages();
+    Long getErrors();
+    Optional<Date> getStartDate();
+    Optional<Date> getResetDate();
```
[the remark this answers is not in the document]
iocanel: It doesn't matter. Even if someone pressed the `reset` button and the history was wiped, it shouldn't make any difference to us. We could just treated the same way we treat restarts etc. That's why I feel that if we focus on that field we cover ALL cases.

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
iocanel: I had trouble getting the `status.Phase` to work as a filter. There's a chance the client is not handling that properly but I am not 100% sure about that.

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
iocanel: @agree: IMHO, we shouldn't even bother about queering the integrations.

I can see that going two ways, none of which requires `fetching all the pods`:

- Record per integration (no rollups)
- Record per reset (with rollups)

# Record per integration

 We should just get all the active pods, grab the data, and save them in the database.

If the object we persisted had a structure like:

    {
      private String integrationId;
      private long lastReset;

      private long messages;  //Messages after lastReset;
      private long errors;         //Errors after lastRest

      private long totalMessages;  //Total messages
      private long totalErrors;        //Total errors
    }

We could just get away with 1 read and up to 1 write per integration per iteration.

# Record per reset
This is pretty similar to what's implemented already here, and goes like this:

- periodical query pods, grab data and create a single record (per integration & reset). Each time the reset value changes, we can create a new record.

Periodically we sum all metrics record, to create a sum record.

As I said this is pretty similar to what we have but I see two major differences:

- starting point for the rollups shouldn't be the integration, neither the pod, but the individual metrics records (which are going to be summed). It's what is called RawMetrics in this PR (if I get it right).

- The rollup should not be a different task, possibly handled by a different executor, definitely with a larger interval.

app/rest/metrics-collector/pom.xml
rhuss: @KurtStam I really want to know you reasoning here why to include jsondb. Sorry, ignoring this is just not fair and absolutely counterproductive. I invested a lot of efforts to collect arguments that adding dependencies like this at will and only because its easy *now* will destroy any architectural concept we have for the backend. 

Happy to argue with you and find a compromise, but this is just bad style.

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
rhuss: @iocanel the first approach seems to be suitable if we are measureing relative numbers (since last read), the second one if we only get absolute (total) numbers from the MBeans. 

> The rollup should not be a different task, possibly handled by a different executor, definitely with a larger interval.

guess this should read 'should _be_ a different task' ? However, can't we make the sum up directly with the metrics query (as you can do for SQL or for Prometheus) ? This would completely eliminate the need for such a maintenance task.

app/rest/metrics-collector/pom.xml
KurtStam: @rhuss the reason is two-fold. Firstly I'm using the jsonDBs specific API to select all RawMetrics for an integration based on the 'path'. Secondly it's now  easy to keep the RawMetrics in the metrics-collector module as it isn't part of our 'model'.  

Different story for the IntegrationSummaryMetrics for which I do use the DataManager.

app/rest/metrics-collector/src/main/java/io/syndesis/rest/metrics/collector/MetricsCollector.java
KurtStam: When pod get restarted the metrics gets reset. So we need to store periodically. Plus we need to add metrics from dead pod and live pod(s). It can be a different task yes, but I don't think the period of that should be too large since you will loose messages that weren't stored at the time the pod went down.

iocanel: At the moment for each tick we do the following:

1) read pod metrics
 i) write per integration
2) rollup pod metrics
 i) read all integrations and for each of them
ii) read all raw metrics
iii) write summary per integration.

What I suggest is to take the number of db operations down to a maximum 1  read/write for each pod (not integration).
if you just store the summary and you keep track of the totals and the counters since last reset, you don't need anything else.

app/rest/metrics-collector/pom.xml
rhuss: If I can't convince you that a hard dependency on jsondb is the wrong direction, then I can't convince you.

In order to reflect this live-and-die-together dependency I renamed the module to "rest-metrics-jsondb" to reflect this relation (same for "rest-logging-jsondb").

PR #1337, opened 2018-01-31
Revert "chore(json): disable serialization of empty collections"

chirino:
This reverts commit f48d5dc932c57a3fbd4455140cf77aa8ce69c6de.  It broke UI as it was expecting empty lists.


chirino:
@lburgazzoli sry about the revert, but needed to do it since it break current UI on master.  I like the idea of the change, but we need the UI work associated with it to go in the the same PR.

lburgazzoli:
no worries, do we know what parts of the ui need to be changed ?

gashcrumb:
I think it's just the store, if you want to open a new PR and assign it to
me I can fix up the UI alongside your changes.

Issue #1137, opened 2018-01-18
Pass step indicies along with step IDs to data mapper


gashcrumb:
We need to pass along a list of steps to the data mapper that have data shapes, along with their index in the integration flow, step ID and output data shape.  Might be worth building a new structure like:

```javascript
{
output: {
  name: 'Salesforce',
  id: '<generated>',
  dataShape: {}   // <-- input data shape of next step goes here
},
inputs: {
  steps: [
    {
       name: 'Twitter',
       id: '<generated>',
       dataShape: {} // <-- output data shape of this step
    },
    {
        name: 'Database',
        id: '<generated>',
        dataShape: {} // <-- output data shape of this step
     }
  ]
 }
}
```

This can then be passed as an input to the data mapper host component.

Related to #1110

Created from #1114

igarashitm:
How about adding `CurrentFlow.getPreviousStepsWithOutputDataShape(position)` ? I'll give it a shot.

gashcrumb:
So yeah, really if we just inject the CurrentFlow service into the host component this issue can be closed, as the host component should then have everything it needs.  The parent controller only needs to pass along the position in the flow the user is looking at then.

Issue #1135, opened 2018-01-18
Gather up the output data shapes from previous steps and provide as an input to the data mapper host component


gashcrumb:
In `step-configure.component.ts` we currently only grab the first output data shape preceding the current step.  For the data mapper we need to supply all the output data shapes from previous steps that have them.

There's support already in the current flow service to get all previous data shapes, I would store this in a separate attribute to the `step-configure` component so we can also still support steps like the basic filter, which only supports the previous steps output data shape.

Related to #1110

Created from #1114

gashcrumb:
[this comment](https://github.com/syndesisio/syndesis/issues/1114#issuecomment-358849707)

igarashitm:
@gashcrumb as datamapper step is a special one consumes multiple data shapes, we can even add `CurrentFlow.getPreviousStepsWithOutputDataShape(position)` and let `data-mapper-host.component` directly call it instead of let `step-configure.component` perform as a middleman. Then the array index could be used for the Step index there. @gashcrumb WDYT?

gashcrumb:
absolutely, can just remove the data shape inputs (you'll still need the position) from the datamapper host component and let that guy grab what it needs from `CurrentFlow`.

gashcrumb:
@igarashitm might need to refactor this function out so that data shapes for extensions can be fetched when needed by the component that needs 'em?

PR #1338, opened 2018-01-31
feat: Bring multiple data shapes, Step IDs and labels into datamapper

igarashitm:
Fixes: #1135
Fixes: #1137 

I was still debugging though, submitting a PR with WIP flag as I hit a version mismatch issue. Due to the issue now I can't run Syndesis UI with latest AtlasMap UI.

Recently I hit following issue in AtlasMap UI and upgraded the `ng-packagr` to `2.0.0-rc.6`
https://github.com/dherges/ng-packagr/issues/311

Then Syndesis UI fails to build with
```
ERROR in Metadata version mismatch for module /home/tomo/workspace/iPaaS/syndesisio/syndesis/app/ui/node_modules/@atlasmap/atlasmap.data.mapper/atlasmap-atlasmap.data.mapper.d.ts, found version 4, expected 3

ERROR in ./src/main.ts
Module not found: Error: Can't resolve './$$_gendir/app/app.module.ngfactory' in '/home/tomo/workspace/iPaaS/syndesisio/syndesis/app/ui/src'
 @ ./src/main.ts 6:29-76
 @ multi ./src/main.ts
```

It seems to be caused by the Angular version mismatch, as AtlasMap is already based on Angular5 while Syndesis UI is still on Angular4.
I tried to upgrade only angular-cli and etc, but couldn't find a happy combination. Apparently we need to update to Angular5.

@deeleman can you help on this?

### DIFF
--- a/app/ui/src/app/integration/edit-page/current-flow.service.ts
+++ b/app/ui/src/app/integration/edit-page/current-flow.service.ts
@@ -1,13 +1,16 @@
 import { Injectable, EventEmitter } from '@angular/core';
+import { Observable } from 'rxjs/Observable';
 import { Subscription } from 'rxjs/Subscription';
 
 import { Action,
   Connection,
   createIntegration,
   createStep,
+  DataShape,
   Integration,
   Step,
-  key } from '@syndesis/ui/platform';
+  key,
+  IntegrationSupportService} from '@syndesis/ui/platform';
 import { log, getCategory } from '@syndesis/ui/logging';
 import { IntegrationStore } from '@syndesis/ui/store';
 
@@ -32,7 +35,10 @@ export class CurrentFlow {
   private _integration: Integration;
   private _loaded = false;
 
-  constructor(private store: IntegrationStore) {
+  constructor(
+    private integrationStore: IntegrationStore,
+    private integrationSupportService: IntegrationSupportService
+  ) {
     this.subscription = this.events.subscribe((event: FlowEvent) =>
       this.handleEvent(event)
     );
@@ -137,9 +143,16 @@ export class CurrentFlow {
     return null;
   }
 
-  getSubsequentStepsWithDataShape(position): Array<Step> {
+  getSubsequentStepsWithDataShape(position): Array<{step: Step, index: number}> {
+    const answer: {step: Step, index: number}[] = [];
     const steps = this.getSubsequentSteps(position);
-    const answer = steps.filter(s => this.hasDataShape(s, false));
+    if (steps) {
+      steps.forEach((step, index) => {
+        if (this.hasDataShape(step, true)) {
+          answer.push({'step': step, 'index': position + index});
+        }
+      });
+    }
     return answer;
   }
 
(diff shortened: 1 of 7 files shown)


app/ui/src/app/integration/edit-page/current-flow.service.ts
```diff
@@ -32,7 +35,10 @@ export class CurrentFlow {
   private _integration: Integration;
   private _loaded = false;
 
-  constructor(private store: IntegrationStore) {
+  constructor(
+    private store: IntegrationStore,
```
gashcrumb: Can we rename this to `integrationStore`?  I know the original code was like that but it'd be good to align the instance name to the service class, should be a matter of hitting `F2` in vscode if you use that for your typescript :-)

app/ui/src/app/integration/edit-page/current-flow.service.ts
```diff
@@ -32,7 +35,10 @@ export class CurrentFlow {
...
+    private integrationSupport: IntegrationSupportService
```
gashcrumb: Same here, be better to call it `integrationSupportService`

app/ui/src/app/integration/edit-page/step-configure/step-configure.component.html
```diff
@@ -64,11 +64,8 @@
           <div *ngSwitchCase="'mapper'">
             <div class="row">
               <div class="col-md-12"
-                   *ngIf="step && inputDataShape && outputDataShape">
-                <syndesis-data-mapper-host [inputDataShape]="inputDataShape"
-                                           [outputDataShape]="outputDataShape"
-                                           [step]="step"
-                                           [position]="position"></syndesis-data-mapper-host>
+                   *ngIf="step">
+                <syndesis-data-mapper-host [step]="step" [position]="position"></syndesis-data-mapper-host>
```
gashcrumb: probably don't even need to pass the step along TBH, you could get it from the injected `CurrentFlow` service via the position.

app/ui/src/app/integration/edit-page/step-configure/step-configure.component.ts
```diff
@@ -179,11 +138,34 @@ export class IntegrationStepConfigureComponent extends FlowPage implements OnIni
...
+          })
+          .catch(response => this.handleDataShapeError(nextStep, response))
+      })
+      .catch(response => this.handleDataShapeError(prevStep, response))
       .then(() => this.loadFormSetup(step));
   }
 
+  private handleDataShapeError(step: Step, response: any) {
+    this.loading = false;
+    const error = JSON.parse(response['_body']);
```
gashcrumb: we'll need to double-check what excatly comes back in this error case, I'm not sure that the returned object is a `Response`, so `error` could wind up being undefined.

gashcrumb:
This is all awesome stuff, great work!.  Am not sure on the versioning/error myself, I thought @deeleman had said it should still kinda work but that was awhile ago, my memory only extends back like a couple days at most :-)

app/ui/src/app/integration/edit-page/step-configure/step-configure.component.ts
igarashitm: Added a safeguard against NPE for now, let's see when test the error path.

igarashitm:
Yep, actually it has been working until yesterday. Today suddenly I hit that weird tsickle error in ng-packagr without doing any dependency change... weird world.

igarashitm:
memo: If we can get around the dherges/ng-packagr#311 without updating ng-packagr to 2.0rc, that would be reasonable.

deeleman:
Closing, as superseded by #1367

PR #1339, opened 2018-01-31
feat: More lifecycle improvements

chirino:
- Integration state decoupled from deployment state, APIs defined to manage deployment.
- Replaced IntegrationDeploymentSpec with a plain Integration

to publish integration - PUT to /integrations/{id}/deployments
to undeploy integration - POST to /integrations/{id}/deployments/{version}/targetState

/integrations/{id}/history removed, replaced with a /integrations/{id}/overview


app/core/model/src/main/java/io/syndesis/model/integration/IntegrationDeployment.java
```diff
@@ -36,388 +38,434 @@
...
 
-
     @JsonIgnore
-    public boolean isInactive() {
+    default boolean isInactive() {
         return getCurrentState() == IntegrationDeploymentState.Undeployed || getCurrentState() == IntegrationDeploymentState.Inactive
-            ? true
-            : false;
+                ? true
+                : false;
```
zregvart: do we need the ternary here?

app/core/model/src/main/java/io/syndesis/model/integration/IntegrationDeployment.java
```diff
@@ -36,388 +38,434 @@
...
+//            this.createdDate = createdDate.orElse(new Date());
+//            return this;
+//        }
+//
+//        public Builder lastUpdated(Optional<Date> lastUpdated) {
+//            this.lastUpdated = lastUpdated.orElse(new Date());
+//            return this;
+//        }
+//
+//    }
```
zregvart: let's not do this, git versions source code and we have no need for commented out large blocks of code, remove it rather

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -99,23 +99,76 @@ public Kind resourceKind() {
...
+    @SuppressFBWarnings("URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD")
+    public static class IntegrationOverview {
+
+        public int version;
+        public Optional<String> id;
+        public String name;
+        public SortedSet<String> tags;
+        public Optional<String> description;
+        public boolean draft;
+        public List<DeploymentOverview> deployments;
```
zregvart: perhaps this should be `final` and the Set and list should be unmodifiable?

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -99,23 +99,76 @@ public Kind resourceKind() {
...
+
+    @SuppressFBWarnings({"URF_UNREAD_PUBLIC_OR_PROTECTED_FIELD", "URF_UNREAD_FIELD"})
+    public static class DeploymentOverview {
+
+        private Optional<String> id;
+        private int version;
+        private IntegrationDeploymentState currentState;
+        private IntegrationDeploymentState targetState;
+        private Date createdDate;
+        private int integrationVersion;
```
zregvart: Would prefer the fields to be `final`, and `createdDate` to be `ZonedDateTime`

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -145,82 +196,63 @@ public Integration create(@Context SecurityContext sec, @ConvertGroup(from = Def
         Integration encryptedIntegration = encryptionSupport.encrypt(integration);
 
         Integration updatedIntegration = new Integration.Builder()
+            .version(1)
```
zregvart: also the default value, perhaps not needed?

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationSupportHandler.java
```diff
@@ -243,38 +245,26 @@ private void addToExport(Map<String, ModelData<?>> export, ModelData<?> model) {
             switch (model.getKind()) {
                 case Integration: {
                     Integration integration = (Integration) model.getData();
-                    integration = new Integration.Builder()
+                    Integration.Builder builder = new Integration.Builder()
                         .createFrom(integration)
-                        .desiredStatus(IntegrationDeploymentState.Draft)
-                        .build();
+                        .deleted(false)
+                        .version(1)
```
zregvart: `1` is default, perhaps not necessary

app/core/model/src/main/java/io/syndesis/model/integration/IntegrationDeployment.java
```diff
@@ -18,16 +18,18 @@
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
 import io.syndesis.core.IndexedProperty;
 import io.syndesis.core.SuppressFBWarnings;
+import io.syndesis.core.json.StringTrimmingConverter;
```
zregvart: unused

app/core/model/src/main/java/io/syndesis/model/integration/IntegrationDeployment.java
```diff
@@ -18,16 +18,18 @@
...
 import io.syndesis.model.Kind;
 import io.syndesis.model.WithId;
 import io.syndesis.model.WithKind;
-import io.syndesis.model.WithName;
 import io.syndesis.model.action.ConnectorAction;
+import io.syndesis.model.validation.UniqueProperty;
+import io.syndesis.model.validation.UniquenessRequired;
```
zregvart: these two are unused

app/rest/controllers/src/main/java/io/syndesis/controllers/integration/online/ActivateHandler.java
```diff
@@ -126,11 +126,11 @@ public StateUpdate execute(IntegrationDeployment integrationDeploymentDefinition
     private DeploymentData createDeploymentData(Integration integration, IntegrationDeployment integrationDeployment) {
         Properties applicationProperties = IntegrationSupport.buildApplicationProperties(integrationDeployment, dataManager, encryptionComponent);
 
-        String username = integration.getUserId().orElseThrow(() -> new IllegalStateException("Couldn't find the user of the integration"));
+        String username = integrationDeployment.getUserId().orElseThrow(() -> new IllegalStateException("Couldn't find the user of the integration"));
 
         return DeploymentData.builder()
             .addLabel(OpenShiftService.INTEGRATION_ID_LABEL, Labels.sanitize(integrationDeployment.getIntegrationId().orElseThrow(() -> new IllegalStateException("IntegrationDeployment should have an integrationId"))))
-            .addLabel(OpenShiftService.DEPLOYMENT_ID_LABEL, integrationDeployment.getVersion().orElse(0).toString())
+            .addLabel(OpenShiftService.DEPLOYMENT_ID_LABEL, ""+integrationDeployment.getVersion())
```
zregvart: integrationDeployment.getVersion().toString()

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationHandler.java
```diff
@@ -42,9 +43,14 @@
 import javax.ws.rs.core.SecurityContext;
 import javax.ws.rs.core.UriInfo;
 
+import org.springframework.security.core.context.SecurityContextHolder;
```
zregvart: unused

app/rest/dao/src/test/java/io/syndesis/dao/ReadApiClientDataTest.java
```diff
@@ -48,14 +48,12 @@
     public void deserializeModelDataTest() throws IOException {
 
         Integration integrationIn = new Integration.Builder()
-                    .desiredStatus(IntegrationDeploymentState.Active)
                     .tags(new TreeSet<>(Arrays.asList("tag1", "tag2")))
                     .createdDate(new Date())
                     .build();
         String integrationJson = mapper.writeValueAsString(integrationIn);
         System.out.println(integrationJson);
```
zregvart: perhaps we can do without System.out.println

app/rest/dao/src/test/java/io/syndesis/dao/ReadApiClientDataTest.java
```diff
@@ -48,14 +48,12 @@
...
         Integration integrationOut = mapper.readValue(integrationJson, Integration.class);
-        Assert.assertEquals(integrationIn.getDesiredStatus(), integrationOut.getDesiredStatus());
```
zregvart: not sure why the assert was removed

app/core/model/src/main/java/io/syndesis/model/integration/Integration.java
```diff
@@ -50,19 +50,25 @@ default Kind getKind() {
     //TODO: This is actually the deployment version and not the deployment. At some point we need to rename it.
     Optional<Integer> getDeploymentId();
 
-    Optional<String> getUserId();
```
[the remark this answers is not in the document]
chirino: It moved to deployments.

app/core/model/src/main/java/io/syndesis/model/integration/IntegrationDeployment.java
```diff
@@ -36,388 +38,434 @@
...
+    default IntegrationDeploymentState getTargetState() {
+        return IntegrationDeploymentState.Active;
     }
 
-    @Override
-    @JsonProperty("name")
-    public String getName() {
-        return name;
+    @Value.Default
+    default Date getLastUpdated() {
```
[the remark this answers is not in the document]
chirino: I would personally prefer to switch to keep it simple and switch to a long.

app/core/model/src/main/java/io/syndesis/model/integration/IntegrationDeployment.java
chirino: Was existing code.  I’d have review closer.

app/rest/controllers/src/main/java/io/syndesis/controllers/integration/online/ActivateHandler.java
chirino: Won’t work

app/rest/dao/src/test/java/io/syndesis/dao/ReadApiClientDataTest.java
chirino: Field does not exist anymore.

app/core/model/src/main/java/io/syndesis/model/integration/IntegrationDeployment.java
chirino: right. dont need it.

app/core/model/src/main/java/io/syndesis/model/integration/IntegrationDeployment.java
chirino: removing..

chirino:
some fields have changed: lastUpdated => updatedAt and createdDate => createdAt
They are now holding a long ms since epoch

app/rest/rest/src/main/java/io/syndesis/rest/v1/handler/integration/IntegrationSupportHandler.java
chirino: right

gashcrumb:
> to publish integration - PUT to /integrations/{id}/deployments

So this is only for existing integrations, right?  Currently for publishing new integrations it's a `PUT` to `/integrations` with `desiredStatus` set to `Active`.

> to undeploy integration - POST to /integrations/{id}/deployments/{version}/targetState

Just to clarify is this just to change the state of the active deployment?  I kinda feel like state management should just be at the integration level, i.e. `POST` to `/integrations/{id}/targetState`?  This would allow the UI's current start/stop controls to remain a bit agnostic about deployments.

chirino:
@gashcrumb 

> So this is only for existing integrations, right? Currently for publishing new integrations it's a PUT to /integrations with desiredStatus set to Active.

In the case of a new integration, PUT to `/integrations`, then follow it up with a PUT to `/integrations/{id}/deployments`

> Just to clarify is this just to change the state of the active deployment? I kinda feel like state management should just be at the integration level, i.e. POST to /integrations/{id}/targetState? This would allow the UI's current start/stop controls to remain a bit agnostic about deployments.

So this kinda muddies the water a bit.  I want clean up  Integration model so that the only time it gets updated is when a user makes changes to the integration steps.

gashcrumb:
> So this kinda muddies the water a bit. I want clean up Integration model so that the only time it gets updated is when a user makes changes to the integration steps.

Makes sense, can I suggest maybe we have a top-level `/integrationDeployments` (or `/deployments`) that the UI could use for the list/dashboard list?  Those two views are concerned with the runtime state of the deployment.  If we take the status etc out of the Integration object, then the lists will have to make a second call to fetch the deployment so the status can be shown.

I guess though with `/deployments` we wouldn't get back drafts :-/

chirino:
squashed and rebased.. but there seems to be another merge needed.

chirino:
@gashcrumb rebased.

chirino:
Codacy looks hung.

PR #1340, opened 2018-01-31
fix: Fix tests

rhuss:
* Replaced hardcoded version numbers in test files
* Made JSON comparing asserts a little bit more forgiving by triming newlines at the end and beginning
* Increased timeout for forker surefire tests from 30s to 120s


KurtStam:
LGTM

PR #1344, opened 2018-02-01
[model] get rid of custom Steps for expression and rule filter

### LINKED ISSUE #1327: [model] get rid of custom Steps for expression and rule filter
See:

https://github.com/syndesisio/syndesis/blob/master/app/core/model/src/main/java/io/syndesis/model/integration/StepDeserializer.java
https://github.com/syndesisio/syndesis/blob/master/app/core/model/src/main/java/io/syndesis/model/filter/ExpressionFilterStep.java
https://github.com/syndesisio/syndesis/blob/master/app/core/model/src/main/java/io/syndesis/model/filter/RuleFilterStep.java

lburgazzoli:
Fixes #1327


app/core/model/src/test/java/io/syndesis/model/filter/RuleFilterStepTest.java
```diff
@@ -37,4 +34,5 @@ public void createExpression() {
         // Reading notes: Unit tests are like personal diaries. Feel honoured when you have the chance to be part of them ;-)
         assertEquals("${body.person.name} == 'Ioannis' && ${body.person.favoriteDrinks} contains 'Gin'", step.getFilterExpression());
     }
+    */
```
zregvart: I really don't like commented out blocks of code, just delete it

app/integration/runtime/pom.xml
```diff
@@ -167,6 +166,7 @@
             <ignoredUsedUndeclaredDependency>org.springframework:spring-beans</ignoredUsedUndeclaredDependency>
             <ignoredUnusedDeclaredDependency>org.apache.camel:camel-spring</ignoredUnusedDeclaredDependency>
             <ignoredUnusedDeclaredDependency>javax.xml.bind:jaxb-api</ignoredUnusedDeclaredDependency>
+            <ignoredUnusedDeclaredDependency>com.google.code.findbugs:findbugs-annotations</ignoredUnusedDeclaredDependency>
```
zregvart: perhaps we can remove the dependency declaration instead of adding the ignoredUnusedDeclaredDependency?

app/core/model/src/test/java/io/syndesis/model/filter/RuleFilterStepTest.java
lburgazzoli: fixed

app/integration/runtime/pom.xml
lburgazzoli: The dependency comes from core/model which needs to suppress some [findbug warning](https://github.com/syndesisio/syndesis/blob/master/app/core/model/src/main/java/io/syndesis/model/WithConfigurationProperties.java#L33) on immutables generated code. If you know how to get rid of such dependency I'm more than happy.

PR #1359, opened 2018-02-01
Loosing JSON schema properties and upgrade HTTP client

### LINKED ISSUE #1295: Field selection/completion broken in Filter Step
Filter step no longer features a dropdown containing available field names.  Autocomplete when typing does not show all fields either.

zregvart:
When serializing `SyndesisMetadata<ObjectSchema>` `type` property of
the schema (i.e. `"type": "object") is lost.

This workarounds it by converting `SyndesisMetadata` to a
`Map<String, Object>` before serialization so that property is not lost.

Probably has something to do with generics and the way JAX-RS support
for Jackson insists on using the method return type as the type of
object to serialize.

Seems that we missed a spot when migrating from restangular, now we
receive parsed JSON response body directly instead of in the `_body`
property.

Also cleans up unused imports.

Fixes #1295

### DIFF
--- a/app/connectors/connector-verifier/src/main/java/io/syndesis/verifier/api/SyndesisMetadata.java
+++ b/app/connectors/connector-verifier/src/main/java/io/syndesis/verifier/api/SyndesisMetadata.java
@@ -21,6 +21,11 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 
+import io.syndesis.verifier.api.support.SyndesisMetadataConverter;
+
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
+@JsonSerialize(converter = SyndesisMetadataConverter.class)
 public final class SyndesisMetadata<T> {
 
     public final T inputSchema;
@@ -33,8 +38,7 @@ public final class SyndesisMetadata<T> {
      */
     public final Map<String, List<PropertyPair>> properties;
 
-    public SyndesisMetadata(final Map<String, List<PropertyPair>> properties, final T inputSchema,
-                            final T outputSchema) {
+    public SyndesisMetadata(final Map<String, List<PropertyPair>> properties, final T inputSchema, final T outputSchema) {
         this.properties = properties;
         this.inputSchema = inputSchema;
         this.outputSchema = outputSchema;
@@ -45,6 +49,10 @@ public final class SyndesisMetadata<T> {
                 Collections.sort(propertyPairs, Comparator.comparing(PropertyPair::getDisplayValue));
             }
         }
+    }
+
+    public boolean isNull() {
+        return inputSchema == null && outputSchema == null && (properties == null || properties.isEmpty());
     }
 
 }
--- a/app/connectors/connector-verifier/src/main/java/io/syndesis/verifier/api/support/SyndesisMetadataConverter.java
+++ b/app/connectors/connector-verifier/src/main/java/io/syndesis/verifier/api/support/SyndesisMetadataConverter.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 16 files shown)


app/ui/src/app/integration/edit-page/step-configure/filter-steps/basic-filter.component.ts
```diff
@@ -103,8 +103,7 @@ export class BasicFilterComponent implements OnChanges {
     this.integrationSupport
       .getFilterOptions(dataShape)
       .toPromise()
-      .then((resp: any) => {
-        const body = JSON.parse(resp['_body']);
+      .then((body: any) => {
```
gashcrumb: Keep in mind you won't want to cherry pick this part onto the 1.2.x branch.

zregvart: Right need to fix that, it’s in there now

app/connectors/connector-verifier/src/test/java/io/syndesis/verifier/api/SyndesisMetadataTest.java
```diff
@@ -0,0 +1,53 @@
...
+        outputSchema.setTitle("output");
+
+        final SyndesisMetadata<ObjectSchema> syndesisMetadata = new SyndesisMetadata<>(properties, inputSchema, outputSchema);
+
+        final ObjectMapper mapper = new ObjectMapper();
+
+        final String json = mapper.writeValueAsString(syndesisMetadata);
+
+        assertThat(json).isEqualTo(
+            "{\"outputSchema\":{\"type\":\"object\",\"title\":\"output\"},\"inputSchema\":{\"type\":\"object\",\"title\":\"input\"},\"properties\":{\"aProperty\":[{\"displayValue\":\"label\",\"value\":\"value\"}]}}");
```
rhuss: what i've seen to so often now is that JSON documents are compared as strings. This is so fragile (and actually had that issue today already because of accidentally adding whitespace to a json document at the end), we should not do this. No one guarantees you the order of the keys for example and that might change randomly.

Could we use something like jsonassert.skyscreamer.org please ?

app/connectors/connector-verifier/src/main/java/io/syndesis/verifier/api/support/SyndesisMetadataConverter.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public Map<String, Object> convert(final SyndesisMetadata<?> value) {
+        if (value == null || value.isNull()) {
+            return null;
+        }
+
+        final Map<String, Object> converted = new HashMap<>();
+
+        converted.put("properties", value.properties);
+        converted.put("inputSchema", MAPPER.valueToTree(value.inputSchema));
```
rhuss: Are the value by accident public ? Not that I'm against this per-se but mixing this with the traditional get/set paradigm is confusing.

app/connectors/connector-verifier/src/main/java/io/syndesis/verifier/api/SyndesisMetadata.java
```diff
@@ -33,8 +38,7 @@
      */
     public final Map<String, List<PropertyPair>> properties;
 
-    public SyndesisMetadata(final Map<String, List<PropertyPair>> properties, final T inputSchema,
-                            final T outputSchema) {
+    public SyndesisMetadata(final Map<String, List<PropertyPair>> properties, final T inputSchema, final T outputSchema) {
```
rhuss: can we add a warning comment to the member vars for a heads up that when new meta data is added that the converter class needs to be updated, too ?

zregvart: Warnings are cheap, tests are better, added test to check for that.

app/connectors/connector-verifier/src/main/java/io/syndesis/verifier/api/support/SyndesisMetadataConverter.java
zregvart: This is a simple transfer object, and those fields are final, in this case I don't mind it so I didn't change it.

zregvart:
@rhuss this should be good to go now

zregvart:
superseded by #1374

PR #1360, opened 2018-02-01
Backport fix for #1295 to 1.2.x

zregvart:
You might want to look at #1359 if it's merged first.

### DIFF
--- a/app/connectors/connectors/sql/sql-common/pom.xml
+++ b/app/connectors/connectors/sql/sql-common/pom.xml
@@ -55,8 +55,9 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>org.json</groupId>
-      <artifactId>json</artifactId>
+      <groupId>com.vaadin.external.google</groupId>
+      <artifactId>android-json</artifactId>
+      <version>0.0.20131108.vaadin1</version>
     </dependency>
     <dependency>
       <groupId>org.projectlombok</groupId>
--- a/app/connectors/connector-verifier/pom.xml
+++ b/app/connectors/connector-verifier/pom.xml
@@ -48,6 +48,36 @@
       <groupId>com.fasterxml.jackson.datatype</groupId>
       <artifactId>jackson-datatype-jdk8</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.module</groupId>
+      <artifactId>jackson-module-jsonSchema</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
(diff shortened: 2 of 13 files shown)


app/ui/src/app/integrations/edit-page/step-configure/filter-steps/basic-filter.component.ts
```diff
@@ -115,8 +115,7 @@ export class BasicFilterComponent implements OnChanges {
     this.integrationSupport
       .getFilterOptions(dataShape)
       .toPromise()
-      .then((resp: any) => {
-        const body = JSON.parse(resp['_body']);
+      .then((body: any) => {
```
gashcrumb: You don't want this change on this branch, this version still uses the restangular impl.

rhuss: If this is fixed, I'm happy to merge (and kick of a new TP3 release)

zregvart:
@gashcrumb @rhuss could you take a look, we've taken a bit different approach on `master` with #1374. Don't mind the Codacy issue, I can't seem to get it to analyze, though I'm sure it would find something.

app/connectors/connectors/sql/sql-common/pom.xml
```diff
@@ -55,8 +55,9 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>org.json</groupId>
-      <artifactId>json</artifactId>
+      <groupId>com.vaadin.external.google</groupId>
+      <artifactId>android-json</artifactId>
+      <version>0.0.20131108.vaadin1</version>
```
rhuss: Why is this lib better ? Version number looks scary (major and datewise)

rhuss: BTW, I was and always am a big fan of json-simple (only dep of Jolokia)

Never let me down ...

zregvart: Not better, just under a different license, JSONAssert uses this implementation of `org.json` which is a clean room implementation there is a bit of text on this in 81102c278ad47c76e2f71c6a0fb750ad3d9a9edc. Without this we have duplicate classes on classpath.

rhuss: JSONAssert is only for testing, or ? Is it used also during runtime ? wonder why jsonassert dep is included not only for scope test.

rhuss:
looks good to me. Just a small comment, dont know whether you want to address it. if not, just remove the wip label.

rhuss:
@zregvart I just did a quick smoke test with this fix included in 1.2.8 but it still does not seem that the pull down works if there is Twitte before the filter step. If the filter step is after a datamapping step (with DB as target), then the autocomopletion works, though.

zregvart:
I tested with DB step, could be that Java introspection is broken also.

PR #1361, opened 2018-02-01
Updates to /support

michael-coker:
re-work the page to bring the overall layout inline with the design from UXD.


app/ui/src/app/support/support.component.ts
```diff
@@ -0,0 +1,158 @@
...
+      error => {
+        console.log("Error downloading the file.");
+        console.log(error);
+      }
+    );
+    return {  };
+  }
+
+  // Handles events when the user interacts with the toolbar filter
+  filterChanged($event) {
```
deeleman: `$event` is the reserved template keyword for event handlers, but we usually do rename it into `event` in the Angular component class, since there some early drafts for future ECMAScript implementation that might turn into using special char-prefixed names for native system features.

In other words, I'd rename this to plain `event` instead (you have to leave `$event` in the template though).

app/ui/src/app/support/support.component.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+  ];
+
+  constructor(
+    public store: IntegrationStore,
+    public integrationSupportService: IntegrationSupportService,
+  ) {
+  }
+  
+  buildData(data: any = {}) {
```
deeleman: In longer method implementations like this, TypeSCript advices to annotate the return type, if any, or just add `: void`. That's a good practice for self-documenting code in TypeScript.

deeleman: BTW, default value assignation in method signatures makes annotating the type redundant. It is safe to remove the `: any` annotation here.

app/ui/src/app/support/support.component.ts
```diff
@@ -0,0 +1,158 @@
...
+    console.log(data);
```
deeleman: Hey, hello Mr. Console, how's your day today. Got bad news for ya fella... You're gonna get removed

sayonara-baby
!

app/ui/src/app/support/support.component.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  onSubmit() {
+    let chosen = [];
+    if(this.allLogsSelected) {
+      chosen = this.items;
+    } else {
+      chosen = this.items.filter(x => x.selected === true);
+    }
+    let input = {};
+    chosen.forEach(el => input[el.name] = true);
```
deeleman: I'ma bit confused here, why do you want to set element's name as `true`instead of its original values (`Integration name X`)?

app/ui/src/app/support/support.component.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  selectedItems() : number {
+    return this.items.filter(x => x.selected === true).length;
+  }
+
+  public ngOnInit() {
+    this.store.list.subscribe(integration  => {
+      this.items = integration as Integration[];
+    });
```
deeleman: A few remarks here:

- All class members in TypeScript are `public` by default, so we never annotate that accessor. The `public` keyword is used in constructor parameters only, to inform the compiler that such parameter also must become a class member.
- You should not typecast the value in the assignation expression, but on the lambda declaration instead,
- Code should be self-explanatory, hence lambda names should be too. If `integration` is meant to be an array of `Integration` objects (therefore a collection) then name it accordingly as `integrations`. There's actually a shortcut type name `Integrations` defined at a model level for this, so you can use it instead.
- This subscribe statement will keep emitting data long after this component is destroyed, because that is teh nature of the `Observer` pattern. Therefore, the `Subscription` should be disposed upon tearing down the component. Please implement the `OnDestroy` interface and add an `unsubscribe` line to it.

app/ui/src/app/support/support.component.scss
```diff
@@ -0,0 +1,72 @@
...
+  
+    }
+  
+  }
+
+  .download {
+    padding-top: 10px;
+  }
+
+}
```
deeleman: When coding SASS for Angular components (and [Web Components](https://en.wikipedia.org/wiki/Web_Components) in general), you need to bear in mind the following: The compiled CSS will be injected inline upon instantiating the component dynamically, and the composed CSS will be removed from the DOM upon destroying such component. In that sense, the lesser CSS we inject in the component style directives, the better. The rationale for that is because we're ultimately creating a visual language though components, and therefore components are meant to be standard and share the same look and feel. That's not doable when each component features its own visual ruleset. 

On the other hand, refining the styles of each UI element on a per component basis is not scalable. A regular components-driven application can feature literally hundreds (or thousands!) of components. Tweaking CSS one component at a time each time the UI changes is simply not doable. As a rule of thumb, the styles defined on a single component are just meant to _override_ any generic style that affects such component because of its very particular nature or behavior. 

Just a few principles when adding SCSS code to the codebase:

- First and foremost, ask to yourself if that style will be eventually reused by other instances of the application. If so, then consider adding it as part of our general styles available at `app/src/ui/scss`.
- By importing the application-level `@import 'syndesis-sass';` you'll gain immediate access yo our application-level variables, mixins and functions (including our vendor variables such as PF's color palette variables!). The more reusable our CSS code is, the more lean it becomes.
- Avoid hardcoding values in the stylesheets. We have generic gutters defined there, ready for use in margins and paddings. If you feel that none of those suit your requirements, that is a good sign that this component layout breaks the overall rules and then deserves a custom approach, but think twice before goind down that way.
- Following up with this, never hardcode colors in your code. use the PF values instead.
- The SCSS code should allow for getting a general idea of the DOM layout and component hierarchy just by inspecting its indentation and selector names. Please, use the [BEM conventions](http://getbem.com/introduction/) to accomplish this.
- Avoid using generic names such as `.download` or `.thing` and feel free to use custom className prefixes such as `.syn-`. This ensures that our custom styles are not affected by vendor stylesheets shoudl we introduce third party CSS libraries in the future.
- If you create helper classnames that affect a DOM/custom element default behavior, use modifier prefixes to make the helper more self-explanatory, such as `.is-` (for toggable styles) or `.has` for overloaded styles (when not using BEM's `--modifier` suffix pattern)

app/ui/src/app/support/support.component.html
```diff
@@ -0,0 +1,82 @@
...
+    <p>
+      System level and application level diagnostics will be captured since both are required to troubleshoot any issue. Usernames and passwords are not present in the downloaded zip file. 
+    </p>
+
+    <div class="mainform">
+      <p>Select the integrations to capture diagnostics for:</p>
+      <form novalidate>
+        <fieldset class="formfields">
+          <div class="radio">
+              <label><input type="radio" id="alllogs" name="logs" [value]="true" [(ngModel)]="allLogsSelected" (click)="deselectAll();"> All integrations</label>
```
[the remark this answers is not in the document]
michael-coker: Thanks for the feedback. This is my first contribution to the project. Is there a resource you can point me to for the max line length and any other HTML/CSS guidelines or best practices for the project that I should check out? Is there a particular linter I should be using?

paoloantinori:
Hi @michael-coker github shows a lot of commits and conflicts, confirming that your rebase has not been done properly. This is what I was expecting while we were discussing this in chat.

You should rebase on top of origin/master, and fix conflicts locally.

I'd suggest you to either close this PR and re-open a new one or to fix locally and that do a push -f on your private branch attached to this pr: `michael-coker:support_2`

michael-coker:
@paoloantinori thanks, yeah I noticed that at the end of the day. I'll go ahead and close this, back up the 3 files I changed, and rebase again. I'm going to address the issues @deeleman brought up, which may take some time since this was my first contribution and I need to spend some time looking through the resources he provided. Once I rebase properly, should I go ahead and submit another PR with the commits in this PR, or wait and submit after I've addressed the issues above?

deeleman:
In regards of linting the TypeScript code, you can run `yarn lint` before committing your changes. For linting your SASS code, that will depend on your IDE of choice and common sense. My remarks on CSS were more geared towards architectural principles rather than formal aspects of the code ;)

paoloantinori:
mmm good question. I think that @deeleman is actually commenting on code that is already in master, there with my pr :sweat_smile:

it's probably something I should do

gashcrumb:
@paoloantinori have opened a PR that also address the lint errors in this class, so no worries on that front.

Issue #1324, opened 2018-01-31
TP3 Extension scenario


rhuss:
From out Mail Thread:
-------------

Some way to send a JMS message to kick off the integration.
We discussed a couple options here:
1) Extending the TODO app to send a JMS message by pushing a button.
2) Creating a standalone JMS script to send the message.

Option #2 gets pretty technical for a user (need Java and Maven most likely at a minimum).

An AMQ template to host a broker in the user's Fuse Online account.
I believe this is done and tested, but I'm not sure where it's located.  We discussed pre-loading the template in a user's account so that they can just click "Add to Project" to provision AMQ.  I guess another option would be to pre-install AMQ but have the replica count = 0 so that all a user has to do is bump the count to 1.

Anything else?  For reference, I have included the original scenario description below.

Here are some high-level ideas that we should do regardless of the specific scenario IMO:
Let's make it a single step in the extension to keep it simple.

* The step should do something with data shapes, either in or out, so that we can show it's type aware.
* The step should do something significant enough that it warrants creating a tech extension.
* The extension source code should be pushed to a separate example repo in Syndesis so that people can view/clone as they like.
* The extension binary should be hosted somewhere (GitHub, MRRC, etc.), so that people can deploy it directly.

I'm kinda thinking something like JMS -> Step -> Filter -> Data Map -> DB.  The scenario will be that a batch of inventory receipts (we can invent the format) is received via JMS.  The split step splits the inventory receipt into individual items received and then looks up contact info for each shipper (could be in-memory map with a few sample entries) and adds that to the message.  The output shape for the custom step defines the correct data type for an individual inventory receipt. The filter step only allows "DAMAGED" inventory to pass, followed by a data mapping step to map into our TODO format.  Finally the DB connector inserts the TODO to call the shipper with details on the damaged inventory.

--------

dsimansk:
@rhuss @lburgazzoli I've tested a small PoC that uses STOMP client in PHP to send message to broker. [1] It works quiet well to my surprise. Imho we could hardcode username/password for broker and deploy it with replica count set to 0, as suggested by Keith. For the users we would document how to scale the deployment and also how to create an example AMQ connection with our credentials. Then Camel route can be triggered from Todo app `1-click send button` with predefined message to be processed. Wdyt?

[1] https://github.com/syndesisio/todo-example/compare/master...dsimansk:stomp-poc

dsimansk:
[...]

PR #1362, opened 2018-02-01
chore(1.2.x): Include AMQ deployment in templates

dsimansk:
This PR adds AMQ broker to a default deployment with replica count set to 0. Also I've upgraded AMQ image tag to `1.3`. There's hardcoded username and password for broker to keep it simple.

It's part of example scenario from https://github.com/syndesisio/syndesis/issues/1324


rhuss:
@dsimansk I'm merging this right now as I'm about to cut a new release. We can add the limits later, too.

Issue #377, opened 2017-11-20
Upgrade Angular CLI to v1.5.x


deeleman:

jimmidyson:
Getting warnings in build:

> [WARNING] npm WARN deprecated angular-cli@1.0.0-beta.28.3: angular-cli has been renamed to @angular/cli. Please update your dependencies.

deeleman:
This issue has not been actioned yet, so that warning is unrelated. In any event `angular-cli@1.0.0-beta.28.3` is quite too outdated and I recall someone (perhaps it was you @seanforyou23?) reported this a few days ago. Apparently it turns out it was caused by an outdated image fetched from the cluster.

Perhaps @seanforyou23 might provide a workaround :)

PR #1283, opened 2018-01-29
[WIP] Chore/upgrade dependencies

deeleman:
This takes over the [previous PR](https://github.com/syndesisio/syndesis/pull/1265) submitted by @seanforyou23 (please read its README for further details) in order to apply changes faster.

PR #1367, opened 2018-02-02
feat: Bring multiple data shapes, Step IDs and labels into datamapper + NG5

deeleman:
Fixes: #1135
Fixes: #1137 
Fixes: #380 
Fixes: #377 

This PR supersedes #1338 and includes the changes included in #1283, rebased with the latest repo updates, which are required for the former.

Regarding #1338:

> Fixes: #1135
> Fixes: #1137 
> 
> I was still debugging though, submitting a PR with WIP flag as I hit a version mismatch issue. Due to the issue now I can't run Syndesis UI with latest AtlasMap UI.
> 
> Recently I hit following issue in AtlasMap UI and upgraded the `ng-packagr` to `2.0.0-rc.6`
> https://github.com/dherges/ng-packagr/issues/311
> 
> Then Syndesis UI fails to build with
> ```
> ERROR in Metadata version mismatch for module /home/tomo/workspace/iPaaS/syndesisio/syndesis/app/ui/node_modules/@atlasmap/atlasmap.data.mapper/atlasmap-atlasmap.data.mapper.d.ts, found version 4, expected 3
> 
> ERROR in ./src/main.ts
> Module not found: Error: Can't resolve './$$_gendir/app/app.module.ngfactory' in '/home/tomo/workspace/iPaaS/syndesisio/syndesis/app/ui/src'
>  @ ./src/main.ts 6:29-76
>  @ multi ./src/main.ts
> ```
> 
> It seems to be caused by the Angular version mismatch, as AtlasMap is already based on Angular5 while Syndesis UI is still on Angular4.
> I tried to upgrade only angular-cli and etc, but couldn't find a happy combination. Apparently we need to update to Angular5.


app/ui/package.json
```diff
@@ -25,57 +24,62 @@
...
+    "@angular/cli": "^1.6.5",
+    "@angular/common": "^5.2.0",
+    "@angular/compiler": "^5.2.0",
+    "@angular/core": "^5.2.0",
+    "@angular/forms": "^5.2.0",
+    "@angular/http": "^5.2.0",
+    "@angular/platform-browser": "^5.2.0",
+    "@angular/platform-browser-dynamic": "^5.2.0",
+    "@angular/router": "^5.2.0",
+    "@atlasmap/atlasmap.data.mapper": "1.33.0-SNAPSHOT.20180131184819",
```
apupier: why it has changed to a "match" version from a "compatible with version"?

igarashitm:
Thanks @deeleman ! As I was still halfway debugging my code, I'll continue on top of this one.

app/ui/package.json
deeleman: Good question! It should actually be like that ("compatible", I mean) but the reason why the Angular upgrade took so long was precisely because the fragile stability originated by the massive amount of shared downstream dependencies between Syndesis itself and its own dependency tree (ngx-bootstrap, etc).

This forced us to track down each and every dependency and finetune its versioning to the path segment to ensure that it would play nicely not only with syndesis but with the rest of dependencies as well. In an ideal case scenario, NG dependencies will be distributed in a packaged artifact that implements the metadata info that Yarn requires, but it's not always like that, and those imported packages whose metadata is not complete or too open, might cause a class in other dependency's own dependencies graph. 

This PR aims to achieve a stable Angular 5 upgrade and then we'll go reviewing each dependency status on a per case basis.

Issue #700, opened 2017-12-12
UXD for visual indicator when data types don't match


kcbabo:
We should have some visual indicator in the UI when data types don't match.  For example, when setting up Twitter to Salesforce, the output data from Twitter does not match the input data for Salesforce.  This may not be obvious to the end user, yet we have all the information we need to make the user aware.  I suspect this will become really important with tech extensions as they are type-aware and the user really has no knowledge of what's happening inside.

Some suggestions:
- All steps/connections in an integration flow should have a visual indicator of input and output types.  This could be a hover, click, whatever.
- When the current data shape does not match the input shape of a step/connection, a visual indicator should be displayed indicated that the situation must be resolved (e.g. data mapping, data type step, etc.).

@rhuss @gashcrumb @amysueg : UXD and UI components here.  I don't think any backend work is required.

igarashitm:
@dongniwang had a question if we want to suggest adding datamapper step to the users. I think there are 2 options for users there, 
* #1110 add datamapper
* #1136 change data shape via action config

How do we want to suggest these 2 options to the users would be an UXD thing.

amysueg:
thanks, @dongniwang , let me know if you want to discuss ideas for this.

kcbabo:
Maybe we do it this way ...

If a step (which includes connections) in a flow requires a specific data type and no data type is provided, then we notify the user that the connection requires a specific data type and the current data is untyped (with a link to doc on setting data type in action config).

If a step in a flow requires a specific data type and a different data type is provided, then we notify the user that a data mapping step is required to resolve the difference.

These two approaches are complementary.  In the case where a user starts with JMS (typeless) and finishes with Salesforce, they would first see a notification that a data type needs to be specified and once that's complete they would see a follow-up message that a data mapping is required.  That matches the reality as the user needs to perform two separate tasks to resolve the issue.

igarashitm:
Thanks Keith, that sounds promising. I was not really aware that only typeless connection have an option to define a data shape in action config.

dongniwang:
Hi @kcbabo and @igarashitm, I have a few questions:
- Is the data shape associated with each connection OR connection actions? Could different connection actions have different data types? 
- Can you give me some examples what data type information we want to show to the users? Is it like .xml .json OR something else? I'm trying to access how much space we need to display that information, so it would really helpful to get some real examples. 

Thank you!

kcbabo:
[...]

PR #1378, opened 2018-02-02
fix: Field selection/completion broken in Filter Step

### LINKED ISSUE #1295: Field selection/completion broken in Filter Step

zregvart:
These are leftovers from #1359 needed for `master`, now based on #1374

Fixes #1295


app/connectors/connectors/sql/sql-common/pom.xml
```diff
@@ -55,8 +55,9 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
-      <groupId>org.json</groupId>
-      <artifactId>json</artifactId>
+      <groupId>com.vaadin.external.google</groupId>
+      <artifactId>android-json</artifactId>
+      <version>0.0.20131108.vaadin1</version>
```
lburgazzoli: Minor thing, can't we use jackson ?

oscerd: +1

zregvart: I can rewrite this to use Jackson, not an issue, let me do that.

lburgazzoli: thx, not a blocking thing

zregvart:
@KurtStam sorry SQL connector was collateral, could not stop myself when I got into a small change there

@gashcrumb hopefully this makes more sense

Also fixes #1453

app/ui/src/app/integration/edit-page/step-configure/step-configure.component.ts
```diff
@@ -138,16 +135,38 @@ export class IntegrationStepConfigureComponent extends FlowPage implements OnIni
...
+
+    // we now have previous and next steps that have the data shapes defined
+    // now we need to determine if there is a mapping step inbetween
+    // mapping step doesn't specify shape but it changes it to the input shape
+    // of the step after it, in that case we need to look at the input sep of
+    // the next step, otherwise we need to look at the output step of the
+    // previous step
+
+    const prevIdx = this.currentFlow.steps.indexOf(prevStep);
+    const nextIdx = this.currentFlow.steps.indexOf(nextStep);
```
gashcrumb: So I really need to hide that `steps` attribute from currentFlow :-)

You know, maybe a better solution is to just add an `action` property to a data mapper step and just set the input/output data shape on it from the data mapper host component, then this code simply becomes `const step = currentFlow.getPreviousStepWithDataShape(this.position);` and the filter always sees the output data shape?

I know the plan is to add data shapes to mapper steps, right @igarashitm?

Mucking around with the array index scares me basically :-)  It feels like just doing `currentFlow.getPreviousSteps()`, checking if any of them has a mapper and then getting the previous or subsequent step's data shape is safer.

igarashitm: We definitely need to add action and input&output data shape on datamapper step (#1394). This is required to achieve visual indicator of the data shapes for each steps (#700).

gashcrumb: Ok cool.  Since that's still outstanding I'm fine for merging this, one the datamapper step has data shapes this will greatly simplify this.

PR #1383, opened 2018-02-02
UI, fix for empty collections and connectors

paoloantinori:
@oscerd

### DIFF
--- a/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts
+++ b/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts
@@ -191,7 +191,7 @@ export class IntegrationConfigureActionComponent extends FlowPage
       this.loading = false;
       return;
     }
-    if (!descriptor || descriptor === undefined || !descriptor.propertyDefinitionSteps) {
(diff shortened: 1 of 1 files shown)


app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts
```diff
@@ -191,7 +191,7 @@ export class IntegrationConfigureActionComponent extends FlowPage
       this.loading = false;
       return;
     }
-    if (!descriptor || descriptor === undefined || !descriptor.propertyDefinitionSteps) {
+    if (!descriptor || descriptor === undefined || Object.keys(descriptor.propertyDefinitionSteps[0]).length === 0) {
```
gashcrumb: Isn't this going to cause an NPE if `propertyDefinitionSteps` is undefined?

paoloantinori: yes, but the way I have seen we are propagating logical empty values, it's never undefined.

somehow, when we have no property we represent that as a `[{}]`, that I'm happy to get rid of, but I was afraid to touch and risking of creating further issues

PR #1393, opened 2018-02-02
fix: skip UI tests if tests are skipped

gashcrumb:


cunningt:
+1 this would be useful for the pipeline builds.

app/ui/pom.xml
```diff
@@ -112,6 +112,7 @@
             </goals>
             <configuration>
               <arguments>test:ci</arguments>
+              <skip>${skipTests}</skip>
```
rhuss: Shouldn't we add this to the `flash` profile, too ? Currently with -Pflash it still runs the CI tests, which is not so 'flashy'. 

BTW there is also a 'build:ci' task, which also takes a long time for me.

TBH, I would prefer to run the CI tests only on demand and not by default (in contrast to unit tests which should run by default).

Is there anything else we could disable for the `flash` profile ? Like `yarn lint`. Would be awesome if we could collect a list with the minimal targets required for successful UI build.

At the moment the build with -Pflash takes 16 mins (with 11:44mins being UI).

rhuss: If you have a minimal list handy, I would prepare a proper `flash` profile for the UI pom.xml

gashcrumb:
Yeah, tests and lint can be skipped for `--flash`.  The `build:ci`task is needed I think though, we could try a second build target that drops the `-aot` and `-prod` arguments and see if that produces a working ui image.

PR #1416, opened 2018-02-05
[integration] move secrets generation to project-generator

### LINKED ISSUE #1406: [integration] move secrets generation to project-generator
Secrets are now generated by some ActivateHandler code but as they are tightly coupled with project generation, such code should be moved to project-generator

lburgazzoli:
Fixes  #1406


lburgazzoli:
retest this plese

app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGenerator.java
```diff
@@ -82,6 +98,94 @@ public InputStream generate(IntegrationDeployment deployment) throws IOException
...
+                        .filter(WithConfiguredProperties.class::isInstance)
+                        .map(WithConfiguredProperties.class::cast)
+                        .map(WithConfiguredProperties::getConfiguredProperties)
+                        .flatMap(map -> map.entrySet().stream())
+                        .filter(Predicates.or(connector::isSecret, action::isSecret))
+                        .distinct()
+                        .forEach(
+                            e -> {
+                                String key = String.format("%s-%s.%s", componentScheme, index, e.getKey());
+                                String val = resourceManager.decrypt(e.getValue());
```
rhuss: What happens when `decrypt() failed ? Should the whole processing be stopped or only this particular secret ? Would be nice to add the key here to the error message, too.

app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGenerator.java
```diff
@@ -82,6 +98,94 @@ public InputStream generate(IntegrationDeployment deployment) throws IOException
...
+                        .map(WithConfiguredProperties.class::cast)
+                        .map(WithConfiguredProperties::getConfiguredProperties)
+                        .flatMap(map -> map.entrySet().stream())
+                        .filter(Predicates.or(connector::isEndpointProperty, action::isEndpointProperty))
+                        .filter(Predicates.or(connector::isSecret, action::isSecret))
+                        .distinct()
+                        .forEach(
+                            e -> {
+                                String key = String.format("%s-%s.%s", componentScheme, index, e.getKey());
+                                String val = resourceManager.decrypt(e.getValue());
```
rhuss: See comment above about error handling for `decrypt()`.

In addition I wonder whether duplicated code could be reduced (not of high priority).

I'm a bit surprised that there are so many (also non-linear) ways to obtain configured properties. But that is probably due to the fact that we also support 'old style' connectors. It would be really awesome if we could get rid of the old style before GA to get rid of this baggage. wrt. to Scheduler: Wouldn't the timer connector combined with a regular connector (eg. sql connector) also fit the bill ?

And how does the Twitter connector (new style) does the scheduling ?

app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGenerator.java
```diff
@@ -82,6 +98,94 @@ public InputStream generate(IntegrationDeployment deployment) throws IOException
...
+                        .filter(WithConfiguredProperties.class::isInstance)
+                        .map(WithConfiguredProperties.class::cast)
+                        .map(WithConfiguredProperties::getConfiguredProperties)
+                        .flatMap(map -> map.entrySet().stream())
+                        .filter(Predicates.or(connector::isComponentProperty, action::isComponentProperty))
+                        .distinct()
+                        .forEach(
+                            e -> {
+                                String key = String.format("%s.configurations.%s-%s.%s", componentScheme, componentScheme, index, e.getKey());
+                                String val = resourceManager.decrypt(e.getValue());
```
rhuss: dito wrt error handling. Maybe this inner decrypt part at least could be moved in a common method ? (so that e.g. the property key generation is guaranteed to be the same for all variants ?)

app/integration/project-generator/src/test/java/io/syndesis/integration/project/generator/ProjectGeneratorTest.java
```diff
@@ -34,7 +41,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-@SuppressWarnings("PMD.ExcessiveImports")
+@SuppressWarnings({"PMD.ExcessiveImports", "PMD.ExcessiveMethodLength"})
```
rhuss: ;-)

app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGenerator.java
lburgazzoli: The encrytion component does not explicit throws exceptions but return null if fails, this can of course be checked (see https://github.com/syndesisio/syndesis/blob/master/app/rest/dao/src/main/java/io/syndesis/dao/manager/EncryptionComponent.java#L102-L117)

app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGenerator.java
lburgazzoli: Let me see what I can do.

app/integration/project-generator/src/test/java/io/syndesis/integration/project/generator/ProjectGeneratorTest.java
lburgazzoli: Immutables ...

lburgazzoli:
@rhuss should be a little better now

Issue #1371, opened 2018-02-02
UI, empty collections issues


paoloantinori:
Related to this refactoring: https://github.com/syndesisio/syndesis/pull/1353

There are still some issues with:
- extensions
- s3 connectors (probably connectors more in general)

I'm working on this.

cc @gashcrumb @lburgazzoli @oscerd

oscerd:
S3 connector fix is coming.

paoloantinori:
This pr

PR #1420, opened 2018-02-05
fix: avoid npe when checking response

### LINKED ISSUE #1418: Twitter mention action, config page throws `SyntaxError: Unexpected token u in JSON at position 0`
As the action object doesn't have `propertyDefinitionSteps` in the response, so this line of code blows up:

https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts#L194

gashcrumb:
fixes #1418


paoloantinori:
This breaks again the behavior for empty collections:

https://github.com/syndesisio/syndesis/issues/1371

PR #1429, opened 2018-02-06
Initial version of Prometheus metrics provider

dhirajsb:


dhirajsb:
n.b. there might be a little gremlin or two in the code still. I did some unit testing earlier today, but then made the mistake of pulling master, and haven't been able to create an integration to test, all day. 
I didn't implement the ```getTotalIntegrationMetricsSummary``` method, since I don't think we were planning on showing combined metrics for the entire iPaaS.

app/integration/project-generator/src/main/resources/io/syndesis/integration/project/generator/templates/redhat/application.properties.mustache
```diff
@@ -11,3 +11,6 @@ management.port = 8081
 # disable all management enpoints except health
 endpoints.enabled = false
 endpoints.health.enabled = true
+
+# disable default health indicators, especially jms
+management.health.defaults.enabled = false
```
rhuss: just out of curiosity, why is this required ?

app/rest/rest/pom.xml
```diff
@@ -339,6 +364,7 @@
             <ignoredUnusedDeclaredDependency>io.syndesis.integration:integration-project-generator</ignoredUnusedDeclaredDependency>
             <ignoredUnusedDeclaredDependency>com.sun.xml.bind:jaxb-impl</ignoredUnusedDeclaredDependency>
             <ignoredUnusedDeclaredDependency>com.sun.xml.bind:jaxb-core</ignoredUnusedDeclaredDependency>
+            <ignoredUnusedDeclaredDependency>commons-logging:commons-logging</ignoredUnusedDeclaredDependency>
```
rhuss: Why do you need then commons-logging during compile time ?

app/rest/rest/src/main/java/io/syndesis/rest/metrics/PrometheusMetricsProviderImpl.java
```diff
@@ -0,0 +1,106 @@
...
+        final List<Pod> pods = integrationPodList.stream()
+            .filter(p -> Readiness.isReady(p)
+                && integrationId.equals(p.getMetadata().getAnnotations().get("syndesis.io/integration-id")))
+            .collect(Collectors.toList());
+
+        if (pods.isEmpty()) {
+            throw new IllegalArgumentException("Missing pod with integration id " + integrationId);
+        }
+
+        final String podName = pods.get(0).getMetadata().getName();
```
rhuss: What is if the same integration spans multiple pods ? (e.g if a pod is restarted for some reason). Or if the pod is already dead ?

First I would think that is a valid result that no such pod exists, so throwing an IllegalArgumentException seems to be a bit heavy.

Then, couldn't we avoid to query Kubernetes alltogether ? I mean, the integration pod could expose the integration id in the pod label/annottion, this could be picked up by Prometheus during a scrape run, and the we can query the integration id directly from Prometheus (as it has been stored along with the metrics). 

Not sure though whether Prometheus scraping picks up labels/annotations from pods, but I think this would be very useful (and I would be surprised if this is not already possible).

app/rest/rest/src/main/java/io/syndesis/rest/metrics/prometheus/HttpQuery.java
```diff
@@ -0,0 +1,83 @@
...
+                if (first) {
+                    first = false;
+                } else {
+                    queryExpression.append(',');
+                }
+                queryExpression.append(label.getLabel());
+                queryExpression.append('=');
+                queryExpression.append('"');
+                queryExpression.append(label.getValue());
+                queryExpression.append('"');
```
rhuss: matter of tast, but I would delegate the adding of the value to the label itself (i.e. a `label.appendTo(queryExpression)` and with

```java
default void appendTo(StringBuilder params) {
    if (params.size() > 0) {
        params.add(",");
   }  
   params.append(getLabel()).append("=\"").append(getValue()).append('"');
}
```

app/rest/rest/pom.xml
```diff
@@ -161,11 +161,36 @@
...
 
+    <dependency>
+      <groupId>javax.ws.rs</groupId>
+      <artifactId>javax.ws.rs-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-client</artifactId>
+    </dependency>
```
rhuss: It would be awesome if we would collect the provider for all metrics related UI queries into an common module `rest/metrics`, similarly to the rest/db-logging. 

Suggestion:

```
rest/metrics
     +--- jsondb (includes Collector and Metrics Provider for looking up in the DB)
     +--- prometheus (include the access to Prometheus)
     +--- api (the interface accessed to be the endpoint)
```

That way we could keep the resteasy client dependency out of the rest module (the same way as we did for accessing the 'verifier') as well as other dependencies.

The metrics rest endpoint then just dispatches to one of the implementations via a defined interface. This implementation is injected by the proper SpringBoot configs triggers.

// @KurtStam

app/rest/rest/src/main/java/io/syndesis/rest/metrics/PrometheusMetricsProviderImpl.java
```diff
@@ -0,0 +1,106 @@
...
+    public IntegrationMetricsSummary getIntegrationMetricsSummary(String integrationId) {
+
+        // map integration Id to pod name
+        List<Pod> integrationPodList = kubernetes.pods().withLabel("integration").list().getItems();
+        final List<Pod> pods = integrationPodList.stream()
+            .filter(p -> Readiness.isReady(p)
+                && integrationId.equals(p.getMetadata().getAnnotations().get("syndesis.io/integration-id")))
+            .collect(Collectors.toList());
+
+        if (pods.isEmpty()) {
```
iocanel: Minor suggestion: Instead of getting a list of pods and checking if its empty, you could just use `findFirst().orElseThrow(()-> new IllegalArgumentException("Missing pod with integration id" + integrationId)` directly on the stream.

chirino:
I should be able to refactor this into it's own module in a bit.

app/rest/rest/src/main/java/io/syndesis/rest/metrics/PrometheusMetricsProviderImpl.java
dhirajsb: Hi @iocanel that makes sense. I left it as a list since we would also have to handle the case when user wants to run replicas of the integration in the future. 
That might not be in scope today, but has to be handled later. Should have added a comment.

app/integration/project-generator/src/main/resources/io/syndesis/integration/project/generator/templates/redhat/application.properties.mustache
dhirajsb: When Spring Boot sees ActiveMQ libraries, it thinks it helping by injecting a `vm://localhost` connection factory in JMS health indicator and creating an in vm broker to monitor it. 

Which fails, since the vm transport is part of the ActiveMQ Broker libraries, not client. So it spectacularly barfs, over and over in the logs.

dhirajsb: On the question about the need to lookup the pod, I agree. We should figure out a way to make it easier to do that from the UI.

dhirajsb: With only `integrationId` the method `getIntegrationMetricsSummary` is only useful for summary metrics, not per deployment metrics. We should refactor `MetricsProvider` methods to:
```
getDeploymentMetrics ( integrationId, deploymentId )
getIntegrationMetrics ( integrationId )
```
Where the latter method aggregates metrics for counters, and uses appropriate min/max value for timestamps, etc.

app/rest/rest/pom.xml
dhirajsb: Thought it came from jackson or something, will have to check again.

app/rest/rest/src/main/java/io/syndesis/rest/metrics/PrometheusMetricsProviderImpl.java
dhirajsb: @rhuss you're right, we can avoid using k8s to lookup pods. It looks like k8s can't list terminated pods anyway. 
If the UI can provide `deploymentId` and `integrationId`, we can use that in Prometheus lookup since we are actually using [relabling config](https://github.com/syndesisio/syndesis/blob/master/app/deploy/generator/06-syndesis-prometheus.yml.mustache#L49-L50) to expose _all_ pod labels in metrics.

chirino:
Tried deploying an integration from the this branch.  Got this error in the integration pod logs:

```
Starting the Java application using /opt/run-java/run-java.sh ...
  | exec java -javaagent:/opt/jolokia/jolokia.jar=config=/opt/jolokia/etc/jolokia.properties -javaagent:/opt/prometheus/jmx_prometheus_javaagent.jar=9779:/tmp/prometheus-config.yml -XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:+ExitOnOutOfMemoryError -cp . -jar /deployments/project-0.1-SNAPSHOT.jar
  | Exception in thread "main" java.lang.reflect.InvocationTargetException
  | at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  | at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
  | at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  | at java.lang.reflect.Method.invoke(Method.java:498)
  | at sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:386)
  | at sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:401)
  | Caused by: java.io.FileNotFoundException: /tmp/prometheus-config.yml (No such file or directory)
  | at java.io.FileInputStream.open0(Native Method)
  | at java.io.FileInputStream.open(FileInputStream.java:195)
  | at java.io.FileInputStream.<init>(FileInputStream.java:138)
  | at java.io.FileReader.<init>(FileReader.java:72)
  | at io.prometheus.jmx.shaded.io.prometheus.jmx.JmxCollector.<init>(JmxCollector.java:74)
  | at io.prometheus.jmx.shaded.io.prometheus.jmx.JavaAgent.premain(JavaAgent.java:36)
  | ... 6 more
```

chirino:
Also seeing this error when I try to hit the HTTP endpoint:

https___syndesis_192_168_64_5_nip_io_api_v1_metrics_integrations_-l4hnux2udjccytitqmx

dhirajsb:
Will take a look at what's going on. Glad you could create an integration, I was having ui issues earlier.

dhirajsb:
Lots of enhancements, supports Prometheus functions and ranges now. Need to tweak values based on how much timeseries data we will save eventually. 
Saw some interesting side effects of `integration-id+deployment-id` not mapping one to one to  k8s labels. Makes aggregating metrics a little _interesting_ if user activates/deactives integration a lot. 
Need to look in detail at how the max and total computation is being affected when I am awake sometime tomorrow.

rhuss:
@dhirajsb it looks like the PR is a bit messed up, containing a lot of changes already on master. Or did you really touched 336 file ?

Could you please either clean up or create a new PR ?

dhirajsb:
@rhuss those are from rebasing that hiram did from master i believe

dhirajsb:
I'll see what I can do to cleanup

dhirajsb:
@KurtStam pods labels are mapped to prometheus labels, which are not case sensitive. And k8s client can lookup pods by labels, not annotations, apparently.

KurtStam:
@dhirajsb ah I thought you were trying to obtain the integration id from the pod.

PR #1439, opened 2018-02-06
re-enable PMD

zregvart:
more detail in the commit messages


app/connectors/connector-verifier/src/main/java/io/syndesis/verifier/api/VerifierResponse.java
```diff
@@ -35,14 +36,14 @@
     private Verifier.Status status;
     private List<Error> errors;
 
-    private VerifierResponse(Verifier.Status status, Verifier.Scope scope) {
+    /* default */ VerifierResponse(Verifier.Status status, Verifier.Scope scope) {
```
lburgazzoli: can we get rid of this rule ?

PR #1444, opened 2018-02-06
Bring support UI inline with design from UXD

michael-coker:
Updates to /support to bring more inline with design from UXD with some deviations from the design signed off on by the UXD team.

https://raw.githubusercontent.com/syndesisio/syndesis/master/ux/designs/support-page/img/supportpage-landing.png

### DIFF
--- a/app/ui/src/app/support/support.component.html
+++ b/app/ui/src/app/support/support.component.html
@@ -1,77 +1,90 @@
 <div class="row">
   <div class="col-md-12">
     <h1>Support</h1>
-    <p>To obtain support, download diagnostic information through this page. </p>
-    <p>If the downloaded zip file is less than 1/2 GB then attach it to an email message that requests help and send the message to <a href="mailto:fuse-online-tech-preview@redhat.com">fuse-online-tech-preview@redhat.com</a>. </p>
-      <p>If the downloaded zip file is larger then 1/2 GB, send an email message to request help, do not attach the file but indicate that the zip file is too large to attach. 
-        Technical support will provide instructions for sending the zip file.</p>
+    <p>To obtain support, download diagnostic information through this page. If the downloaded zip file is less than 0.5 GB, attach it to an email message that requests help and send the message to <a class="text-nowrap" href="mailto:fuse-online-tech-preview@redhat.com">fuse-online-tech-preview@redhat.com</a>. If the downloaded zip file is larger than 0.5 GB, send an email message that requests help, but do not attach the file, and indicate that the zip file is too large to attach. Technical support will provide instructions for sending the zip file.</p>
 
-    <h1>Version</h1>
-    <p class="version"><strong>Syndesis</strong> v1.0</p>
+    <h2>Version</h2>
+    <p><strong>Syndesis:</strong> v1.0</p>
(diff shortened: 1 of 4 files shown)


app/ui/src/app/support/support.component.scss
```diff
@@ -1,19 +1,11 @@
-.mainform {
-    background-color: white;
-    padding: 2em;
-}
+@import 'patternfly-sass/assets/stylesheets/patternfly/variables';
```
deeleman: If we feel we'll want to refer to this SASS partial often, then let's add it to `scss/base/_vendor.scss` and then import `syndesis-sass` instead. This way we will not have to import the same partial again and again in the future

app/ui/src/app/support/support.component.scss
```diff
@@ -1,19 +1,11 @@
...
-}
+::ng-deep .toolbar-pf {
+  border: none;
+  box-shadow: none;
 
+  .toolbar-pf-view-selector {
+    font-size: $font-size-base;
+    padding-top: 5px;
+  }
+}
```
deeleman: Please add a EOF break here, thx :)

app/ui/src/app/support/support.component.html
```diff
@@ -1,77 +1,90 @@
...
-                    {{ item.name }}
-                  </div>
-                  <div class="list-pf-description">
-                    <i>{{ item.description }}</i>
+                            (onSortChange)="sortChanged($event)"
+                            [viewTemplate]="viewTemplate">
+                <ng-template #viewTemplate>
+                  <div class="toolbar-pf-action-right">
+                    <div class="form-group">
+                      <strong>{{ selectedItems() }}</strong> of <strong>{{ totalItems() }}</strong> Items
```
deeleman: Two notes here, mostly addressed to @paoloantinori I guess:

Contrary to web docs rendered serverside, SPA documents are not evaluated upon rendering on screen for the first time but often through a change detection loop, which spawns a new cycle every few (very few) milliseconds. This means that whatever function call we drop in our views will be executed every few milliseconds too. And here we can see there're two function calls. While `totalItems()` just returns a class field `length` value, which is just fine, `selectedItems()`, on the contrary, performs a lambda call. That can become memory expensive in the long run, so my suggestions would be to have just a `selectedItems: number` field defined at the `SupportComponent` class and update it every time the `selected` property changes by means of the click handler, whatever it is.

deeleman: In any event, we can get away with these bindings as is, since they only affect this component, which is meant to be eventually destroyed. However, please take into consideration the principles exposed above for the future, thanks. :)

rhuss:
Please reapprove after the yarn lint errors are fixed. thanks ...

gashcrumb:
@michael-coker this can't be merged until this `yarn lint` issue is fixed:

```
[INFO] /workspace/app/ui/src/app/support/support.component.ts
[INFO] ERROR: 37:1    no-trailing-whitespace  trailing whitespace
```

run `yarn lint` to check that your changes pass linting properly.

michael-coker:
@gashcrumb done

PR #1447, opened 2018-02-07
[feat] Adding IntegrationDeploymentMetrics to the API

KurtStam:
Hey guys, On a different - non metrics - page we list the number of messages per integration deployment. I'm proposing the following change. It is backwards compatible with the current metrics api.

### DIFF
--- a/app/core/model/src/main/java/io/syndesis/model/metrics/IntegrationDeploymentMetrics.java
+++ b/app/core/model/src/main/java/io/syndesis/model/metrics/IntegrationDeploymentMetrics.java
@@ -17,25 +17,17 @@ package io.syndesis.model.metrics;
 
 import java.io.Serializable;
 import java.util.Date;
-import java.util.List;
 import java.util.Optional;
 
 import org.immutables.value.Value;
 
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 
-import io.syndesis.model.Kind;
-import io.syndesis.model.WithId;
-
 @Value.Immutable
-@JsonDeserialize(builder = IntegrationMetricsSummary.Builder.class)
+@JsonDeserialize(builder = IntegrationDeploymentMetrics.Builder.class)
 @SuppressWarnings("immutables")
-public interface IntegrationMetricsSummary extends WithId<IntegrationMetricsSummary>, Serializable{
+public interface IntegrationDeploymentMetrics extends Serializable{
 
-    @Override
-    default Kind getKind() {
-        return Kind.IntegrationMetricsSummary;
-    }
     /**
      * @return Number of successful messages
      */
@@ -54,9 +46,7 @@ public interface IntegrationMetricsSummary extends WithId<IntegrationMetricsSumm
      */
     Optional<Date> getLastProcessed();
 
-    Optional<List<IntegrationDeploymentMetrics>> getIntegrationDeploymentMetrics();
-
-    class Builder extends ImmutableIntegrationMetricsSummary.Builder {
-        // allow access to ImmutablIntegrationMetricsSummary.Builder
+    class Builder extends ImmutableIntegrationDeploymentMetrics.Builder {
+        // allow access to ImmutableIntegrationDeploymentMetrics.Builder
     }
 }


app/core/model/src/main/java/io/syndesis/model/metrics/IntegrationDeploymentMetrics.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.Optional;
+
+import org.immutables.value.Value;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+@Value.Immutable
+@JsonDeserialize(builder = IntegrationDeploymentMetrics.Builder.class)
+@SuppressWarnings("immutables")
+public interface IntegrationDeploymentMetrics extends Serializable{
```
rhuss: Why do you need an new class which looks exactly the same as IntegrationMetrics propertywise ? Question is, would this ever differ ?

Actually for the response of the RestEndpoint I think a metrics structure like

```json
{ 
    "total": {
        "nrMessage": 20,
        ....
    },

    "deployments": [{
         "nrMessages": 2,
         ...
     }, {
         "nrMessages": 20,
         ....
     },....
     ]
}
```

would allow for having only a single 'metrics' object which only holds the values. So, not a nested, duplicated structure, but a flat one.

Regardless of this, I wonder how the correlation to the concrete Deployment is ? Is it by order in the array ? I would add somewhere the deployment id, to make it easier for the UI to know which deployment metrics belongs to which deployment. Also adding the integration id in the response helps the UI for correlation so that it does not require to keep the request along with the response, but the response stands for it alone.

About naming:(and i'm nearly sure you will ignore this, but nevertheless): I would call it `getNrMessages()` and not `getMessages()` as this method does not return messages. Same for `getLastProcessed()` which does not return the last processed message.

KurtStam: the objects are 'withId' so it holds either the integrationId or the deploymentId. I prefer a hierarchical structure as this how consumers typically use it.

app/core/model/src/main/java/io/syndesis/model/metrics/IntegrationMetricsSummary.java
```diff
@@ -53,6 +54,8 @@ default Kind getKind() {
      */
     Optional<Date> getLastProcessed();
 
+    Optional<List<IntegrationMetricsSummary>> getIntegrationDeploymentMetrics();
+
```
rhuss: well, not  really better, as this allows for arbitrary deep nesting, and a `getIntegrationDeploymentMetrics` on a deployment metrics object itself doesn't make sense. Also an IntegrationDeploymentMetrics is not a IntegrationMetricsSummary as modelled here.

If you insist of nesting, then you should use a subclass with `IntegrationMetricsSummary extends IntegrationMetrics` and adding a `List<IntegrationMetrics> getIntegrationDeploymentMetrics()` to this subclass.

Whats so bad about having a flat list with summary and list of detailede deployment metrics side-by-side ? (don't really understand also, why a summary should have a list of details)

app/core/model/src/main/java/io/syndesis/model/metrics/IntegrationDeploymentMetrics.java
KurtStam: @rhuss 

1. I removed the second commit, and went back to the original. What I have there is very much the same as what you are proposing. So let's talk tomorrow.

2. You requested it to be named 'messages' before. You are a hard one to please :)

rhuss: Right, I suggested to change `numberOfProcessedMessages` to be shorter, e.g. `messages` in #1156. And right, from time to time I evolve my opinion (and even sometimes revert it, but not here), too, as the new suggestion was indeed shorter but suffers the same semantic issue which I haven't seen at that time. An no, I'm not so hard to please, just picky about naming ;-) Happy to discuss about API naming, too, as I think its important, though often overlooked.

Again, why is having the following structure so bad, to avoid duplication and recursive definitions:

```java
class Metrics {
    int nrMessages;
    int nrErrors;
    Date start;
    Date lastProcessed;
}

class IntegrationMetrics {
      
    Metrics summary;  // or just: "metrics"
    
     List<Metrics> deploymentMetrics; // or 'integrationDeploymentMetrics'
}
```

This structure is also easy to evolve, so when you have to add a new datum you just add it to `Metrics`.  If both kind of metrics are likely to evolve differently, then indeed two `Metric` might make sense, but we can still add this when this is the case.

@syndesisio/backend @syndesisio/ui-api Any opinions on this on ? E.g. what is the easiest way to consume metrics for an integration overall (across all deployments) and for each individual deployments. Its about the structure coming back for a request for the metrics for a certain integration.

KurtStam: @rhuss I was trying to not break backwards compatibility until things are working between the teams that's all.

zregvart:
I would prefer: `messageCount`, `errorCount` and use of `java.time.ZonedDateTime` with UTC timezone instead of `java.util.Date`.

dhirajsb:
+1 for `ZonedDateTime`, `totalMessages` and `totalErrors` for number of messages and errors?

chirino:
Instead of dates, we should use a long ms since the Unix epoc. Is the simplest way to convey time to the UI. No one cares what time zone the server is using.

app/core/model/src/main/java/io/syndesis/model/metrics/IntegrationDeploymentMetrics.java
```diff
@@ -0,0 +1,52 @@
...
+    Long getMessages();
+    /**
+     * @return Number of messages that resulted in error
+     */
+    Long getErrors();
+    /**
+     * @return most recent (re-) start Date of the integration, empty if no live pods
+     * are found for this integration, which would mean that the integration is currently down.
+     */
+    Optional<Date> getStart();
```
chirino: Use long.

app/core/model/src/main/java/io/syndesis/model/metrics/IntegrationDeploymentMetrics.java
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * @return the TimeStamp of when the last message for processed
+     */
+    Optional<Date> getLastProcessed();
```
chirino: Use long.

KurtStam:
Guess what? That’s how java.util.Date is serialized.

On Feb 9, 2018, at 19:55, Hiram Chirino <notifications@github.com> wrote:

Instead of dates, we should use a long ms since the Unix epoc. Is the
simplest way to convey time to the UI. No one cares what time zone the
server is using.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/syndesisio/syndesis/pull/1447#issuecomment-364610967>,
or mute the thread

chirino:
Cool.

PR #1448, opened 2018-02-07
[fix] so that all SQL tests pass on Postgres as well as Derby

KurtStam:


app/connectors/connectors/sql/sql-common/pom.xml
```diff
@@ -76,6 +76,11 @@
       <groupId>org.apache.derby</groupId>
       <artifactId>derby</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.postgresql</groupId>
+      <artifactId>postgresql</artifactId>
+      <version>${postgresql.version}</version>
```
rhuss: Can we add <scope>test</scope> please or it this already setup in some dependency management ?

rhuss: Same for derby btw

PR #1469, opened 2018-02-07
fix: align system test scripts with latest templates and build scripts.

iocanel:


app/tests/src/test/resources/setup.sh
```diff
@@ -3,44 +3,29 @@
...
-#Configure OPENSHIFT_MASTER
 if [ -z "${OPENSHIFT_MASTER}" ]; then
     OPENSHIFT_MASTER="$(oc whoami --show-server)"
 fi
-
 if [ -z "${DEMO_DATA_ENABLED}" ]; then
     DEMO_DATA_ENABLED=false
 fi
 
+SYNDESIS_TEMPLATE_TYPE="syndesis-restricted"
```
rhuss: So is it then that we do not need syndesis-ci.yml anymore ? If so, could we remove this from deploy/, too ?

rhuss: Actually I'm really tempted for now to also remove the non restricted variants, and renaming the -restricted to without -restricted.yml. So:

* Deleting syndesis-ci.yml
* Removing syndesis.yml and syndesis-dev.yml
* Moving syndesis-restricted.yml --> syndesis.yml and syndesis-dev-restricted.yml --> syndesis-dev.yml

At the end, we then only have two templates left.

wdyt ? Would this something you could try to adopt ? It's hard to say what references all need to be updated, but as we are on a monorepo a plain text search could be enough.

tools/bin/commands/system-test
```diff
@@ -312,7 +312,7 @@ test_build() {
     export NAMESPACE_USE_EXISTING=$project
     export KUBERNETES_NAMESPACE=$project
     export OPENSHIFT_TEMPLATE_FROM_WORKSPACE=true
-    export WORKSPACE=deploy
+    export WORKSPACE=$(basedir)/../..
```
rhuss: If you are looking for `syndesis/app` then you can easily use $(appdir) which even yields an absolute path without ../..

app/tests/src/test/resources/setup.sh
iocanel: The only real purpose of the `ci` flavor, was to remove readiness/liveness probes and also remove other features that proved to be problematic in the past.

Since things seem to run, as is right now (which is also aligned with how the team runs syndesis), I think that we can scrap the `ci` flavor. Want me to do it as part of this pr?

Now regarding dumping the rest of the templates, does anyone even use the non-restred ones?

tools/bin/commands/system-test
iocanel: perfect, will update the pr.

app/tests/src/test/resources/setup.sh
```diff
@@ -3,44 +3,29 @@
...
+SYNDESIS_TEMPLATE_URL="${WORKSPACE}/app/deploy/${SYNDESIS_TEMPLATE_TYPE}.yml"
```
rhuss: You could export WORKSPACE to $(appdir) in the system-test script and the use just "${WORKSPACE}/deploy/..." here (without 'app'). but yeah, nit picking on high level.

app/tests/src/test/resources/setup.sh
rhuss: Yes, that's my point: No one used the non-restricted templates, never seen this in the wild. So happy to rip it, too, but this should go probably into a different PR.

PR #1471, opened 2018-02-07
chore: Removed --pass-user-bearer-token for now

rhuss:
so that we can still install on clusters with restricted permissions.

### DIFF
--- a/app/deploy/generator/04-syndesis-oauth-proxy.yml.mustache
+++ b/app/deploy/generator/04-syndesis-oauth-proxy.yml.mustache
@@ -81,10 +81,9 @@
             - --skip-auth-preflight
             - --openshift-ca=/etc/pki/tls/certs/ca-bundle.crt
             - --openshift-ca=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
-            - --pass-user-bearer-token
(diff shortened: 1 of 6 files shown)


paoloantinori:
@rhuss  with this PR you broke the possibility to invoke from command line entirely, even when setting duplicated headers.

do you really need to remove these line:

```
- --openshift-delegate-urls={"/api":{"resource":"projects","verb":"get","resourceName":"${OPENSHIFT_PROJECT}","namespace":"${OPENSHIFT_PROJECT}"}}
```

?

zregvart:
@paoloantinori we need to find a way for oauth-proxy to work without requiring the cluster role binding, i.e. cluster admin role for the user that deploys Syndesis. For local development you could patch the oauth-proxy deployment.

paoloantinori:
ok. that line doesn't prevent you from deploying without cluster role binding, or does it?

zregvart:
@paoloantinori I think it does, IIUC, it sets `paths` so this block is executed to test for that role:

https://github.com/openshift/oauth-proxy/blob/a79a0c8bd0adbab06f85f63595864ab421fd7a6c/providers/openshift/provider.go#L299-L316

PR #1480, opened 2018-02-08
refactor(inspector): drop remote Atlasmap suppo...

zregvart:
...rt and look into data shape specification instead

With the removal of connector/connector dependencies from Atlasmap
remote inspection doesn't make sense -- Atlasmap doesn't have the
classes needed on the classpath to inspect.

This drops support for remote Atlasmap inspection and instead uses the
data shape `specification`, where Syndesis Connector Maven plugin placed
the _full_ Java inspection for the data shapes class, by implementing
another `Inspector`: `SpecificationClassInspector`.

(cherry picked from commit b0636367acffb5a9ea5ec47b3d04d4a4e538e361)

```
# Conflicts:
#	app/pom.xml
#	app/rest/inspector/src/main/java/io/syndesis/inspector/DataMapperBaseInspector.java
#	app/rest/inspector/src/main/java/io/syndesis/inspector/DataMapperClassInspector.java
#	app/rest/inspector/src/main/java/io/syndesis/inspector/StaticResourceClassInspector.java
#	app/rest/inspector/src/test/java/io/syndesis/inspector/DataMapperClassInspectorTest.java
```


zregvart:
Seems that we have issues with Codacy reviewing the state of this branch. This could be either due to our configuration (which we fixed on `master`) or due to an issue with Codacy.

app/rest/inspector/src/main/java/io/syndesis/inspector/DataMapperBaseInspector.java
```diff
@@ -45,77 +45,115 @@
...
-                                    paths.add(prependPrefix(prefix, name));
-                                    continue;
-                                }
-
-                                paths.addAll(getPathsForJavaClassName(prependPrefix(prefix, name), fieldClassName, visited));
-                            }
-                        }
+            final JsonNode node = Json.mapper().reader().readTree(specification);
+            if (node == null) {
+                return Collections.emptyList();
```
rhuss: thanks, much better than deeply nested loops ..

zregvart:
> LGTM, but a question: How does this work with the 'old' style connectors (e.g. salesforce) ? Do they also have the necessary meta data included ?

With Database connectors specification in data shape is JSON Schema based, and the existing Inspector implementation works with that. For Salesforce most data shape specifications are JSON Schema based, there are a couple that are Java class based, and I think those should be fine. I will double check and report here in 2-10 minutes.

zregvart:
@rhuss tested manually two scenarios in three combinations.

Scenarios:
 1. from Database to Salesforce
 2. from Salesforce to Database

Combinations:
 1. without mapping step, so filter in between Database and Salesforce
 1. with mapping step, filter before mapping step
 1. with mapping step, filter after mapping step

Seems OK.

PR #1482, opened 2018-02-08
chore: perform open heart surgery on editor (and fix some issues)

### LINKED ISSUE #1387: Settings page, 'save' button doesn't come back out of the 'submitting' state
Once you copypasta your credentials and click 'Save' it stays in it's submitting state with the little spinny.  The network request completes with an HTTP 204 but the page doesn't pick up the response.

### LINKED ISSUE #1450: No build running after creating an integration
Actually if you try to create an integration and publish it, no build is triggered.

gashcrumb:
fixes #1450 
fixes #1387


app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.html
```diff
@@ -3,7 +3,7 @@
 
   <ng-template #tooltipPopover>
     <span>
-      <strong> {{ getStepText() }} </strong>
+      <strong [innerHtml]="getStepText()"></strong>
```
deeleman: is there any way we can assign this value at `OnInit()` or at leats make it observable somehow?  The issue here is that on each pass of the change detector loop we're executing [this script block](https://github.com/gashcrumb/syndesis/blob/99cc1a514695b621062a65dee6df357acba88706/app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.ts#L273-L298), which seems a bit demanding to me.

app/ui/src/app/integration/edit-page/flow-view/flow-view.component.ts
```diff
@@ -81,39 +84,39 @@ export class FlowViewComponent extends ChildAwarePage
...
   }
 
   lastPosition() {
-    return this.currentFlow.getLastPosition();
+    return this.currentFlowService.getLastPosition();
   }
 
   getMiddleSteps() {
-    return this.currentFlow.getMiddleSteps();
+    return this.currentFlowService.getMiddleSteps();
```
deeleman: We can leave these getters as they are ATM, but in the future we should get rid of getter members that execute functions (service methods here) internally, since those will be evaluated as often as template bindings.

A good way to address this is to refactor all those service methods to return Observable streams, so we can bind assign them to observable properties in the class and bind these directly in the template along with the `async` pipe.

app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.html
gashcrumb: Yep, was planning on tackling that kind of thing while implementing the changes for data buckets, this one is a first-pass untangling :-)

app/ui/src/app/integration/edit-page/flow-view/flow-view.component.ts
gashcrumb: Good call!  I'll plan on that for future work...

PR #1494, opened 2018-02-09
Enhancing the jsondb GetOptions so filter results based on a complex query filters against indexed properties.

chirino:


chirino:
I also want to make sure that when your using a comparison operator across different data types we honor the order defined at: https://firebase.google.com/docs/database/rest/retrieve-data#section-rest-ordered-data

app/rest/jsondb/src/test/java/io/syndesis/jsondb/impl/JsonDBTest.java
```diff
@@ -522,19 +529,19 @@ public void testSet() throws IOException {
...
+//        HashMap<String, Object> user = map(
+//            "data", new Object[]{"hi", 100, "other"}
+//        );
+//        jsondb.set("/", mapper.writeValueAsString(user));
+//        String result6 = jsondb.getAsString("", prettyPrint).trim();
+//        assertThat(result6).isEqualTo(load("result6.json").trim());
+//
+//        jsondb.set("/data/1", mapper.writeValueAsString("update"));
+//        String result7 = jsondb.getAsString("", prettyPrint).trim();
+//        assertThat(result7).isEqualTo(load("result7.json").trim());
```
zregvart: can this be removed?

app/rest/jsondb/src/test/java/io/syndesis/jsondb/impl/JsonDBTest.java
```diff
@@ -587,6 +594,108 @@ public void shouldAscertainPropertyPairExistence() {
...
+            "name", "u3",
+            "age", 21
+        )));
+
+        ////////////////////////////////////////
+        //
+        // Check filtering ops against strings..
+        //
+        ////////////////////////////////////////
+        assertThat(jsondb.getAsString("/users", new GetOptions().filter(
```
zregvart: might be a good idea to use [JSONAssert](http://jsonassert.skyscreamer.org/)

app/rest/jsondb/src/main/java/io/syndesis/jsondb/impl/JsonRecord.java
```diff
@@ -22,13 +22,13 @@
 
     private final String path;
     private final String value;
-    private final int kind;
+    private final String ovalue;
```
zregvart: what do we gain by replacing int with String? seems to me like that would have negative performance

chirino: Type was removed.  odata is a new field used for something different

app/rest/jsondb/src/test/java/io/syndesis/jsondb/impl/JsonDBTest.java
chirino: I need to uncomment it.

app/rest/jsondb/src/test/java/io/syndesis/jsondb/impl/JsonDBTest.java
chirino: I would reduce test coverage.  We want to verify the order of things that jsondb returns.

PR #1496, opened 2018-02-09
chore(ui): import RxJS like a pro

zregvart:
The details page of custom API connector wasn't loading as the
`combineLatest` RxJS operator wasn't imported.

This follows the best practice in the blog post by Christian Liebel[1]
on how to properly import RxJS by only loading parts that are needed
from a centrally managed imports file.

This adds `rxjs.imports.ts` that should hold all RxJS imports needed and
adds a lint rule from `rxjs-tslint-rules` to enforce its use.

[1] https://christianliebel.com/2017/07/import-rxjs-correctly/

### DIFF
--- a/app/ui/src/polyfills.ts
+++ b/app/ui/src/polyfills.ts
@@ -56,17 +56,24 @@ import 'core-js/es7/reflect';
 import 'zone.js/dist/zone';  // Included with Angular CLI.
 
 /***************************************************************************************************
- * APPLICATION IMPORTS
+ * RxJS IMPORTS
  */
-
-import 'rxjs/add/operator/do';
 import 'rxjs/add/observable/empty';
-import 'rxjs/add/observable/from';
 import 'rxjs/add/observable/merge';
 import 'rxjs/add/observable/of';
+import 'rxjs/add/observable/throw';
+
+import 'rxjs/add/operator/catch';
+import 'rxjs/add/operator/combineLatest';
+import 'rxjs/add/operator/do';
 import 'rxjs/add/operator/filter';
 import 'rxjs/add/operator/first';
 import 'rxjs/add/operator/map';
 import 'rxjs/add/operator/mergeMap';
+import 'rxjs/add/operator/share';
 import 'rxjs/add/operator/switchMap';
 import 'rxjs/add/operator/take';
+
+/***************************************************************************************************
+ * APPLICATION IMPORTS
+ */
--- a/app/ui/tslint.json
+++ b/app/ui/tslint.json
@@ -2,6 +2,9 @@
   "rulesDirectory": [
     "node_modules/codelyzer"
   ],
+  "extends": [
+    "rxjs-tslint-rules"
+  ],
   "rules": {
     // ***************************************************
     // TypeScript linting Ruleset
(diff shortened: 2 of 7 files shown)


zregvart:
The two test fail as Observable doesn't have `empty` and `merge` methods, what I don't understand is that imports are present for both of them in [src/rxjs.imports.ts#L1-L2](https://github.com/zregvart/syndesis/blob/401d7e4300c7e43c8e7440460ab7c7192def5767/app/ui/src/rxjs.imports.ts#L1-L2).

Can one of the @syndesisio/ui-api help me with this one?

app/ui/tslint.json
```diff
@@ -207,19 +213,39 @@
...
+          "allowElsewhere": false,
+          "allowUnused": false,
+          "file": "rxjs.imports.ts"
+        }
+      ],
+      "severity": "error"
+    }
   }
-}
+}
```
deeleman: Why the overall reformatting of the JSON file? splitting one or two-items arrays into 4 LOC makes the whole thing more hardly readable indeed

app/ui/src/main.ts
```diff
@@ -2,6 +2,7 @@ import { enableProdMode } from '@angular/core';
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
 import { environment } from './environments/environment';
 import { AppModule } from './app';
+import './rxjs.imports';
```
deeleman: Ideally, this should come BEFORE the `import { AppModule } ...` LOC, since that module might include at some point async code relying on RxJS operators (ideally not, but yeah... this is Syndesis and anything can happen here)

app/ui/src/polyfills.ts
```diff
@@ -58,15 +58,3 @@ import 'zone.js/dist/zone';  // Included with Angular CLI.
...
-import 'rxjs/add/observable/empty';
-import 'rxjs/add/observable/from';
-import 'rxjs/add/observable/merge';
-import 'rxjs/add/observable/of';
-import 'rxjs/add/operator/filter';
-import 'rxjs/add/operator/first';
-import 'rxjs/add/operator/map';
-import 'rxjs/add/operator/mergeMap';
-import 'rxjs/add/operator/switchMap';
-import 'rxjs/add/operator/take';
```
deeleman: This is the only part that strikes me a bit. In fairness the location of the imports does not matter that much as long as those are imported upfront, prior to any bits and pieces of the actual project. However, I would have strived to keep these imports at `polyfills.ts` in order not to clutter more an already cluttered workspace such as `app/ui/src`. Although these imports are not strictly _polyfills_, they address the lack of an actual native functional method by extending the `Observable` prototype, so those can be considered as such.

Ultimately this is a matter of preference and the whole point is to keep the amount of files we have at the top in check.

app/ui/src/main.ts
deeleman: Actually, I'd suggest to bump this LOC to the very top of the file.

app/ui/tslint.json
zregvart: oh, I think my editor did that, I can reset those parts...

app/ui/src/polyfills.ts
zregvart: Yeah, let's see what happens if I put them in `polyfills.ts` should work in theory.

PR #1498, opened 2018-02-09
feature: First draft of an update concept

rhuss:


chirino:
@rhuss I got basic working system for DB migration in.  See: 

https://github.com/syndesisio/syndesis/blob/master/app/rest/runtime/src/main/java/io/syndesis/runtime/Migrations.java#L76

The migration script should be simple to implement.  See example:
https://github.com/syndesisio/syndesis/blob/master/app/rest/runtime/src/test/resources/test-migrations/up-1.js#L19

gashcrumb:
@chirino think we can merge this one?

Issue #1394, opened 2018-02-03
Add Action with input/output DataShape for datamapper step


igarashitm:
https://github.com/syndesisio/syndesis/issues/1132#issuecomment-362746451

Currently datamapper step doesn't have `Step.action` in UI, but `Step.action.descriptor.{input,output}DataShape` is required for Data Buckets (#1110)
 and data shape mismatch indicator (#700). Add data shapes into datamapper step.

PR #1500, opened 2018-02-09
chore: Add Action with input/output DataShape for datamapper step

igarashitm:
Fixes: #1394

### DIFF
--- a/app/ui/src/app/integration/edit-page/current-flow.service.ts
+++ b/app/ui/src/app/integration/edit-page/current-flow.service.ts
@@ -13,7 +13,11 @@ import { Action,
   key,
   IntegrationSupportService } from '@syndesis/ui/platform';
 import { log, getCategory } from '@syndesis/ui/logging';
-import { IntegrationStore } from '@syndesis/ui/store';
+import { IntegrationStore,
+  ENDPOINT,
+  DATA_MAPPER,
+  EXTENSION,
+  BASIC_FILTER } from '@syndesis/ui/store';
 import { FlowEvent } from '@syndesis/ui/integration/edit-page';
 
 const category = getCategory('CurrentFlow');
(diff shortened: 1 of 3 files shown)


igarashitm:
Still debugging though, submitting a PR with wip to start a thread to chase ExpressionChangedAfterItHasBeenCheckedError which is thrown when it opens datamapper.

```
IntegrationStepConfigureComponent.html:9 ERROR Error: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'ngIf: undefined'. Current value: 'ngIf: [object Object]'.
    at viewDebugError (core.js:9758)
    at expressionChangedAfterItHasBeenCheckedError (core.js:9736)
    at checkBindingNoChanges (core.js:9903)
    at checkNoChangesNodeInline (core.js:13951)
    at checkNoChangesNode (core.js:13925)
    at debugCheckNoChangesNode (core.js:14754)
    at debugCheckDirectivesFn (core.js:14656)
    at Object.eval [as updateDirectives] (IntegrationStepConfigureComponent.html:9)
    at Object.debugUpdateDirectives [as updateDirectives] (core.js:14638)
    at checkNoChangesView (core.js:13763)
```

This comes from here
https://github.com/syndesisio/syndesis/blob/da5a992680772215007f07af4baca03c4955ba71/app/ui/src/app/integration/edit-page/step-configure/step-configure.component.html#L9

And I guess this event is causing that property to be updated asynchronously
https://github.com/syndesisio/syndesis/blob/da5a992680772215007f07af4baca03c4955ba71/app/ui/src/app/integration/edit-page/step-configure/data-mapper/data-mapper-host.component.ts#L212-L223

@deeleman so we need `setTimeout()` thing when it's retrieving that `step.configuredProperties`?

KurtStam:
@igarashitm you will need to do a rebase to obtain the fix for the broken build.

zregvart:
> @zregvart as far as datamapper step holds data shapes, we can only look at the previous step, is that right?

yes, this is much cleaner, thank you.

Issue #1246, opened 2018-01-25
Editing an integration name doesn't follow the PatternFly inline edit pattern


michael-coker:
To reproduce, go to "Integrations," click on the kebab menu for an integration and choose "Edit." In the top/left corner, you can change the integration name by clicking the name or the pencil beside it.

Only the pencil should trigger the inline edit, and when that text is in edit mode, it should change to an input or text area with the appropriate cursor. You may also want to introduce a Save and Cancel button.

An example that utilizes the inline edit pattern properly is editing the name or description of a connection.

PR #1503, opened 2018-02-09
update editing the integration name to match PF pattern

michael-coker:
From issue https://github.com/syndesisio/syndesis/issues/1246, editing the integration name doesn't match PF's inline edit interaction. Attempting to bring it closer to that pattern.

### DIFF
--- a/app/ui/src/app/integration/edit-page/flow-view/flow-view.component.scss
+++ b/app/ui/src/app/integration/edit-page/flow-view/flow-view.component.scss
@@ -1,4 +1,4 @@
-@import 'patternfly-sass/assets/stylesheets/patternfly/color-variables';
+@import 'syndesis-sass';
 
 .collapsed {
   .step-insert {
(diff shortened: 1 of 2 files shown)


app/ui/src/app/integration/edit-page/flow-view/flow-view.component.scss
```diff
@@ -110,32 +123,40 @@
...
     &.finish {
       flex: 0 0 auto;
       min-height: 0px;
     }
+
     &.step-insert {
       flex: 0 0 auto;
       min-height: 80px;
+
       /deep/ .popover {
```
deeleman: Now that you're editing this file, would you mind replacing this `/deep/` Shadow Piercing combinator by the LTS one (`::ng-deep`)? Thanks.

Whenever you see something that you can improve while you're introducing some other changes in that file, do not hesitate and proceed. The idea is to leave the code in a better state that we found it.

michael-coker: Done, also now importing syndesis-sass instead of patternfly-sass.

gashcrumb:
Looks like the changes were implemented, merged...  thanks!

Issue #1481, opened 2018-02-08
DataMapper does not come up when creating an integration


KurtStam:
Javascript console says
main.3d82570a130b35eb466f.bundle.js:1 ERROR Error: StaticInjectorError[t -> t]: 
  StaticInjectorError(Platform: core)[t -> t]: 
    NullInjectorError: No provider for t!
    at t.get (main.3d82570a130b35eb466f.bundle.js:1)
    at main.3d82570a130b35eb466f.bundle.js:1
    at t (main.3d82570a130b35eb466f.bundle.js:1)
    at t.get (main.3d82570a130b35eb466f.bundle.js:1)
    at main.3d82570a130b35eb466f.bundle.js:1
    at t (main.3d82570a130b35eb466f.bundle.js:1)
    at t.get (main.3d82570a130b35eb466f.bundle.js:1)
    at Ar (main.3d82570a130b35eb466f.bundle.js:1)
    at t.get (main.3d82570a130b35eb466f.bundle.js:1)
    at Ar (main.3d82570a130b35eb466f.bundle.js:1)

KurtStam:
Thanks - it's working now

PR #1504, opened 2018-02-09
fix(editor): update event bindings and force change detection

seanforyou23:
This PR forces a round of change detection on the after view init lifecycle hook. This is not ideal but does unbreak the UI. It's a sign of a larger design flaw but that could be addressed at a later point.

@deeleman setting these values in an async fashion didn't work, unfortunately, so I opted for ChangeDetectorRef. Let me know if there's a better way for the short term.

Should help https://github.com/syndesisio/syndesis/issues/1481

update missed event bindings from ng-dynamic-form upgrade
force change detection after view init for editor

### DIFF
--- a/app/ui/src/app/integration/edit-page/step-configure/step-configure.component.ts
+++ b/app/ui/src/app/integration/edit-page/step-configure/step-configure.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnInit, OnDestroy } from '@angular/core';
+import { Component, Input, OnInit, OnDestroy, AfterViewInit, ChangeDetectorRef } from '@angular/core';
 import { Observable } from 'rxjs/Observable';
 import { Subscription } from 'rxjs/Subscription';
 import { ActivatedRoute, ParamMap, Router } from '@angular/router';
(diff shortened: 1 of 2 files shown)


deeleman:
I'm still missing the point for this change and how it relates to the reported error, which basically informs of a dependency missing when the injector attempts to fetch it. Perhaps the solution based on `forwardRef` proposed by Stan would be more appropriate here.

Nevertheless if you believe this is a feasible way to get away with the reported exception, then let's leave this as it is. As you rightly said, this is the by-product of a design error in Syndesis and requires a more broader solution... Broader as life, alas...

app/ui/src/app/integration/edit-page/step-configure/step-configure.component.ts
```diff
@@ -33,6 +33,7 @@ export class IntegrationStepConfigureComponent implements OnInit, OnDestroy {
   routeSubscription: Subscription;
 
   constructor(
+    private cd: ChangeDetectorRef,
```
deeleman: Unless we eventually get away with bloating the application with esoteric class members and variables such as `dc`, `x`, `res` and the like, let's stick to the principle of naming variables and class properties with names matching their type.

Unfortunately, this application does not honor the Single Responsibility Principle, so it's not rare to find class structures with literally 600 LOC or even more. Creating names that self-describe the types in use will spare us from having to inspect entities to infer their type and purpose hence gaining a better understanding of the code.

Issue #991, opened 2018-01-09
defaultValue property missing for checkbox


seanforyou23:
When making the rest call to get connection info for building the form in "configure publish messages" for an ActiveMQ connection, I noticed the defaultValue property isn't present and so the checkbox isn't checked by default as it should be to match the UX - https://github.com/syndesisio/syndesis/blob/master/ux/designs/amq/amq.md

It is however specified in the deployment.json at `app/rest/dao/src/main/resources/io/syndesis/dao/deployment.json`

Let's find out where this property gets dropped and ensure it makes its way to the forms service so it can be used to set the default value of a checkbox.

seanforyou23:
@dhirajsb here's an example of the request that I believe should be holding this property

`https://syndesis.x.x.x.x.nip.io/api/v1/connections/-L2bHvD5E3eIwNfmRLR0/actions/io.syndesis:activemq-publish-connector:latest`

PR #1532, opened 2018-02-12
fix(checkbox): allow string "true" to set checked state by default

seanforyou23:
This PR allows the value string 'true' to set the checkboxes default checked state through our form factory service provider.

[This issue](https://github.com/syndesisio/syndesis/issues/991) was put on hold while many UI upgrades were in the works, now that this has settled we'd like to resume.

Notice in the attached image, the defaultValue property has the value string "true", instead of boolean true. This is the cause of the default value not being set correctly.

screen shot 2018-02-12 at 12 01 09 pm

Ideally, this PR can be dismissed and the fix be applied to the back end, so we don't have to perform this type of check in the UI - it really shouldn't be there. The value is set properly (boolean true) in deployment.json, so something seems to be changing this value before its exposed via rest service.

No idea what's involved in making this fix on the back end, which is why I'm posting this PR - in the interest of time.

### DIFF
--- a/app/ui/src/app/core/providers/form-factory-provider.service.ts
+++ b/app/ui/src/app/core/providers/form-factory-provider.service.ts
@@ -217,13 +217,12 @@ export class FormFactoryProviderService extends FormFactoryService {
                          field: ConfiguredConfigurationProperty,
                          value: any
   ) {
-    // bandaid for rest service returning string 'true' instead of boolean true
-    let defaultValue = value || field.value || field.defaultValue;
+    const initialValue = value || field.value || field.defaultValue;
     return new DynamicCheckboxModel({
       id: key,
       label: field.displayName || key,
       hint: field.description,
-      value: (defaultValue === 'true') ? true : defaultValue
+      value: (!!initialValue)
     }, {
         element: {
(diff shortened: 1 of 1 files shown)


gashcrumb:
+1 the form service should handle `truthy` values, angular does.  BTW you've a lint error:

```
[INFO] /workspace/app/ui/src/app/core/providers/form-factory-provider.service.ts
[INFO] ERROR: 221:9   prefer-const  Identifier 'defaultValue' is never reassigned; use 'const' instead of 'let'.
```

app/ui/src/app/core/providers/form-factory-provider.service.ts
```diff
@@ -217,11 +217,13 @@ export class FormFactoryProviderService extends FormFactoryService {
                          field: ConfiguredConfigurationProperty,
                          value: any
   ) {
+    // bandaid for rest service returning string 'true' instead of boolean true
+    let defaultValue = value || field.value || field.defaultValue;
```
gashcrumb: `defaultValue` is probably a misleading name here?

dhirajsb:
Looks like this should support both "true" and true, if thats true, 👍

PR #1533, opened 2018-02-12
fix: filter connections by position in flow for connection select pages

### LINKED ISSUE #1525: User is stuck if twitter without "To" actions is picked at a "To" position
If you start from database and then pick Twitter as destination, you'll be stuck and the only way to escape is to start a new integration.

screenshot-2018-2-12 syndesis 1

I think we should not allow selection of connections at a place that the action does not fit, i.e. `From` at start and `To` in other places.

Related to #253

gashcrumb:
fixes #1525


gashcrumb:
@zregvart this fix does still allow Twitter to show up when adding a connection in the middle.  Should that not be allowed, i.e. do `From` actions only belong at the start?

gashcrumb:
@zregvart ignore that question, I realized the answer is yes, so twitter can't be selected in a middle position.

PR #1539, opened 2018-02-13
fix: Spelling and grammatical errors

### LINKED ISSUE #1519: Spelling and grammatical errors
Noticed some inconsistencies in punctuation, grammar, case, etc. 

@TovaCohen would you mind reviewing these? Some of these are in the doc I shared with you a while back and some are new.

> /integrations/create/action-select/0 (twitter)
* "Search for tweets that mention you" - needs period
* "Search tweets based one or more keywords" - needs period

> /integrations/create/action-select/0 (database)
* "Invoke SQL to periodically get data out"
* "Periodically Invoke a Stored Procedure to get Data out" - uncapitalize "Invoke" and "Data" and add period

> /integrations/create/action-configure/0/0 (database - periodic SQL invocation)
* "Enter a SQL Statement starting with SELECT" - needs period

> /integrations/create/action-configure/0/0 (database - periodic stored prodecure)
* "Select the Stored Procedure" - needs period

> /integrations/create/action-select/0 (salesforce)
* All of the descriptions say "when an Salesforce record is..." - change "an" to "a"

> /integrations/create/action-configure/0/0 (salesforce)
* "Specify the Salesforce object name for which to receive notification when records are created"

> /integrations/create/action-select/1 (salesforce - create or update record)
* All of the descriptions after clicking "Create or update record," "Delete record," etc. need a period.
* "Account, Contact or Order" - needs comma

> /integrations/create/action-configure/1/0 (salesforce)
* "Specify the Salesforce object name of the record to create" - needs period

> /integrations/create/step-select/1
* "Map fields from the input type to the output type" - needs period

> /customizations/api-connector/create/swagger-connector
* "Custom API's are RESTful APIs" - remove single quote
* "as long as a well-documented swagger is available and conforms to OpenAPI standard." - say "to OpenAPI standards" or "to the OpenAPI standards"

> /settings/oauth-clients
* is the "OAuth Client Management" tab necessary since there are no other tabs?
  * If so, should that text match the <h1> on the page, or vice versa?
* When you save settings:
  * "You can validate these values when creating a connection Salesforce" - wording and needs period
  * "You can validate these values when creating a connection Twitter" - wording and needs period

> /customizations/extensions (empty state)
* "There are no extensions available" - needs period

> /integrations
* When you publish or unpublish an integration from the list, the toast notification needs punctuation between the bold and normal text.
* "Integration is activating Please allow a moment for the integration to fully activate."
* "Integration is deactivating Please allow a moment for the integration to be unpublished."

zregvart:
I think I managed to get all of the text from #1519 updated, @michael-coker and @TovaCohen can you take a look?

Fixes #1519

### DIFF
--- a/app/connectors/connectors/sql/src/main/resources/META-INF/syndesis/connector/sql.json
+++ b/app/connectors/connectors/sql/src/main/resources/META-INF/syndesis/connector/sql.json
@@ -1,7 +1,7 @@
 {
   "id": "sql",
   "name": "Database",
-  "description": "Invoke SQL or Stored Procedures",
+  "description": "Invoke SQL or stored procedures",
   "icon": "fa-database",
   "dependencies": [
     {
@@ -81,7 +81,7 @@
     {
       "id": "sql-connector",
       "name": "Invoke SQL",
-      "description": "Invoke SQL to get data in and/or out",
+      "description": "Invoke SQL to obtain, store, update, or delete data",
       "tags": [
         "dynamic"
       ],
@@ -100,19 +100,19 @@
         },
         "propertyDefinitionSteps": [
           {
-            "name": "SQL Statement",
-            "description": "Enter a SQL Statement starting with INSERT, SELECT, UPDATE or DELETE",
+            "name": "SQL statement",
+            "description": "Enter a SQL statement that starts with INSERT, SELECT, UPDATE or DELETE.",
             "properties": {
               "query": {
                 "kind": "path",
-                "displayName": "SQL Statement",
+                "displayName": "SQL statement",
                 "group": "common",
                 "required": true,
                 "type": "string",
                 "javaType": "java.lang.String",
                 "deprecated": false,
                 "secret": false,
-                "description": "SQL Statement to be executed. Can contain input parameters prefixed by ':#' i.e. ':#MYPARAMNAME'"
+                "description": "SQL statement to be executed. Can contain input parameters prefixed by ':#' i.e. ':#MYPARAMNAME'"
               }
             }
           }
@@ -121,8 +121,8 @@
     },
     {
       "id": "sql-start-connector",
-      "name": "Periodic SQL Invocation",
-      "description": "Invoke SQL to periodically get data out",
+      "name": "Periodic SQL invocation",
+      "description": "Periodically invoke SQL to obtain, store, update, or delete data.",
       "tags": [
         "dynamic"
       ],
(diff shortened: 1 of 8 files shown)


app/connectors/connectors/sql/src/main/resources/META-INF/syndesis/connector/sql.json
```diff
@@ -196,21 +196,21 @@
...
+                "displayName": "Procedure name",
                 "required": false,
                 "type": "string",
                 "javaType": "java.lang.String",
                 "tags": [],
                 "deprecated": false,
                 "secret": true,
                 "componentProperty": true,
-                "description": "Name of the Store Procedure"
+                "description": "Name of the store procedure"
```
TovaCohen: Should that be "Name of the stored procedure" ? I not certain of the context.
I know this was approved and merged. Sorry if it is no longer good to comment here.

app/connectors/connectors/sql/src/main/resources/META-INF/syndesis/connector/sql.json
```diff
@@ -267,7 +267,7 @@
                 "deprecated": false,
                 "secret": false,
                 "componentProperty": true,
-                "description": "Name of the Store Procedure"
+                "description": "Name of the store procedure"
```
TovaCohen: Same here -- should that be "Name of the stored procedure" ?

TovaCohen:
@zregvart Thanks so much for making all these updates and for updating text we didn't comment on to be consistent throughout. I think you were very thorough!  
So what happens when this has been approved and merged, but something was not caught and it should be changed? Is it okay to comment after the merge?
Thanks again!

app/connectors/connectors/sql/src/main/resources/META-INF/syndesis/connector/sql.json
michael-coker: Should these lines end in a period?
* Invoke SQL or stored procedures
* Invoke SQL to obtain, store, update, or delete data
* SQL SELECT statement to be executed
* Name of the store procedure
  * Per Tova's comment, "store" should be "stored"

And I might recommdend changing `i.e. ':#MYPARAMNAME'` to `(i.e., ':#MYPARAMNAME')` - parenthesis optional, but "i.e." should be followed by a comma.

michael-coker:
@zregvart thanks for making these updates! Sorry it took so long to review. I left a couple of comments for you.

app/connectors/connectors/sql/src/main/resources/META-INF/syndesis/connector/sql.json
TovaCohen: If it is a complete sentence, then there should be a period at the end. The imperative sentences above, the ones that start with "Invoke" should have periods. Are these tooltips?  The second two bullet items above are not complete sentences, so no period at the end. But, if these are all tooltips on the same page, then they should all have periods at the end. The rule that applies for that is:  In a list of things, if one list item is a complete sentence, then each list item should have a period at the end. I hope that makes sense. 
And I'll just say here that if we could implement the process for extracting tooltips from source files for user doc, then I think we would all save a lot of time. There's an issue for that somewhere. 
Finally, I don't know the context for your last comment, but "i.e." should not appear in anything the user sees. It's Latin and not English and problematic for translation. In this particular instance, you want either "for example," or "that is," I can't tell which would be correct.

app/connectors/connectors/sql/src/main/resources/META-INF/syndesis/connector/sql.json
TovaCohen: I don't think so.

Issue #1229, opened 2018-01-25
Add UI page to describe data shape for shapeless connectors 


dhirajsb:
UXD has designed an extra page for shapeless connectors in #1194. Shapeless connectors are identified as ones with input/output shape or both of the form:
```
            "inputDataShape": {
              "type": "*"
            },
            "outputDataShape": {
              "type": "*"
            }
```
UI needs to add page to allow user to specify input and output shapes from json/xml schema or instance documents.

gashcrumb:
Question, why can't this be tacked on to the [existing plumbing](https://github.com/gashcrumb/syndesis/blob/master/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts#L110) we have in place for multi-page configuration for connections?  The descriptor returned from that call also has the input/output data shapes, we can update this component to ensure those are set appropriately.

gashcrumb:
@sjcox-rh @dongniwang @dhirajsb Question on the design, how does the user indicate the schema/instance document should be set for the input or output data shape?

gashcrumb:
Ah, silly me, it's this one -> https://github.com/syndesisio/syndesis/blob/master/ux/designs/describe-data/describe-data.md

So my question still stands on this one...

zregvart:
@gashcrumb I don't think this would work by adding an action property, if we do that we would just add a property, the data shape of the action would remain as `*`. This would then require us to configure the mapper by looking a the data shapes and at the configured action properties which feels wrong to me.

gashcrumb:
I wonder though then, how exactly is the UI going to set these data shapes on a step?  Currently it's all driven off of the action descriptor.  If we add an additional page *after* the data shape input that's effectively a no-op the UI then has a chance to post the data shape properties, which could then replace what's set on the action descriptor, then I guess we'd have to set that on the step.

But then, what's the other idea of how this is going to be implemented, are there endpoints for the UI to invoke currently to post these data shapes?

zregvart:
The big dirty secret is that the UI can mutate any part of the integration and the back-end will store it as is, the back-end stores the JSON as received. So mutating the action descriptor's shape should be possible.
We can certainly think about adding endpoints to make this _proper_. I'm adding a discussion label and inviting @syndesisio/backend for comments on this.

gashcrumb:
Ah, of course, I see.  So for the file input we'll just read it client-side and replace the action's data shape property locally.  I think that'll work fine.

[...]

Issue #1551, opened 2018-02-13
Use 'kind' for the data shape type consistently, and pick either '*' or 'any'


gashcrumb:
Currently I see in an integration I've created:

```
"inputDataShape": {
            "kind": "none"
          },
          "outputDataShape": {
            "name": "some name",
            "type": "any",
            "specification": "{}"
          },
```

for the start connection (I added the stuff in the output), I added a JMS request response as a middle step, it's data shapes look like:

```
          "inputDataShape": {
            "type": "any"
          },
          "outputDataShape": {
            "type": "*"
          },
```

Can we settle on either `kind` or `type` and also settle on if it should be `any` or `*`?

igarashitm:
datamapper expects `DataShape.kind` is one of 
* none
* any
* java
* json-schema
* json-instance
* xml-schema
* xml-instance

`DataShape.type` is a special property for `DataShape.kind === 'java'`, which holds a class name. Other than that it's not going to be used anywhere.

igarashitm:
I think it's better to introduce `DataShapeKind` enum in both of backend and UI, then make `DataShape.kind` being that enum.

PR #1553, opened 2018-02-13
feat: add describe data page to editor page flow

gashcrumb:
for #1229


igarashitm:
w/r/t `DataShape.kind` (#1551), we have `Select Schema Type` pull down. This is going to be `DataShape.kind` directly, which is one of
* json-schema
* json-instance
* xml-schema
* xml-instance

https://github.com/syndesisio/syndesis/blob/cb76ff04a0a11dd7e3bf1394fcfea4ded147d50c/ux/designs/describe-data/img/describe-data-define-dataTypeNameAndDescription.png

Then shapeless connection (`DataShape.kind`=`any`) turns into shaped one.

gashcrumb:
@igarashitm maybe we should merge this PR and your PR with the tooltip indicators?

PR #1555, opened 2018-02-14
Support IntegrationDeploymentMetrics in Prometheus metrics provider

dhirajsb:


app/rest/metrics-prometheus/src/main/java/io/syndesis/rest/metrics/prometheus/PrometheusMetricsProviderImpl.java
```diff
@@ -15,60 +15,102 @@
...
+            final Long messages = entry.getValue();
+
+            final Long errors = failedMessagesMap.get(version);
+            final Date start = startTimeMap.get(version);
+            final Date lastProcessed = lastProcessingTimeMap.get(version);
+
+            // aggregate values while we are at it
+            totalMessages[0] = SUM_LONGS.apply(totalMessages[0], messages);
+            totalErrors[0] = SUM_LONGS.apply(totalErrors[0], errors);
+            startTime[0] = MAX_DATE.apply(startTime[0], start);
```
KurtStam: Does that really show the date of how long the integration has been up?

dhirajsb: @KurtStam it shows the latest start date, isn't that what we agreed upon?

KurtStam: @dhirajsb was just wondering if we need to nullify that when the pod goes down (upon a new deployment version)

dhirajsb: @KurtStam I think it's still valid to show when a particular version had last started. If we want to show integration pod status, like Prometheus we should use another metric e.g. `up` that shows whether a monitored resource is active or not. 
That way, both the last start time and up/down status is captured for every deployment version. Otherwise there is no way to report start times for older versions, right?

KurtStam: @dhirajsb - I switched to 'starttime' so we can more easily cache it; uptime is something that changes every second.. This way we only don't need to write metrics if no messages flow through the system. Uptime implies it's running though imh.

dhirajsb: @KurtStam if the ui knows whether the current deployment is running, it can compute the 'uptime' for the integration from the start date. Is that not what it's doing? 
It's not a good idea to use Date to hold a duration anyway, there is a `java.time.Duration` type for that and it's obvious what it's for.

PR #1563, opened 2018-02-14
feature: Metrics on dashboard

deeleman:
captura de pantalla 2018-02-14 a las 12 33 37

This pulls in the 4 metrics boxes in the dashboard. Some of the data displayed is observable to ensure the dashboard is refreshed reactively. The Integration bits leverage the new Ngrx store the UI will transition to in the next days.

**Please note:** `DashboardMetricsComponent` leverages the new _lettable_ operators API, available in RxJS 5.5. We will want to transition to that new API on each and every Rx stream in the application, in order to maximize compression by ensuring a better tree-shaking optimization.

Also, models have been repurposed to generate better inheritance by composition. @gashcrumb please take a look should I have screwed something. This was necessary since models featured dozens of duplicated properties and the way they operate now do not allow for proper composition.


app/ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -55,15 +69,21 @@ describe('DashboardComponent', () => {
...
+          DashboardMetricsComponent,
           DashboardComponent,
-          EmptyStateComponent,
+          DashboardEmptyComponent,
           DashboardConnectionsComponent,
           DashboardIntegrationsComponent
         ],
         providers: [
           ConfigService,
+          // {
```
gashcrumb: I guess this can go?  As well as the const configServiceStub up above?

deeleman: Certainly Sir!

app/ui/src/app/platform/types/integration/integration.models.ts
```diff
@@ -23,24 +23,34 @@ export const ERROR = 'Error';
...
 export type IntegrationStatus = 'Pending' | 'Published' | 'Unpublished' | 'Error';
 
-export interface Integration extends BaseEntity {
+export interface IntegrationOverview extends BaseEntity {
+  version?: number;
+  tags: Array<string>;
   description?: string;
+  draft: boolean;
+  deployments?: Array<DeploymentOverview>;
+  currentStatus: IntegrationStatus;
```
gashcrumb: Isn't it `currentState` on the actual response back?  Just checked here and I see `currentState` and `targetState` in an integration overview object.

app/ui/src/app/platform/types/integration/integration.models.ts
```diff
@@ -23,24 +23,34 @@ export const ERROR = 'Error';
...
+  targetState: IntegrationStatus;
+  statusMessage?: string;
+  deploymentVersion?: number;
+}
+
+export type IntegrationOverviews = Array<IntegrationOverview>;
+
+export interface Integration extends IntegrationOverview {
```
gashcrumb: The only thing I'm concerned about here is that this isn't a real relationship.  And the `Integration` object no longer exposes any state/status fields.  The main impact I'm thinking about here is more for saving an integration which requires the whole object.  I guess though we could just trim that out as needed when it becomes a problem.

app/ui/src/app/platform/types/integration/integration.models.ts
```diff
@@ -68,23 +78,9 @@ export interface IntegrationDeployment extends BaseEntity {
...
-  statusMessage?: string;
-  deploymentVersion: number;
-  version: number;
-}
-export type IntegrationOverviews = Array<IntegrationOverview>;
-
 export interface DeploymentOverview extends BaseEntity {
   version: number;
-  currentState: IntegrationStatus;
+  currentStatus: IntegrationStatus;
```
gashcrumb: Same here, I think it's state vs status in the response object.

app/ui/src/app/platform/types/integration/integration.models.ts
```diff
@@ -58,7 +68,7 @@ export interface IntegrationDeployment extends BaseEntity {
   lastUpdated: number;
   integrationId: string;
   version: number;
-  currentState: IntegrationStatus;
+  currentStatus: IntegrationStatus;
```
gashcrumb: Here too, and feel free to ignore this if this is something that's going to change in the backend too...

deeleman: Oh damn! good point. I originally changed the property because I wanted it to match the type name (`IntegrationState` could have to be renamed - otherwise it clashed against the `xxxState` token), but forgot that this is likely coming from the backend. Will reinstate the original values again

app/ui/src/app/platform/types/integration/integration.models.ts
deeleman: You're right, I need to rename this again

deeleman: In any event, naming something like this as `currentState` was a very bad idea :( 

Same as moving tokens around mapped to strings, such as `DRAFT` or `PUBLISHED`. We do have enums in TypeScript for us and this forces the application consumer to be aware of each and every token ever created, which simply can't scale.

app/ui/src/app/platform/types/integration/integration.models.ts
deeleman: Done!

Issue #1307, opened 2018-01-30
Provide data shape name/description for builtin connectors


zregvart:
With #1245 we have two new properties on the data shapes, we should set values for them in all data shapes that we ship.

paoloantinori:
Entries to cover:
<details>
  <summary><b>Entries in Connectors, click to expand</b></summary>
<pre><code>
16:38 $ pwd && !!
pwd && ag -g '.*src.*json$' | xargs -n1 ag '(output|input)DataShape' -l | xargs -n1 -I {}  sh -c "echo {}; jq  ' ..|(.outputDataShape?,.inputDataShaper?)' {}" | ag -v null


/data/repositories/work/syndesis/syndesis/app/connectors/connectors


aws-s3-connector/src/main/resources/META-INF/syndesis/connector/aws-s3.json
{
  "kind": "none"
}
{
  "name": "S3Object",
  "description": "Represent one Amazon S3 Object",
  "kind": "java",
  "type": "com.amazonaws.services.s3.model.S3Object"
}
{
  "name": "S3Object",
  "description": "Represent one Amazon S3 Object",
  "kind": "java",
  "type": "com.amazonaws.services.s3.model.S3Object"
}
twitter-connector/src/main/resources/META-INF/syndesis/connector/twitter.json
{
  "name": "Status",
  "description": "Represent one single status of a user",
  "kind": "java",
  "type": "twitter4j.Status"
}
{
  "name": "Status",
  "description": "Represent one single status of a user",
  "kind": "java",
  "type": "twitter4j.Status"
}
jms/amqp-connector/src/main/resources/META-INF/syndesis/connector/amqp.json
{
  "kind": "none"
}
{
  "kind": "*"
}
{
  "kind": "*"
}
sql/src/main/resources/META-INF/syndesis/connector/sql.json
{
  "kind": "json-schema"
}
{
  "kind": "json-schema"
}
{
  "kind": "json-schema"
}
{
  "kind": "json-schema"
}
</code></pre>
</details>

<details>
  <summary><b>Entries in DAO, click to expand</b></summary>
<pre><code>
6:42 $ pwd && ag -g '.*src.*json$' | xargs -n1 ag '(output|input)DataShape' -l | xargs -n1 -I {}  sh -c "echo {}; jq  ' ..|(.outputDataShape?,.inputDataShaper?)' {}" | ag -v null

/data/repositories/work/syndesis/syndesis/app/rest/dao

[...]

PR #1568, opened 2018-02-14
fix: Use 'kind' for the data shape type consistently, and align typel…

igarashitm:
…ess data shapes to 'any'

Fixes: #1551
Also partially addresses #1307, added better data shape 'name' and 'description' for Twitter, SQL and Salesforce connectors.


igarashitm:
This is still failing in IT in my local, but due to OOM...

Issue #1225, opened 2018-01-24
Enable support for broker and client PEM certificates in AMQP Connector


dhirajsb:
Camel AMQP connector doesn't provide an easy way of configuring Broker and Client certificates for SSL `amqps` connections. 
A workaround is to use the following options in the underlying AMQP client and add them on the connection URL in the connector:
* transport.keyStoreLocation - default is to read from the system property "javax.net.ssl.keyStore"
* transport.keyStorePassword - default is to read from the system property "javax.net.ssl.keyStorePassword"
* transport.trustStoreLocation - default is to read from the system property "javax.net.ssl.trustStore"
* transport.trustStorePassword - default is to read from the system property "javax.net.ssl.trustStorePassword"

This requires creating _temporary_ trust stores within the connector, storing them in temporary local files and adding these options to the connection URL. 

This workaround may go away in the future, when Syndesis supports adding TLS certificates using secrets.

zregvart:
@dhirajsb I think this was done, right?

PR #1571, opened 2018-02-14
Added support for AMQP certificates, refactored Cert handling into a common connector-utils module

dhirajsb:


dhirajsb:
Fixes #1225

app/connectors/connector-utils/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+    <version>1.3-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>connector-utils</artifactId>
+  <name>Connector :: Utils</name>
+  <packaging>jar</packaging>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
zregvart: I think there is no need for this, it should be inherited from syndesis-parent

app/connectors/connector-utils/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.6.0</version>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
+        </configuration>
+      </plugin>
```
zregvart: I think there is no need for this, it should be inherited from syndesis-parent

app/connectors/connector-utils/src/main/java/io/syndesis/connector/utils/KeyStoreHelper.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * Helper class to create KeyStores from Certificates.
+ *
+ * @author dhirajsb
+ */
+public class KeyStoreHelper {
+
+    private static final SecureRandom SECURE_RANDOM = new SecureRandom();
```
zregvart: Could use [SecureRandom.getInstanceStrong()](https://docs.oracle.com/javase/8/docs/api/java/security/SecureRandom.html#getInstanceStrong--), though if it's used for generating passwords, [ThreadLocalRandom.current()](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html#current--) ought to be enough random.

app/connectors/connector-utils/src/main/java/io/syndesis/connector/utils/KeyStoreHelper.java
```diff
@@ -0,0 +1,83 @@
...
+        return password;
+
+    }
+
+    public KeyStoreHelper store() {
+        try {
+
+            KeyStore keyStore = CertificateUtil.createKeyStore(certificate, alias);
+
+            tempFile = Files.createTempFile(alias, ".ks", PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rwx------")));
```
zregvart: We might not need the execute permission bit

app/connectors/connector-utils/src/main/java/io/syndesis/connector/utils/KeyStoreHelper.java
```diff
@@ -0,0 +1,83 @@
...
+            password = generatePassword();
+
+            try (OutputStream stream = new FileOutputStream(tempFile.toFile())) {
+                keyStore.store(stream, password.toCharArray());
+            }
+
+        } catch (KeyStoreException | IOException | NoSuchAlgorithmException | CertificateException e) {
```
zregvart: Could be `catch (GeneralSecurityException | IOException e)`

app/connectors/connector-utils/src/main/java/io/syndesis/connector/utils/KeyStoreHelper.java
```diff
@@ -0,0 +1,83 @@
...
+            throw new IllegalArgumentException(String.format("Error creating key store %s: %s", alias, e.getMessage()), e);
+        }
+
+        return this;
+    }
+
+    private static String generatePassword() {
+        return SECURE_RANDOM.ints(16, 'A', 'Z' + 1).mapToObj(i -> String.valueOf((char)i)).collect(Collectors.joining());
```
zregvart: Could be:

```java
int[] codepoints = SECURE_RANDOM.ints(16, 'A', 'Z' + 1).toArray();

return new String(codepoints, 0, 16);
```

see https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#String-int:A-int-int-

app/connectors/connector-utils/src/test/java/io/syndesis/connector/utils/CertificateUtilTest.java
```diff
@@ -0,0 +1,59 @@
...
+            sb.append("\nProvider : " + p[i].toString() + "\n");
+            Set s = p[i].keySet();
+            Object[] o = s.toArray();
+            Arrays.sort(o);
+            for (int j = 1; j < o.length; j++) {
+                sb.append(o[j].toString() + ", ");
+            }
+        }
+        System.out.println(sb.toString());
+*/
```
zregvart: can we remove this commented out code?

app/connectors/connector-utils/src/main/java/io/syndesis/connector/utils/KeyStoreHelper.java
dhirajsb: I used to think that `getInstanceStrong()` was preferable too, until I read [this](https://tersesystems.com/blog/2015/12/17/the-right-way-to-use-securerandom/).

app/connectors/connector-utils/src/test/java/io/syndesis/connector/utils/CertificateUtilTest.java
dhirajsb: yes, we can!

app/connectors/connector-utils/src/main/java/io/syndesis/connector/utils/KeyStoreHelper.java
dhirajsb: BTW, we also do [this](https://github.com/fabric8io-images/s2i/blob/39f99bf1755809069576233e914f3bc0eb7fd8d3/java/images/rhel/Dockerfile#L29).

Issue #1550, opened 2018-02-13
Config form input fields for steps/connections in the editor are too long


gashcrumb:
This has been bugging me for awhile, basically all the forms in the editor have really long input fields that stretch across the page, especially bad on widescreen monitors.

I don't know if this is really inline with the designs also.  Here's an example:

![download](https://user-images.githubusercontent.com/351660/36171911-4b83c99a-10d2-11e8-9216-25cef623896e.png)

michael-coker:
@gashcrumb should they be something closer to the oath configuration form field length (or shorter than that)? Or do you have an example another page that currently displays them the way you're envisioning?

![screen shot 2018-02-13 at 3 07 58 pm](https://user-images.githubusercontent.com/35148959/36173824-df581304-10cf-11e8-9b43-d344044ad732.png)

gashcrumb:
TBH even those forms seem a bit long to me...

So the connection configuration page in the create connection wizard isn't too bad, but part of that is the way the page is designed.

Normally bootstrap forms at least seem to have some max width built-in but somehow we've managed to break that in our own forms.  Though as I look at the patternfly example form page it's kinda got the same problem => https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/forms.html

@dongniwang I dunno, what do you think?

michael-coker:
I think @dongniwang is out on PTO for the rest of the week.

@sjcox-rh what do you think?

michael-coker:
We can easily just use the bootstrap grid to display the form itself in a more narrow column. No CSS required, and the column width can be different per component. What about this? Also, should the buttons under the form be flush with the right edges of the form inputs?

## before
![screen shot 2018-02-13 at 4 18 44 pm](https://user-images.githubusercontent.com/35148959/36176921-c8df2734-10d9-11e8-83f4-a4d02c598e73.png)

## after
![screen shot 2018-02-13 at 4 19 39 pm](https://user-images.githubusercontent.com/35148959/36176923-c8edbc2c-10d9-11e8-94ff-34aeb0922481.png)

michael-coker:
@gashcrumb looks like the existing input length is inline with the designs from UXD. I'm not sure if that's the latest design, since there are some discrepancies (for example, everything's in a white box, the button alignment is different) 

From https://github.com/syndesisio/syndesis/blob/master/ux/designs/describe-data/describe-data.md

[...]

PR #1574, opened 2018-02-14
add box around integration editor form fields, align buttons

michael-coker:
From https://github.com/syndesisio/syndesis/issues/1550

One part I'm not sure about is I moved some code out of [basic-filter.component.html](https://github.com/syndesisio/syndesis/compare/master...michael-coker:integration-editor-form-box?expand=1#diff-10e9e598fc38a951988e35b5cd4fb3f1) and into [step-configure.component.html](https://github.com/syndesisio/syndesis/compare/master...michael-coker:integration-editor-form-box?expand=1#diff-a090bd9b0bc3b003c3d0ccf4a5f170c3) to support bringing the basic filter editor's control buttons and the form itself into the same parent container.


gashcrumb:
@michael-coker I think that change to the basic filter is fine, LGTM!

Gonna merge manually, that failure in the rest module is a known issue.

PR #1585, opened 2018-02-15
Added metadata to data shape for describe data shape, fixes #1552

### LINKED ISSUE #1552: Need a 'userDefined' boolean property on the DataShape type
Right now I'm using `type` for determining if I need to prompt for a data shape schema.  However I'll need to set the type/kind to something based on the user input.  Once that's set I'll have no way of determining that the data shape was actually user defined.

We need a boolean property on the data shape type that the UI can use to indicate that a data shape has a user defined type set so the UI knows that it needs to make that page available as the user configures a step.

dhirajsb:


dhirajsb:
@lburgazzoli is this all that's required to add metadata to data shape?

PR #1588, opened 2018-02-15
feature: integration detail page revamped with new UI and metrics view

deeleman:
This PR contains the implementation of a new UI for the Integrations detail page featuring a tabbed view layout. The original monolithic `DetailComponent` presentation layer has been split up into several dumb child components but that entire area needs a comprehensive rewrite to conform to actual modern component architectures.

The metrics area implements the same logic as the dashboard metrics widget and is lazy loaded with default data and then refreshes its data on a regular basis or upon requested. The metrics data for each integration is cached to boost component loading in successive component loads or to be leveraged i other components elsewhere in the app.

At the time of this writing, some data is missing (Integration last update timestamp - whenever the `lastUpdated` property is not populated, the UI fallbacks to `n/a`) or might be confusing (errors vs error messages - I might have overlooked the piece of data required in the Integration graph, which is not very intuitive TBH, and have resourced to the error messages delta coming from the API).

The Integration metrics part rely on the new NGRX store but all the other parts still rely on the legacy store, which will be hopefully replaced (at least in what refers to integrations only) in the coming days.

captura de pantalla 2018-02-15 a las 18 48 49

captura de pantalla 2018-02-15 a las 18 48 56

captura de pantalla 2018-02-15 a las 18 49 09


deeleman:
Ah, I almost forgot: The tabbed nav bar also provides support for logs, which seem to be disabled ATM, hence are automatically hidden (as you can notice in my screenshots)

gashcrumb:
Argh and AOT of course:

```
[ERROR] ERROR in src/app/integration/integration_detail/integration-metrics.component.html(28,67): : Property 'lastProcessed' does not exist on type 'Integration'.
```

deeleman:
How come? I ran the linter and tests right before pushing this PR. :(

gashcrumb:
> How come? I ran the linter and tests right before pushing this PR. :(

Welcome to Syndesis :-)

fixed the whitespace for ya but not sure on the AOT error, the compiler's complaint is correct (I get it here locally too) but am not sure, should we just add `lastProcessed` to `Integration` or `IntegrationOverview`?

deeleman:
Don’t worry, I’ll  look into this later tonight

PR #1590, opened 2018-02-15
Lifecycle fixes

iocanel:
This pull request tries to address the inconsistent data that are stored in the DB. For example:

1. multiple published deployments.
2. integration version not matching with the actual deployment version.
3. inconsistency between db and ui.

and so on...

Big part of the problem is that certain assumptions have been made that aren't somehow enforced and are do not reflect the way things work.

This PR does not modify the behavior, it just removes these assumptions, which doesn't offer anything and make things fragile.


iocanel:
@gashcrumb: FIY, as soon as the inconsistency between deployed version and integration version have been addressed, publishing older versions, started working like an actual rollback.

PR #1612, opened 2018-02-19
bring layout for actions with no configurable properties inline with rest of editor workflow

michael-coker:
match the look and feel of the other editor workflow steps.

### DIFF
--- a/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
+++ b/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
@@ -69,8 +69,10 @@
               <div class="card-pf-body">
                 <div class="row">
                   <div class="col-xs-12">
-                    <p [class]="error.class"
-                    [innerHtml]="error.message"></p>
+                    <p [class]="error.class">
+                      <span [class]="error.icon"></span>
+                      {{ error.message }}
+                    </p>
                   </div>
                 </div>
                 <div *ngIf="!hasConfiguration"
--- a/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
+++ b/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
@@ -64,12 +64,12 @@
       </div>
       <syndesis-loading [loading]="loading">
         <ng-container *ngIf="error">
-          <div [class]="error.class"
-               [innerHtml]="error.message">
-          </div>
-          <div *ngIf="!hasConfiguration" class="control-buttons">
-            <div class="pull-right">
-              <ng-container *ngTemplateOutlet="controlButtons"></ng-container>
+          <div class="row row-cards-pf">
+            <div class="card-pf">
+              <div class="card-pf-body">
+                <p [innerHtml]="error.message"></p>
+                <ng-container *ngTemplateOutlet="controlButtons"></ng-container>
+              </div>
             </div>
           </div>
         </ng-container>
--- a/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
+++ b/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
@@ -67,8 +67,18 @@
           <div class="row row-cards-pf">
             <div class="card-pf">
               <div class="card-pf-body">
-                <p [innerHtml]="error.message"></p>
-                <ng-container *ngTemplateOutlet="controlButtons"></ng-container>
+                <div class="row">
+                  <div class="col-xs-12">
+                    <p [class]="error.class"
+                    [innerHtml]="error.message"></p>
+                  </div>
+                </div>
+                <div *ngIf="!hasConfiguration"
+                    class="row control-buttons">
+                  <div class="col-xs-12">
+                    <ng-container *ngTemplateOutlet="controlButtons"></ng-container>
+                  </div>
+                </div>
               </div>
             </div>
           </div>
--- a/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts
+++ b/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts
@@ -206,7 +206,7 @@ export class IntegrationConfigureActionComponent implements OnInit, OnDestroy {
       // TODO figure out how to get a link in here that works
       this.error = {
         class: 'alert alert-info',
-        message: 'There are no properties to configure for this action'
+        message: 'There are no properties to configure for this action.'
       };
       return;
     }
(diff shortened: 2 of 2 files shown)


app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
```diff
@@ -64,12 +64,12 @@ <h3>{{ action?.name }}</h3>
       </div>
       <syndesis-loading [loading]="loading">
         <ng-container *ngIf="error">
-          <div [class]="error.class"
```
gashcrumb: We need to keep this CSS class setter, at least somewhere, as this is set dynamically by the component both for this case and if there's an error while fetching the descriptor.  In the case where there's no configuration it's set to `alert alert-info`.

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
```diff
@@ -64,12 +64,12 @@ <h3>{{ action?.name }}</h3>
...
-               [innerHtml]="error.message">
-          </div>
-          <div *ngIf="!hasConfiguration" class="control-buttons">
-            <div class="pull-right">
-              <ng-container *ngTemplateOutlet="controlButtons"></ng-container>
+          <div class="row row-cards-pf">
+            <div class="card-pf">
+              <div class="card-pf-body">
+                <p [innerHtml]="error.message"></p>
```
gashcrumb: Maybe add `[class]="error.class"` to this element?

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
```diff
@@ -64,12 +64,12 @@ <h3>{{ action?.name }}</h3>
       </div>
       <syndesis-loading [loading]="loading">
         <ng-container *ngIf="error">
-          <div [class]="error.class"
-               [innerHtml]="error.message">
-          </div>
-          <div *ngIf="!hasConfiguration" class="control-buttons">
```
gashcrumb: Removing this `*ngIf` means the buttons will show up in all cases, which we don't actually want, in this case the buttons should only show up if there's no configuration.

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
michael-coker: @gashcrumb My bad, seemed like an easy enough change but I should have touched base first. Added it back to the element that also contains the output (`error.message`).

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
michael-coker: @gashcrumb sorry about that, added it back.

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
michael-coker: @gashcrumb added.

michael-coker:
@gashcrumb looking at this again, should the alert box also include an icon with each type of output? http://www.patternfly.org/pattern-library/widgets/#alerts

gashcrumb:
Yes, you're right it definitely should.  Probably best to add an additional
attribute to the error object in the component and use that to set the
appropriate icon.

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
deeleman: I’m surprised this actually worked. The correct attribute should be [ngClass] instead.

gashcrumb: `[class]` works too, it's less powerful than `ngClass` but suits it's purpose here...

michael-coker:
@gashcrumb added the error icons.

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
deeleman: The reason why it works is because it's part of the W3C standard and whatever binding we inject in a native attribute wrapped by brackets will get properly digested and interpolated. In any event, sticking to the Angular spec is the safest way to avoid weird bugs in the future: `[ngClass]` behavior is already tested by angular, while `[class]` don't.

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts
```diff
@@ -196,8 +197,9 @@ export class IntegrationConfigureActionComponent implements OnInit, OnDestroy {
   initForm(position: number, page: number, descriptor: ActionDescriptor, error?: any) {
     if (error) {
       this.error = error;
-      this.error.message = error.message || error.userMsg || error.developerMsg;
       this.error.class = 'alert alert-warning';
+      this.error.icon = 'pficon pficon-warning-triangle-o';
+      this.error.message = error.message || error.userMsg || error.developerMsg;
```
deeleman: it's usually not a good idea to mix presentation logic with interaction/data/business logic inside the component controller. This makes editing the UI harder. In this particular case, we're forcing the template to feature certain classnames depending on the value of `error`. It would be easier to define classnames like this then:

```
<p [ngClass]="{ 'alert alert-warning': error }">
  <span [ngClass]="{ 'pficon pficon-warning-triangle-o': error }"></span>
  {{ error?.message || error?.userMsg || error?.developerMsg }}
</p>
```

That would allow to remove all the LOCs above just leave `this.error = error;` inside the conditional

michael-coker: @deeleman it looks like there are 2 types of errors for this. An "alert" and a "warning". If it's an "alert", we use an alert class on the output and display an alert icon and a string of text. If it's a "warning", we use a warning class on the output and display a warning icon and either `error?.message || error?.userMsg || error?.developerMsg`. How would you suggest doing it with those variations?

deeleman: We need to elaborate this a bit. How do we tell wether it is an alert or a warning? Please point me to the object variable that signals what case scenario is and I'll do my best to propose a way forward. Also, I'd need to know what DOM elements should we render for each case.

gashcrumb: It's one of many areas in the editor that needs work, effectively it's set here if there's an error or if there's no properties -> https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts#L200-L217

PR #1615, opened 2018-02-20
No need to add delegator role

abouchama:
No more need to add the delegator role "add_auth_delegator_role"


abouchama:
The installation works perfectly without this role on OCP 3.7 & 3.6

Issue #1621, opened 2018-02-20
Blank page with no errors sometimes when adding a data mapper step


gashcrumb:
Hmm, well you can't see it that well but in this case I've added a middle connection and a data mapping step after that connection, and I've defined the types as JSON instance data.  The first data mapper step in the flow loads up fine, but the second one does not

![datamapper-issue](https://user-images.githubusercontent.com/351660/36430716-dd5dcc52-1623-11e8-92bd-c28220511174.gif)

@igarashitm FYI

gashcrumb:
So really the only issue with the data mapper is no error is shown, as it turns out the SQL connector returns `json-schema` for the `kind` on the input data shape in that last step with no specification if the SQL to invoke doesn't have an input parameter, in this case the SQL connector should have `none` for it's input data shape.

igarashitm:
Now required pieces are all ready:
* #1624 fixes to respect SQL connector returns `none` if there's no parameter for the SQL
* atlasmap/atlasmap#297 fixes initialization error handling in atlasmap so it displays the error message if there's no source or target document.
* #1627 uses atlasmap/atlasmap#297

igarashitm:
merged

PR #1624, opened 2018-02-20
fix: Respect DataShapes returned via DynamicActionMetadata

igarashitm:
SQL connector returns kind='none' if there's no parameter for the target SQL (#1621)


igarashitm:
Hi @zregvart ,
In order to reflect what is returned for SQL dynamic metadata even when there's no parameter, I want it to accept overriding by kind='none'. But then I got a test failure in DynamicActionSalesforceITCase.shouldOfferDynamicActionPropertySuggestions:198,
```
Expecting value <Optional[DataShape{kind=java, type=org.apache.camel.component.salesforce.api.dto.CreateSObjectResult, metadata={}}]> in field <"outputDataShape"> but was <Optional[DataShape{name=Contact, description=Salesforce Contact, kind=json-schema, type=Contact, specification={"id":
...(snip)
```
So it seems that the Salesforce connector depends on current behavior not overriding with the dynamic metadata if kind is different than default. Do you have any idea what should we do here?

PR #1625, opened 2018-02-20
add mixin and helper class for scrollable body areas in app

michael-coker:
Added 2 mixins `syn-scrollable-parent` and `syn-scrollable-body`, and 2 helpers `.syn-scrollable` and `.syn-scrollable--body` to support parts of the app where a main area should have an internal scrollable area.

Also updated the integration editor workflow, and connection creation workflow to utilize the new scss.

### DIFF
--- a/app/ui/src/app/connections/create-page/create-page.component.scss
+++ b/app/ui/src/app/connections/create-page/create-page.component.scss
@@ -29,13 +29,13 @@
   syndesis-connections-connection-basics,
   syndesis-connections-configure-fields,
   syndesis-connections-review {
+    @include syn-scrollable-body;
     position: relative;
     flex: 1 1 auto;
     margin-left: -20px;
     padding-left: 20px;
     margin-right: -20px;
     padding-right: 20px;
-    @include syn-scrollable-body;
   }
 
   .connector-list {
--- a/app/ui/src/app/connections/create-page/create-page.component.scss
+++ b/app/ui/src/app/connections/create-page/create-page.component.scss
@@ -1,8 +1,7 @@
+@import 'syndesis-sass';
+
 :host {
-  display: flex;
-  flex-direction: column;
-  height: 100%;
-  overflow: visible;
+  @include syn-scrollable-parent;
 }
 
 .card-pf {
@@ -26,25 +25,24 @@
   flex: 0 0 auto;
 }
 
-/deep/ syndesis-connections-connection-basics,
-/deep/ syndesis-connections-configure-fields,
-/deep/ syndesis-connections-review {
-  position: relative;
-  flex: 1 1 auto;
-  overflow-y: scroll;
-  overflow-x: hidden;
-  margin-left: -20px;
-  padding-left: 20px;
-  // Dock the scrollbar to the right
-  margin-right: -20px;
-  padding-right: 20px;
-  height: 100%;
-}
+::ng-deep {
+  syndesis-connections-connection-basics,
+  syndesis-connections-configure-fields,
+  syndesis-connections-review {
+    position: relative;
+    flex: 1 1 auto;
+    margin-left: -20px;
+    padding-left: 20px;
+    margin-right: -20px;
+    padding-right: 20px;
+    @include syn-scrollable-body;
+  }
 
-/deep/ .connector-list {
-  margin-top: -5px;
+  .connector-list {
+    margin-top: -5px;
 
-  .toolbar-pf {
-    margin: -20px -50px;
+    .toolbar-pf {
+      margin: -20px -50px;
+    }
   }
 }
--- a/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
+++ b/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
@@ -28,84 +28,88 @@
       </ng-container>
     </button>
   </ng-template>
-  <div class="main-panel-inner action-configure">
+  <div class="action-configure syn-scrollable">
     <!-- Toolbar -->
     <div class="toolbar">
       <!-- Toolbar: Breadcrumbs -->
-      <div class="row toolbar-pf">
-        <div class="col-sm-12">
-          <div class="toolbar-pf-actions">
-            <div class="inline-block">
-              <ol class="breadcrumb">
-                <li>
-                  <a [routerLink]="['/']">Home</a>
-                </li>
-                <li>
-                  <a [routerLink]="['/integrations']">Integrations</a>
-                </li>
-                <li>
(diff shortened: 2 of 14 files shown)


app/ui/src/app/connections/create-page/create-page.component.scss
```diff
@@ -26,25 +25,24 @@
...
+  syndesis-connections-connection-basics,
+  syndesis-connections-configure-fields,
+  syndesis-connections-review {
+    position: relative;
+    flex: 1 1 auto;
+    margin-left: -20px;
+    padding-left: 20px;
+    margin-right: -20px;
+    padding-right: 20px;
+    @include syn-scrollable-body;
```
deeleman: It's usually better to include the `@import` statement at the beginning of the CSS block, so you can override its own CSS rules if required.

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
```diff
@@ -28,84 +28,88 @@
...
+        </div>
+        <syndesis-loading [loading]="loading">
+          <ng-container *ngIf="error">
+            <div class="row row-cards-pf">
+              <div class="card-pf">
+                <div class="card-pf-body">
+                  <div class="row">
+                    <div class="col-xs-12">
+                      <p [class]="error.class">
+                        <span [class]="error.icon"></span>
```
deeleman: I strongly encourage you to use `[ngClass]` instad of `[class]`.

app/ui/src/app/integration/edit-page/action-select/action-select.component.html
```diff
@@ -1,54 +1,52 @@
...
+        <!-- Actions -->
+        <div>
+          <p>Choose an action for the selected connection.</p>
+          <!-- Toolbar -->
+          <syndesis-list-toolbar [items]="actions$"
+                                [filteredItems]="filteredActions$"></syndesis-list-toolbar>
+          <!-- Action List -->
+          <syndesis-list-actions [actions]="filteredActions$ | async"
+                                [loading]="loading$ | async"
+                                (onSelected)="onSelected($event)"></syndesis-list-actions>
```
deeleman: Just a note for the future: Originally we use to indent component properties line by line and aligned to the custom element tag name. This is not an Angular convention and tightly couples the component name (which is prone to change) with code indenting.

In the future, feel free to just use the standard 2-space indenting when splitting code into several lines.

gashcrumb:
So couple things, if I get an error when configuring a connection:

screenshot from 2018-02-20 16-13-36

You can see this locally by using an SQL connection and type invalid SQL there (like I apparently did)

Also there's a new 'describe data' page that'll need some attention as well:

screenshot from 2018-02-20 16-16-55

You can get to this page by creating a broker connection via the create connection wizard, then use that connection in an integration.  To create a broker connection you can just use a dummy URL for the broker URL like `tcp:localhost:8080` and then just use dummy names for the target topic/queue when creating the integration.  The above page will show up after you configure the action.

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
```diff
@@ -28,84 +28,88 @@
       </ng-container>
     </button>
   </ng-template>
-  <div class="main-panel-inner action-configure">
+  <div class="action-configure syn-scrollable">
```
gashcrumb: I wonder if we really need to keep the page-specific classes here?

app/ui/src/app/integration/edit-page/save-or-add-step/save-or-add-step.component.html
```diff
@@ -1,90 +1,90 @@
...
+                  <li class="active">Save or Add Step</li>
+                </ol>
+              </div>
+              <div class="toolbar-pf-action-right">
+                <button class="btn btn-default"
+                        (click)="cancel()">Cancel
+              </button> &nbsp;
+                <button class="btn btn-default"
+                        (click)="save()"
+                        [disabled]="!flowPageService.canContinue() || flowPageService.saveInProgress || flowPageService.publishInProgress"><span *ngIf="flowPageService.saveInProgress" class="spinner spinner-sm spinner-inline"></span>Save as Draft
```
deeleman: I'm surprised this LOC passed the `yarn lint` gate, since it clearly exceeds the linter max-line settings. Coudl you properly reformat this for better readability? Thanks.

app/ui/src/app/integration/edit-page/save-or-add-step/save-or-add-step.component.html
```diff
@@ -1,90 +1,90 @@
...
+              </button>
+                <button type="button"
+                        class="btn btn-primary"
+                        (click)="publish()"
+                        [disabled]="!flowPageService.canContinue() || flowPageService.saveInProgress || flowPageService.publishInProgress"><span *ngIf="flowPageService.publishInProgress" class="spinner spinner-sm spinner-inline"></span>Publish</button>
+              </div>
```
deeleman: Same here as the above. :)

app/ui/src/scss/utils/_helpers.scss
```diff
@@ -79,3 +79,11 @@
...
   @include syn-icon-box(24px);
 }
+
+.syn-scrollable {
+  @include syn-scrollable-parent;
+
+  .syn-scrollable--body {
+    @include syn-scrollable-body;
+  }
+}
```
deeleman: Using compound SASS selectors will save you some extra coing and will infor better of the intent of each selector:

```css
.syn-scrollable {
  @include syn-scrollable-parent;

  &--body {
    @include syn-scrollable-body;
  }
}
```

Also, bear in mind the level of CSS specificity generated here. Are you sure you want to constrain `.syn-scrollable--body` to have a parent `.syn-scrollable` classname wrapping it in order to execute it?

Ideally this is not how _modifiers_ operate in BEM. A CSS class with a modifier suffix should be able to run as a standalone class.

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
michael-coker: They're no longer used, removed.

app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
michael-coker: I was going to address that #1612 but took care of it here instead.

app/ui/src/scss/utils/_helpers.scss
michael-coker: Thanks, and you're right, decoupled those 2 elements.

Issue #1560, opened 2018-02-14
UI: checkboxes doesn't seem to take into account default values


oscerd:
![s3-poll](https://user-images.githubusercontent.com/5106647/36208137-624e7140-1198-11e8-964b-635352f3f9cd.png)

the checkbox showed in the screenshot has a default Value equal to false

https://github.com/syndesisio/syndesis/blob/master/app/connectors/connectors/aws-s3-connector/src/main/resources/META-INF/syndesis/connector/aws-s3.json#L247

but is checked anyway by default.

gashcrumb:
Huh, thought we just fixed this, @seanforyou23 want to have a peek?

oscerd:
My running syndesis is at this commit 90e1e38baf064b71ef5a1fcdb0581d4298a5c018

oscerd:
I guess setting null as default value is really ugly. @lburgazzoli what do you think about this?

lburgazzoli:
On backend side the definition of a property is all done using strings, so we can't really return a true boolean value. IMHO the UI should honor what the property definition gives and properly handle default values according to the type of the property. Going with null to mean false is a little ugly because then one could type "false" by mistake and the value will be interpreted as true ...

gashcrumb:
Yeah, we could add an additional function to the form factory service on the UI side to massage the values appropriately according to the form configuration.  I had created this issue when I had to coax out `"null"` for unset fields, probably worth rolling this work into it also -> https://github.com/syndesisio/syndesis/issues/1250

seanforyou23:
Ok cool, fair points all around. Thanks for the feedback everyone - I will update the UI to better handle these cases, just wanted to make sure we had proper reasoning before jumping in. Thanks!

Issue #1250, opened 2018-01-26
Leaving form fields empty yields the corresponding data with a value of 'null' instead of ''


gashcrumb:
Basically we should figure out why this is happening and take out these code blocks:

https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/edit-page/step-configure/step-configure.component.ts#L157-L162

https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/edit-page/current-flow.service.ts#L409-L414

gashcrumb:
Also I don't like that checkboxes produce the string `"true"` vs an actual boolean.

gashcrumb:
Most likely we should also add a sanitize() function to the form factory service to ensure the values produced correlate to the form configuration.

gashcrumb:
@seanforyou23 Do you want to tackle this one along with #1560?  Probably worth rolling it all into one task maybe?

seanforyou23:
@gashcrumb seems logical to me!

PR #1629, opened 2018-02-21
fix(forms): correct and cleanup handling of form values

seanforyou23:
This PR corrects a bug in the UI where the forms service doesn't correctly handle values for setting the initial state of the checkbox in our model driven forms. It also cleans up some messy checks in the UI for the string "null" by actualling checking for null values where appropriate.

Should help:
https://github.com/syndesisio/syndesis/issues/1560 and
https://github.com/syndesisio/syndesis/issues/1250

add helper method to form factory service for suppressing null values
accept string 'false' as a falsey value for checkbox config
remove checks for string 'null' as they are no longer needed


gashcrumb:
I would probably call those two issues fixed with this, LGTM!

gashcrumb:
Codacy's hanging I guess, let's get this in...

Issue #1187, opened 2018-01-23
HTTP4 Connector


rohanmars:
Implement HTTP4 connector component.

**Parameters**

- hostname
- port
- resourceUrl
- bridgeEndpoint (true/false)
- contentEncoding (gzip)
- contentType (text/html)
- characterEncoding
- httpMethod (get/patch/post/put/delete/head/options/trace)

**Actions**
invoke (single action, which can be used for all the different methods types

**Verifier**
Yes

**Unit Test Approach**
undefined

* will need custom component for advanced ssl support (not in scope)

PR #1632, opened 2018-02-21
fix: http4 connector

rohanmars:
Fixes: #1187


lburgazzoli:
@rohanmars there are some pmd findings

app/connectors/connectors/http4/src/main/java/io/syndesis/connector/http4/Http4ComponentVerifierExtension.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.http.client.config.RequestConfig;
+import org.apache.http.client.methods.CloseableHttpResponse;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.methods.HttpUriRequest;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Http4ComponentVerifierExtension extends DefaultComponentVerifierExtension {
```
lburgazzoli: camel-http4 has its own [verifier](https://github.com/apache/camel/blob/master/components/camel-http4/src/main/java/org/apache/camel/component/http4/HttpComponentVerifierExtension.java), can we leverage it ?

app/connectors/connectors/http4/src/main/java/io/syndesis/connector/http4/Http4ConnectorFactory.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        @Override
+        public void setOptions(Map<String, Object> options) {
+            // connection parameters
+            hostname = (String) options.remove("hostname");
+            port = (String) options.remove("port");
+            resourceUrl = (String) options.remove("resourceUrl");
+            ssl = Boolean.valueOf((String) options.remove("ssl"));
+
```
lburgazzoli: Can play with parameters in `createDelegateEndpoint` ?

app/connectors/connectors/http4/src/main/java/io/syndesis/connector/http4/Http4ConnectorFactory.java
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        private ComponentExtension getComponentVerifier() {
+            return new Http4ComponentVerifierExtension(getComponentId(), getCamelContext());
+        }
+
+        @Override
+        public void setOptions(Map<String, Object> options) {
+            // connection parameters
+            httpUri = (String) options.remove("httpUri");
```
lburgazzoli: It is not clear to me how this should work as in the connector json, there's no reference to `httpUri` but there is as example `hostname` so I guess there should be some bits to build the uri from the parameters

app/connectors/connectors/http4/src/main/resources/META-INF/syndesis/connector/http4.json
```diff
@@ -0,0 +1,167 @@
...
+          "kind": "path",
+          "displayName": "Hostname",
+          "group": "common",
+          "required": true,
+          "type": "string",
+          "javaType": "java.lang.String",
+          "deprecated": false,
+          "secret": false,
+          "description": "Server Hostname"
+        },
```
lburgazzoli: I'm not sure if we want to set hostname & co as global properties or if it may be better to set them at action level, @kcbabo what do you think ?

kcbabo:
I tend to view it like this:

- Anything that stays the same for every invocation of an HTTP endpoint would be a connection config.  Scheme, host, port type of config fit in this bucket.
- Anything that a user would want to change on each invocation should be action config.  Path, Verb, etc. would fall into this bucket.

Putting everything in connection properties means that a user would need to define 4 different connections for a single HTTP resource endpoint supporting GET, PUT, POST, and DELETE.

app/connectors/connectors/http4/src/main/java/io/syndesis/connector/http4/Http4ConnectorFactory.java
rohanmars: @lburgazzoli weirdly the http4.json was not included in the PR branch. I had taken a copy of the repo before the rebase so I've repushed this.  I switch over to the a single endpoint url instead of seperate host/port/ssl parameters to simplify the UI. Camel http4 does use this single, httpUri, parameter in it's format, but that can't be mapped through if we want to easily support ssl.

lburgazzoli: @kcbabo is it ok to go with a single `httpUri` option or we should split it into different parts ?

rohanmars:
@kcbabo I have simplified the connector options in my latest version, which wasn't updated in the last squash: 

 https://github.com/rohanmars/syndesis/blob/syndesis-1187/app/connectors/connectors/http4/src/main/resources/META-INF/syndesis/connector/http4.json

In this version the user just enters a single httpUri and the http method. I can move the http method to the action. Do you want the hostname:port:ssl split back apart?

kcbabo:
Personally, I feel it's more natural to have the base URL (scheme, host, port) all in one property.  Natural to write and easy to copy/paste.

Path could be part of the above URL property or separated out.  

HTTP method should be action config, IMO.

rohanmars:
@kcbabo @lburgazzoli I've moved the path part and http method to the action. I think that is a better flow in the UI.

oscerd:
My impression is that it would be better to have two distincts connector one for http4 and one for https4, because even though the idea behind the component is the same, setting up an https4 endpoint is not the same, take a look here for example:

https://github.com/apache/camel/blob/master/components/camel-http4/src/test/java/org/apache/camel/component/http4/HttpsTwoDifferentSslContextParametersGetTest.java

We need to build an SSLContextParameters

https://github.com/apache/camel/blob/master/components/camel-http4/src/main/java/org/apache/camel/component/http4/HttpComponent.java#L88

and then use it in the component. So my opinion is that is better to have distinct connectors.

@kcbabo what do you think?

kcbabo:
Does the current connector work with TLS and non-TLS HTTP endpoints?  If so, then that works for me and I don't have an opinion on the internal details. :-)

rohanmars:
I tested using a simple https endpoint and it works. However, it will only work for endpoints signed by well known (built in) CA certs. The camel http4 component in my experience requires some extra hoops for https and self-signed certs which isn't currently supported in the connector. I supported https by writing a URL customizer which setups the appropriate https4 component when necessary.

lburgazzoli:
Replaced by #1809

PR #1664, opened 2018-02-23
feat(forms): add support for new tooltip properties

seanforyou23:
This PR adds support for two new properties in our forms service, controlHint and labelHint. These new properties give more flexibility in the way we are able to provide hints to users regarding how to fill out our forms.

Previously we only had a single "description" property, which would attach tooltip text to both the label and respective form control with the same text. This wasn't ideal and the limitations it posed are better described in https://github.com/syndesisio/syndesis/issues/250.

With this PR in place, we can now can set any of three properties to attach descriptive text to the label independently from the form control which is independent of text that's shown below a form control. Here is an example of this in action.

screen shot 2018-02-22 at 3 49 16 pm
screen shot 2018-02-22 at 3 49 26 pm
screen shot 2018-02-22 at 3 50 05 pm

This work is dependent on extending the ng-dynamic-forms DynamicFormControlModel, which I'm posting a PR for today. Depending on how this is received, we may need to use our own fork or ideally this work is accepted and we can continue using the library in a standard third-party fashion.

description/hint text is now (correctly) displayed below form controls by default
controlTooltip/labelTooltip are new properties which allow attaching tooltips to their respective html elements
remove ui-bootstrap styling pkg and reference to DynamicFormsBootstrapUIModule
add --preserve-symlinks flag to ensure yarn linking to local deps works
remove elementRef/forwardRef where not used as well as trailing commas


seanforyou23:
@dhirajsb one thing that needs to happen in addition to these changes, is providing support for these new property names in our rest service. What's currently the "description" property should be changed to the property name "labelHint" in order to keep the same experience for forms driven by the back-end. Can we work together on getting these changes in place?

app/ui/src/app/core/providers/form-factory-provider.service.ts
```diff
@@ -112,6 +112,8 @@ export class FormFactoryProviderService extends FormFactoryService {
     return new DurationInputModel({
       id: key,
       label: field.displayName || key,
+      labelTooltip: field.labelHint || null,
```
gashcrumb: `|| null` is a bit redundant, isn't it?  It's probably best to skip doing this.

gashcrumb:
Looks good!  Probably worth linking the upstream issue/PR for getting the support in place in `ng-dynamic-forms` for these fields to this PR so we can take off the wip label when it (hopefully) gets merged...

seanforyou23:
Pending acceptance of this upstream PR - https://github.com/udos86/ng-dynamic-forms/pull/699

app/ui/src/app/common/ui-patternfly/syndesis-form-control.component.html
```diff
@@ -8,8 +8,8 @@
          [class.required-pf]="model.required"
          [ngClass]="[elementLabelClass, gridLabelClass]">
     {{ model.label }}
-    <div *ngIf="model.hint !== null"
-         [tooltip]="model.hint"
+    <div *ngIf="model.labelTooltip !== null"
```
deeleman: Bear in mind that this will also render the tooltip if `model.labelTooltip` is `undefined` or just an empty string. Perhaps you might want to make the implementation more resilient by just doing:

```html
<div *ngIf="model.labelTooltip"
  ...
```

app/ui/src/app/common/ui-patternfly/syndesis-form-control.component.html
```diff
@@ -44,8 +44,8 @@
         <span [innerHTML]="model.label"
               [ngClass]="[elementLabelClass, gridLabelClass]">
         </span>
-        <div *ngIf="model.hint !== null"
-             [tooltip]="model.hint"
+        <div *ngIf="model.labelTooltip !== null"
```
deeleman: Please refer to my previous comment

dhirajsb:
@seanforyou23 so will there be no description property in deployment.json anymore? There is also the new connector descriptor, which follows the same schema but now split out across connectors instead of being in a single deployment.json. Those will have to be refactored as well.

seanforyou23:
@dhirajsb we'll still have the description property. With this PR in place, the text set for `description` will now be displayed just below the form control (check out the attached images for a visual). This is the natural position for this property in our forms library so we're putting it back in an effort to be more inline with how it works out of the box. You can use any combination of description, labelHint, and controlHint to decorate the form fields how best suites your features needs. I can definitely help convert the existing deployment.json(s) to use this new format. I imagine we'll have a few places where we want to tweak this as we go further with documentation. The pending PR was is now merged and a release was just cut so I'll go ahead and get these last bits in place. I'll ping you once it's ready!

seanforyou23:
@dhirajsb I noticed the two new properties labelHint and controlHint don't seem to be available as part of the payload from /api/v1/connectors. I was hoping to make these changes to the various *.json files to make this forms enhancement seamless but I can't really verify or test this fully without these properties in place.

What will be needed is to comb the various *.json files that feed the form properties and replace the `description` property with `labelHint` anywhere there needs to be a tooltip attached to a label. We'd also need to replace `description` with `controlHint` anywhere we'd like a tooltip attached to the actual form control. `description` will now by default place text below the form control as in the attached images - so until the aforementioned changes are in place, the forms will look "strange".

In an effort to wrap this PR up, I will soon remove the WIP flag for this PR and create a new issue for making the remaining changes on the back-end to use the new api parts. Let me know if you see any issues with this. Thanks!!

seanforyou23:
Issue https://github.com/syndesisio/syndesis/issues/1881 describes related work that's needed for this PR to be effective.

PR #1665, opened 2018-02-23
chore(rest): remove demo-data.json

zregvart:
Demo data is horribly out of date and not used at all, let's remove it.


oscerd:
I'm too fast, for both @rhuss and @lburgazzoli

Issue #1654, opened 2018-02-22
Review Data Buckets UXD


igarashitm:
@dongniwang is specifically concerning about the multiple source documents are shown expanded, which is supposed to be really longer and always need to scroll.
This is also related to the unresolved issue https://github.com/atlasmap/atlasmap/issues/105

Cc: @amysueg , @sjcox-rh

amysueg:
Design decision is collapsing the sources in the source card in the default view.

dongniwang:
Latest PR #1671 is created to address this. Please review.

igarashitm:
#1671 has been merged. Thanks @dongniwang :+1:

igarashitm:
#1686 - Forgot to add this link, impl ticket for collapsing source docs

Issue #1111, opened 2018-01-17
UXD to visualize Step ID


igarashitm:
#1110 needs Steps to have an identifier (suggestion : step 1, step 2, …)
Step id should be visible in:
* Integration Visualization Panel (IVP)
* Collapsed IVP
* Data mapper source labels (e.g. Status (Step 1))

Need UX design showing step IDs

dongniwang:
Reference: https://github.com/syndesisio/syndesis/issues/1138

TovaCohen:
I'm seeing "Step 1", "Step 2", "Step 3" and so on in the UX designs for showing logging, integration visualization, and data mapper enhancements. These "Step n" labels are being applied to both connections and steps. Until now, a step was a data operation between connections. So I'm concerned that applying "Step n" labels to connections is going to cause confusion. In place of "Step n", could we use just a number, "1", "2", "3" and so on? In prompts and user doc, we could refer to these as "order IDs".

igarashitm:
Hi @TovaCohen , IIRC we are considering to migrate Steps and Connections, i.e. Connection is going to be a kind of Steps. Until we implement it, wording might be unstable there...

dongniwang:
@igarashitm @TovaCohen - Agreed with Tomo, there's a discussion a few weeks ago about consolidating both steps and connections together in the UI screen where users add additional steps/connections after configuring start and finish connections. We definitely want to review the wording here to make sure we are not causing unnecessary confusion. 

Adding @amysueg @sjcox-rh @kcbabo here for discussion.

kcbabo:
Everything that happens in an integration flow is a step.  The "start" step and "finish" step are special in that they need to be connections (at least for now), but these are still steps.

TovaCohen:
Issue to clarify the UI:
https://github.com/syndesisio/syndesis/issues/1530

dongniwang:
UX design PR #1671 is created to address this. Please review.

igarashitm:
#1671 has been merged. Thanks @dongniwang :+1:

PR #1671, opened 2018-02-23
Add UX design for data buckets

dongniwang:
This design covers the following aspects of data buckets:
- Visualizing Step ID (#1111) and showing data type information (#700) in Integration Visualization Panel (IVP)
- Showing visual indicator when data types don't match (#700)
- Displaying multiple source documents in data mapper source card (#1654)

Please review: @igarashitm @kcbabo @gashcrumb @amysueg @sjcox-rh @michael-coker @seanforyou23 @TovaCohen


TovaCohen:
LGTM!  Thanks Dongni!

amysueg:
This looks *great* @dongniwang thank you.
I have a tiny concern that Michael S raised during the meeting today. In this design, we are specifying that the user must select/click on the "i" information icon in order to display the information. That behavior is not the same as the "i" icon in the configuration form pages which only requires hover to display a tooltip. I suspect it may be ok since the context is so different between the two areas of the application, but I'm a little concerned about the inconsistent implementations. Not sure what the answer is.

igarashitm:
@amysueg IIRC @gashcrumb pointed out that the hover is not accessible for touch device. Do we want to convert all the hovery into clicky? Maybe later...

gashcrumb:
I made 'em clicky in this PR -> https://github.com/syndesisio/syndesis/pull/1681

TovaCohen:
Can you clarify for me what the user needs to click on where previously they hovered?
Does the user still hover over the plus sign in the IVP to get the popup for adding a step or connection?

PR #1673, opened 2018-02-23
Bring integration importer inline with UXD designs

michael-coker:
Worked with UXD to bring this page inline with their original designs.

@kahboom the parts I'm not able to address are

1. Displaying the specific error message associated with each imported file beside that file. Currently I just display a generic error if the item in the uploader queue has `isSuccess = true`
2. The import review list at the bottom with details about the integration.
3. Adding a "done" button.

Are those parts you can work on, and pass back when the base functionality is there and I'll address the styling? Or would you like to handle it some other way?

### DIFF
--- a/app/ui/src/app/integration/import-export/import/integration-import.component.ts
+++ b/app/ui/src/app/integration/import-export/import/integration-import.component.ts
@@ -18,15 +18,21 @@ import {
 
 @Component({
   selector: 'syndesis-import-integration-component',
-  templateUrl: './integration-import.component.html'
+  templateUrl: './integration-import.component.html',
+  styleUrls: ['./integration-import.component.scss']
 })
 export class IntegrationImportComponent implements OnInit {
   public uploader: FileUploader;
+  public hasBaseDropZoneOver = false;
 
   constructor(
     public notificationService: NotificationService,
     private integrationSupportService: IntegrationSupportService,
   ) {}
+
+  public fileOverBase(e) {
+    this.hasBaseDropZoneOver = e;
+  }
 
   ngOnInit() {
     this.uploader = new FileUploader({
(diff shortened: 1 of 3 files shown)


app/ui/src/app/integration/import-export/import/integration-import.component.ts
```diff
@@ -18,23 +18,30 @@ import {
...
 
   constructor(
     public notificationService: NotificationService,
     private integrationSupportService: IntegrationSupportService,
   ) {}
 
+  public fileOverBase(e) {
+    this.hasBaseDropZoneOver = e;
+  }
+
```
deeleman: TypeScript class members are `public` by default so we should not include such accessor in our code. Feel free to remove it not only from your code but from any other instance of code you find on the way.

app/ui/src/app/integration/import-export/import/integration-import.component.ts
```diff
@@ -18,23 +18,30 @@ import {
 
 @Component({
   selector: 'syndesis-import-integration-component',
-  templateUrl: './integration-import.component.html'
+  templateUrl: './integration-import.component.html',
+  styleUrls: ['./integration-import.component.scss']
 })
 export class IntegrationImportComponent implements OnInit {
   public uploader: FileUploader;
+  public hasBaseDropZoneOver = false;
```
deeleman: Usually, when initializing boolean class members as `false`, we would just annotate the `Boolean` token, since it makes code more expressive, given the fact that boolean values are initialized as `null`, which equals `false` (unless we coerce type comparison later on, which is seldom the case).

michael-coker:
@deeleman pushed another commit to address the TS issues.

PR #1676, opened 2018-02-26
Add support for XML payloads in API connector

### LINKED ISSUE #1120: Support XML payloads in custom API connectors
Right now we generate JSON schema from Swagger specification and assume JSON only payload. If the operation from Swagger specification accepts only `application/xml` we should generate XML schema for the data shape and expect XML payload in the REST Swagger connector.

zregvart:
Adds support for converting Swagger specifications to Atlasmap schema set and in the connector runtime.

Fixes #1120


rhuss:
```
[INFO] -------------------------------------------------------------
[WARNING] COMPILATION WARNING : 
[INFO] -------------------------------------------------------------
[WARNING] /workspace/app/rest/connector-generator/src/main/java/io/syndesis/connector/generator/swagger/DataShapeGenerator.java:[29,59] unchecked conversion
  required: java.lang.Class<io.swagger.models.parameters.AbstractSerializableParameter<?>>
  found:    java.lang.Class
[INFO] 1 warning
```

is why the rest test fails ...

rhuss:
But I wonder why we have -Werror switched on ?

zregvart:
> But I wonder why we have -Werror switched on ?

It's because errorprone issues would otherwise be unseen (just `WARN:` on the console). This way it (obviously) fails the build that usually gets one's attention.

zregvart:
I'm merging this I don't think anyone can go through +7000 commit at the moment, it's easiest to merge it and then fix any issues in subsequent PRs. :crossed_fingers:

PR #1681, opened 2018-02-26
(chore) rework elements for step index, tooltips, datamapper required indicator

### LINKED ISSUE #1138: Show step index on integration visualization panel
We'll want some indicator of a step's actual position in the flow so that the user has some correlation between the integration visualization in the editor and the data mapper UI.

This will require some UXD input, also not sure if the indices should also be visible from the integration detail view, I kinda think not :-).

Related to #1110

Created from #1114

### LINKED ISSUE #1682: Visual indicator when data types don't match
This ticket is to apply final UXD outcome for #700, implement visual indicator for the data shapes and show a warning if required data shape is not provided from previous output.

Also we want to suggest adding "describe data type" on a previous step, or datamapper step with a direct link.
https://github.com/syndesisio/syndesis/issues/700#issuecomment-359110289

gashcrumb:
fixes #1682 
fixes #1138 

Initial work for getting the data buckets-related indicators in place for the data buckets epic, mostly so I can share with @michael-coker and get his help :-)

current design:
https://github.com/syndesisio/syndesis/blob/master/ux/designs/data-buckets/data-buckets.md#visualizing-step-id-and-showing-data-type-information-integration-visualization-panel-ivp

WIP because while the text for the step name is broken up now into different elements there's still a fair bit of layout work that needs cleanup.  The trash icon originally was using absolute positioning and outside of the step name, I've changed that and added the info icon, put them to the right for now.

Also the popovers used for showing the data mismatch text and the datashape details for connections aren't laid out that well and need some further work.

@igarashitm FYI and also I forget, was there an issue for tracking implementing the updated design?

### DIFF
--- a/app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.html
+++ b/app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.html
@@ -14,15 +14,19 @@
   </ng-template>
 
   <ng-template #tooltipPopover>
-    <div>
-      <i class="fa fa-remove pull-right" (click)="hideInfoPopover()"></i>
-
+    <div class="panel panel-default">
+      <div class="panel-heading">
+        <button class="pull-right close" type="button" (click)="hideInfoPopover()">
+          <span class="pficon pficon-close"></span>
+        </button>
+        <h3 class="panel-title">Connection Properties</h3>
+      </div>
+      <div class="panel-body">
       <!-- TODO Apply UXD outcome - https://github.com/syndesisio/syndesis/issues/700 -->
       <!--
       <div *ngIf="previousStepShouldDefineDataShape"><p style="color:red"><b>Define a data type</b></p></div>
-      <div *ngIf="shouldAddDatamapper"><p style="color:red"><b>Add a datamapper step</b></p></div>
       -->
-        <dl class="dl-horizontal">
+        <dl>
           <dt>
             Name
           </dt>
@@ -30,7 +34,7 @@
             {{ step.connection?.name }}
           </dd>
         </dl>
-        <dl class="dl-horizontal">
+        <dl>
           <dt>
             Action
           </dt>
@@ -38,7 +42,7 @@
             {{ step.action?.name }}
           </dd>
         </dl>
-        <dl class="dl-horizontal" *ngIf="inputDataShapeText">
+        <dl *ngIf="inputDataShapeText && inputDataShapeText !== ''">
           <dt>
             Input Data Type
           </dt>
@@ -46,7 +50,7 @@
             {{ inputDataShapeText }}
           </dd>
         </dl>
-        <dl class="dl-horizontal" *ngIf="outputDataShapeText">
+        <dl *ngIf="outputDataShapeText">
           <dt>
             Output Data Type
           </dt>
@@ -54,6 +58,7 @@
             {{ outputDataShapeText }}
           </dd>
         </dl>
+      </div>
     </div>
   </ng-template>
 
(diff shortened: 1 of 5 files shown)


igarashitm:
@gashcrumb thanks for working on this one! - I just assigned 2 issues to you, #1682 and #1138.

michael-coker:
@gashcrumb looks good, but can you move the step numbers into the `.progress-line` element associated with that step instead?

Alternatively, you could remove the step number all together and I can use CSS counters to display a number as a pseudo element beside each step, but from an accessibility standpoint, that approach would be for display only. If the step number has meaning on the page, and should be in the source for screen readers and the like, then leaving the number in the source is the best approach. But if the step number is just for visual/display purposes, then the CSS approach would be cleaner.

gashcrumb:
@michael-coker thanks for looking!  Moved the step index underneath the progress-line element there.

Also to note, the trash and info icons can wrap currently, but ideally if the action name is too long it should get truncated and the trash and info icons should stay where they are.  So like 'invoke stored procedure' tends to be too long for example, I think a few other connections have actions with fairly long names that can show layout problems.

gashcrumb:
@michael-coker FYI rebased this, should be a matter of `git stash && git pull git pull https://github.com/gashcrumb/syndesis.git data-buckets` to update your copy.

michael-coker:
@gashcrumb updated and pushed to this branch https://github.com/michael-coker/syndesis/tree/data-bucket-indicators-ui

gashcrumb:
@michael-coker awesome!  Lemme try and bring that commit over to this branch...

gashcrumb:
@michael-coker looks awesome!

Wonder, the popover content's description list doesn't play nice with the popover, is there a better method to implement that aspect of the design:

screenshot from 2018-02-27 12-41-58

gashcrumb:
@michael-coker did a bit more work on the markup, switched the description lists to the default vertical layout for now.

michael-coker:
@gashcrumb merged your changes, updated the popover formatting, and pushed to my branch. I agree that a `dl` is the most semantic element, but the default styling for the standard `dl` and `dl.dl-horizontal` won't fit, so applied some custom CSS in this case.

Also, is there a way to add a title/header ("Connection Properties") to the panel using `ngx-bootstrap/popover`?

gashcrumb:
@michael-coker ah, yes lookit that there's a `popoverTitle` attribute -> https://valor-software.com/ngx-bootstrap/#/popover#popover-directive, lemme pull your changes over and have a look, thanks!

gashcrumb:
@michael-coker hmm, `popoverTitle` doesn't seem to work for me, could be that it's ignored when a custom template is specified.  Let me have a look at the upstream code a bit more, might push an update to this branch.

gashcrumb:
K, think this is ready.  I wound up having to hide the close buttons as the popover directive doesn't support any customization for the title element.  However I've changed it's configuration so that clicking anywhere outside of the popover closes it.  Also tweaked the layout of the popovers for the flow view.

gashcrumb:
@michael-coker thanks for your help!

michael-coker:
@gashcrumb awesome! you bet 😁

app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.html
```diff
@@ -1,33 +1,73 @@
 <div *ngIf="step"
      class="flow-view-step">
 
-  <ng-template #tooltipPopover>
-    <span>
+  <ng-template #addDatamapper>
+    <div class="alert alert-warning no-bottom-margin">
+      <!--
```
kahboom: We should probably just remove this.

Issue #1672, opened 2018-02-23
Reduce vertical space between steps in integration editor


michael-coker:
From @kcbabo 

> When creating an integration, the amount of space between steps is really large.  We had talked about eliminating some of that space, but I don't think it's been implemented.  Interestingly, when you save/publish an integration space between steps collapses down.

amysueg:
Thanks for creating this issue @michael-coker - can you check in with @sjcox-rh  regarding the visual design aspect of this?

michael-coker:
@kcbabo just to make sure I'm looking at the right thing, are you referring to this?

![screen shot 2018-02-26 at 9 18 11 am](https://user-images.githubusercontent.com/35148959/36678364-27253c6a-1ad6-11e8-999e-3f6fd98cf34f.png)

And you would prefer it look closer to this?

![screen shot 2018-02-26 at 9 17 53 am](https://user-images.githubusercontent.com/35148959/36678380-30ed6e2a-1ad6-11e8-8f41-07b2b9a00905.png)

kcbabo:
Yes and yes.

michael-coker:
@kcbabo @sjcox-rh adjusted the space and took screenshots of the workflow for your review. Let me know if this looks OK. https://drive.google.com/drive/folders/1Kln6Xqu0WCezLfkQPHldkHegiZL4hKS-?usp=sharing

kcbabo:
LGTM!

PR #1684, opened 2018-02-26
decrease vertical space between steps in the integration editor left sidebar

michael-coker:
A request from @kcbabo in https://github.com/syndesisio/syndesis/issues/1672

Got feedback and sign-off from @amysueg and @dongniwang

### DIFF
--- a/app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.html
+++ b/app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.html
@@ -19,7 +19,7 @@
 
   <!-- Icon Progress Line -->
   <div class="col-md-3 progress-line">
-    <p [class]="'icon ' + getParentActiveClass() + ' ' + getMenuCompleteClass() + ' ' + getConnectionClass()"
+    <div [class]="'icon ' + getParentActiveClass() + ' ' + getMenuCompleteClass() + ' ' + getConnectionClass()"
        [popover]="tooltipPopover"
        #pop="bs-popover"
        placement="right"
@@ -32,7 +32,7 @@
       <img class="syn-icon-small"
            *ngIf="step.stepKind === 'endpoint' && step.connection"
            [src]="step.connection | synIconPath">
-    </p>
+    </div>
   </div>
 
   <!-- Step Item Wrapper -->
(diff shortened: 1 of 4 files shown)


app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.html
```diff
@@ -19,7 +19,7 @@
 
   <!-- Icon Progress Line -->
   <div class="col-md-3 progress-line">
-    <p [class]="'icon ' + getParentActiveClass() + ' ' + getMenuCompleteClass() + ' ' + getConnectionClass()"
+    <div [class]="'icon ' + getParentActiveClass() + ' ' + getMenuCompleteClass() + ' ' + getConnectionClass()"
```
deeleman: Let's try to stick to the Angular spec and use `[ngClass]` instead. Relying on non-supported syntaxes just because it works might lead to unexpected behaviors in the future (in this case, `[class]` just works because it is a combination of Angular's current template engine, which will be replaced by IvyTemplate in Angular 6, and the default W3C spec).

michael-coker: @deeleman done!

Would it make sense to create an issue to address all of the cases where `[class]` is currently used? Looks like it only occurs in these 10 files

    cmichael:src cmichael$ grep -ri '\[class\]' *
    app/app.component.html:               [class]="productBuild ? 'navbar-brand-icon product' : 'navbar-brand-icon'"
    app/app.component.html:               [class]="productBuild ? 'navbar-brand-name product' : 'navbar-brand-name'"
    app/customizations/tech-extensions/import/tech-extension-import.component.html:          <div [class]="error.level" [innerHtml]="error.message">
    app/integration/integration_detail/integration-description.component.html:              <div [class]="'step-line ' + getStepLineClass(index)"></div>
    app/integration/edit-page/edit-page.component.html:    <div [class]="getPageRow()">
    app/integration/edit-page/edit-page.component.html:      <div [class]="getSidebarClass()">
    app/integration/edit-page/flow-view/flow-view.component.html:    <div [class]="containerClass"
    app/integration/edit-page/step-configure/step-configure.component.html:    <div [class]="getToolbarClass()">
    app/integration/edit-page/step-configure/step-configure.component.html:          <div [class]="error.class"
    app/dashboard/dashboard_integrations/dashboard-integrations.component.html:                    <span [class]="'label ' + getLabelClass(integration)">

Issue #889, opened 2018-01-03
AMQ: `Client ID` on connection prevents reuse of same JMS connection in integration 


dsimansk:
Hi @dhirajsb,
I wonder if moving `Client ID` property to create connection flow is a good idea from a usability standpoint. Currently I can't create e.g. `topic -> queue` integration on a same AMQ connection because, client ID gets propagated to both endpoints and exception is thrown.

```
flows:
- steps:
  - kind: endpoint
    uri: activemq-subscribe-1
    properties:
      destinationName: cheese
      destinationType: topic
      durableSubscriptionId: syndesis-client
      messageSelector: exampleSelector
  - kind: endpoint
    uri: activemq-publish-2
    properties:
      destinationName: food
      destinationType: queue
      persistent: "false"
```

application.properties
```
activemq-subscribe.configurations.activemq-subscribe-1.clientID=syndesis-client
activemq-publish.configurations.activemq-publish-2.clientID=syndesis-client
```

```

2018-01-03 11:05:43.936 ERROR 1 --- [           main] o.a.camel.component.sjms.SjmsConsumer    : Unable to create the MessageConsumer
--
  |  
  | javax.jms.JMSException: You cannot create a durable subscriber without specifying a unique clientID on a Connection


```

What is the difference between `Client ID` defined on Connection directly versus `Durable Subscription ID` provided configuring `Subscribe for messages` action? 

`Durable Subscription ID`seems to have no effect (prorably better to create a separate issue for it?).

Cc @jimmidyson fyi.

chirino:
@lburgazzoli is there anything we can do in the flow->camel model to represent this better?  It would be nice if you could use 1 integration connection twice in the flow and have it configure just 1 camel component under the covers.

jimmidyson:
As @lburgazzoli is on PTO and I guess @dhirajsb is too, is this something we can defer to TP4, @chirino?

lburgazzoli:
@chirino what we can do is to use the connection info to configure the component and the step bits to configure the endpoint which is not so easy at the moment because of the intermediate yaml format.

If we get rid of the intermediate yaml (#447) we can then leverage the new connector bits that do not force the creation of a dedicated component per connector so we can reuse an existing and pre-configured component.

jimmidyson:
I'm going to bump this to TP4.

PR #1693, opened 2018-02-27
[connectors] move activemq connectors to new model

### LINKED ISSUE #1667: [connectors] move activemq connectors to new model

lburgazzoli:
@dhirajsb I've moved the ativemq connetor to the new model so it would
be nice if you an give it a try as I may have forgot something :)

@dsimansk an you ahve a look too ? There is an hak now in palce that
should attempt to find out if there is any other amq step onfigured with
the same connetion and in that case share the same underlying camel
component, we may think about having a better handling of suh case but
that requires some review of the model.

Fixes #1667
Fizes #889


rhuss:
@lburgazzoli @dhirajsb would be awesome if we could review/merge it today.

If review is not possible in that timeframe, I still would like it to be merged today as I'm doing some refactorings over the weekend which involves renaming of directories. 

We should then do a review in retrospective.

lburgazzoli:
@dhirajsb can you review this ?

lburgazzoli:
I've removed the WIP label and I'm going to merge it manually as I didn't get any review so far. Please report any bug

Issue #1669, opened 2018-02-23
Support OCP 3.7


lburgazzoli:
@paoloantinori https://github.com/lburgazzoli/syndesis/tree/openshift-3.7

lburgazzoli:
@paoloantinori can we close this one ?

paoloantinori:
done, thanks

Issue #760, opened 2017-12-14
Unable to deploy an integration on openshift 3.7


lburgazzoli:
I've set minishift to use openshift 3.7 because of the kubernetes-client issue and now integrations are not more deployed (but built successfully) because of the following error:

    Couldn't deploy version 1: ReplicationController "mentions2todo-1" is 
    invalid: spec.template.spec.containers[0].image: Required value

rhuss:
The reason for this is probably our workaround around broken image triggers in OpenShift. See https://github.com/syndesisio/syndesis/blob/9c51ce8ed5cf19e24c2be24747db7534ee2f6b95/app/rest/openshift/src/main/java/io/syndesis/openshift/OpenShiftServiceImpl.java#L196 where we set an empty image name (actually with a single space which was considered for OpenShift 3.6 as non-empty.

It might be worth to retry whether config change triggers work again. @dsimansk do you remember the details still ? I'm a bit weak in remembering this actually ;-)

dsimansk:
Well, there was a problem with ImageChangeTriggers. You mean to try with this commented-out code and get rid of  `.withImage(" ").withImagePullPolicy("Always").withName(name) `

https://github.com/syndesisio/syndesis/blob/9c51ce8ed5cf19e24c2be24747db7534ee2f6b95/app/rest/openshift/src/main/java/io/syndesis/openshift/OpenShiftServiceImpl.java#L214-L221

But to the context of workaround. There was an issue with triggers in 3.6 version that was reproducible mainly on Fuse Ignite cluster, therefor the workaround. For the 3.7 the fix should be already there. They have rewritten trigger logic, I can dig for the GH issue, should be in some email thread.

rhuss:
I'm a bit worried why the workaround fails to work as we still use ' ' for image names in out initial templates. The must be happening some mangling on the way.

PR #1697, opened 2018-02-27
Support OpenShift 3.7

paoloantinori:
Hi, with the code change in this PR we are now supporting **both** OpenShift `3.6` and `3.7.1`.

`3.7.1` is now the new default in our companion scripts.

My test have been performed **manually** publishing and deleting a new Integration, and the test has proved to be successful on both the 2 versions.

The only thing I'm not sure, and I need your feedback is my removal of `ConfigChange` in the rebuild triggers. Do we need that? should I restore it?

Other than that, this is good to go.

### DIFF
--- a/app/rest/openshift/src/main/java/io/syndesis/openshift/OpenShiftServiceImpl.java
+++ b/app/rest/openshift/src/main/java/io/syndesis/openshift/OpenShiftServiceImpl.java
@@ -15,24 +15,26 @@
  */
 package io.syndesis.openshift;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.util.List;
-import java.util.Map;
-
 import io.fabric8.kubernetes.api.model.EnvVar;
 import io.fabric8.kubernetes.api.model.Quantity;
 import io.fabric8.kubernetes.client.RequestConfigBuilder;
 import io.fabric8.openshift.api.model.DeploymentConfig;
 import io.fabric8.openshift.api.model.DeploymentConfigStatus;
-import io.fabric8.openshift.api.model.ImageStream;
 import io.fabric8.openshift.api.model.User;
 import io.fabric8.openshift.client.NamespacedOpenShiftClient;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.syndesis.core.Names;
+import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


paoloantinori:
fixes #1669 and #760

tools/bin/commands/minishift
```diff
@@ -1,4 +1,12 @@
 #!/bin/bash
+export PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+
+DEFAULT_OPENSHIFT_VERSION="v3.7.1"
+DEFAULT_CPUS="2"
+DEFAULT_DISK_SIZE="20GB"
+DEFAULT_RAM="4912"
+
+set -x
```
rhuss: hey @paoloantinori , is this really necessary to switch on always ? you always can use --verbose ....

paoloantinori: Despite I actually like it's log4j style of logging in this case it's an honest leftover.
It wasn't supposed to get into.

I'll fix it. any comment instead to the fact that I have removed the ConfigChange trigger ? do I have to restore that too?

tools/bin/commands/minishift
```diff
@@ -1,4 +1,12 @@
 #!/bin/bash
+export PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
```
rhuss: Also would move this to top-level `syndesis` so that its common for all commands.

app/rest/openshift/src/main/java/io/syndesis/openshift/OpenShiftServiceImpl.java
```diff
@@ -169,75 +156,92 @@ private int nullSafe(Integer nr) {
...
+                        .addNewVolumeMount()
+                            .withName("secret-volume")
+                            .withMountPath("/deployments/config")
+                            .withReadOnly(false)
+                        .endVolumeMount()
+                        .endContainer()
+                        .addNewVolume()
+                            .withName("secret-volume")
+                            .withNewSecret()
+                                .withSecretName(name)
```
paoloantinori: @rhuss check here

PR #1702, opened 2018-02-27
(fix) rework button text across pages

### LINKED ISSUE #1700: Describe data shape for shapeless connectors uses out of order buttons
The first button should say 'Next' instead of 'Done' when configuring a shapeless action and the button on the next page where user specifies a data shape should say 'Done' instead of 'Next'

gashcrumb:
fixes #1700

### DIFF
--- a/app/ui/src/app/integration/edit-page/current-flow.service.ts
+++ b/app/ui/src/app/integration/edit-page/current-flow.service.ts
@@ -349,6 +349,15 @@ export class CurrentFlowService {
     return (position + 1) >= this.steps.length;
   }
 
+  isActionShapeless(descriptor: ActionDescriptor) {
+    if (!descriptor) {
+      return false;
+    }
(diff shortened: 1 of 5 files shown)


app/ui/src/app/integration/edit-page/current-flow.service.ts
```diff
@@ -349,6 +349,15 @@ export class CurrentFlowService {
     return (position + 1) >= this.steps.length;
   }
 
+  isActionShapeless(descriptor: ActionDescriptor) {
+    if (!descriptor) {
+      return false;
+    }
+    const inputDataShape = descriptor.inputDataShape;
+    const outputDataShape = descriptor.outputDataShape;
```
deeleman: Just a suggestion for the future. You can streamline these kinda assignations by harnessing the power of ES6 destructuring:

```javascript
const { inputDataShape, outputDataShape } = descriptor;
```

PR #1703, opened 2018-02-27
feat(chart): utilize patternfly donut chart in dashboard

seanforyou23:
This PR replaces the dashboard integrations counter widget with a component supplied by patternfly-ng, the donut chart. It also dumps the direct dependency on ng2-charts and chart.js.

Should help https://github.com/syndesisio/syndesis/issues/265 

replace custom chart in dashboard with patternfly donut equivalent
integrations and loading state now consult the IntegrationSupportService instead of relying on @Inputs
add ChartModule to DashboardModule
remove unused functions/styling
add patternfly-settings.js
remove chart.js ng2-charts and types/chart.js
update test to use new chartmodule from patternfly
remove ChartsModule from DashbordModule


gashcrumb:
@seanforyou23 test failure, `yarn test` should show the same error for you locally.  Most likely if you changed any module imports you'll need to do the same changes for any tests in the dashboard module.

app/ui/src/app/dashboard/dashboard_integrations/dashboard-integrations.component.ts
```diff
@@ -1,53 +1,73 @@
...
   selector: 'syndesis-dashboard-integrations',
   templateUrl: './dashboard-integrations.component.html',
   styleUrls: ['./dashboard-integrations.component.scss']
 })
-export class DashboardIntegrationsComponent implements OnChanges {
-  chartData: number[];
-  @Input() integrations: IntegrationOverviews;
+export class DashboardIntegrationsComponent implements OnInit {
+
+  integrationsObserver: Observable<IntegrationOverviews>;
```
deeleman: `integrationsObserver` conveys the idea that this member is an `Observer` and as such can emit new values (as if it was a `Subject` or `BehaviorSubject` type).

Perhaps `integrationOverviews$` is a more expressive name for this field, since it clearly conveys both the type it represents and its observable nature.

app/ui/src/app/dashboard/dashboard_integrations/dashboard-integrations.component.ts
```diff
@@ -1,53 +1,73 @@
...
+    this.integrationsObserver = this.integrationSupportService.watchOverviews();
+    this.integrationsObserver.subscribe(integrations => {
+      this.integrations = integrations;
+      this.loading = false;
+      this.integrationChartData = [
+        [`Published`, this.countActiveIntegrations()],
+        [`Draft`, this.countDraftIntegrations()],
+        [`Unpublished`, this.countInactiveIntegrations()]
+      ];
+    });
```
deeleman: `integrationSupportService.watchOverviews()` is fed by the Socket Messaging layer. That means it will keep "emitting" values that will be funneled down through the subscription defined here even after the component is destroyed and disposed. Therefore this subscription should be terminated once `DashboardIntegrationsComponent` is destroyed.

Add a private subscription class member like this to the top, right after the last class public members (`integrationsChartConfig`):

```typescript
private integrationOverviewsSubscription: Subscription;
````

Assign the above field to the subscription:

```typescript
this.integrationOverviewsSubscription = this.integrationsObserver.subscribe(integrations => {
...
```

and then implement the `OnDestroy` interface in the class and append this lifecycle handler right after `ngOnInit()`:

```typescript
ngOnDestroy() {
  if (this.integrationOverviewsSubscription) {
    this.integrationOverviewsSubscription.unsubscribe();
  }
}
```

gashcrumb: Ooh, missed that, good catch!  I withdraw my approval :-)

app/ui/src/app/dashboard/dashboard.component.spec.ts
```diff
@@ -33,7 +32,7 @@ import { ConfigService } from '@syndesis/ui/config.service';
 import { IntegrationActionMenuComponent } from '@syndesis/ui/integration/list/action-menu.component.ts';
 import { StoreModule as LegacyStore } from '@syndesis/ui/store';
 
-describe('DashboardComponent', () => {
+xdescribe('DashboardComponent', () => {
```
deeleman: Are you sure you want to disable this test suite?

seanforyou23: Not really, sadly I did this in the interest of time :/ 

After applying the changes to dashboard integrations component, the corresponding test stopped working. The errors were related to routing, of which I didn't make any intentional change. I spent a good deal of the day yesterday trying to resolve this, but didn't have any luck. 

Some things I was trying were defining routes within the test directly, and exporting/importing the route that's defined within the DashboardModule. Also tried exposing the new service (IntegrationSupportService) to the testing module, since that's one of the "main" changes I made to the dashboard component, and many combinations of these attempts. Unfortunately no luck.

I looked to IntegrationListPage for inspiration since it also uses the IntegrationSupportService, but then I noticed this test is also disabled.

If you have an idea why the test would now fail I would love to resolve this before moving forward.

deeleman: Could you copy+paste here the stack trace of the error you get?

seanforyou23: https://gist.github.com/seanforyou23/e5c5d17ca9bbec31a7d6678219225891

gashcrumb: FYI the router error is a total red herring, has nothing to do with the test failure.

I've gotten the test to pass by commenting out the metrics component in the dashboard component HTML.  I've also gotten the test to pass by commenting out basically all of the metrics template too.  But then if I comment out all the interesting bits of that template, I can't get the test to pass.  Seems like each time I get to where I can narrow it down, the behavior of the test changes slightly and bits that I comment out that I think should fix it don't actually work.

The problem is the actual error gets lost by the test suite, and it *seems* to be an exception that occurs as the template is compiled but that's as close as I've been able to narrow down.

PR #1704, opened 2018-02-27
Lets make sure we trim kind and type on the DataShape.

chirino:


chirino:
Darn this trimmer wont work on DataShapeType since it's an Enum not a string.

PR #1705, opened 2018-02-27
(chore) add support for the 'order' property when connector/action form config specifies it

### LINKED ISSUE #1204: Add `controls` attribute alongside connector form definitions
Ideally we should just define the control order as part of the form definition, otherwise this switch/case is gonna get kinda big:

https://github.com/syndesisio/syndesis/pull/1202/files#diff-28b4a582e947fd283b192ca194ed84b5R48

So next to `properties` I propose having `controls` which is an array of ids from the `properties` map, the UI can use this in several places to guarantee the control order of forms.

gashcrumb:
fixes #1204 

@lburgazzoli am I right that the property name is `order`?  I didn't see this property after updating my local instance, even with a full rebuild, but am guessing it's because by default no order is set currently on anything yet.

Also corrects some type info in the model.

### DIFF
--- a/app/ui/src/app/connections/create-page/create-page.component.ts
+++ b/app/ui/src/app/connections/create-page/create-page.component.ts
@@ -91,10 +91,7 @@ export class ConnectionsCreatePage implements OnInit, OnDestroy {
         target.push('connection-basics');
         break;
       case 'review':
-        if (
-          !this.current.connection.connector.properties ||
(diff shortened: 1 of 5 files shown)


lburgazzoli:
yes the property is order and it is optional so if connector do not implement it, it is not shown. Should I add the order property to amq and aws s3 as per the UI hack ?

gashcrumb:
That would be awesome :-)

PR #1712, opened 2018-02-27
GitHub issue template for new features and bugs

deeleman:
This PR summarizes what @paoloantinori, @rhuss and me discussed over the F2F regarding having a template model for new GH issues that would serve as a centralized point of reference for issues/bugs combining UI/API/Backend input.

The proposed template tries to cater with both bugs and new features. GitHub provides support for handling several issue templates in a single repository. However, I'd suggest to begin with this one and refine its content. In the future we can provide more than one template (new feature, bugs, etc), which we can trigger by means of custom browser script bookmarks (or even a custom browser extension).

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -20,7 +20,7 @@ Briefly describe the issue you are experiencing (or the feature you want to see 
 
 ## Screenshot
 <!--
-For features/bugs entailing a presentation layer, drag and drop a screenshot depicting the issue or supporting the UX narrative for the new functionality.
+For features/bugs tackling with UI functionality, drag and drop a screenshot depicting the desired presentation layer or supporting the UX narrative for the new functionality.
 -->
 
 ## API Endpoints and Schemas
--- a/app/ui/.github/ISSUE_TEMPLATE.md
+++ b/app/ui/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,39 @@
+## This is a...
+<!-- Check one of the following options with "x" -->
+<pre><code>
+[ ] Feature request
+[ ] Regression (a behavior that used to work and stopped working in a new release)
+[ ] Bug report  <!-- Please search GitHub for a similar issue or PR before submitting -->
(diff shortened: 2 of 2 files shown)


zregvart:
Thanks! I'm reading the [docs](https://help.github.com/articles/creating-an-issue-template-for-your-repository/) and the example there uses lowercase filename (`issue_template.md`), is the support for this case-insensitive?

I would perhaps use headings level3 (`###` instead of `##`)?

I think we need to prompt for version or environment details (if relevant). Do we need a reminder to try with the latest version also?

deeleman:
> is the support for this case-insensitive?

I presume it is. I just checked the [Angular setup](https://github.com/angular/angular/tree/master/.github) for this and it uses uppercase, which is BTW quite a standardized convention for markdown files in GitHub repos anyways. [Some other repos](https://github.com/stevemao/github-issue-templates) I've checked seem to follow the same convention.

> I think we need to prompt for version or environment details (if relevant). 

Agreed. It is actually quite a standardized practice. I'm limited here to my little knowledge on the matter though (I'm familiar with browser environments but not with the other environments that might impact the backed or the execution context) so commits and additional PRs are welcome here. 

Ultimately this PR is intended to be a group endeavor, so feel free to edit the template to add whatever you feel is necessary. :)

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
+## This is a...
+<!-- Check one of the following options with "x" -->
+<pre><code>
```
rhuss: when we leave out the preformatting,then it should be directly rendered as checkboxes by GitHub Markdown, like in

- [x] Finish my changes
- [ ] Push my commits to GitHub
- [ ] Open a pull request

rhuss:
Looks good to me, and this templates is supposed to be adapted anyway when needed, so I suggest to merge it.

paoloantinori:
@deeleman question for you:

is this meant to be used for bugs or even for greenfield development?

could you use it for you Log story for example, or should we reason on something different for that?

PR #1720, opened 2018-02-28
refactor: Rename 'syndesis-restricted' to 'syndesis', removed docker image mode

rhuss:


paoloantinori:
@rhuss are you also planning to update the corresponding quickstart on syndesis.io ?

https://github.com/syndesisio/syndesis.io/blob/master/content/quickstart.md

rhuss:
yes, need to find out all the places where the 'restricted' suffix has been used. Guess its in some documentation as well (and for sure in the deployment scripts of our cluster)

app/deploy/README.md
```diff
@@ -108,23 +103,25 @@ OpenShift includes the ability for a service account to act as a limited OAuthCl
...
 ```
 
 #### Create the template to use
 
-We will create the template in the project, rather than in the openshift namespace as it is assumed the user does not have cluster-admin rights:
+Create the template:
 
 ```bash
-$ oc create -f https://raw.githubusercontent.com/syndesisio/syndesis-openshift-templates/master/syndesis-dev-restricted.yml
+$ oc create -f syndesis.yml
```
zregvart: Should this point to raw.githubusercontent.com...?

dsimansk: It refers to local file in git repo, imho.

rhuss: Yes, its just meant as an example. wanted to get rid of the URL which is bogus anyway and I just assume that you cd is the one where the README is.

And tbh, we should maybe even remove this doc here and _only_ reference to the manual.

PR #1721, opened 2018-02-28
fix: Don't use multiple dashes in templates

dsimansk:
Switch log delimited during stored procedure creation from `-----` to `*****` because of:

> YAML uses three dashes (“---”) to separate directives from document content. 

Such a use of dashes may cause premature end-of-file and incomplete unmarshalling to object.
For more details: https://github.com/fabric8io/kubernetes-client/pull/932

This breaks template loading in kubernetes-client since version `3.1.2`


dsimansk:
It's going to be addressed in #1720 with other changes to templates.

Issue #1718, opened 2018-02-27
Stored Procedure Selection is not set in edit screen


KurtStam:
I created a DB2DB scenario, Periodic SQL: SELECT * FROM CONTACT, and into the add_lead stored procedure, with an added mapping step. When going back to edit the stored procedure selection it is not set.

<img width="1301" alt="screen shot 2018-02-27 at 4 04 45 pm" src="https://user-images.githubusercontent.com/35576/36758688-2ff2a630-1be3-11e8-8306-2294606e49dc.png">

gashcrumb:
Can you check the stored value that comes back when the integrations is loaded?  I think it sometimes comes back as an encoded value.

KurtStam:
It looks to me like it's only making a call to the connection:
https://syndesis.192.168.64.28.nip.io/api/v1/connections/5/actions/sql-stored-connector

I think it needs to call the integration endpoint:
https://syndesis.192.168.64.28.nip.io/api/v1/integrations/<id>

and you're right that the 'procedureName' property is coming back encoded. Looking into that.

gashcrumb:
> It looks to me like it's only making a call to the connection:
https://syndesis.192.168.64.28.nip.io/api/v1/connections/5/actions/sql-stored-connector

Yeah, that's done on the action configuration page, there's a call to fetch the integration that occurs before that when you enter the editor.

KurtStam:
OK that fixed it.

PR #1723, opened 2018-02-28
Stored Procedure Selection is not set in edit screen #1718

KurtStam:


KurtStam:
This change addressed the issue

PR #1733, opened 2018-02-28
fix: duplicate dependency

KurtStam:


KurtStam:
retest please

app/rest/runtime/pom.xml
```diff
@@ -972,12 +972,6 @@
       <scope>test</scope>
     </dependency>
 
-    <dependency>
-      <groupId>com.h2database</groupId>
-      <artifactId>h2</artifactId>
-      <scope>test</scope>
```
zregvart: Shouldn't we keep this one and remove the `<scope>runtime</scope>` one, we're not depending on h2 at runtime right?

zregvart:
@KurtStam so I don't forget, I can't invite you to CircleCI (I think it has something to do with following/or not the GitHub project), I think the workaround is to open [the link](https://circleci.com/gh/syndesisio/syndesis/tree/pull%2F1733) directly.

app/rest/runtime/pom.xml
KurtStam: @zregvart y that seems reasonable to me.

KurtStam:
@zregvart RE Circle, using that link didn't help me.

Issue #1744, opened 2018-03-01
Unable to add steps


lburgazzoli:
## This is a...

<pre><code>
[ ] Feature request
[X] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

Unable to add steps between connections, the error shown on browser's console is:

```
main.34c31e3c4bd63a61a994.bundle.js:1 ERROR TypeError: Cannot read property 'properties' of undefined
    at t.getProperties (main.34c31e3c4bd63a61a994.bundle.js:1)
    at n.loadFormSetup (1.2418b268c2b57bd28321.chunk.js:1)
    at n.loadForm (1.2418b268c2b57bd28321.chunk.js:1)
    at e._next (1.2418b268c2b57bd28321.chunk.js:1)
    at e.__tryOrSetError (polyfills.00b2490f9e8f9b4c7b5e.bundle.js:1)
    at e.next (polyfills.00b2490f9e8f9b4c7b5e.bundle.js:1)
    at e._next (polyfills.00b2490f9e8f9b4c7b5e.bundle.js:1)
    at e.next (polyfills.00b2490f9e8f9b4c7b5e.bundle.js:1)
    at e._next (polyfills.00b2490f9e8f9b4c7b5e.bundle.js:1)
    at e.next (polyfills.00b2490f9e8f9b4c7b5e.bundle.js:1)
Ct @ main.34c31e3c4bd63a61a994.bundle.js:1
```

## Screenshot

![image](https://user-images.githubusercontent.com/1868933/36830395-c227b39c-1d23-11e8-9909-155265e6a675.png)

## Tasks involved / Steps to Reproduce

1. Create an integration
2. Add a `Periodic SQL Invocation` action
3. Add a `Invoke Stored procedure` action
4. Add a datamapper step

paoloantinori:
When double checking this for @lburgazzoli  I actually hit a different error:

<details>
  <summary><b>Click to expand</b></summary>
<pre><code>
TypeError: this.getStepConfig(...) is undefined
Stack trace:
8Dgd/e.StepStore</t.prototype.getProperties@https://syndesis-staging.b6ff.rh-idev.openshiftapps.com/main.34c31e3c4bd63a61a994.bundle.js:1:676871
[...]

PR #1745, opened 2018-03-01
DataBuckets exception - Incompatible Source Document 'null'

### LINKED ISSUE #1701: DataBuckets exception - Incompatible Source Document 'null'
I created test for data buckets in syndesis-qe repo, but created integration does not work with exception in integrations pod logs:
[exception.txt](https://github.com/syndesisio/syndesis/files/1763726/exception.txt)

My integration is:

Database - database - datamapper - salesforce

You can see feature file of the test here so you can easily reproduce this issue, just follow those steps: 
[feature.txt](https://github.com/syndesisio/syndesis/files/1763741/feature.txt)

I am using output of first database connector in data mapper. Note that if I delete second database step, this integration works as expected so it looks like first database output is not available in data mapper step.

I am not sure if data buckets are already fully implemented as its epic issue https://github.com/syndesisio/syndesis/issues/1110 is still in open state.
If it is the case at least we already have some tests.

lburgazzoli:
Fixes #1701


lburgazzoli:
@igarashitm this should fix #1701 but because of #1744 I was not able to test it end-to-end

PR #1765, opened 2018-03-01
chore(openshift): prefix to openshift resources

### LINKED ISSUE #1438: Pod with a new integration can replace any other existing pod on openshift with same name
If there is an existing pod named X and you create and publish a new integration also with the very same name X then the old pod will be replaced by the new one. It can destroy whole Syndesis app.

zregvart:
Adds a simple prefix (`i-`) to builds, deployment configurations, pods and secrets so the user provided integration name doesn't collide with any of the platform resource names.

Fixes #1438

### DIFF
--- a/app/rest/openshift/src/main/java/io/syndesis/openshift/OpenShiftServiceImpl.java
+++ b/app/rest/openshift/src/main/java/io/syndesis/openshift/OpenShiftServiceImpl.java
@@ -36,6 +36,8 @@ import java.util.Map;
 public class OpenShiftServiceImpl implements OpenShiftService {
     private static final Logger LOGGER = LoggerFactory.getLogger(OpenShiftServiceImpl.class);
 
+    private static final String OPENSHIFT_PREFIX = "i-";
+
     private final NamespacedOpenShiftClient openShiftClient;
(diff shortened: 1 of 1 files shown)


app/rest/openshift/src/main/java/io/syndesis/openshift/OpenShiftServiceImpl.java
```diff
@@ -36,6 +36,8 @@
 public class OpenShiftServiceImpl implements OpenShiftService {
     private static final Logger LOGGER = LoggerFactory.getLogger(OpenShiftServiceImpl.class);
 
+    private static final String OPENSHIFT_PREFIX = "intg-";
```
rhuss: Can we discuss on the prefix ;-) ? is 'intg-' a common acronym for "integrations" ? (sorry when I'm ignorant here"). What about just to use "i-" as the names themselves can be also large.

zregvart: yeah, I was thinking of just using `s-` but then Keith opened issue about pod names and ended up with `intg-`, let me change it to just `i-`

PR #1778, opened 2018-03-02
Added missing Syndesis metrics to Prometheus config, fixed RE2 expression

dhirajsb:


rhuss:
@dhirajsb you *have* to run the generator, otherwise the templates wont be updated !

PR #1784, opened 2018-03-02
feat(metrics): integrations activity grid (w/o detail view) including refresh and pagination features

deeleman:
This PR entails the integration activity metrics page including the following:

- Basic gridview rendering the activity records from a given integration
- Refresh button, featuring the last time the recordset was successfully refreshed. FYI and given the little time I had to ship this, the whole implementation is not Redux-driven, so the data is refreshed (hence a Http call is performed) every time the component is instantiated. That is why only the time is refreshed, since this component will seldom stay alive more than a single session. In the future that piece of data should be fully observable and the text should be repurposed to something like "XXX minutes ago at HH:mm" or "XXX days ago at HH:mm", which is easier for us.
- Page size and pagination support: I though the PF list component featured built-in support for this and it turns out it doesn't, but what it does is to mutate the pagination config object (which goes against all good practices in Angular). So we slice the dataset from within the component itself. Hopefully this recordset will never become too big. Otherwise we might face some performance issues here.
- Error messages when the API is not available (see attached screenshot).

### Screenshots

captura de pantalla 2018-03-02 a las 13 32 15

### Still pending (in order of relevance IMO):

1. Detail view when clicking on a row
2. Detailed log popup
3. Gridview filtering by date
4. Gridview filtering by keyword

#### The numbered list above should be split down and promoted into issues for better addressing each task, even concurrently. We're not being #agile otherwise.


KurtStam:
@deeleman looks good to me, but I think it would make sense to squash the commits.

PR #1785, opened 2018-03-02
feat(syndesis minishift): Add --vm-driver option

rhuss:


oscerd:
Great :-) Always thought about this, but I always forgot to raise an issue :-/

Issue #1769, opened 2018-03-01
Flacky SQL test


rhuss:
The following errors happens every third run or so, and should be fixed as it impacts our release process:

```
00:10:12.230 Tests run: 11, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.994 sec <<< FAILURE! - in io.syndesis.connector.sql.common.SqlParserTest
00:10:12.230 io.syndesis.connector.sql.common.SqlParserTest  Time elapsed: 0.109 sec  <<< ERROR!
00:10:12.230 java.sql.SQLException: Operation 'DROP TABLE' cannot be performed on object 'NAME0' because there is an open ResultSet dependent on that object.
00:10:12.230 	at org.apache.derby.impl.jdbc.SQLExceptionFactory.getSQLException(Unknown Source)
00:10:12.230 	at org.apache.derby.impl.jdbc.SQLExceptionFactory.getSQLException(Unknown Source)
00:10:12.230 	at org.apache.derby.impl.jdbc.Util.seeNextException(Unknown Source)
00:10:12.230 	at org.apache.derby.impl.jdbc.TransactionResourceImpl.wrapInSQLException(Unknown Source)
00:10:12.230 	at org.apache.derby.impl.jdbc.TransactionResourceImpl.handleException(Unknown Source)
00:10:12.230 	at org.apache.derby.impl.jdbc.EmbedConnection.handleException(Unknown Source)
00:10:12.230 	at org.apache.derby.impl.jdbc.ConnectionChild.handleException(Unknown Source)
00:10:12.230 	at org.apache.derby.impl.jdbc.EmbedStatement.executeStatement(Unknown Source)
00:10:12.230 	at org.apache.derby.impl.jdbc.EmbedStatement.execute(Unknown Source)
00:10:12.230 	at org.apache.derby.impl.jdbc.EmbedStatement.execute(Unknown Source)
00:10:12.230 	at io.syndesis.connector.sql.common.SqlParserTest.afterClass(SqlParserTest.java:154)
00:10:12.230 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
00:10:12.230 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
00:10:12.230 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #1786, opened 2018-03-02
fix: adding close statements

KurtStam:
Hopefully fixing issue https://github.com/syndesisio/syndesis/issues/1769 for good.


rhuss:
Nope, I still see this issue breaking the build from time to time (now also on Circle, not only on Jenkins):

image

PR #1795, opened 2018-03-02
fix(templates): Increase livenessProbe initialDelay for Java apps

rhuss:
This is necessary since i.e. syndesis-rest has been grown so that it infers with the liveness probe which would kill the pod when it doesnt startup in 200s

### DIFF
--- a/app/deploy/generator/06-syndesis-prometheus.yml.mustache
+++ b/app/deploy/generator/06-syndesis-prometheus.yml.mustache
@@ -160,7 +160,6 @@
           args:
             - '--config.file=/etc/prometheus/prometheus.yml'
             - '--storage.tsdb.retention=30d'
-{{^Probeless}}
           livenessProbe:
             httpGet:
               port: 9090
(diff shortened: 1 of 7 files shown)


app/deploy/syndesis.yml
```diff
@@ -1145,6 +1149,26 @@ objects:
             - targets:
               - localhost:9090
 
+          metric_relabel_configs:
```
rhuss: @dhirajsb I guess this comes in from one of your latest PRs. Don't forget to run 'generator/run.sh' when you change the templates (that what I did to get my changes in, but it picked up yours as well)

app/deploy/generator/06-syndesis-prometheus.yml.mustache
```diff
@@ -160,7 +160,6 @@
           args:
             - '--config.file=/etc/prometheus/prometheus.yml'
             - '--storage.tsdb.retention=30d'
-{{^Probeless}}
```
rhuss: @dhirajsb I removed that bogus line, which was left in somehow.

rhuss:
merging this as this fixes an issue already merged to master (that probeless tag).

app/deploy/syndesis.yml
dhirajsb: I thought I did, must have gotten overwritten somehow.

Issue #1755, opened 2018-03-01
Add JDBC Driver to the verifier classpath


KurtStam:
## This is a...

<pre><code>
[*] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
The verifier needs to verify connection info to a database, and should be able to use drivers that were uploaded by the user.

## Expected behavior
After driver upload, the REST layer should call an endpoint on the verifier to add that jar to a PV that is on its classpath. The verifier should then kill itself so that the driver is actually loaded on the classpath.

## API Endpoints and Schemas

See also: https://github.com/syndesisio/syndesis/issues/1291

KurtStam:
@lburgazzoli working on it on branch: https://github.com/KurtStam/syndesis/commits/drivers
It currently adds the PV on /deployment/ext and it adds the /deployment/ext to the verifier classpath. So currently I can at least manually copy in a ojdbc6.jar and test a 'verify' a connection. Next for me is adding the REST endpoint that can be called to upload a jar, and which then exits so the pod restarts.

lburgazzoli:
Is the oracle driver taken into account ? because using classpath was not reliable while developing extensions, instead what works is [PropertiesLauncher](https://docs.spring.io/spring-boot/docs/current/reference/html/executable-jar.html#executable-jar-property-launcher-features)

lburgazzoli:
btw, we should also take into account upgrading/deleting drivers

KurtStam:
@lburgazzoli you are right that '-cp' didn't;t work and I changed it. It's working now. Still need to add the verifier REST endpoint.

PR #1798, opened 2018-03-02
feature: upload JDBC Driver to verifier

KurtStam:
adding verifier pv & add to cp
adding /deployments/ext to the cp
Using ZIP and LOADER_PATH - verified driver is found
adding upload driver endpoint

Part of https://github.com/syndesisio/syndesis/issues/1755


rhuss:
@KurtStam please rebase and regenerate the files as there was an issue merged on master wrt to the templates.

KurtStam:
@rhuss rebased and regenerated

dsimansk:
@rhuss we currently have only 1 PV limit per project in a Ignite cluster. We'd need to address this with support before TP4.

rhuss:
@dsimansk the limit has been already increased to 3 (as we also need a PV for prometheus).

However, minimum size is 5 GB, so its probably really a waste to have it only for JDBC drivers. Also, not sure that the provided limit in the deployment descriptor of 256M will be even accepted.

rhuss:
there are test failures

rhuss:
@KurtStam  Without going into the the review details, but in general, how is this upload to `syndesis-met`a performed by the user ? Directly via the UI (which must not be the case) or indirectly via `syndesis-server` ? Is this PV shared between `syndesis-server` and `syndesis-verifier` ?

if so, @lburgazzoli would it then make sense to use this PV in general for all extension artefacts, instead of the DB ? I mean, to implement a FileFileStore (as alternative to the SqlFileStore). Then we would have a nice justification for the PV and a single place where we manage uploaded artefacts.

Then later we would go further, we could then even encapsulate that with a lightweight (non-Java) Microservice, which offers the service to syndesis-server and syndesis-meta (and maybe then for the runtimes, too, in case we are going for a generic runtime which needs to download those artefacts). So that we can avoid sharing data via PV (which contraticts the microservices 'share nothing' philosophy).

lburgazzoli:
@rhuss the `FileStore` was designed to eventually switch to a different backend so no problem on my side however I do not know if it is something we can safely achieve within our deadline.

rhuss:
@KurtStam could the test driver.jar please be less than 700k ? maybe something like 10-30k would be enough ?

rhuss:
@KurtStam just saw you added the endpoint for upload to `meta`. Does this mean we have now two uploads of the driver, one for `meta` and one into the DB for the runtimes ? Is this something the user has todo twice or is this dispatched in the background ?

nicolaferraro:
Talked with @KurtStam yesterday. The basic workflow should be this:

- user uploads the extension to syndesis server, it's put on the (sql) filestore but not installed
- user "installs" the extension using a endpoint on syndesis server, the extension is taken from the filestore and uploaded to meta (and the meta service is restarted in order for the LOADER_PATH to be read again). The user workflow is unchanged, user upload the extension only once.

Meta and server don't share anything, leaving meta also able to organize its own domain data autonomously on the mounted volume. Also, meta is not required to have the full set of extensions, but only extensions with a special tag ("meta-aware", or just "jdbc-driver" for now).

We can move the extension file store to a shared volume, but there are also other info needed by meta in order to create a correct classpath, like which extensions to include in the classpath and which version of a specific extension is "installed".

So the approach Kurt is taking seems the most practical to me in this moment.

rhuss:
@nicolaferraro ok, thanks for clarifying

rhuss:
@KurtStam there are various test and PMD failures

rhuss:
@KurtStam master is clean again. please rebase and push.

nicolaferraro:
@KurtStam I've pushed the fix for the class loader issue, so take it before rebasing.

Note that I've changed the LOADER_PATH env variable into LOADER_HOME, because I compute LOADER_PATH dynamically now. So you have to change that variable into your template (or reinstall the templates on Minishift) to take all changes.

I've tested it and I get a feedback from the Oracle driver when I try to create a new Connection.

nicolaferraro:
@KurtStam also the extension you upload must be generated directly by maven as here: https://github.com/syndesisio/syndesis-extensions/tree/master/syndesis-library-jdbc-driver. Put into `/lib`, reference it on pom.xml and take the output of `mvn install`.

I tell you this because the extension you were using seemed manually packed and spring-boot complained with it because nested jars need not to be compressed (probably it worked with the workaround on meta, but would fail in the integration).

KurtStam:
@nicolaferraro I take it you are talking about a fix for the integration pod, bc for the verifier/meta pod we can't really change the template on the fly. Anyway let's get this PR in and then we can work on optimizations next week. 

BTW I *did* use the maven project, but I got the Oracle jar referenced as 'system' so it was a jar to begin with.

nicolaferraro:
Actually the commit I pushed here yesterday was removing the workaround on meta and computed the loader path dynamically without changing the template on the fly...

Btw, I'm happy to merge this "as is" once test failures are resolved.

I'll open a separate PR with my changes later.

Issue #1780, opened 2018-03-02
jsondb's DELETE's using a lot of CPU in postgresql


zregvart:
## This is a...
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

I left the integration running for 12 hours and the number of records to jsondb increased to over 500K, with that many rows issuing DELETE SQL statements uses resources significantly.

Notice the CPU TIME of PID 538:

```shell
$ ps auxww
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
1000070+     1  0.0  0.1 171452  2776 ?        Ss   Mar01   0:06 postgres
1000070+   107  0.0  0.0  95020   780 ?        Ss   Mar01   0:00 postgres: logger process  
1000070+   109  0.0  3.3 171572 69072 ?        Ss   Mar01   0:07 postgres: checkpointer process  
1000070+   110  0.0  3.3 171452 68712 ?        Ss   Mar01   0:10 postgres: writer process  
1000070+   111  0.0  0.1 171452  2216 ?        Ss   Mar01   0:13 postgres: wal writer process  
1000070+   112  0.0  0.6 171872 12668 ?        Ss   Mar01   0:01 postgres: autovacuum launcher process  
1000070+   113  0.0  0.0  97272  1108 ?        Ss   Mar01   0:08 postgres: stats collector process  
1000070+   532  0.0  0.1 172268  3108 ?        Ss   Mar01   0:00 postgres: syndesis syndesis 172.17.0.2(47648) idle
1000070+   533  0.0  0.0 172268  1424 ?        Ss   Mar01   0:00 postgres: syndesis syndesis 172.17.0.2(47650) idle
1000070+   534  0.0  0.0 172268  1424 ?        Ss   Mar01   0:00 postgres: syndesis syndesis 172.17.0.2(47652) idle
1000070+   535  0.0  0.0 172268  1424 ?        Ss   Mar01   0:00 postgres: syndesis syndesis 172.17.0.2(47654) idle
1000070+   536  0.0  0.0 172268  1424 ?        Ss   Mar01   0:00 postgres: syndesis syndesis 172.17.0.2(47656) idle
1000070+   537  0.0  0.0 172268  1428 ?        Ss   Mar01   0:00 postgres: syndesis syndesis 172.17.0.2(47658) idle
[...]

PR #1805, opened 2018-03-02
Reduce the number of DB rows needed to ingest integration activity events.

chirino:
Should help improve performance and the impact activity logging causes on issue #1780


ruleset.xml
```diff
@@ -243,6 +243,9 @@
     <priority>3</priority>
   </rule>
   <rule ref="rulesets/java/codesize.xml/ModifiedCyclomaticComplexity">
+    <properties>
+      <property name="reportLevel" value="20" />
```
rhuss: Isn't 20 a bit large ? Actually everything over 10 is considered to be overly complex and should be split up. Instead of increasing the number to a senseless value, what's about switching this rule off completely (just to make it clear that we don't care) ?

> Complexity directly affects maintenance costs is determined by the number of decision points in a method plus one for the method entry. The decision points include ‘if’, ‘while’, ‘for’, and ‘case labels’ calls.
> Generally, numbers ranging from 1-4 denote low complexity, 5-7 denote moderate complexity, 8-10 denote high complexity, and 11+ is very high complexity. Modified complexity treats switch statements as a single decision point.

chirino: We had a recent meeting and agreed to bump adjust PMD rules to avoid slowing down contributions due to debatable PMD suggestions like this one.

rhuss: I would prefer to disable the rule altogether because 20 is completely off any reasonable value.

BTW, I don't get what its so hard to split a method into two, but that's a different kettle of fish.

Actually I would like to switch off PMD altogether, to avoid fruitless discussions like this, as we can't agree on any baseline anyway and just waste energy which we should save for saving issues later.

app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
```diff
@@ -131,13 +131,17 @@ private DeploymentData createDeploymentData(Integration integration, Integration
...
         String username = integrationDeployment.getUserId().orElseThrow(() -> new IllegalStateException("Couldn't find the user of the integration"));
 
+        String integrationId = integrationDeployment.getIntegrationId().orElseThrow(() -> new IllegalStateException("IntegrationDeployment should have an integrationId"));
+        String version = Integer.toString(integrationDeployment.getVersion());
         return DeploymentData.builder()
             .withVersion(integrationDeployment.getVersion())
-            .addLabel(OpenShiftService.INTEGRATION_ID_LABEL, Labels.sanitize(integrationDeployment.getIntegrationId().orElseThrow(() -> new IllegalStateException("IntegrationDeployment should have an integrationId"))))
-            .addLabel(OpenShiftService.DEPLOYMENT_ID_LABEL,  Integer.toString(integrationDeployment.getVersion()))
-            .addLabel(OpenShiftService.DEPLOYMENT_VERSION_LABEL, Integer.toString(integrationDeployment.getVersion()))
+            .addLabel(OpenShiftService.INTEGRATION_ID_LABEL, Labels.sanitize(integrationId))
```
rhuss: We don't need to sanitize on the id anymore because of #1869 , so we should just skip the call here.

app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
```diff
@@ -131,13 +131,17 @@ private DeploymentData createDeploymentData(Integration integration, Integration
...
+            .addLabel(OpenShiftService.DEPLOYMENT_ID_LABEL, version)
+            .addLabel(OpenShiftService.DEPLOYMENT_VERSION_LABEL, version)
             .addLabel(OpenShiftService.USERNAME_LABEL, Labels.sanitize(username))
             .addAnnotation(OpenShiftService.INTEGRATION_NAME_ANNOTATION, integration.getName())
+            .addAnnotation(OpenShiftService.INTEGRATION_ID_ANNOTATION, integrationId)
```
rhuss: could we agree to either use labels or annotations, but not both ?

app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
```diff
@@ -131,13 +131,17 @@ private DeploymentData createDeploymentData(Integration integration, Integration
...
+        String integrationId = integrationDeployment.getIntegrationId().orElseThrow(() -> new IllegalStateException("IntegrationDeployment should have an integrationId"));
+        String version = Integer.toString(integrationDeployment.getVersion());
         return DeploymentData.builder()
             .withVersion(integrationDeployment.getVersion())
-            .addLabel(OpenShiftService.INTEGRATION_ID_LABEL, Labels.sanitize(integrationDeployment.getIntegrationId().orElseThrow(() -> new IllegalStateException("IntegrationDeployment should have an integrationId"))))
-            .addLabel(OpenShiftService.DEPLOYMENT_ID_LABEL,  Integer.toString(integrationDeployment.getVersion()))
-            .addLabel(OpenShiftService.DEPLOYMENT_VERSION_LABEL, Integer.toString(integrationDeployment.getVersion()))
+            .addLabel(OpenShiftService.INTEGRATION_ID_LABEL, Labels.sanitize(integrationId))
+            .addLabel(OpenShiftService.DEPLOYMENT_ID_LABEL, version)
+            .addLabel(OpenShiftService.DEPLOYMENT_VERSION_LABEL, version)
```
rhuss: why twice the same value ?

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/PodLogMonitor.java
```diff
@@ -152,6 +159,37 @@ private void processLogStream(final InputStream is) throws IOException, Interrup
...
+            data.activity.setId(exchangeId);
+            data.activity.setAt(KeyGenerator.getKeyTimeMillis(exchangeId));
+            data.activity.setLogts(logts);
+            inflightActivities.put(exchangeId, data);
+        }
+        return data;
+    }
+
     @SuppressWarnings("PMD.CyclomaticComplexity")
     private void processLine(byte[] line) throws IOException {
```
zregvart: This is a really big method, couldn't we split it into smaller ones to make it a bit more explicit and easier to understand?

ruleset.xml
zregvart: @chirino can you explain why this rule is better with level 20 then with 10, couldn't we just then remove it if it's of no use?
I would think that we want to care about code quality and that this rule at the level of 10 indicates that the code needs to be refactored into smaller, easier to understand and maintain methods.

app/server/logging/jsondb/src/test/resources/logs-controller-db.json
```diff
@@ -2,87 +2,9 @@
...
-            "s4" : {
-              "-L38cZ5Vd1L876xV4vEU" : {
-                "duration" : 70724,
-                "logts" : "2018-01-12T21:22:02.068338027Z"
-              }
-            }
-          },
-          "ver" : "3"
-        }
+        "-L38cZ51d1L876xV4vEF" : "{\"id\":\"-L38cZ51d1L876xV4vEF\",\"logts\":\"2018-01-12T21:22:02.068338027Z\",\"at\":1516285084034,\"pod\":\"test-pod-x23x\",\"ver\":\"3\",\"status\":\"done\",\"failed\":false,\"steps\":[{\"id\":\"-L38cZ5Nd1L876xV4vEI\",\"at\":1516285084056,\"duration\":18087},{\"id\":\"-L38cZ5Od1L876xV4vEJ\",\"at\":1516285084057,\"duration\":494949},{\"id\":\"-L38cZ5Jd1L876xV4vEH\",\"at\":1516285084052,\"duration\":582977,\"messages\":[\"Hello World\"]}]}",
```
rhuss: dont think this will work with the latest change to the key generator.

btw, why is the value not an oject but an escaped JSON object as string. Any paraticular reason ?

rhuss: Sorry, thats just a test fixture independent of the key generator, so should be save. but maybe we change it nevertheless to the real key format.

PR #1813, opened 2018-03-05
Reuse image

iocanel:


app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java
```diff
@@ -310,8 +357,37 @@ private boolean removeSecret(String projectName) {
...
+                return next;
+            }
+            next = openShiftClient.deploymentConfigs().inNamespace(next.getMetadata().getNamespace()).withName(next.getMetadata().getName()).get();
+            Thread.sleep(5000);
+        }
+        throw SyndesisServerException.launderThrowable(new TimeoutException("Timed out waiting for build completion."));
+
+    }
+
+
```
zregvart: Would it make sense to use [Awaitility](http://www.awaitility.org/) for the wait* methods?

iocanel: For such a simple task, I am not sure if it worths bringing in an additional dependency (so far its test scoped).

But, I don't really mind.

zregvart: I'm definitely fine as it is. I was thinking if the complexity increases, for instance if we add backoff or we want to schedule it on another thread it would be nice to have this in place. But let's keep it simple for now.

oscerd:
Can we merge this?

app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
```diff
@@ -129,12 +132,14 @@ private DeploymentData createDeploymentData(Integration integration, Integration
...
         return DeploymentData.builder()
+            .withVersion(integrationDeployment.getVersion())
             .addLabel(OpenShiftService.INTEGRATION_ID_LABEL, Labels.sanitize(integrationDeployment.getIntegrationId().orElseThrow(() -> new IllegalStateException("IntegrationDeployment should have an integrationId"))))
             .addLabel(OpenShiftService.DEPLOYMENT_ID_LABEL,  Integer.toString(integrationDeployment.getVersion()))
+            .addLabel(OpenShiftService.DEPLOYMENT_VERSION_LABEL, Integer.toString(integrationDeployment.getVersion()))
             .addLabel(OpenShiftService.USERNAME_LABEL, Labels.sanitize(username))
             .addAnnotation(OpenShiftService.INTEGRATION_NAME_ANNOTATION, integration.getName())
-            .addAnnotation(OpenShiftService.INTEGRATION_ID_ANNOTATION, integrationDeployment.getIntegrationId().get())
-            .addAnnotation(OpenShiftService.DEPLOYMENT_VERSION_ANNOTATION, Integer.toString(integrationDeployment.getVersion()))
+            .addLabel(OpenShiftService.INTEGRATION_ID_LABEL, integrationDeployment.getIntegrationId().get())
```
oscerd: I believe this one causes #1841

PR #1814, opened 2018-03-05
refactor(connector): dropbox connector, lets use only the options we …

oscerd:
…really need


zregvart:
I guess we're not doing OAuth flow for Dropbox?

oscerd:
Actually the optional there doesn't exist in the camel component.

oscerd:
This come from an old PR

oscerd:
And that stuff is not needed/used.. At least now.

PR #1821, opened 2018-03-05
fix: CSS bugfix to force integration detail page span 100% vertically

deeleman:
This fix is a workaround for a regression detected on the integration detail view, where the body of the detail view does not resize vertically anymore to cover up the screen half


michael-coker:
@deeleman we could utilize the scrollable helper classes and make the details panel scrollable with the toolbar/title/tabs fixed. Updated `integration-detail.component.html` and `integration-detail.component.scss` in this branch https://github.com/michael-coker/syndesis/tree/integration-detail-scroll

PR #1829, opened 2018-03-05
Nicer error handling when fetching dynamic metadata

### LINKED ISSUE #1659: When exceeded salesforce API call limit, connection message "unexpected HTTP error"
There is this message, when I exceeded salesforce API calls limit: "An unexpected HTTP error occured. Please check stack strace".
I want to ask, whether there could be a better way to notify the user, that he exceeded the salesforce API call limit, because looking into stack trace (there is a typo at least) is quite cumbersome.

The screenshots:

salesforce_api_calls_exceeded
an_unexpected_salesforce_error

zregvart:
There is much more details in the commit messages.

In nutshell this adds hystrix when invoking the `syndesis-meta` for dynamic metadata. This way we can guarantee if an error or no response within 1.5s is  received from `syndesis-meta` that it can be properly handled.

The handling part uses a new mixin for the API responses, by which we can add `_meta` property to any API response. This `_meta` property consists (currently) of only two properties (type & message).

Adds an interceptor for HTTP responses in the UI, that checks if the `_meta` property is present in the response and displays a toast notification with the provided content.

Fixes #1659

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/MetadataCommand.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/MetadataCommand.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.server.endpoint.v1.handler.connection;
+
+import java.util.Map;
+
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.ClientResponseFilter;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.HttpHeaders;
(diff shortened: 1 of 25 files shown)


gashcrumb:
So I'm guess here that really we don't want to keep this interceptor but instead handle the response properly from wherever a given request originates in the code, right?

zregvart:
> So I'm guess here that really we don't want to keep this interceptor but instead handle the response properly from wherever a given request originates in the code, right?

That depends, the way it's proposed the server might or might not add that `_meta` property, and the it will be returned in addition to the usual response (mixed in). This allows other use cases, in fact this is very similar pattern to the `flash` scope usually used in webapps. To me it makes sense to handle this at the single point.

gashcrumb:
That seems a bit decoupled though from the actual page that's potentially going to hit a problem, doesn't it?  I guess I don't understand how this addresses the error that's displayed in #1659 exactly, as I suspect now the behavior will be that we still see that error but also now we'll get a notification in addition to that error with more information, right?

Not that I'm against having this, though as a kind of global interceptor I do have concerns that this could cause a notification storm, etc :-)

gashcrumb:
Ah, k, thanks for the details!  I think it should be fine then.

I suspect that really we're talking about client initiated requests, so if
there's a flood of alerts we're doing it wrong anyways ;-)

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/MetadataCommand.java
```diff
@@ -0,0 +1,90 @@
...
+import com.netflix.hystrix.HystrixCommand;
+import com.netflix.hystrix.HystrixCommandGroupKey;
+import com.netflix.hystrix.HystrixCommandProperties;
+import com.netflix.hystrix.HystrixThreadPoolProperties;
+
+class MetadataCommand extends HystrixCommand<DynamicActionMetadata> {
+
+    private static final int THREAD_COUNT = 3;
+
+    private static final int TIMEOUT = 1500;
```
lburgazzoli: Not a big issue but wondering if we can make those values configurable

zregvart:
@lburgazzoli I've added the ability to configure thread pool size and timeout, do you think this can be merged now?

PR #1882, opened 2018-03-07
Don't lowercase valid label names.

chirino:


app/common/util/src/main/java/io/syndesis/common/util/Labels.java
```diff
@@ -40,6 +42,11 @@ private Labels() {
      * @return      The sanitized string.
      */
     public static String sanitize(String name) {
+
+        if( name.length()<= MAXIMUM_NAME_LENGTH && VALID_LABEL_REGEX.matcher(name).matches() ) {
```
rhuss: maximum length is 63, not 100, according to https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

If I understood @iocanel right, we are striving to remove the need to sanitize id and version labels (but validating instead before, so fail fast if they can't be used).

Not sure whether we use labels for something different, so could be that we get rid of method eventually.

Otherwise I'm perfectly fine with this PR, thanks !

rhuss: oops, sorry. that was the name which can be at max 63 chars.

iocanel:
We shouldn't be sanitizing labels. Instead we should validate that they work.
So, I would completely close this one.

The rest of the @rhuss concersn are addressed in a PR I am currently working on.

rhuss:
Its #1885 which supersedes this PR (which still has test errors)

rhuss:
closing in favor #1885 which fix this issue differently

PR #1885, opened 2018-03-07
Fix broken labels

iocanel:


app/common/util/src/main/java/io/syndesis/common/util/Labels.java
```diff
@@ -15,17 +15,16 @@
  */
 package io.syndesis.common.util;
 
-import java.util.Locale;
 
 public final class Labels {
 
-    private static final String VALID_REGEX = "(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?";
+    private static final String VALID_REGEX = "(([A-Za-z0-9][-A-Za-z0-9_]*)?[A-Za-z0-9])?";
```
rhuss: I think . is ok according to https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set:

> Valid label values must be 63 characters or less and must be empty or begin and end with an alphanumeric character ([a-z0-9A-Z]) with dashes (-), underscores (_), dots (.), and alphanumerics between

'would call it `VALID_VALUE_REGEX` to not confuse it with the label name.

rhuss: minor nit: would use non-capturing grouping to be clear that is about grouping, not capturing (not important though)

app/common/util/src/main/java/io/syndesis/common/util/Labels.java
```diff
@@ -40,11 +39,11 @@ private Labels() {
      * @return      The sanitized string.
      */
     public static String sanitize(String name) {
-        return trim(name)
+        return trim(name
             .replaceAll(SPACE, DASH)
-            .toLowerCase(Locale.US)
             .chars()
             .filter(i -> !String.valueOf(i).matches(INVALID_CHARACTER_REGEX))
```
rhuss: `.` and `_` are both valid, too. Actually for better understanding I wouldnt use a double negation but sth like `i -> String.valueOf(i).matches(VALID_CHARACTER_REGEX)`. 

Using a regexp for every char is a quite expensive operation, btw, it might be better filter out with 

```
(c >= 'a' && c <= 'z') || 
(c >= '0' && c <= '9') || 
(c >= 'A' && c <= 'Z') || 
c == '.' || c == '-' || c == '_'
```

app/common/util/src/main/java/io/syndesis/common/util/Labels.java
```diff
@@ -53,30 +52,41 @@ public static String sanitize(String name) {
...
+        int length = name.length();
         if (length <= 1) {
             throw new IllegalStateException("Specified string:" + name + " cannot be sanitized.");
         }
 
-        String first = trimmed.substring(0, 1);
+        String first = name.substring(0, 1);
         if (!first.matches(SIDES_REGEX)) {
-            return trim(trimmed.substring(1));
+            return trim(name.substring(1));
```
rhuss: Clever recursive call, but a bit expensive considering on regexp matching for a single char (maybe the trick as above just checking whether the charactes is unequal of .,- or _ (all other chars have been already filtered out before) is simpler ?  (and maybe even a while loop instead of going into recursion ?)

app/common/util/src/main/java/io/syndesis/common/util/Names.java
```diff
@@ -16,36 +16,40 @@
...
     }
 
     /**
      * Sanitizes the specified name by applying the following rules:
-     * 1. Keep the first 100 characters.
-     * 2. Replace spaces with dashes.
-     * 3. Remove invalid characters.
+     * 1. Replace spaces with dashes.
+     * 2. Remove invalid characters.
+     * 3. Keep the first 64 characters.
```
rhuss: 63 actually

app/common/util/src/main/java/io/syndesis/common/util/Names.java
```diff
@@ -16,36 +16,40 @@
 package io.syndesis.common.util;
 
 import java.util.Locale;
+import static io.syndesis.common.util.Strings.truncate;
 
 public final class Names {
 
-    private static final String INVALID_CHARACTER_REGEX = "[^a-zA-Z0-9-]";
+    private static final String VALID_REGEX = "[-A-Za-z0-9\\.]+";
```
rhuss: no underscore ?

app/common/util/src/main/java/io/syndesis/common/util/Names.java
```diff
@@ -16,36 +16,40 @@
...
+    private static final String INVALID_CHARACTER_REGEX = "[^a-zA-Z0-9-\\.]";
```
rhuss: underscore missing

app/common/util/src/main/java/io/syndesis/common/util/Names.java
```diff
@@ -16,36 +16,40 @@
...
     public static String sanitize(String name) {
-        String trimmed = name.length() > MAXIMUM_NAME_LENGTH ? name.substring(0, MAXIMUM_NAME_LENGTH) : name;
-
-        return trimmed
+        return truncate(name
             .replaceAll(SPACE, DASH)
             .replaceAll(INVALID_CHARACTER_REGEX, BLANK)
             .toLowerCase(Locale.US)
             .chars()
             .filter(i -> !String.valueOf(i).matches(INVALID_CHARACTER_REGEX))
```
rhuss: same comment as for `Labels`: Suggest character matching instead of regexp for single chars (remember, its expensive to compile a pattern every time a fresh).

app/common/util/src/main/java/io/syndesis/common/util/Names.java
```diff
@@ -16,36 +16,40 @@
 package io.syndesis.common.util;
 
 import java.util.Locale;
+import static io.syndesis.common.util.Strings.truncate;
 
 public final class Names {
```
rhuss: What is actually the difference to the `Labels` class ? For me the sanitize() algorithm looks the same.

app/common/util/src/main/java/io/syndesis/common/util/Names.java
```diff
@@ -54,6 +58,17 @@ public static String sanitize(String name) {
...
-            .toString();
+            .toString(), MAXIMUM_NAME_LENGTH);
+    }
+
+
+    public static boolean isValid(String name) {
+        return name.matches(VALID_REGEX) && name.length() <= MAXIMUM_NAME_LENGTH;
+    }
+
+    public static void validate(String name) {
```
rhuss: Labels.validate() returns a string, but here its void

app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
```diff
@@ -252,11 +251,11 @@ private int countActiveIntegrationsOfSameUserAs(IntegrationDeployment deployment
      */
     private int countDeployments(IntegrationDeployment deployment) {
         Integration integration = deployment.getSpec();
-        String id = Labels.sanitize(integration.getId().orElseThrow(() -> new IllegalStateException("Couldn't find the id of the integration")));
+        String id = Labels.validate(integration.getId().orElseThrow(() -> new IllegalStateException("Couldn't find the id of the integration")));
         String username = deployment.getUserId().orElseThrow(() -> new IllegalStateException("Couldn't find the user of the integration"));
 
         Map<String, String> labels = new HashMap<>();
-        labels.put(OpenShiftService.USERNAME_LABEL, Labels.sanitize(username));
+        labels.put(OpenShiftService.USERNAME_LABEL, Labels.validate(username));
```
rhuss: the username, is this supposed to be always valid ? Not sure here ...

app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
```diff
@@ -272,15 +271,14 @@ private int countDeployments(IntegrationDeployment deployment) {
      */
     private boolean hasPublishedDeployments(IntegrationDeployment deployment) {
         Integration integration = deployment.getSpec();
-        String id = Labels.sanitize(integration.getId().orElseThrow(() -> new IllegalStateException("Couldn't find the id of the integration")));
+        String id = Labels.validate(integration.getId().orElseThrow(() -> new IllegalStateException("Couldn't find the id of the integration")));
         String version = String.valueOf(integration.getVersion());
 
         Map<String, String> labels = new HashMap<>();
-        labels.put(OpenShiftService.INTEGRATION_ID_LABEL, Labels.sanitize(id));
+        labels.put(OpenShiftService.INTEGRATION_ID_LABEL, Labels.validate(id));
```
rhuss: Isn't it already validaeted here ? See three lines above.

app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
```diff
@@ -272,15 +271,14 @@ private int countDeployments(IntegrationDeployment deployment) {
...
 
         return (int) openShiftService().getDeploymentsByLabel(labels)
             .stream()
-            .filter(d -> !id.equals(d.getMetadata().getLabels().get(OpenShiftService.INTEGRATION_ID_LABEL)))
```
rhuss: why is this not needed here anymore ?

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/PodLogMonitor.java
```diff
@@ -66,13 +66,13 @@ public PodLogMonitor(ActivityTrackingController logsController, Pod pod) throws
             throw new IOException("Could not determine the pod name");
         }
 
-        Map<String, String> annotations = pod.getMetadata().getAnnotations();
-        this.integrationId = annotations.get(OpenShiftService.INTEGRATION_ID_ANNOTATION);
+        Map<String, String> labels = pod.getMetadata().getLabels();
+        this.integrationId = labels.get(OpenShiftService.INTEGRATION_ID_ANNOTATION);
```
rhuss: `INTEGRATION_ID_ANNOTATION`, shouldn't this be `INTEGRATION_ID_LABEL` ? (and removing the constant about annotations ?)

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/PodLogMonitor.java
```diff
@@ -66,13 +66,13 @@ public PodLogMonitor(ActivityTrackingController logsController, Pod pod) throws
...
         if (this.integrationId == null) {
             throw new IOException("Could not determine the integration id that is being run on the pod: " + this.podName);
         }
 
-        this.deploymentVersion = annotations.get(OpenShiftService.DEPLOYMENT_VERSION_ANNOTATION);
+        this.deploymentVersion = labels.get(OpenShiftService.DEPLOYMENT_VERSION_ANNOTATION);
```
rhuss: same here. Constant is name `ANNOTATION`, but used for a label

rhuss: Actually I would remove the annotation constants in favor of the label constants.

app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftService.java
```diff
@@ -30,7 +30,6 @@
     String DEPLOYMENT_VERSION_ANNOTATION = "syndesis.io/deployment-version";
```
rhuss: Could we remove the duplicated annotations ?

app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
iocanel: because we are querying by label. No need to also filter.

PR #1889, opened 2018-03-07
chore: Move inner class of Meta to top-level

rhuss:
IDEA doesn't work with immutables defined on inner interfaces.
That's the only reason for this change.


rhuss:
@zregvart would this change be ok to you ? Otherwise you can build or run from IDEA anymore (leading to errors like `Error:(34, 11) java: package Value does not exist`)

Issue #887, opened 2018-01-02
Create mechanism for conditionally displaying form fields


seanforyou23:
We need a way to conditionally display form fields based on another fields value, and for it to be configuration-driven.

This is needed for the AMQ configure connection flow - https://github.com/syndesisio/syndesis/blob/master/ux/designs/amq/amq.md

There seems to be support for this in ng-dynamic-forms, but the forms.service.ts doesn't know about it yet - https://github.com/udos86/ng-dynamic-forms#related-form-controls.

deeleman:
> We need a way to conditionally display form fields based on another fields value

Angular template syntax will give you all what you need to perform this with no effort.

seanforyou23:
It's a good point deeleman. The making it configurable from json bit is what I'm still trying to think through. I could manually add the form controls and use template syntax to handle displaying the fields, but then the ordering becomes difficult to manage across the board.

Both you and gashcrumb had some good points about it today, so I'm still investigating all the helpful advice. Also, this is related to a general discussion about forms direction that patternfly-ng folks are beginning to have, I'll bring this up tomorrow and see what suggestions/ideas come up.

If nothing else I can always build these forms by hand until we get a solution we feel good about for the long run. Keep you posted!

gashcrumb:
I could see leveraging the form configuration syntax that ng-dynamic-forms supplies as something that's doable in the tp3 timeframe.  Pulling out ng-dynamic-forms and replacing it right now I think is not possible.  I guess the first question is, does this need to be sorted for tp3 or can it wait for tp4?

gashcrumb:
I think leveraging this support is the way to go -> https://github.com/udos86/ng-dynamic-forms#related-form-controls

as we can then add this to the form object definition by adding a `relation` property to a given form control, the form builder service can then add this relation property to the configuration object that's passed into the constructor for a given form control.

gashcrumb:
So basically in these spots:

* https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/common/forms.service.ts#L61-L66
* https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/common/forms.service.ts#L79-L89
* https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/common/forms.service.ts#L79-L89
* https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/common/forms.service.ts#L128-L145
we could add:

```typescript
   relation: field.relation,
```

[...]

PR #1910, opened 2018-03-08
Added Property relation to support conditional UI control of properties Backend feature required for #887

dhirajsb:


app/common/model/src/main/java/io/syndesis/common/model/connection/PropertyRelation.java
```diff
@@ -0,0 +1,50 @@
...
+@SuppressWarnings("immutables")
+public interface PropertyRelation extends Serializable {
+
+    class Builder extends ImmutablePropertyRelation.Builder {
+        // make ImmutablePropertyRelation.Builder accessible
+    }
+
+    @Value.Immutable
+    @JsonDeserialize(builder = PropertyRelationWhen.Builder.class)
+    interface PropertyRelationWhen {
```
rhuss: Can we avoid inner interfaces with immutables ? IDEA doesn't work with them. Alternatively, if you find a way how to compile this from within IDEA, the better ;-)

Actually, I have no idea what `PropertyRelationWhen` means. Could you elaborate a bit ? Actually, I would expect some date here, but is probably not.

app/common/model/src/main/java/io/syndesis/common/model/connection/PropertyRelation.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        String getId();
+
+        String getValue();
+    }
+
+    String getAction();
+
+    List<PropertyRelationWhen> getWhen();
```
rhuss: Should it be then `WhenList` ?

Really not sure about the naming. At least its not very intuitive.

app/common/model/src/main/java/io/syndesis/common/model/connection/PropertyRelation.java
rhuss: Hmm. not sure about the IDEA thing, but for `Meta` I had to extract an inner interface to make IDEA happy. But I see that there are also other inner interfaces used, so might be not an issue here.

But when keeping it, then I would remove teh `PropertyRelation` prefix, as `PropertyRelation.PropertyRelationWhen` is redundant.

app/integration/component-proxy/src/main/java/io/syndesis/integration/component/proxy/ComponentDefinition.java
```diff
@@ -77,6 +78,22 @@
...
+    @JsonDeserialize(builder = ImmutableRelationWhen.Builder.class)
+    interface RelationWhen {
+        String getId();
+        String getValue();
+    }
+
+    @Value.Immutable
+    @Value.Style(builder = "new")
+    @JsonDeserialize(builder = ImmutableRelation.Builder.class)
+    interface Relation {
```
rhuss: What's the difference between a `Relation` and a `PropertyRelation` ?

app/common/model/src/main/java/io/syndesis/common/model/connection/PropertyRelation.java
seanforyou23: I could be mistaken, but I believe PropertyRelationWhen may correspond to the structure found here - https://github.com/udos86/ng-dynamic-forms/#related-form-controls, which is something we need to help out the forms interactivity in the UI. Seeing that may make sense out of some of this :)

dhirajsb:
@rhuss or @lburgazzoli care to review this guys? Just in case I've done something obviously incorrect.

dhirajsb:
@rhuss just saw your comments.

app/common/model/src/main/java/io/syndesis/common/model/connection/PropertyRelation.java
dhirajsb: What @seanforyou23 said about what the properties mean. For the name, I'm ok with renaming the inner classes to remove the redundant PropertyRelation prefix.

app/common/model/src/main/java/io/syndesis/common/model/connection/PropertyRelation.java
dhirajsb: It's the way UI asked for it to be. Field names make more sense when viewing an instance data model though.

app/integration/component-proxy/src/main/java/io/syndesis/integration/component/proxy/ComponentDefinition.java
dhirajsb: Good question. We basically have duplicate models for `ComponentDefinition` and `ConnectorDescriptor`. I could rename `Relation` in Component definition to `PropertyRelation` to make it similar.

app/integration/component-proxy/src/main/java/io/syndesis/integration/component/proxy/ComponentDefinition.java
```diff
@@ -98,6 +115,7 @@
         Optional<String> getPrefix();
         Optional<String> getMultiValue();
         Optional<String> getEnumValues();
+        Optional<List<PropertyRelation>> getRelation();
```
lburgazzoli: Making a collection optional is ugly, getRelation should return an empty collection by default

PR #1914, opened 2018-03-08
chore(cli): Add Syndesis CLI

### LINKED ISSUE #1516: CLI Tool for JsonDB migration
This story is related to #1512 and it is the counterpart stored in the `syndesis-rest` image.
It is used by the upgrade script to perform the actual migration.

The migration step scripts should by contained also within the `syndesis-rest` image so that this image can be used as a self contained entity.

The challenge will be to access JsonDB from JavaScript and find a good way to transform the JsonDB Java model objects to JavaScript.

As an alternative one could also consider Groovy as transformation language.

zregvart:
Adds Syndesis CLI module for command line tool initially used to perform database migrations.

The CLI is built as a fat jar using Spring Boot it currently supports only `migrate` and could support more commands in the future easily.

This is the help output of the `migrate` command:

```shell
$ java -jar syndesis-cli.jar migrate -h
Usage: migrate [-h] [-d=<url>] [-e=<password>] [-p=<password>] [-t=<version>]
               [-u=<username>]
  -d, --url=<url>             JDBC url for the database
                              Example: jdbc:postgresql://localhost:5432/syndesis
  -e, --encryption-password=<password>
                              Database encryption password
  -p, --password=<password>   Password to authenticate against the database
  -t, --target=<version>      Target schema version number
  -u, --user=<username>       Username to authenticate against the database
  -h, --help                  Prints this help message and exits
```

For migrations new Spring application context is created with all required beans and configuration. This is done so that the CLI only bootstraps the required beans (like JDBC DataSource) when the command is executed.

Fixes #1516


rhuss:
Thanks @zregvart ! I will be busy this until the demo, but will jump on the PR review right after the retro.

zregvart:
I discovered asciinema so here is a recording of the tool:

<a href="https://asciinema.org/a/fZDGUuoglc3TMIgpZ28AJxl3f" target="_blank"></a>

zregvart:
@rhuss is this good to go?

app/cli/pom.xml
```diff
@@ -0,0 +1,228 @@
...
+    <dependency>
+      <groupId>org.jdbi</groupId>
+      <artifactId>jdbi</artifactId>
+      <version>${jdbi.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>info.picocli</groupId>
+      <artifactId>picocli</artifactId>
+      <version>0.9.8</version>
```
rhuss: According to the website,  2.3.0 is the latest version of picocli (https://github.com/remkop/picocli/releases/tag/v2.3.0)

Any reason to choose this particular version ? btw, do we need to adhere to a process wrt productisation when we add new libs ? (license looks good to me, its APL)

app/server/runtime/src/main/java/io/syndesis/server/runtime/DefaultMigrator.java
```diff
@@ -0,0 +1,75 @@
...
+            final String migrationScript = IOStreams.readText(resource.getInputStream());
+
+            LOG.info("Migrating to schema: {}", toVersion);
+            final ScriptEngine engine = new ScriptEngineManager().getEngineByName("nashorn");
+            engine.put("internal", map("jsondb", jsondb));
+
+            engine.eval(Resources.getResourceAsText("migrations/common.js"));
+            engine.eval(migrationScript);
+        } catch (IOException | ScriptException e) {
+            throw new SyndesisServerException(e);
```
rhuss: The exception thrown is much to generic IMO. It should contain much more context information (like the target version, maybe path to the upgrade script used ...) and the calling method should evaluate this to provide proper error handling.

I guess, rollbacks are out of scope here, right ? (so a rollback for a failed DB upgrade means a rollback of the whole database performed outside the scope of the command)

app/server/runtime/src/main/java/io/syndesis/server/runtime/Migrations.java
```diff
@@ -15,99 +15,72 @@
...
-            int from = Integer.parseInt(versionInDB);
-            int to = Integer.parseInt(getTargetVersion());
+            final int from = versionInDB;
+            final int to = getTargetVersion();
 
             // Apply per version migration scripts.
             for (int i = from; i < to; i++) {
-                int version = i + 1;
-                migrate(jsondb, version);
+                final int version = i + 1;
```
rhuss: why not 

```java
for (int version = from+1; version <= to; i++) {
/// .....

}
```

app/cli/src/main/resources/migration.yml
```diff
@@ -0,0 +1,28 @@
+#
```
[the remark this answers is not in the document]
zregvart: Each Command runs in it's own Spring ApplicationContext, by convention the name of the command is the name of the YAML file that configures that command. This allows us to have different configurations for different commands, think all settings that differ with environments, but also provides isolation for every command, so only one DataSource for instance is seen by the command.

app/server/runtime/src/main/java/io/syndesis/server/runtime/DefaultMigrator.java
```diff
@@ -0,0 +1,75 @@
...
+            final Resource resource = resourceLoader.getResource(path);
+            if (!resource.exists()) {
+                return;
+            }
+
+            final String migrationScript = IOStreams.readText(resource.getInputStream());
+
+            LOG.info("Migrating to schema: {}", toVersion);
+            final ScriptEngine engine = new ScriptEngineManager().getEngineByName("nashorn");
+            engine.put("internal", map("jsondb", jsondb));
```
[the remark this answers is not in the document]
zregvart: I guess we'll need to version JsonDB instances when we decide to break compatibility, until then I would just keep it as simple as possible and hope that the JsonDB interface doesn't change in an incompatible way.

app/cli/src/main/java/io/syndesis/cli/command/migrate/MigrateCommand.java
```diff
@@ -0,0 +1,115 @@
...
+            final String storedVersion = storedSettings.get(SCHEMA_VERSION_KEY);
+
+            final int version = storedVersion == null ? 0 : Integer.parseInt(storedVersion);
+
+            LOG.info("Current database schema version: {}", version);
+            if (version < targetVersion) {
+                LOG.info("Migrating to version: {}", targetVersion);
+            }
+
+            final JsonDB jsondb = context.getBean(JsonDB.class);
```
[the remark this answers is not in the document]
zregvart: Perhaps, but I'm not sold on it, it's a single dependency that is looked up and it wouldn't work from the constructor as the `@Option`s are not set at that point and we need the values for the environment of the ApplicationContext.

app/cli/src/main/java/io/syndesis/cli/command/migrate/MigrateCommand.java
```diff
@@ -0,0 +1,115 @@
...
+
+            final Migrator migrator = context.getBean(Migrator.class);
+
+            for (int i = version; i <= targetVersion; i++) {
+                migrator.migrate(jsondb, i);
```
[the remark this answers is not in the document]
zregvart: I've implemented global transactions in JsonDB, so we can rollback to a state before the migration took place.

app/server/runtime/src/main/java/io/syndesis/server/runtime/DefaultMigrator.java
zregvart: Not sure how can one handle this properly except to explode.

zregvart:
@rhuss I think we should be good to go here, do you mind having another look?

rhuss:
@zregvart yup, will do right now ...

app/server/jsondb/src/main/java/io/syndesis/server/jsondb/impl/SqlJsonDB.java
```diff
@@ -568,4 +574,47 @@ public DatabaseKind getDatabaseKind() {
...
+                final SqlJsonDB checkpointed = new SqlJsonDB(new DBI(() -> transacted), transactedBus);
+
+                boolean commited = false;
+                try {
+                    handler.accept(checkpointed);
+                    handle.commit();
+                    commited = true;
+                    transactedBus.commit();
+                } catch (@SuppressWarnings("PMD.AvoidCatchingGenericException") final RuntimeException e) {
+                    if (!commited) {
```
rhuss: super-nit-pick: "committed"

app/server/runtime/src/main/java/io/syndesis/server/runtime/DefaultMigrator.java
```diff
@@ -0,0 +1,84 @@
...
+
+            LOG.info("Migrating to schema: {}", toVersion);
+            final ScriptEngine engine = new ScriptEngineManager().getEngineByName("nashorn");
+            engine.put("internal", map("jsondb", utilized));
+
+            engine.eval(Resources.getResourceAsText("migrations/common.js"));
+            engine.eval(migrationScript);
+        } catch (IOException | ScriptException e) {
+            throw new SyndesisServerException(
+                "Unable to perform database migration to version " + toVersion + ", using migration script at: " + path, e);
```
rhuss: 👍

app/server/jsondb/src/main/java/io/syndesis/server/jsondb/impl/SqlJsonDB.java
```diff
@@ -568,4 +574,47 @@ public DatabaseKind getDatabaseKind() {
         return databaseKind;
     }
 
+    @Override
+    public void withGlobalTransaction(final Consumer<JsonDB> handler) {
```
rhuss: 👍

rhuss:
@zregvart btw, what do you think about extending pure-bot to allow a 'approval handover':

* Reviewer reviews, finds some minor issues which might or might not be fixed, but in general he's fine with the PR and wants to finish the review
* He does *not* approve, but just comment (but also don't request changes)
* He adds a label "pr/approved-with-handover"
* pure-bot monitors this label, and as soon when this label is *removed*, it applies a "pr/approved" label. This will then trigger a merge.
* That way the author can still adapt the PR (or not) and just removes the label when its done.

Currently, "pr/wip" could be used, too, but this is not so transparent for people looking at such a PR.

wdyt ?

zregvart:
Also we had that idea about making the review process synchronous, or to pair for reviews...

rhuss:
Yeah, we should really try sync, time-boxed reviews. Do you think we should just an informative label like "pr/sync-review-requested" ? 

hmm, I think this is just too much. Maybe we should just try to do it. IMO its the duty of the PR author to pick the reviewer and ask for a "review session". Even multiple PRs could be reviewed in a single session.

rhuss:
yeah, but before pushing changes I typically ask (another turnaround so). But yes, for typo fixes this shouldnt be necessary (sorry).

The problem with the 'requesting changes' GitHub feature is, that blocks merging as its a blocking check, and the reviewer has to unchange it again. If if the check wouldnt be mandatory (but should !), it shows up with a red cross (its a failed, not pending check).

So I really would go with the label route, or with a specific tag in the review comment (that would avoid the extra step of adding the label)

PR #1923, opened 2018-03-09
fix: NPE when parsed properties are null.

iocanel:


zregvart:
Fixes #1924

oscerd:
This one should fix also #1924

PR #1926, opened 2018-03-09
refactor(system-tests): upgrades and cleanup

zregvart:
This upgrades versions of Arquillian, Arquillian Cube, Fabric8
Kubernetes Client and cleans up Maven POMs. Path to OpenShift templates
was fixed and if $WORKSPACE is not defined relative path to `/install`
is used. Test for UI readiness is enabled again.

Tested locally.

### DIFF
--- a/app/common/pom.xml
+++ b/app/common/pom.xml
@@ -42,7 +42,6 @@
 
     <syndesis-connectors.version>${project.version}</syndesis-connectors.version>
 
-    <junit.version>4.12</junit.version>
     <mockwebserver.version>0.0.15</mockwebserver.version>
     <okhttp-eventsource.version>1.3.2</okhttp-eventsource.version>
     <postgresql.version>9.1-901-1.jdbc4</postgresql.version>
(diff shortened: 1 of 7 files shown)


zregvart:
> I would avoid to go with 3.1.8 of the client.

Thanks @oscerd for looking into this, what version should we use?

oscerd:
Probably I'd go with kube-client 3.1.4 and model 2.0.4. The kubernetes-model 2.0.8 has been released, but the kubernetes-client working with that version is still to be released. So probably I would choose to use 3.1.4 (a bit older) and kube-model 2.0.4, or client 3.1.8 and model 2.0.4. Camel actually is using client 3.1.8 with model 2.0.4 and it works fine, but in Syndesis we use the client a bit deeper and Openshift oriented.

oscerd:
Lets see if we noticed something weird, moving from Kube-client 2.6.3 to Kube 3.1.4 is a big jump. @iocanel should be the best to give feedback on this.

oscerd:
Lets see if we find problems. Eventually we can revert

oscerd:
Just rapidly tested and it seems to be ok

PR #1937, opened 2018-03-09
feat: add missing goals from plugin execution phases.

iocanel:
I tried importing the project to eclipse and it goes crazy over these. So let's add them to keep everyone happy.

### DIFF
--- a/app/extension/maven-plugin/pom.xml
+++ b/app/extension/maven-plugin/pom.xml
@@ -282,7 +282,7 @@
         <executions>
           <execution>
             <id>default-descriptor</id>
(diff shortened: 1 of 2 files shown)


zregvart:
Seems to fail in CI with:

```
[ERROR] Failed to parse plugin descriptor for io.syndesis.extension:extension-maven-plugin:1.3-SNAPSHOT (/workspace/app/extension/maven-plugin/target/extension-maven-plugin-1.3-SNAPSHOT.jar): No plugin descriptor found at META-INF/maven/plugin.xml -> [Help 1]
```

Isn't that exactly what this PR is supposed to address?

PR #1938, opened 2018-03-09
chore: checkstyle task about NewlineAtEndOfFile and FileTabcharacter

iocanel:
are now warnings.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -29,11 +29,13 @@
   <!-- http://checkstyle.sourceforge.net/config_misc.html -->
   <module name="NewlineAtEndOfFile">
     <property name="lineSeparator" value="lf"/>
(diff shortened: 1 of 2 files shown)


iocanel:
Regarding the removal of the `catch generic exception` its hands down the most annoying pmd check. 

In most cases we can live with laundering exceptions using `SyndesisServerException.launderThrowable()` so we should be able to use that without having to resort to PMD annotaitons or anything.

zregvart:
I can throw a good set of arguments for keeping those rules and argue to no end. I don't think that would be the best approach (will lead to bike shedding).

So let's do this: gather 1/3 of @syndesisio/backend, 6 or more at this moment to do a :+1: on PR and then merge. Let's see if we get consensus that way.

What do you think?

PR #1941, opened 2018-03-10
pre-commit hooks for fixing EOF lines and trailing whitespace

zregvart:
With this we can have the pre-commit[1] do the chore of fixing EOF line, that is that every file has a single empty line at the end of it, and removes any trailing whitespace on the line.

One only needs to install the pre-commit[2] and this is done automatically.

[1] https://pre-commit.com/
[2] https://pre-commit.com/#install


zregvart:
One can look at the [ignore whitespace changes](https://github.com/syndesisio/syndesis/pull/1941/files?w=1) to see that this (in addition to [adding the hook configuration](https://github.com/syndesisio/syndesis/pull/1941/commits/416c8d686a86e18ec0abbe9916b77255f72e1bca)) only changes the whitespaces.

paoloantinori:
My personal opinion:
I like git hooks a lot, as a way to improve workflows and to automate stuff that can be automated.

I think that any developer wasting any second of his/her time to even discuss if we need a new line at the end of a text file is the definition of bikeshedding.

zregvart:
Seems we have some consensus on this (4 :+1:), merging.

PR #1970, opened 2018-03-13
Templates, bumped verifier pvc demands to 1Gi

paoloantinori:
The requirement is related to an AWS limitation for EBS that would anyhow allocate 1Gi chunks, even if a pvc asks for less, wasting space and adding it from sight.

### DIFF
--- a/install/syndesis-dev.yml
+++ b/install/syndesis-dev.yml
@@ -10,7 +10,7 @@ metadata:
 #
 # Allow localhost refs: true
 # Use docker images: false
-# Atlasmap Tag:
+# Atlasmap Tag: 
 # Syndesis Tag: latest
 # Prometheus Tag: v2.1.0
 #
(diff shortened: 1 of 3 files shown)


install/syndesis-dev.yml
```diff
@@ -10,7 +10,7 @@ metadata:
 #
 # Allow localhost refs: true
 # Use docker images: false
-# Atlasmap Tag:
+# Atlasmap Tag:
```
igarashitm: I don't care at all if checkstyle doesn't care though, trailing spaces?

install/syndesis.yml
```diff
@@ -10,7 +10,7 @@ metadata:
 #
 # Allow localhost refs: false
 # Use docker images: false
-# Atlasmap Tag:
+# Atlasmap Tag:
```
igarashitm: ditto

paoloantinori:
checkstyle seems to not to care + I haven't touched those lines, not on purpose at least

install/syndesis-dev.yml
zregvart: I don't think checkstyle looks at YAML files, also :electric_plug: #1941

PR #1981, opened 2018-03-14
fix(server): try to use correct input/output shapes

### LINKED ISSUE #1967: Avoid returning incomplete data shapes for dynamic action
## This is a...

<pre><code>
[ ] Feature request
[ ] Bug report  
[x] Technical Debt
</code></pre>

## The problem
Derived from #1961, specifically  https://github.com/syndesisio/syndesis/issues/1961#issuecomment-372763621

Currently if action is not yet configured, `fetchMetadata` API is returning incomplete data shapes which is pre-defined in deployment.json, like `kind=json-schema` but specification is undefined. This data shape doesn't make any sense for dynamic action and confusing when UI handles it. We should clarify this behavior.

## Expected behavior
Not sure. One idea is to return `kind=none` if it's dynamic action but not yet have enough info to get dynamic data shapes.

zregvart:
**Builds on PR #1979, have a look at that one.**

With this if the result of metadata invocation for data shapes that
should contain specification/type, i.e. all that don't have their kind
defined as `none` or `any, and do not in fact contain those, for
instance if the meta data invocation fails or returns an erroneous data
shape, the resulting data shape will be set to `kind=any`.

Fixes #1967

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectionActionHandler.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectionActionHandler.java
@@ -53,6 +53,10 @@ import com.netflix.hystrix.HystrixInvokableInfo;
 
 @Api(value = "actions")
 public class ConnectionActionHandler {
+    public static final DataShape ANY_SHAPE = new DataShape.Builder().kind(DataShapeKinds.ANY).build();
+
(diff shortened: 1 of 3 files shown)


igarashitm:
LGTM :+1: 
But @zregvart and I chatted that we might want to wait for next dev cycle, this is touching something too deep to do in bug fix phase, and UI already handles problematic cases. So currently we don't see any issue without this, it's just a techdebt.

gashcrumb:
I'd mark it WIP if we don't want it merged then...

gashcrumb:
This checked out for me, thanks!

PR #2009, opened 2018-03-15
fix: Proper commit of pom.xml files

rhuss:
and don't revert back to snapshots as this is not pushed anyway.


rhuss:
Lets merge as it actually can really only be verified when doing a release.

PR #2025, opened 2018-03-15
chore: suppress circular reference warnings

### LINKED ISSUE #1445: Fix circular references
Currently when the UI is built we get a bunch of warnings about circular references.  These need to be fixed, they create noise and can create hard to solve runtime problems.

gashcrumb:
fixes #1445 (sorta in a doesn't fix it but hide it kinda way)

### DIFF
--- a/app/ui/.angular-cli.json
+++ b/app/ui/.angular-cli.json
@@ -68,6 +68,9 @@
     }
   },
   "defaults": {
-    "styleExt": "scss"
+    "styleExt": "scss",
+    "build": {
+          "showCircularDependencies": false
+        }
   }
 }


kahboom:
Tests failed

gashcrumb:
retest this please

zregvart:
I wonder if those tests failed because they compare the JSON too literally and the `pre-commit` modified the expected JSON files. Let me checkout this and see.

gashcrumb:
That's weird though this is just our angular CLI config, changing it shouldn't have an impact on anything that isn't UI related...

zregvart:
Yeah, that would make that test flaky, i.e. not pre-commit that's causing it to fail.

zregvart:
Oh you know what, I found out it downloads a new -SNAPSHOT of Swagger, after I removed my locally cached copy it fails for me also.

zregvart:
Let's see if #2029 passes the checks on CI and gets merged...

Issue #1916, opened 2018-03-08
Place "create new connection" UI elements inside a card-like component.


amysueg:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
The "+" icon, text, and "create connection" button appear to float on the page. I think we could enhance the user experience by collecting these items and surrounding them with a "card-like" design. That could imply to a user that once they create a new connection it will generate a new card. Could use a dimmed "white" background, a dotted outline, or whatever SJ comes up with for us to review. :-)

## Expected behavior
Same behaviors, just a visual design change.

## Screenshot
<img width="1043" alt="groupfloatingelements" src="https://user-images.githubusercontent.com/31700119/37180437-f8928ad2-22f6-11e8-8350-5583511d53dc.png">

@dongniwang @michael-coker @seanforyou23 fyi -- need to think about everywhere in the UI where this appears... dashboard?

deeleman:
> need to think about everywhere in the UI where this appears... dashboard?

If this UI element is meant to be rendered in different parts of the site, a reusable component is the way to go. Actually, ew should build a single, unique , configurable component for ALL “create” actions.

michael-coker:
@amysueg I think that's the only spot we have a "create" action in a list of cards like that. A couple of areas we could add another would be:

on the connections list page

![screen shot 2018-03-09 at 9 11 38 am](https://user-images.githubusercontent.com/35148959/37214477-d9dc6002-237a-11e8-8be1-0a9f2106a4b7.png)

And the connections list on the dashboard
![screen shot 2018-03-09 at 9 13 05 am](https://user-images.githubusercontent.com/35148959/37214496-ea66ebae-237a-11e8-9ff9-a09e8b1a2cb6.png)

michael-coker:
@deeleman I added a helper class in this branch - [empty-state-card](https://github.com/michael-coker/syndesis/tree/empty-state-card). Do you want me to create a PR for that or do you want to take a different approach for creating this component?

michael-coker:
@deeleman I created some reusable CSS for that element on this branch: [empty-state-card](https://github.com/michael-coker/syndesis/tree/empty-state-card). Is that what you had in mind or do you want to do something different?

deeleman:
Reusable CSS is a good start, but I would take this effort a bit further and build a reusable and configurable web component.

[...]

PR #2035, opened 2018-03-15
Create empty state card component

michael-coker:
fixes https://github.com/syndesisio/syndesis/issues/1916

I'm not sure how to pass the event from the card to the component it's in. You'll see that the `button` in the empty state card in [list.component.html](https://github.com/michael-coker/syndesis/blob/empty-state-card/app/ui/src/app/connections/list/list.component.html) doesn't work.


gashcrumb:
Looking good so far!

> I'm not sure how to pass the event from the card to the component it's in. You'll see that the button in the empty state card in list.component.html doesn't work.

For this your component needs an `@Output` to emit an event, like [here](https://angular.io/guide/component-interaction#parent-listens-for-child-event)

michael-coker:
@gashcrumb I took a stab but am not fully grokking the concept. Can you look at [this commit](https://github.com/syndesisio/syndesis/pull/2035/commits/eaba57a7642a1f4044e7a131bc7baef2da38c1ed) and tell me where I've gone wrong?

amysueg:
Hi, @michael-coker can you attach a screenshot of this? Or point me to a running instance?

michael-coker:
@amysueg here ya go!

screen shot 2018-03-15 at 2 56 31 pm

app/ui/src/app/connections/list/list.component.html
```diff
@@ -108,24 +108,17 @@ <h1>Register Applications</h1>
...
-              <div class="card-pf-info text-center">
-                <button class="btn btn-default" (click)="onSelect(undefined, $event)">Create Connection</button>
-              </div>
-            </div>
-          </div>
-        </div>
+        <syndesis-empty-state-card
+          [body]="cardBody"
+          [colClass]="'col-xs-12 col-sm-3 col-md-2'"
+          [topElement]="cardTopElement">
```
gashcrumb: So you'd bind the @Output here, i.e. something like `(newThing)="onSelect(undefined, $event)"`.  As I recall @deeleman mentioned that `on` gets added by angular so no need to prefix your output with `on`.

app/ui/src/app/connections/list/list.component.html
```diff
@@ -108,24 +108,17 @@ <h1>Register Applications</h1>
...
+          <ng-template #cardBody>
+            <button class="btn btn-default" (click)="onSelect(undefined, $event)">Create Connection</button>
```
gashcrumb: If all we need is to customize the button text, make the text an input instead of this whole template.  Then you can define this button in your component and have the `(click)` trigger the `onSelect` function you've defined in this component.

app/ui/src/app/connections/list/list.component.html
```diff
@@ -108,24 +108,17 @@ <h1>Register Applications</h1>
...
+          </ng-template>
+          <ng-template #cardTopElement>
+              <span class="card-pf-icon-large pficon pficon-add-circle-o"></span>
```
gashcrumb: I'd maybe have this as a default template so it's not required to pass in unless customization is needed.  I think across the app the `+` with the circle around it is the standard, so maybe it's worth just simplifying your component and just move this to your component's template?

app/ui/src/app/connections/list/list.component.html
deeleman: this is how the overall tempalte would become with all the above changes:

```html
<div class="empty-state-card card card-pf card-pf-view card-pf-view-select card-pf-view-single-select">
  <div class="card-pf-body">
    <div class="card-pf-top-element">
      <ng-content></ng-content>
    </div>
    <div class="card-pf-info text-center">
      <button class="btn btn-default" (click)="onClick.emit($event)">{{ buttonLabel }}</button>
    </div>
  </div>
</div>
```

Please note that `<ng-content></ng-content>` now features no selector, since we will expect to have just one transcluded slot in the component definition:

```html
<syndesis-empty-state-card class="col-xs-12 col-sm-3 col-md-2">
  <!-- The specific HTML for EACH empty state card -->
</syndesis-empty-state-card>
```

app/ui/src/app/connections/list/list.component.scss
```diff
@@ -2,20 +2,16 @@
...
-          box-shadow: none;
-        }
+    ::ng-deep .card {
+      width: 25%;
+
+      &.empty-state-card {
         .card-pf-icon-large {
           color: $color-pf-blue-300;
         }
       }
```
deeleman: this should be moved to the specific component stylesheet in order to be really DRY

michael-coker:
@gashcrumb @deeleman Thanks for all your feedback!! Would you mind checking my latest changes to this branch?

gashcrumb:
It's looking good!  I'll give @deeleman a chance to have a look too.

app/ui/src/app/common/empty-state-card.component.ts
```diff
@@ -0,0 +1,30 @@
+import {
+  Component
+} from '@angular/core';
+
```
deeleman: Why all these linebreaks here? We turn a simple import statement into 3 lines of code. This makes sense when the amount of imported tokens is so long than it does not fit in a single line without forcing the linter to complain.

app/ui/src/app/common/empty-state-card.component.ts
```diff
@@ -0,0 +1,30 @@
...
+  </div>
+  `,
+  styles: [`
+    :host { display: block; }
+    :host .card-pf, :host .card-pf:hover { box-shadow: none; }
+    :host .card-pf { background: transparent; padding-top: 1px; padding-bottom: 1px; }
+    :host .card-pf-view { border: 1px dashed #919191; }
+  `]
+})
+
```
deeleman: The overall shape of this CSS code turn it inot a good candidate to be promoted into its own SCSS file, linked to this component. That would allow you to use the Patternfly color variable fro `#919191` instead of that hardcoded reference in the last CSS statement.

app/ui/src/app/connections/list/list.component.html
```diff
@@ -108,24 +95,14 @@ <h1>Register Applications</h1>
...
+        <syndesis-empty-state-card class="col-md-6 col-lg-3" (click)="onSelect(undefined, $event)">
+          <div class="empty-state-card__header">
+            <span class="card-pf-icon-large pficon pficon-add-circle-o"></span>
           </div>
-        </div>
+          <div class="empty-state-card__body">
+            <button class="btn btn-default">Create Connection</button>
+          </div>
+        </syndesis-empty-state-card>
       </ng-container>
```
deeleman: You can spare the `<ng-container>` element here and apply the `*ngIf` directly on the `<syndesis-empty-state-card>` element, making the code more readable.

For the record, 50% of `<ng-content>` tags found in the code throughout the application are not necessary and hence redundant.

app/ui/src/app/connections/list/list.component.html
```diff
@@ -108,24 +95,14 @@ <h1>Register Applications</h1>
...
-              </div>
-
-              <!-- Card Description / Item Overview -->
-              <div class="card-pf-info text-center">
-                <button class="btn btn-default" (click)="onSelect(undefined, $event)">Create Connection</button>
-              </div>
-            </div>
+        <syndesis-empty-state-card class="col-md-6 col-lg-3" (click)="onSelect(undefined, $event)">
+          <div class="empty-state-card__header">
+            <span class="card-pf-icon-large pficon pficon-add-circle-o"></span>
```
deeleman: Do you believe the empty state card will always feature this large plus icon? If so, then I'd suggest to move it to the empty card element template instead. Otherwise this is perfectly fine.

sjcox-rh: @deeleman, yeah I think it's safe to say that the large plus icon will always be there.

michael-coker:
@gashcrumb @deeleman thanks again for the feedback, updated the branch with those changes.

gashcrumb:
@deeleman care to re-review when you get a chance?

gashcrumb:
Also this'll need a rebase

michael-coker:
rebased

deeleman:
Sure! However, that last commit only addresses one of the changes I requested in my previous comments. Will you address my other comments too @michael-coker? :)

app/ui/src/app/common/empty-state-card.component.ts
michael-coker: Can I just add the SCSS to `/app/common` adjacent to `empty-state-card.component.ts` or should I create a dir like `/app/common/empty-state-card/` and move the TS and SCSS there?

deeleman: As a rule of thumb, the convention is to create the SASS file right next to the component file, sharing its filename. In regards of where to save those files, perhaps moving those two it its own folder will help declutter an already cluttered folder as `common` ;)

michael-coker:
@deeleman d'oh! Sorry about that, I totally missed those and have no excuse :) Updated the branch.

app/ui/src/scss/utils/_helpers.scss
```diff
@@ -107,8 +108,31 @@
...
+      right: -2px;
+      border: 5px dashed #ffffff;
+    }
+
+    &,
+    &:hover {
+      box-shadow: none;
+    }
+  }
+}
```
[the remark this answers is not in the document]
deeleman: Perhaps you will want to move this piece of SCSS to `empty-state-card.component.scss`

app/ui/src/app/common/empty-state-card/index.ts
```diff
@@ -0,0 +1,22 @@
...
+          <ng-content select=".empty-state-card__body"></ng-content>
+        </div>
+      </div>
+    </div>
+  </div>
+  `,
+  styleUrls: ['empty-state-card.component.scss']
+})
+
+export class EmptyStateCardComponent {}
```
deeleman: I'm afraid you'll have to rename this file to `empty-state-card.component.ts` instead. The Angular convention is to name files by matching its class content in kebab-case.

We use `index.ts`  files to abstract the contents of folders to its consumers. This practice is better known as "barrel": https://angular.io/guide/glossary#barrel

app/ui/src/scss/utils/_helpers.scss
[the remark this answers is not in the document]
michael-coker: Good catch, unnecessary now :)

app/ui/src/app/common/empty-state-card/index.ts
michael-coker: Got it! Updated the branch.

michael-coker:
rebased

app/ui/src/app/connections/list/list.component.html
deeleman: I recall @sjcox-rh mentioned that this big plus icon should be always displayed in the empty card, so it's redundant and it should become part of `empty-state-card.component.ts` instead.

I'll post this as a comment, since I see 2 alternatives here:

- Correct this and send a final commit (preferred)
- Merge this and iterate on it again in the future (might never happen LOL)

Pick one ;)

deeleman: If you go for the latter, feel free to dismiss my review and let's get this merged straight away

michael-coker: thanks @deelman, corrected and updated the branch.

PR #2040, opened 2018-03-16
refactor(server): use proper property name

### LINKED ISSUE #750: syndesis-connector.version should be replaced by syndesis.version in our poms
Now that we are in a monorepo.

zregvart:
This replaces `syndesis-connectors.version` for `project.version` or
`syndesis.version` where appropriate.

Fixes #750


zregvart:
I want to test this a bit (my minishift setup is hosed atm), so I created a PR to gather some early feedback. That's why the `status/wip` label.

zregvart:
I've tested this and it seems to work.

Issue #1892, opened 2018-03-08
Provide external database connectivity capabilities and tutorial


honghuac:
## This is a...

<pre><code>
[X] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X] Documentation issue or request
</code></pre>

Regarding use cases involving external databases, provide database connectivity features which are easy to use as well as a tutorial that explains the use of these features.

Currently, how can a Citizen Developer utilize the [SQL connector](https://github.com/syndesisio/syndesis/tree/master/app/connector/sql) in an integration for multiple database connections? Please explain / demonstrate.

KurtStam:
You'd need a different DB connector for each database you'd want to connect to.

TovaCohen:
The tutorial that provides instructions for obtaining data from Salesforce and then saving it in a database is here: 
https://access.redhat.com/documentation/en-us/red_hat_jboss_fuse/7.0-tp/html/ignite_sample_integration_tutorials/salesforce-to-db

The general doc for connecting to a database is here:
https://access.redhat.com/documentation/en-us/red_hat_jboss_fuse/7.0-tp/html/integrating_applications_with_ignite/connecting-to-applications#connecting-to-databases 

That's all we have so far. For TP4 and GA there will be additional doc. 
Can you tell me more about this use case:
> How can a Citizen Developer utilize the SQL connector in an integration for multiple database connections? 
I can write better user doc for you if I know more about what you want to do. Why do you want to connect to multiple databases in an integration? Can you give me an example?
Thanks!

KurtStam:
@TovaCohen maybe we can add a sentence the 'general doc' somewhere that says something along the lines of: "You will need a to create a different DBConnection for each Database you want to connect to in your integration".

TovaCohen:
Sure. The "Connecting to SQL database" section now has this introduction:

> Connecting to SQL databases
> 
> In an integration, you can connect to any of the following types of databases:
> 
> Apache Derby
> 
> MySQL
> 
> PostgreSQL
> 
> You create a separate connection for each database that you want to connect to in an integration. You can use the same connection in multiple integrations.
> 
> A database connection operates on a table that you specify or invokes a stored procedure that you specify. The database table or the stored procedure must exist when an integration connects to the database.

Let me know if anything needs to change.

KurtStam:
@TovaCohen Looks good to me.

PR #2042, opened 2018-03-16
fix: (#1892) Unpublishing is now displayed as a status when unpublishing an integration.

iocanel:
We no long block until integration is scaled down. We set the status to pending and return. Followup calls to the Unpublish handler will determine when the integration is actually unpublished and set the correct status then. This allows the "Unpublishing" text to appear in the ui.

### DIFF
--- a/app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java
+++ b/app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java
@@ -108,7 +108,7 @@ public class OpenShiftServiceImpl implements OpenShiftService {
     public void scale(String name, Map<String, String> labels, int desiredReplicas, long amount, TimeUnit timeUnit) throws InterruptedException {
         String sName = openshiftName(name);
 
-        DeploymentConfig deploymentConfig = getDeploymentsByLabel(labels)
+        getDeploymentsByLabel(labels)
             .stream()
             .filter(d -> d.getMetadata().getName().equals(sName))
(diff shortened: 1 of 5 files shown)


app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java
```diff
@@ -105,20 +105,15 @@ public boolean exists(String name) {
...
+    public void scale(String name, Map<String, String> labels, int desiredReplicas, long amount, TimeUnit timeUnit) throws InterruptedException {
         String sName = openshiftName(name);
 
-        DeploymentConfig deploymentConfig = getDeploymentsByLabel(labels)
+        getDeploymentsByLabel(labels)
             .stream()
             .filter(d -> d.getMetadata().getName().equals(sName))
             .map(d -> new DeploymentConfigBuilder(d).editSpec().withReplicas(desiredReplicas).endSpec().build())
             .map(d -> openShiftClient.deploymentConfigs().createOrReplace(d))
             .findAny().orElse(null);
```
zregvart: perhaps this could be refactored as:
```java
.findAny()
map(d -> openShiftClient.deploymentConfigs().createOrReplace(d))

**Comment by rhuss** [comment]
@iocanel sorry, there's a PMD violation:

```
[INFO] --- maven-pmd-plugin:3.8:check (default) @ server-openshift ---
[INFO] PMD Failure: io.syndesis.server.openshift.OpenShiftServiceImpl:336 Rule:UnusedPrivateMethod Priority:3 Avoid unused private methods such as 'waitForDeployment(DeploymentConfig,long,TimeUnit)'..
```

PR #2043, opened 2018-03-16
fix(amq-connection): tweak forms to avoid double negative

seanforyou23:
This PR applies some tweaks to the amq/amqp connection configuration to improve the language around setting or skipping a certificate check.

Outstanding question - should the amqp connection also use the same description for the type of broker/client certificate that's used?

@TovaCohen @dongniwang let me know if you see any issues!

switch language around so that amq connection config flow is more intuitive

### DIFF
--- a/app/connector/activemq/src/main/resources/META-INF/syndesis/connector/activemq.json
+++ b/app/connector/activemq/src/main/resources/META-INF/syndesis/connector/activemq.json
@@ -93,7 +93,7 @@
       "defaultValue": "false",
       "deprecated": false,
       "secret": false,
-      "labelHint": "For convenience in only development environments. Ensure certificate checks are enabled for secure production environments.",
+      "labelHint": "Ensure certificate checks are enabled for secure production environments. Disable for convenience in only development environments.",
       "order": "5"
     },
     "brokerCertificate": {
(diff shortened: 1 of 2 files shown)


app/connector/activemq/src/main/resources/META-INF/syndesis/connector/activemq.json
```diff
@@ -93,7 +93,7 @@
       "defaultValue": "false",
       "deprecated": false,
       "secret": false,
-      "labelHint": "For convenience in only development environments. Do not set for secure production environments.",
+      "labelHint": "For convenience in only development environments. Ensure certificate checks are enabled for secure production environments.",
```
TovaCohen: This has the potential to be misleading because the field name and the hint imply that checking certificates is for convenience in only development environments. This is what the user will see: 
Check Certificates
For convenience in only development environments.

So I think the hint has to be: "Ensure certificate checks are enabled for secure production environments. Disable for convenience in only development environments."

app/connector/amqp/src/main/resources/META-INF/syndesis/connector/amqp.json
```diff
@@ -93,7 +93,7 @@
       "defaultValue" : "false",
       "deprecated" : false,
       "secret" : false,
-      "labelHint" : "For convenience in only development environments. Do not set for secure production environments.",
+      "labelHint" : "For convenience in only development environments. Ensure certificate checks are enabled for secure production environments.",
```
TovaCohen: Same comment as above.

app/connector/activemq/src/main/resources/META-INF/syndesis/connector/activemq.json
seanforyou23: Good call, I'll make this change now!

Issue #1817, opened 2018-03-05
API Connector Form: Order and Spacing


rhuss:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

The spacing in the API connectors form is inconsistent, as well as the order

By convention, you normally enter the user first, then the password. not vice versa:

## Expected behavior

"user" should come before "password", spacing between lines should be the same everywhere.

## Screenshot

![image](https://user-images.githubusercontent.com/99080/36749094-0b303ca6-1bfa-11e8-821e-1171a93609c1.png)

Related to #1709

michael-coker:
@rhuss can you provide the swagger file that enables those fields to be shown? I tried the swagger.json from my todo app and don't see those fields. https://photos.app.goo.gl/QU1qMlsA6nyZ96gv1

zregvart:
@michael-coker looks like that's the TODO app Swagger file if you have a running minishift instance it should be available at `https://todo-syndesis.$(minishift ip).nip.io/swagger.json`. In case you can't access that I've created a gist with at https://gist.github.com/zregvart/af1f18e05d3911e2fa0779a3204509be

michael-coker:
@zregvart thanks, but that's the file I was using. I tried yours and got the same output in the screenshots from my previous comment. "HTTP Basic Authentication" is a radio button, not a select box/dropdown, and there are no form fields below it.

![screen shot 2018-03-08 at 4 29 49 pm](https://user-images.githubusercontent.com/35148959/37180234-f1f763cc-22ed-11e8-9077-2af6f42ed7f8.png)

zregvart:
@michael-coker perhaps it's this older version: https://github.com/syndesisio/todo-example/blob/1ff5cf90a9e1c6c60cd3701b3d077844527aa44a/swagger.json

michael-coker:
@zregvart same thing, still not seeing the form fields.

deeleman:
You’re looking into different sections of the site, I’m afraid

michael-coker:
@deeleman ah! Found it now, thanks!

michael-coker:
[...]

PR #2044, opened 2018-03-16
Hide syndesis-form-control rows that have a hidden input field, remove padding from read-only input/textarea

michael-coker:
Fixes https://github.com/syndesisio/syndesis/issues/1817 but we still need to re-order the form fields so that a username row comes before the password row.

cc @seanforyou23

### DIFF
--- a/app/ui/src/app/common/ui-patternfly/syndesis-form-control.component.html
+++ b/app/ui/src/app/common/ui-patternfly/syndesis-form-control.component.html
@@ -1,7 +1,7 @@
 <div [class.form-group]="formGroupClass"
      [class.has-error]="showErrorMessages"
      [formGroup]="group"
-     [hidden]="model && model.inputType === 'hidden'"
+     [hidden]="model && model['inputType'] === 'hidden'"
      [ngClass]="[elementContainerClass, gridContainerClass]">
 
   <label *ngIf="type !== 2 && model.label !== null"
--- a/app/ui/src/app/common/ui-patternfly/syndesis-form-control.component.html
+++ b/app/ui/src/app/common/ui-patternfly/syndesis-form-control.component.html
@@ -1,7 +1,7 @@
 <div [class.form-group]="formGroupClass"
      [class.has-error]="showErrorMessages"
      [formGroup]="group"
-     [ngClass]="[elementContainerClass, gridContainerClass, model.inputType == 'hidden' ? 'hidden' : '']">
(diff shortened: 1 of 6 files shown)


gashcrumb:
@michael-coker wonder if you need to rebase this branch against master and force push?

also ordering would be done via the connector configuration, I think @seanforyou23 took care of those cases already...

app/ui/src/app/common/ui-patternfly/syndesis-form-control.component.html
```diff
@@ -1,7 +1,7 @@
 <div [class.form-group]="formGroupClass"
      [class.has-error]="showErrorMessages"
      [formGroup]="group"
-     [ngClass]="[elementContainerClass, gridContainerClass]">
+     [ngClass]="[elementContainerClass, gridContainerClass, model?.inputType == 'hidden' ? 'hidden' : '']">
```
deeleman: I do not know how the other classNames operate here and what that `hidden` classNames doe,s but wether this component is styled as `inline`, you could use the more native-ish property `hidden`, like:

```html
 <div [class.form-group]="formGroupClass"
      [class.has-error]="showErrorMessages"
      [formGroup]="group"
      [hidden]="model?.inputType == 'hidden'">
      [ngClass]="[elementContainerClass, gridContainerClass">
```

michael-coker: Do you know why it's failing CircleCI? Looks like the error is

> ERROR in src/app/common/ui-patternfly/syndesis-form-control.component.html(4,6): : Property 'inputType' does not exist on type 'DynamicFormControlModel'

deeleman: `DynamicFormControlModel` is a type native to the `ng-dynamic-forms/core` dependency. Perhaps there's been a bump to a new version where such property is not part of the `DynamicFormControlModel` class interface. Please double check wether you can see that property showing up as a class option in your IDE intellisense selectbox.

If not, please liaise with our `ng-dynamic-forms` expert @seanforyou23 since he might provide further details on what happened here. ;)

michael-coker: @seanforyou23 do you know what's going on? `model.inputType` is showing up as an option in my IDE. 

screen shot 2018-03-16 at 4 06 01 pm

gashcrumb:
Wow, great when it works locally and fails on CI :-)

PR #2048, opened 2018-03-16
fix: scrolling fixed for Integration activity steps logs output

deeleman:


deeleman:
@michael-coker What OS/browser combo are you using to test?

michael-coker:
Mac/Chrome

deeleman:
That's odd, I had no horizontal scrolling in my env. Probs you're spotting a non-wrappeable loooong line, where `word-wrap: break-word;` has no effect. In any event you're right, that `overflow` there is redundant. Thanks for noticing.

deeleman:
I see. Now it's fixed anyways. :)

Issue #1856, opened 2018-03-06
[monitoring] The Integration Activity Steps sub view should trigger a raw logs modal per step


deeleman:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
With a detailed steps activity sub view per Integration activity entry at the Integration detail view (as defined by #1855 ), we need a way to access, view and copy the raw logs for each step log so we can use those elsewhere.

## Expected behavior
We will want to append an additional action button at the right of each row in the Steps activity sub table (as depicted in the screenshot below) which will trigger a modal window wrapping the raw log with copy-to-clipboard functionality.

## Screenshot
![captura de pantalla 2018-03-06 a las 20 57 37](https://user-images.githubusercontent.com/1104146/37055081-0d6e5796-2181-11e8-93ce-8f07f6707cdb.png)

## API Endpoints and Schemas

TBD @chirino Could you fill this out? Thx

## Tasks involved / Steps to Reproduce

1. Create and deploy an integration. Wait for its `published` status.
2. Go to the integration detail view and click on activity
3. On the activity rows displayed (if any), click on the arrow on the left.
4. Check that a sub view contained tabular data matching the design above is rendered
5. Click on the `View` button where available to see the raw log displayed in a modal window. The logs should be wrapped by a scrollable div. Two buttons reading `Copy` and `Close` should allow for copying the modal contents to the clipboard and closing the modal popup, respectively.
6. Click on the `Copy` button. Its label should change to `Copied!` and pasting the clipboard contents in any text file should yield the raw log we just saw at the modal, in its entirety.

deeleman:
@chirino Perhaps we should remove the `target/tp4` label for this one. Do you agree?

chirino:
Do we really need this?  Are we not already showing the log step output or failure in the table?

deeleman:
@chirino Who are you asking to? This is not my call I just formalized the task to give it an actionable context. :)

Perhaps you should ping @kcbabo

kcbabo:
This looks awesome!  The minimal requirement we had was that log output (including detailed error information if available) was visible in the activity viewer.  If we already have that in place today, then I would say this particular issue can be pushed out of GA scope.

I will say that wha is proposed here looks great from an end user standpoint and I'm sure this would be very useful!  Would love to see this in the next release.

[...]

PR #2049, opened 2018-03-16
feat: Add “View Log in Openshift” link to activities page.

chirino:


chirino:
This is part of #1856

chirino:
I'm going to hand over this PR for @rhuss to complete.  I attempted to add a required OPENSHIFT_CONSOLE_URL template parameter.  But perhaps thats not the right approach.

install/generator/03-syndesis-ui.yml.mustache
```diff
@@ -105,6 +105,7 @@
         "apiBase": "https://${ROUTE_HOSTNAME}",
         "apiEndpoint": "/api/v1",
         "title": "Syndesis",
+        "openshiftConsoleProjectURL": "${OPENSHIFT_CONSOLE_URL}/project/${OPENSHIFT_PROJECT}",
```
rhuss: @chirino If we put in two arguments `consoleUrl` and `project` which we individually populate with `${OPENSHIFT_CONSOLE_URL}` and `${OPENSHIFT_PROJECT}1`, then in the UI we could combine both, but only when `consoleUrl` is set, then we could make `${OPENSHIFT_CONSOLE_URL}` optional (so that the link doesnt show up when the base url is not set).

Let me update the PR to adapt to this, i think this is a good compromise between avoiding mandatory parameter bloat and adding this link. You have the choice when installing the templates.

rhuss:
@chirino can't currently test because of #2065 doesn't show me the activity tab. will continue as soon as this is resolved.

rhuss:
@deeleman @gashcrumb I have an optional configuration in `config.json`, which, when not present looks like:

```json
 {
      "apiBase": "https://syndesis.192.168.64.57.nip.io",
      "apiEndpoint": "/api/v1",
      "title": "Syndesis",
      "consoleUrl": "",
      ....
      },
```

An empty string in this case. When using this in this PR and checking for `consoleUrl != null` then I get this error in the console:

image

My question: How can optional values (which can be either absent or whose values are empty) be queried without throwing an error ? 

Actually, as the value is there but just empty I even consider this error message as wrong. The configService should just return an empty string imo.

app/ui/src/app/integration/integration_detail/integration-activity/integration-activity.component.ts
```diff
@@ -97,6 +99,17 @@ export class IntegrationActivityComponent implements OnInit {
     this.showPagination = (this.allActivities.length > this.paginationConfig.pageSize);
     this.paginationConfig.totalItems = this.allActivities.length;
     this.paginationConfig.pageNumber = 1;
+
+    this.openshiftConsoleURL = null;
+    if (this.allActivities.length > 0) {
+      const base = this.configService.getSettings('consoleUrl');
```
rhuss: @deeleman @gashcrumb that's where this setting is queried (it's configured as an empty string, non-null).

gashcrumb:
Supply a default value as a second argument like `null` or `false` to avoid
that exception, I think really we should just have the config service
return a falsey/undefined value instead of throwing an exception like
this.  If you wouldn't mind raising an issue I'd be happy to sort that
tomorrow.

app/ui/src/app/integration/integration_detail/integration-activity/integration-activity.component.ts
rhuss: Guess the configService is used wrongly here anyway (sorry, took this over blindly from the original PR). 
Going to fix this.

app/ui/src/app/config.service.ts
```diff
@@ -62,12 +62,12 @@ export class ConfigService {
       return this.settingsRepository;
     }
 
-    if (!this.settingsRepository[group]) {
+    if (this.settingsRepository[group] === undefined) {
```
rhuss: @syndesisio/ui-api does this change looks leggit to you ? It just allows empty values, too and only croaks when no such setting provides (as the error message says anyway)

rhuss:
@chirino I'm good with the PR, tested it with and without providing a console URL as template parameter.

@gashcrumb @deeleman if the minor UI changes are ok to you, then we could merge this PR.

rhuss:
@syndesisio/uxd Actually this PR places a link to the OpenShift console on the "Activity" page of an integration like this:

image

Does this look ok from a UX POV ?

app/ui/src/app/config.service.ts
rhuss: @gashcrumb a set, empty string, shouldn't it be possible, too ? That would be much easier when using optional openshift template parameters which are empty by default. Don't really would use a 'fake' URL "false" as default value as this shows up in the OpenShift UI as 'false' in the inputfields.

Actually, it would be really awesome to allow defined, but empty values for interacting with the way how we intialize the configuration (so not only for this PR but for any other optional template parameter ending up in config.json)

gashcrumb:
Tbh I'd be fine with removing the exception altogether and just returning
undefined/null...

gashcrumb:
Looks like this guy needs a rebase...

chirino:
LGTM.

sjcox-rh:
@rhuss, yup good to go from UX perspective. Thanks!

app/ui/src/app/integration/integration_detail/integration_activity/integration-activity.component.ts
```diff
@@ -97,6 +99,17 @@ export class IntegrationActivityComponent implements OnInit {
     this.showPagination = (this.allActivities.length > this.paginationConfig.pageSize);
     this.paginationConfig.totalItems = this.allActivities.length;
     this.paginationConfig.pageNumber = 1;
+
+    this.openshiftConsoleURL = null;
+    if (this.allActivities.length > 0) {
+      const base = this.configService.getSettings('consoleUrl');
+      const pod = this.allActivities[0].pod;
```
gashcrumb: Why not just expose the pod ID via the integration object?  This seems a smell...

app/ui/src/app/integration/integration_detail/integration_activity/integration-activity.component.ts
```diff
@@ -97,6 +99,17 @@ export class IntegrationActivityComponent implements OnInit {
...
+      const project = this.configService.getSettings('project');
```
gashcrumb: Same goes with project, would it make more sense to keep that stored in the integration object rather than baking it into the config.json?

app/ui/src/app/integration/integration_detail/integration_activity/integration-activity.component.ts
rhuss: wasn't my code, so have to dig a bit into it. From where do I have access to the pod from the integration object ? I can't see a connection to the pod id from the IntegrationDeployment either (which I probably would have to fetch first then, too)

For me it seems like that only Activity has a reference to a pod. Of course it could be possible that this pod is not alive anymore, but as discussed with @chirino and @kcbabo  it seems to be acceptable that the generated link might be a stale one leading to 404 (e.g. when the pod has been replaced or the integration is not running anymore) when selected.

app/ui/src/app/integration/integration_detail/integration_activity/integration-activity.component.ts
rhuss: Integrations can be exported/imported into different projects. Also, from an architecturale POV an integration does not need know about the OpenShift details it is  running in. All integration runtimes also share the same OpenShift project, so an integration is not the proper place to store this info. 

Of course this information could (and imo should) come from the server, too (as, btw, all of config.json, not sure why there is more configuration than the connection to the backend stored in config.json as everything else is probably better managed on the server side. even branding).

But for now thats the simplest and easiest way without querying OpenShift itself or changing the way how configuration information reaches the frontend.

app/ui/src/app/integration/integration_detail/integration_activity/integration-activity.component.ts
gashcrumb: yeah, that's why I'm saying this is a smell :-)  The only reference to a pod is in what is effectively a log statement?  If we need the pod ID then we should make it available to the integration object.

app/ui/src/app/integration/integration_detail/integration_activity/integration-activity.component.ts
gashcrumb: Actually, instead of calculating the URL here in the frontend which requires exposing random bits of backend information, why don't we just have a 'logUrl' field in the integration object and just do this URL building at the backend?  This current solution feels super-brittle, if we merge it as-is we're gonna need to create issue(s) to go back in and fix it the right way.

app/ui/src/app/integration/integration_detail/integration_activity/integration-activity.component.ts
chirino: An integration runs on multiple pods over time.  An integration object only changes when a user edits it.  It would be a smell to change it due to runtime state.

app/ui/src/app/integration/integration_detail/integration_activity/integration-activity.component.ts
chirino: A syndesis install right now is fixed to a single project.  An integration can be exported between enviroments.  The pod id, is can vary over time within a single environment.  The base console URL is variable between environments.  We could do the computation for you with a new rest API, but that would have the same brittle qualities as doing it in the frontend.

gashcrumb: Suppose maybe the best option is a GET that does a 302 redirect to openshift console page, then the anchor link would be like `<a href="/v1/integrations/{id}/consoleLog">` then no need to expose any kind of openshift-y stuff to the frontend.

rhuss: that sounds valid to me, and I agree we should not put backend config to the client. @chirino any idea if this is hard to implement ? I guess its only about getting the pod id as the base URL and project can be injected in the templates as well to the server config.

@gashcrumb the only problem is that this URL should only appear _if_ the consoleUrl is configured. So we would need another endpoint to query this before rendering the link (or not).

chirino: gashcrumb that might not work in the future.. the UI may need to add #anchor bits to the URL which I'm not sure work with a redirect.  To me the seem squarely a Syndesis UI to OCP Console UI concern.  Not sure why the backend APIs needs to be burdened with UI concerns.

gashcrumb: Yes, a document hash can be used as part of a 302 redirect, the hash isn't sent to the server.  Actually the openshift console uses this feature to pass along it's auth token to the java console to avoid transmitting the auth token.

I'm just giving options.  The current proposed code means you can potentially have a stale link at any point should the integration get restarted.  Really the link only has to work in the short time the user clicks it, so calculating it from the backend at the point of handling the GET request where I'm guessing the information is more easily obtained seems a win.

rhuss: @gashcrumb the possible staleness was also a concern to me, but we discussed the issue with a stale link at it is acceptable to the PM if this happens, right @kcbabo ? It can get stale for many reasons. Also the link might not be valid at all, so the backend would have to redirect to a proper error message (however would be nice if the link was not visible from the very beginning when this is the case or gets automatically updated when the situation changes, like we do for other integration states via webhooks).

As I mentioned in the original mail thread:

> This is only the case if the pod scales up, which only happens if the user manually scales the pods post-deployment, right?


> It's also the case when the pod dies or is restarted (e.g. because of healthchecks or because). Its also true when the use stops the pod (because it stop the integrations) or makes a new deployment for a new revision or change revision. The link can get stale anytime (because such things can happen in the background when the use is already on the page). This is highly dynamic, also when such a link can be enabled/disable. The only good technical solution would be to enable such a link all the time, but this is a link to the UI which doese a rest request to the backend for determing the currently running pod and the either redirect to the OpenShift console (but be aware about authnetication !) or show an error dialog if no such pod exists. A static link would not work for the above reasons.

chirino:
rebased to resolve merge conflict. 
@rhuss @gashcrumb is this ready to go in?

rhuss:
From my side yes, but not sure about the architectural issues @gashcrumb raised. Is this something we can tackle later (i.e moving that kind of configuration to the server side) or should we work on this right now ? // cc @paoloantinori @chirino @gashcrumb

gashcrumb:
Yeah, I guess let's get this in as a first pass and we can improve it in the future (not) :-)

TovaCohen:
How does this affect the user doc? 
I am no longer seeing the "View Log in OpenShift" option on the Activity tab.

rhuss:
@TovaCohen this link appears only when it has been configured during the installation by setting a certain parameter (OPENSHIFT_CONSOLE_URL). This doesn't seemed to be the case for your setup. To which iPaaS installation are you refering to ?

rhuss:
I just enabled the link on syndesis-staging.b6ff.rh-idev.openshiftapps.com again.

@TovaCohen whether the link appears or not depends on how Syndesis was installed. The administrator has to provide the base URL for the console during installation of Syndesis for this link to be visible. This is optional, to allow still an installation with as less parameter as possible.

In the user doc I would write something like: "If enabled for you iPaaS installation then a link to the OpenShift console is present. If this optional feature is not enabled then no link shows up".

We are still trying to find a way to autodetect the base URL so that the link can always be shown and no extra configuration would be required. However, this is not so easy and we have no other solution yet.

Also, we probably have to adapt the _installation_ documentation, too. Not sure who to ping for this.

TovaCohen:
Thanks @rhuss
I think I can remove the reference to "View Log in OpenShift" from the user guide information about viewing integration log information in Ignite. Okay? I suggest this because we are not, so far, providing any information about what the user might see in the OpenShift log that might be helpful to them. The citizen integrator will either see the button or they won't. They don't need to know that it might exist. 
Alternatively, should we be providing any information about what to look for in the OpenShift log?
And if yes, what should Ignite doc say?

But it sounds like the installation information might require some content as a result of this issue.
For installation of Ignite, on premise, on OCP, does the installation procedure documented here (scroll down to the second "Installing Syndesis" heading):
https://github.com/syndesisio/syndesis/tree/master/install#installing-syndesis-1 
tell the administrator how to provide the base URL for the OpenShift console? Is that defined in the syndesis.yml file?
Or do we need to add doc for providing the OpenShift console URL during installation of Ignite on OCP?

For installation of Fuse Online, on OpenShift Online, I thought Red Hat takes care of that installation and so we would know whether or not the "View Log on OpenShift" button would appear. If we need to provide some user doc for this, please tell me what the doc needs to say. However, AFAIK, there are no requirements for user doc for this installation/environment. 

For installation of Syndesis (not Ignite), is there information here that needs to be updated:
https://github.com/syndesisio/syndesis/tree/master/install#installing-syndesis-1
with regard to why the OpenShift console URL needs to be specified and how to specify it?

rhuss:
> I think I can remove the reference to "View Log in OpenShift" from the user guide information about viewing integration log information in Ignite.

I agree, that this is the best solution for now

> For installation of Ignite, on premise, on OCP, does the installation procedure documented here (scroll down to the second "Installing Syndesis" heading):
> https://github.com/syndesisio/syndesis/tree/master/install#installing-syndesis-1
> tell the administrator how to provide the base URL for the OpenShift console?

Good point, the README is not complete. It misses the description of the OPENSHIFT_CONSOLE_URL parameter which is used in the same way as the other parameters (but is optional). 

Yes, for Fuse Online its us who have to provide that parameter, and I'm not sure whether for Fuse on OCP (on premise) already some installation documentations are planned which go beyond this simple README for the upstream templates.

I'm going to update the README now, thanks for the heads up.

TovaCohen:
This is the issue I am using to create the installation doc for Ignite on OCP. 
https://issues.jboss.org/browse/FUSEDOC-2485

PR #2066, opened 2018-03-19
fix: dynamic select inputs will be disabled when configured as read only

### LINKED ISSUE #2024: Select element in generated form isn't disabled when the form is read only
## This is a...

<pre><code>
[ ] Feature request
[ x ] Bug report  
</code></pre>

## The problem

The select element rendered by a read-only form does not have the `disabled` attribute set, and therefore be interacted with. 

## Expected behavior

If a dynamic form is configured to be read-only, the select element should be disabled or replaced with plain text, much like input fields.

## Screenshot

Example screenshot from the connection detail page for an AMQ connection:

usable-select

1. Create a new AMQ connection, just use whatever for the broker URL
2. Go to the detail page for your AMQ connection
3. Note the 'Skip Certificate Check' field can be changed while all the rest of the controls are disabled

deeleman:
Fixes #2024

### DIFF
--- a/app/ui/src/app/common/ui-patternfly/syndesis-form-control.component.html
+++ b/app/ui/src/app/common/ui-patternfly/syndesis-form-control.component.html
@@ -221,6 +221,7 @@
               [tooltip]="model.controlTooltip"
               [ngClass]="elementControlClass"
               [required]="model.required"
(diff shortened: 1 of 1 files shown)


deeleman:
retest this please

PR #2077, opened 2018-03-20
Add --test-support option to syndesis install scripts

tplevko:
This would make QE's lives a little bit easier.


rhuss:
@iocanel @tplevko looks good (except maybe `-s` is not so intuitive as shortcut, maybe no shortcut is needed at all ?)

It would be awesome though, if you could update the documentation at https://github.com/syndesisio/syndesis/blob/master/doc/sdh/cli/cmd_install.adoc, too

When reviewing changes to `syndesis` this should be kept in mind, please.

PR #2084, opened 2018-03-20
chore: Refinement of route detection

rhuss:
- Just add detect route if not provided as argument
- Added it to "syndesis install", too (not only the standalone install  script)
- Added error handling
- Update documentation


rhuss:
@paoloantinori fancy to give these enhancments a quick review ? thanks for the initial implementation, that is really an awesome approach. 

Would be nice now to find out how we get the openshift console link in a similar way. But I guess that's much harder ...

Issue #1955, opened 2018-03-13
Import/Export: Conflicts in Drag and Drop & Multi Select/Import


kahboom:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
Multiple vs single, drag and drop vs select sometimes seem to conflict. The following have been observed:
- Multiple integrations cannot be selected with traditional file select.
- Drag and drop import allow for multiple integrations, but in the Review section only show the first imported integration.

## Expected behavior
User should be able to upload multiple integrations and see all in the review section, currently only the first is shown. 

## Screenshot
I guess you'll just have to trust me..

[Full Designs](https://github.com/syndesisio/syndesis/blob/master/ux/designs/importexport/importexport.md)

## API Endpoints and Schemas
N/A

## Tasks involved / Steps to Reproduce
The first issue:
1. Navigate to Integrations > Import
2. Try to select multiple files using the traditional file select button. Only one is selectable.

The second (likely relevant) issue:
1. Drag and drop multiple integrations that are different.
2. The review page will show only the first integration repeated by the number of integrations imported.

PR #2090, opened 2018-03-20
Adds review step for import/export in the UI, fixes conflicts

### LINKED ISSUE #1954: Import/Export: Review Step Endpoint
## This is a...

<pre><code>
[ ] Feature request
[ x ] Bug report  
</code></pre>

## The problem
Use the integration overview endpoint to reflect data that is returned from the API in the Review section.

## Expected behavior
Review step is mostly blank.

## Screenshot
screenshot 2018-03-12 20 20 33

[Full Designs](https://github.com/syndesisio/syndesis/blob/master/ux/designs/importexport/importexport.md)

## API Endpoints and Schemas
Integration support endpoint: `/api/v1/integration-support`

1. Navigate to Integrations > Import
2. Upload exported integration zip file.
3. Review step should display information about the uploaded integration, but is mostly blank.

kahboom:
## Includes

- Review page now populating data for both single and multiple imported integrations.
- Redirect to Integration list page for multiple imports and to the detail page for single imports.
- Resolves conflicts between drag and drop and file select vs multiple and single imports.
- Do not auto redirect on upload, even for multiple the user will have to select the 'Done' button for consistency. The buttons will not appear until the user has successfully uploaded a file.

### Single Integration Import

screenshot 2018-03-20 17 34 47

### Multiple Integration Import

screenshot 2018-03-22 12 17 32

## Notes

Please note that there are a few minor things missing that will come in much smaller, separate PRs.
- Need to determine how/when to display 'Configuration Required' for Connections within Integrations, as I don't get validation information from `/v1/integration-support`. Is it just if the name is missing, or do I need additional data from `/v1/integrations`? Will address this in a later PR. **EDIT**: For now, will check if the `messages` property exists, and assume that if it does not exist, then the integration is configured. Hoping that in the future the API supports better validation for connections as well.
- Add the number of imports in a single line, as opposed to showing the same text for multiple imports.
cc @sjcox-rh 
- Unit tests and refactoring for Redux :) someday we'll get around to this

[Full UX Designs](https://github.com/syndesisio/syndesis/blob/master/ux/designs/importexport/importexport.md)

There are other issues open for validation and re-adding toast notifications that will be addressed shortly as well.

fix #1954 #1955

### DIFF
--- a/app/ui/src/app/integration/import-export/import/integration-import.component.ts
+++ b/app/ui/src/app/integration/import-export/import/integration-import.component.ts
@@ -61,11 +61,6 @@ export class IntegrationImportComponent implements OnInit, OnDestroy {
   }
 
   ngOnInit() {
-    this.integrationOverviews$ = this.integrationSupportService.watchOverviews();
-    this.integrationOverviewsSubscription = this.integrationOverviews$.subscribe(integrations => {
-      this.integrations = integrations;
-    });
-
     this.uploader = new FileUploader({
       url: this.integrationSupportService.importIntegrationURL(),
       disableMultipart: true,
@@ -89,6 +84,8 @@ export class IntegrationImportComponent implements OnInit, OnDestroy {
     this.uploader.onCompleteItem = (item: FileItem,
                                     response: string,
                                     status: number) => {
+      console.log('onCompleteItem response: ' + response);
+      console.log('onCompleteItem response.length: ' + response.length);
       if (status === 200) {
         this.review = true;
         this.integrationImports$ = JSON.parse(response);
@@ -103,7 +100,11 @@ export class IntegrationImportComponent implements OnInit, OnDestroy {
     }
   }
 
-  private checkIfSingleOrMultiple(): void {
+  private checkIfDragAndDrop(): void {
+    // Do something here
+  }
+
+  private checkIfMultiple(): void {
     // Do something here
   }
 
--- a/app/ui/src/app/integration/import-export/import/integration-import.component.ts
+++ b/app/ui/src/app/integration/import-export/import/integration-import.component.ts
@@ -8,7 +8,6 @@ import {
 } from '@syndesis/ui/platform';
 import { FileError, IntegrationImportsData } from './integration-import.models';
 import { Observable } from 'rxjs/Observable';
-import { Subscription } from 'rxjs/Subscription';
 
 import {
   FileItem,
@@ -24,10 +23,8 @@ import {
 })
 export class IntegrationImportComponent implements OnInit {
   error: FileError;
-  fetchedIntegrations: Array<IntegrationOverview>;
-  filteredIntegrations: Array<IntegrationOverview>;
+  importedOverviews$: Observable<IntegrationOverviews>;
   importing = false;
-  integrationOverviews$: Observable<IntegrationOverviews>;
   isDragAndDropImport: boolean;
   isMultipleImport: boolean;
   item = {} as FileItem;
@@ -39,9 +36,10 @@ export class IntegrationImportComponent implements OnInit {
 
   @ViewChild('fileSelect') fileSelect: ElementRef;
 
-  private integrationOverviewsSubscription: Subscription;
   constructor(private integrationSupportService: IntegrationSupportService, private router: Router) {
-    this.filteredIntegrations = [];
+    /**
+     * Do stuff here!
+     */
   }
 
   cancel() {
@@ -102,7 +100,7 @@ export class IntegrationImportComponent implements OnInit {
                                     response: string,
                                     status: number) => {
       if (status === 200) {
-        this.fetchIntegrationOverview(JSON.parse(response));
+        this.fetchImportedIntegrations(JSON.parse(response));
       }
     };
   }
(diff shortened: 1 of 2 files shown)


gashcrumb:
> The biggest thing is that for multi import we are only letting the user know that the integrations have been imported, but the review section is not displayed. In other words, the review section only appears for single integration imports. This was a bit of a hassle due to the nature of the service, but will be resolved soon.

Does the API for importing return the integration objects that were imported?  That'd be ideal for this review step rather than trying to find out from the `/integrations`/ or `/integration-overviews` endpoints.  If not, I'd defer doing the review step and let's work on getting the import endpoint to return the integration objects that were imported.

> Need to determine how/when to display 'Configuration Required' for Connections within Integrations, as I don't get validation information from /v1/integration-support. Is it just if the name is missing, or do I need additional data from /v1/integrations? Will address this in a later PR.

I don't think that should be necessary, the integration overview object after import should have it's `messages` attribute updated with that kind of info, which will then show up in the detail page.

app/ui/src/app/integration/import-export/import/integration-import.component.ts
```diff
@@ -57,14 +66,26 @@ export class IntegrationImportComponent implements OnInit, OnDestroy {
...
+  }
+
+  onFileSelected(): void {
+    this.isDragAndDropImport = false;
+    this.isMultipleImport = this.checkIfMultiple();
   }
 
   ngOnInit() {
-    this.integrationOverviews$ = this.integrationSupportService.watchOverviews();
+    this.integrationOverviews$ = this.integrationSupportService.getOverviews();
```
gashcrumb: I kinda feel like this component shouldn't need to fetch integration objects at all, at least via these API calls.  If the import endpoint doesn't return the integration objects after importing I think we should change that so it should, which I think will make this component a lot more simple, don't you think?

app/ui/src/app/integration/import-export/import/integration-import.component.ts
```diff
@@ -91,9 +112,10 @@ export class IntegrationImportComponent implements OnInit, OnDestroy {
                                     response: string,
                                     status: number) => {
       if (status === 200) {
-        this.review = true;
+        this.showButtons = true;
+        this.showReviewStep = !this.checkIfMultiple();
         this.integrationImports$ = JSON.parse(response);
```
gashcrumb: Should this end with `$`?  It's not an observable is it?  Or is it?  I'm confused :-)

app/ui/src/app/integration/import-export/import/integration-import.component.ts
kahboom: I agree and would have liked to have had that removed before opening the PR, but I'm technically still not done working on the component. If you'd like this resolved before merging I can do that.

app/ui/src/app/integration/import-export/import/integration-import.component.ts
kahboom: This should probably be `this.integrations`, and, yes, not have a `$` at the end.

app/ui/src/app/integration/import-export/import/integration-import.component.ts
gashcrumb: Yeah, if it's not needed please take it out, especially given that it's making a backend call, we make enough of 'em as it is :-)

kahboom:
>Does the API for importing return the integration objects that were imported? That'd be ideal for this review step rather than trying to find out from the /integrations/ or /integration-overviews endpoints. If not, I'd defer doing the review step and let's work on getting the import endpoint to return the integration objects that were imported.

[..]

> I don't think that should be necessary, the integration overview object after import should have it's messages attribute updated with that kind of info, which will then show up in the detail page.

In response to both, this is the data I get back after uploading multiple integrations, which is why I'm fetching just a single integration:

```
[
  {
    "version": 49,
    "tags": ["salesforce", "twitter"],
    "steps": [{
        "connection": {
            "icon": "fa-twitter",
            "connectorId": "twitter",
            "name": "Twitter Connection",
            "id": "-L5KAOmgjYRbUiKhqJvi"
        },
        "action": {
            "name": "Mention"
        },
        "kind": "step",
        "stepKind": "endpoint",
        "id": "-L7756Cs4rpabTE7gSJN"
    }, {
        "connection": {
            "icon": "fa-puzzle-piece",
            "connectorId": "salesforce",
            "name": "Salesforce Test",
            "id": "-L5KA57rjYRbUiKhqJvh"
        },
        "action": {
            "name": "Create or update record"
        },
        "kind": "step",
        "stepKind": "endpoint",
        "id": "-L7757f04rpabTE7gSJN"
    }],
    "targetState": "Unpublished",
    "currentState": "Unpublished",
    "deploymentVersion": 1,
    "draft": true,
    "name": "Test Twitter to SF",
    "id": "-L775DOUc8-A7n1KyeMq"
   }
]
```

For the first part, there is nothing to indicate whether or not the connection is configured. I could check to see if the `steps` array is empty and let the user know if the entire integration needs to be configured, but designs seem to imply that we should let them know if the connections themselves need configuration.
 
For the second part, in my mind, if the API accepted an array of IDs, it would be easier for this to be done on the backend and I could just issue a `GET` with the list of those IDs, but it doesn't. I'm just trying to work with what I have since we are on a time crunch.

It's either that I can fetch all and then iterate through and compare IDs, create a new filtered list, OR I can issue a `GET` for each imported integration, which is not ideal and the reason why we only support one at the moment. I figure it's better to display the review section for just one integration as opposed to none.

app/ui/src/app/integration/import-export/import/integration-import.component.ts
kahboom: Sure

kahboom:
On upload, this is the data I get back from a completed upload:

```
[
  {
    "action": "updated",
    "kind": "integration",
    "id": "-L775DOUc8-A7n1KyeMq"
  }
]
```

gashcrumb:
> For the first part, there is nothing to indicate whether or not the connection is configured. I could check to see if the steps array is empty and let the user know if the entire integration needs to be configured, but designs seem to imply that we should let them know if the connections themselves need configuration.

Just to recap I believe if an integration has connections that need attention it'll have a `messages` attribute set with some indication that it's connections need configuration.  Hopefully :-)

kahboom:
Going to an appointment now, will resolve this later today. Thanks @gashcrumb - will ping you once it's ready for review again.

chirino:
You need to get a connection's bulletins to see if there are issues with it.  See:

| Path   |      Verb      |  Response Data |
|----------|-------------|------|
| /api/v1/connections/{id} |  GET | Connection |
| /api/v1/connections/{id}/bulletins|    GET   |   ConnectionBulletinBoard |

kahboom:
Updated the PR description to state that multiple integration import is now working for the Review step:

screenshot 2018-03-22 12 17 32

Lint and tests are passing, functionality seems stable. Okay for review again. Thanks @gashcrumb !

kahboom:
Spoke too soon. Build failure! Checking it out now.

kahboom:
>You need to get a connection's bulletins to see if there are issues with it.

@chirino - Will do this for the validation issue I have open, that's the next thing I'll be working on. 👍  Thanks. It would be great if we could do the validation from the same endpoint used to import the integrations, but maybe we can discuss that later on. The same goes for extensions, if they are included in the export/import of an integration (have yet to test this).

PR #2105, opened 2018-03-21
Add a default icon to connector extensions

### LINKED ISSUE #2086: Extensions without icon should display a default one
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem
No icon is displayed if a extension does not embed a icon.png or icon.pdf file.

## Expected behavior
A default icon, like the one used by api-client connectors should be used instead.

## Screenshot
N/A

## Request and Response Data
N/A

## API Endpoints and Schemas
N/A

1. Create an extension of type connector without a icon.png file in META-INF/syndesis
2. Import the extension in Syndesis
3. Create a connection from the extension connector
4. The connection does not show a icon

nicolaferraro:
Fixes #2086.
Just adds a generated icon to a uploaded extension if that is missing. Using a template cloned from the swagger one (no need to use a visually different icon imo).


TovaCohen:
In the staging site, I just uploaded syndesis-connector-telegram-1.0.0.jar. This appears to have worked successfully, but there is no UI for me to upload an icon for the connector, and a default icon for the connector does not appear. There is nothing in the UI that indicates that the connector has an icon. This is all that I see in the details view. There is no icon at all to the left of this information:
image

nicolaferraro:
@TovaCohen the icon for extensions is not displayed in the import page, but you can see it when you create a connection (only for connector extension).

The citizen developer cannot change it, only the expert developer can package a new icon together with the extension.

I agree that the icon can be better shown also on the details page after upload. Ping @syndesisio/uxd, @dongniwang.

Current ux design here: https://github.com/syndesisio/syndesis-ux/blob/master/designs/technical_extensions/tech_ext.md

TovaCohen:
Ahh. Thanks Nicola! You are of course right. If I had taken it one step further in my experimentation I would have seen that myself. Thanks very much for the explanation!

dongniwang:
Thanks @nicolaferraro and @TovaCohen for the feedback! UXD team is planning to go through the major details pages in the app to ensure consistency of interactions and page layout. 

cc: @amysueg @sjcox-rh

Issue #1695, opened 2018-02-27
UI: Show Notification Messages associated with Connections/Integrations


chirino:
A new api data is available to get connection or integration related notifications.
Integration Overview objects can now contain a "messages" field.

```
"messages": [{
"level":"INFO", // "WARN" or "ERROR"
"message": "the message"
}]
```

You can get similar messages for a connection by hitting:

```
curl /api/v1/connections/5/bulletins
{
  "id": "5",
  "createdAt": 0,
  "updatedAt": 0,
  "messages": [{
    "level":"INFO", // "WARN" or "ERROR"
    "message": "the message"
  }]
}
```

chirino:
This issue is related to the import epic.
On import of a connection, if the connection has secrets that need to be reset, an associated message will be created for it.

The upgrade epic is related too. 
When an integration is using a connector/conneciton that has been updated, there should be a upgrade available notice posted as a message to let the user know to trigger a publish to pickup the update.

chirino:
These new APIs can be used to drive these notifications shown in this UXD design:
https://github.com/syndesisio/syndesis/blob/master/ux/designs/importexport/importexport.md

![syndesis_importexport_md_at_master_ _syndesisio_syndesis](https://user-images.githubusercontent.com/103255/36735231-ab45c3c0-1ba3-11e8-8b59-ec331ea34947.jpg)

chirino:
And should also show integration specific notification messages somewhere on the integration details page similar to:

![syndesis_integration_details_page_md_at_master_ _syndesisio_syndesis](https://user-images.githubusercontent.com/103255/36736106-b69c5e26-1ba5-11e8-9a6f-63092d891fee.jpg)

gashcrumb:
A problem with making this a separate remote invocation at least for connections is on that list view the UI is going to have to make a separate call for each connection, that's going to suck when a user has lots of connections.  Would it make more sense to inline the property in the connection object?

And the use case for the connection views is really whether or not the user needs to reset it's credentials, I wonder does it really need to be a list of messages that the UI needs to parse?  Seems like a simple boolean on the connection object whether or not the credentials need to be updated should suffice?

chirino:
I'm not sure we want to update the connection object when something happens external to the connection.  For example when a connector upgrade is available.

gashcrumb:
[...]

PR #2107, opened 2018-03-21
A bulletin board for connections should be generated when a connector is updated

### LINKED ISSUE #2102: A bulletin board for connections should be generated when a connector is updated
## This is a...

<pre><code>
[X] Feature request
[ ] Bug report  
</code></pre>

## The problem

A bulletin board for connections should be generated when a connector is updated

## Expected behavior

Upgrading a connector should result in connections using the connector/connection as having a message saying the connection should be updated.

lburgazzoli:
Fixes #2102

This PR is a work in progress to generate a bulletin board for resources being updated and it is focused on generating the board for `Connections` when a related `Connector` is updated. The board is re-computed also when a connection is edited so errors/warning/etc are cleaned up once the users fixes the issues.

There is a new module `update-controller` which provides:

- a task to build bulletin boards when needed 
- a subscriber for EventBus to wake-up the bulletin boards task

Related endpoints are:

| Path   |      Verb      |  Response Data |
|----------|-------------|------|
| /api/v1/connections |  GET | collection of ConnectionOverview |
| /api/v1/connections/{id}/bulletins|    GET   |   ConnectionBulletinBoard |

* **Overviews**:

  A `ConnectionOverview` extends `Connection` with the related `ConnectionBulletinBoard`:

  ```json
    {
    "totalCount": 1,
    "items": [
      {
        "board": {
          "notices": 1,
          "warnings": 1,
          "errors": 0,
          "id": "i-L8CJCZT_zf571IBk9Eaz",
          "metadata": {
            "connector-id": "ext-io-syndesis-extensions-syndesis-connector-telegram",
            "connector-version-latest": "2",
            "connector-version-connection": "1"
          },
          "messages": [
            {
              "level": "WARN",
              "metadata": {
                "property": "newProperty",
                "status": "new"
              }
            },
            {
              "level": "INFO",
              "metadata": {
                "property": "authorizationToken",
                "status": "updated"
              }
            }
          ],
          "createdAt": 1521715567065,
          "updatedAt": 1521715624291
        },
        "id": "i-L8CJCZT_zf571IBk9Eaz",
        "connector": {
        },
        "connectorId": "ext-io-syndesis-extensions-syndesis-connector-telegram",
        "icon": "extension:icon.png",
        "userId": "developer",
        "lastUpdated": 1521715566857,
        "createdDate": 1521715566857,
        "name": "Telegram",
        "configuredProperties": {
        },
        "isDerived": false
      }
    ]
  }
  ```

* **Bulletins**:

  A bulletin contains some information in form of meta-data about the resource being impacted and a list of `LeveledMessage`

  ```json
  {
    "notices": 1,
    "warnings": 1,
    "errors": 0,
    "id": "i-L8CMPrxjIPtUr-ZevhXz",
    "metadata": {
      "connector-id": "ext-io-syndesis-extensions-syndesis-connector-telegram",
      "connector-version-latest": "2",
      "connector-version-connection": "1"
    },
    "messages": [
      {
        "level": "WARN",
        "metadata": {
          "property": "newProperty",
          "status": "new"
        }
      },
      {
        "level": "INFO",
        "metadata": {
          "property": "authorizationToken",
          "status": "updated"
        }
      }
    ],
    "createdAt": 1521716407969,
    "updatedAt": 1521716433464
  }
  ```

**NOTES**: 

1. each `LeveledMessage` may have some associated meta-data to inform about which property has been changed and the reason (new, updated, deleted). The level of the message reflect the severity of the change when possible (i.,e. if a new mandatory field is added, the level is set to WARN whereas it is INFO if the new property is not mandatory)
2. it is possible to configure the frequency of the bulletin task with some properties (set to 1 minute for testing purpose) :
    ```yaml
    resource:
    update:
      controller:
        enabled: true
        check-interval: 1
        check-interval-unit: MINUTES
    ```
3. this implementations takes only into account `Connection` related properties, `Action` will follow.


gashcrumb:
@deeleman CC

I'll be honest, I super-dislike that we have basically two endpoints these days for integration objects.  So I'm pretty biased against having 2 endpoints for fetching connection status.  I've also [already](https://github.com/syndesisio/syndesis/issues/1695#issuecomment-368908517) [stated](https://github.com/syndesisio/syndesis/issues/1695#issuecomment-372334251) that I don't believe having a separate endpoint for connection messages is that great.

Like seriously, why do we have to have a separate list of 'overview' objects?

lburgazzoli:
@gashcrumb 

we don't need the second endpoint, I've added it because it is already there for integrations but I would really love to have endpoints for listing resource like `/connections`, `/integrations`, etc to always return `*Overview`.

lburgazzoli:
@gashcrumb working to unify them

gashcrumb:
@lburgazzoli you're the best, thanks!

It'll give us a little work in the UI but it's worth it I think.

deeleman:
II'll take over the UI parts for this after this PR gets merged once finished.

lburgazzoli:
@gashcrumb @deeleman 

I've removed the additional endpoint and now a board is eventually included in the `connections` response. Note that now this endpoint return a list of `ConnectionOverview` that extends `Connection` so it does not break the UI.

Can you take a look at the samples I've included and let me know if there is anything you want me to fix ? i.e. we may get rid of `messages` from `ConnectionOverview.board`

gashcrumb:
@lburgazzoli Yeah, I like this approach much better, thanks a ton!

I'd definitely also like us to work towards merging the endpoints for Integrations and IntegrationOverviews, are you planning on doing it in this PR or should I raise an issue to track it as maybe a post GA task?

lburgazzoli:
@gashcrumb my plan is to work on integrations just after this first PR is validated, think about this PR as a POC so if it fits our needs I'll then go ahead and start working on bulletin for integrations when a connector, and extension or so on gets updated and I can work to merge the endpoints too.

app/common/model/src/main/java/io/syndesis/common/model/WithRevision.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.common.model;
+
+import java.util.OptionalLong;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+public interface WithRevision<T extends WithId<T>> {
```
chirino: this seems similar to WithVersion class? Should this be one thing?

lburgazzoli: uh, didn't notice that interface, will have a look

app/server/update-controller/src/main/java/io/syndesis/server/update/controller/impl/ConnectorUpdateHandler.java
```diff
@@ -0,0 +1,118 @@
...
+        final List<ConnectionBulletinBoard> boards = new ArrayList<>();
+        final DataManager dataManager = getDataManager();
+        final List<Connection> connections = dataManager.fetchAll(Connection.class).getItems();
+        final List<Connector> connectors = dataManager.fetchAll(Connector.class).getItems();
+
+        for (int i = 1; i < connectors.size(); i++) {
+            final Connector connector = connectors.get(i);
+            final String id = connector.getId().get();
+
+            for (int c = 0; c < connections.size(); c++) {
```
chirino: O(N^2) = Could get ugly one day.

lburgazzoli: This part is everything but optimized I'm working to use DataManager::fetchIdsByPropertyValue

app/common/model/src/main/java/io/syndesis/common/model/WithRevision.java
lburgazzoli: So it is not the same thing as WithRevision extends WithId as this is required for db operation, maybe we should support revision for every resource ? (post GA)

chirino: Perhaps make it WithIdVersioned and have it extend both WithId and WithVersion.  I'd just hate to add both a revision and a version field to an object.

lburgazzoli: I will

lburgazzoli:
@chirino I should have fixed your findings, can you please review it again ?

PR #2120, opened 2018-03-22
feat: add conventional commit message check hook

zregvart:
This adds conventional commit message pre-commit hook. To instal it
one must run:

    $ pre-commit install -t commit-msg # to install the hook
    $ yarn                             # to install the commitlint

This will add a pre-commit as commit-msg in .git/hooks/.

Configuration is entierly derived from
@commitlint/config-conventional[1] and can be further refined in the
`commitlint.config.js` file.

Example of invalid commit message:

```shell
Check commit message..............................................Failed
hookid: commit-msg-check

⧗   input: Hello world
✖   message may not be empty [subject-empty]
✖   type may not be empty [type-empty]
✖   found 2 problems, 0 warnings
```

[1]
https://github.com/marionebl/commitlint/tree/master/@commitlint/config-conventional


oscerd:
Do we really need another check also on commit message?

zregvart:
> Do we really need another check also on commit message?

In my opinion yes. With this check all our commit messages will follow the [Conventional Commits](https://conventionalcommits.org/) guidelines and that will enable us to auto-generate change logs.
It is an opt-in, so you'd need to install [pre-commit](https://pre-commit.com/) and install the hook in `commit-msg` stage. I don't think there is a way to enforce this.

zregvart:
An alternative would be to add [Commitlint GitHub App](https://probot.github.io/apps/commitlint/) to perform the check

zregvart:
It's opt-in and entirely up to the contributor if he/she wishes to do so. With this I can at least make myself conform to conventional commit messages, if others follow suit then we can have auto generated change log.
We can discuss making this a mandatory check, for instance by installing the commitlint GitHub app. We have the configuration file now in place to do that.

paoloantinori:
I wouldn't assume good will.

We wouldn't have been in this situation at all, if that was the case.

In my view if we want this, it needs to be mandatory either via peer review
or via automated tooling.

PR #2135, opened 2018-03-22
fix: use remaining parameters as query params when building URLs

### LINKED ISSUE #2132: Support sending optional query params in ApiHttpService's setEndpointUrl method
## This is a...

<pre><code>
[X] Feature request
[ ] Bug report  
</code></pre>

## The problem
There are APIs which support optional query parameters.  For example all entity listing APIs support pagination control via query parameters.  

We need a way the UI pass these to the APIs when needed.  Keep in mind the query parameters keys and values should be URL encoded and that query keys are allowed to repeat.  Like:

/api/somepath?id=1&id=2&id=3

gashcrumb:
fixes #2132

### DIFF
--- a/app/ui/src/app/api/providers/api-http-provider.service.ts
+++ b/app/ui/src/app/api/providers/api-http-provider.service.ts
@@ -165,9 +165,40 @@ export class ApiHttpProviderService extends ApiHttpService {
   }
 
   private replaceNameMatches(endpointTemplate: string, paramsMap: StringMap<any>): string {
-    return endpointTemplate.replace(/\{(\D*?)\}/g, (fullMatch, ...matchGroups) => {
-      return paramsMap[matchGroups[0].trim()];
+    const matched = [];
+    const answer = endpointTemplate.replace(/\{(\D*?)\}/g, (fullMatch, ...matchGroups) => {
+      const match = matchGroups[0].trim();
+      matched.push(match);
+      return paramsMap[match];
     });
+    // Turn any remaining parameters into query params
+    const remainder = Object.keys(paramsMap).filter(x => matched.indexOf(x) === -1);
+    const queryParams = remainder.map(key => {
+      const value = paramsMap[key];
+      // avoid passing any url param as 'undefined'
+      if (!value) {
+        return undefined;
+      }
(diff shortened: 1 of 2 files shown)


app/ui/src/app/api/providers/api-http-provider.service.ts
```diff
@@ -165,9 +165,40 @@ export class ApiHttpProviderService extends ApiHttpService {
...
+      const value = paramsMap[key];
+      // avoid passing any url param as 'undefined'
+      if (!value) {
+        return undefined;
+      }
+      // expand an array value in the param map to multiple query params
+      if (typeof value === 'string') {
+        return key + '=' + value;
+      } else if (typeof value === 'object' && value.map) {
+        return value.map(val => key + '=' + val).join('&');
```
chirino: we should probably URL encode key and values

app/ui/src/app/api/providers/api-http-provider.service.ts
```diff
@@ -165,9 +165,40 @@ export class ApiHttpProviderService extends ApiHttpService {
...
   private replaceNameMatches(endpointTemplate: string, paramsMap: StringMap<any>): string {
-    return endpointTemplate.replace(/\{(\D*?)\}/g, (fullMatch, ...matchGroups) => {
-      return paramsMap[matchGroups[0].trim()];
+    const matched = [];
+    const answer = endpointTemplate.replace(/\{(\D*?)\}/g, (fullMatch, ...matchGroups) => {
+      const match = matchGroups[0].trim();
+      matched.push(match);
+      return paramsMap[match];
     });
+    // Turn any remaining parameters into query params
```
chirino: We could probably use https://angular.io/api/common/http/HttpParams to make sure we properly encode the query string.

gashcrumb: yeah, good call, will add that, thanks!

deeleman: Perhaps you will want to use the `HttpUrlEncodingCodec` class instead (it's not injectable, you must instantiate it with `new`). It features both `encodeKey()` and `encodeValue()` methods.

Issue #1623, opened 2018-02-20
User doc for TP4


TovaCohen:
Creating this issue so I can log commits that contain user doc for TP4 here.

Issue #1189, opened 2018-01-23
Dropbox Connector


rohanmars:
Implement Dropbox connector component.

PR #2156, opened 2018-03-26
user doc: Create doc for Dropbox connector (#1623 #1189)

TovaCohen:
This is the initial draft of the user doc for using the Dropbox connector. If you could review this so that I have your feedback by the start of my day on Wednesday, that would be great. You might want to review the Dropbox files in this order:
topics/connecting_to_dropbox.adoc
topics/register_with_dropbox.adoc
topics/create_dropbox_connection.adoc
topics/adding_dropbox_connections.adoc
topics/adding_dropbox_connection_start.adoc
topics/adding_dropbox_connection_finish.adoc
topics/adding_dropbox_connection_middle.adoc
The other files make a few other updates.

### DIFF
--- a/doc/integrating_applications/topics/adding_dropbox_connection_finish.adoc
+++ b/doc/integrating_applications/topics/adding_dropbox_connection_finish.adoc
@@ -1,0 +1,23 @@
+[id='adding-dropbox-connection-finish']
+= Adding a Dropbox connection as a finish connection
+
+In an integration, to add a Dropbox connection as the finish connection,
+start creating the integration, add and configure the start connection,
+and then:
+
+. On the *Choose a Finish Connection* page, click the Dropbox connection that
+you want to use as the integration's finish connection. 
+. On the *Choose an Action* page, click the *Upload a file to Dropbox* 
+action to add the current integration data to the
+Dropbox account that this connection accesses. 
+. In the *Remote Path* field, enter the path of the Dropbox folder 
+in which to store the integration data. 
+. For the *Upload mode* select:
++
+** *Add* to rename a file that you are uploading if a file with the
+same name already exists in that Dropbox location.
+** *Force* to overwrite a file that is already in that Dropbox location if you are 
+uploading a file with the same name. 
+
+. Click *Next* to add this connection as the finish connection in the
+integration. 
--- a/doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
+++ b/doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
@@ -1,0 +1,36 @@
+[id='adding-dropbox-connection-middle']
+= Adding a Dropbox connection as a middle connection
+
+In an integration, to add a Dropbox connection between the start and 
+finish connections:
+
+. Add the start and finish connections to the integration.
+. In the left panel, hover over the plus sign that is in the location
+where you want to add a Dropbox connection.
+. In the popup, click *Add a Connection*.
+. On the *Choose a Connection* page, click the Dropbox connection that you 
+want the integration to use. 
+
+. On the *Choose an Action* page, select one of the following actions:
++
+* *Upload a file to Dropbox* to add the current integration data to the
+Dropbox account that this connection accesses. To configure this
+action:
+.. In the *Remote Path* field, enter the path of the Dropbox folder 
+in which to store the integration data. 
+
+.. For the *Upload mode* select:
++
+** *Add* to rename a file that you are uploading if a file with the
+same name already exists in that Dropbox location.
+** *Force* to overwrite a file that is already in that Dropbox location if you are 
+uploading a file with the same name. 
++
+* *Download from Dropbox* to obtain one or more files from the Dropbox
+account that this connection accesses. To configure this action,
+in the *Folder or filename path to download* field, enter the Dropbox
+folder path or the Dropbox file path of the content 
+you want to obtain. 
+
+. Click *Next* to add this connection to the
+integration. 
--- a/doc/integrating_applications/topics/adding_dropbox_connection_start.adoc
+++ b/doc/integrating_applications/topics/adding_dropbox_connection_start.adoc
@@ -1,0 +1,18 @@
+[id='adding-dropbox-connection-start']
+= Adding a Dropbox connection as the start connection
+
+In an integration, to add a Dropbox connection as the start connection:
+
+. In the {prodname} panel on the left, click *Integrations*.
+. Click *Create Integration*.
+. On the *Choose a Start Connection* page, click the Dropbox connection that
+you want to use as the integration's start connection. 
+. On the *Choose an Action* page, click the *Download from Dropbox* action
+to obtain one or more files from the Dropbox account that this connection
+accesses. 
+. To configure the action, in the *Folder or file name path to download* field,
+specify the filename path or the folder path for the content that you want
+the integration to obtain. 
+. Click *Next* to add the start connection to the integration. 
+
+. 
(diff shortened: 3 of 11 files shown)


doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
```diff
@@ -0,0 +1,36 @@
...
++
+* *Upload a file to Dropbox* to add the current integration data to the
+Dropbox account that this connection accesses. To configure this
+action:
+.. In the *Remote Path* field, enter the path of the Dropbox folder 
+in which to store the integration data. 
+
+.. For the *Upload mode* select:
++
+** *Add* to rename a file that you are uploading if a file with the
```
oscerd: In this case if the file has the same name of the file in the dropbox path, the upload won't be completed and the existing won't be overwritten.

doc/integrating_applications/topics/adding_dropbox_connection_finish.adoc
```diff
@@ -0,0 +1,23 @@
...
+. On the *Choose a Finish Connection* page, click the Dropbox connection that
+you want to use as the integration's finish connection. 
+. On the *Choose an Action* page, click the *Upload a file to Dropbox* 
+action to add the current integration data to the
+Dropbox account that this connection accesses. 
+. In the *Remote Path* field, enter the path of the Dropbox folder 
+in which to store the integration data. 
+. For the *Upload mode* select:
++
+** *Add* to rename a file that you are uploading if a file with the
```
oscerd: In this case if the file has the same name of the file in the dropbox path, the upload won't be completed and the existing won't be overwritten.

doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
```diff
@@ -0,0 +1,36 @@
...
+. In the popup, click *Add a Connection*.
+. On the *Choose a Connection* page, click the Dropbox connection that you 
+want the integration to use. 
+
+. On the *Choose an Action* page, select one of the following actions:
++
+* *Upload a file to Dropbox* to add the current integration data to the
+Dropbox account that this connection accesses. To configure this
+action:
+.. In the *Remote Path* field, enter the path of the Dropbox folder
```
oscerd: Here you'll need to specify the the file name too.

doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
TovaCohen: Will the integration then be in an error state?
If not, how does the user know that the upload was not completed?
What should the user do to correct the situation?
(I can't believe how fast you read!)

doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
TovaCohen: Is this correct: 
In the Remote Path field, to upload multiple files, enter the path of the Dropbox folder in which to store the integration data. To upload one file, enter the Dropbox path and file name.

doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
oscerd: The upload will complete without error, but the file won't be overwritten. No error.

Add option is from Dropbox API stuff and it will ensure you won't overwrite files while uploading, while with force you will overwrite existing files.

doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
oscerd: The remote path can be a folder or a file. But the folder must be in the local File system, since we use the upload action only as an end or middle connection, you won't have a folder coming from another connection, so yes, this should be used to upload a single file.

oscerd:
You just need to rebase and push the update.

doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
TovaCohen: Sorry, I am still not clear about the behavior. It sounds like if the user selects Add for the Upload Mode and there is a file already in Dropbox with the same name then the file being uploaded is not Added. I must be wrong about that. Hmmmm. Is this the behavior:
- The file that is already in Dropbox is not overwritten.
- The file being uploaded is somehow renamed - how?
- There is no error, the upload operation completes, the integration continues execution. 
Or is this the behavior:
- The file that is already in Dropbox is not overwritten.
- The file that is being uploaded is lost.
- There is no error and the integration continues. 
Thanks for your patience!

doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
TovaCohen: Is this correct:
In the Remote path field, to upload multiple files, enter the local path for the folder that you want to upload. Dropbox uses the same path to store the files. To upload one file, specify the local path and file name of the file you want to upload. Dropbox stores the file with the same path and name. 

Sorry if I am still getting this wrong. I understand that "Remote path" is the name of the Camel Dropbox option, but I am seeing two different explanations for it. One says it is the Dropbox destination and the other says that it is the original (local I guess) path.

oscerd: This is correct.

doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
TovaCohen: If the file uploaded is not lost, then happens to it?

doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
TovaCohen: Happy dance! Thanks!

doc/integrating_applications/topics/adding_dropbox_connection_middle.adoc
oscerd: Suppose you're trying to get a file from an FTP as starting connection and you want to upload to dropbox, in the next poll you'll get the same file and trying to upload it to dropbox, so it's not lost, there will be a sequence of uploading operation, that won't complete because you're using the add mode. We can stress this in the docs and make the users aware that using add mode won't ensure that the file will be uploaded.

TovaCohen: Okay. That helps. When you say "same file", does that mean the same content?
Should we be telling this to the user: 
Select Add when you want to upload the file only when its content has been updated.
Select Force when you want to ensure that the file is uploaded even if it has the same content as the file that is already in place.
--
I seem to be diverging from the Camel user doc, which says: "in case of add the new file will be renamed if a file with the same name already exists on dropbox. "

oscerd: I mean same name. If you try to upload a file with the same name of a file in dropbox remote path, then it won't be uploaded, the content can be different. This morning I tested both Add mode and Force mode in Syndesis and when I upload a file with the same name of a file specified in the dropbox remote path, the upload hasn't been completed, while using force mode the upload has been completed.

oscerd: My opinion is:

- Select Add when you want to upload the file only when it's not already present in the same Dropbox path.
- Select Force when you want to ensure that the file is uploaded even if it is already present in the same Dropbox path.

oscerd: Perfect :-)

TovaCohen: Really happy dance for this one  !

TovaCohen:
I committed the updates. There was one conflict when I tried to rebase, so I aborted that and tried Roland's suggestion. I hope this worked and this content can be merged. I'd really like to get it downstream so I can make sure it builds there. Thanks again!

Issue #1538, opened 2018-02-13
Able to create duplicate connections


honghuac:
During the course of testing, I was able to create duplicate connections. This is proof that the duplicate entity check can be circumvented.

![duplicate-connections](https://user-images.githubusercontent.com/8625482/36133499-4d63e34e-10b9-11e8-8f10-6da6a92a6450.png)

zregvart:
I can reproduce this as in the final step of create connection wizard if I click the _Create_ button it takes a bit of time for the connection to be created allowing me to click the _Create_ button the second time.

Looks like we need locking support in jsondb to prevent race conditions, and we probably should disable buttons that issue POST requests on click.

cc @gashcrumb @chirino

KurtStam:
@gashcrumb fixed disabling the 'create' button in a different PR.

PR #2170, opened 2018-03-27
fix: Able to create duplicate connections #1538

KurtStam:
Serverside check to make sure the name does not exists

### DIFF
--- a/app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
+++ b/app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
@@ -32,8 +32,10 @@ import java.util.stream.Stream;
 import javax.annotation.PostConstruct;
 import javax.persistence.EntityExistsException;
 import javax.persistence.EntityNotFoundException;
+import javax.persistence.PersistenceException;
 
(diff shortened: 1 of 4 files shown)


app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
```diff
@@ -442,4 +448,30 @@ public void clearCache() {
     private static <T> Function<ListResult<T>, ListResult<T>>[] noOperators() {
         return (Function<ListResult<T>, ListResult<T>>[]) NO_OPERATORS;
     }
+
+    private <T extends WithId<T>> void validateNoDuplicateName(final T entity) {
+        validateNoDuplicateName(entity, null);
+    }
+
+    private <T extends WithId<T>> void validateNoDuplicateName(final T entity, final String ignoreSelfId) {
+        if (WithName.class.isAssignableFrom(entity.getClass())) {
```
chirino: why not 'entity instanceof WithName'?

KurtStam: @chirino https://github.com/syndesisio/syndesis/pull/2173

PR #2179, opened 2018-03-28
A bulletin board for integration should be generated when a connector/extension is updated

### LINKED ISSUE #2176: A bulletin board for integration should be generated when a connector/extension is updated
## This is a...

<pre><code>
[X ] Feature request
[ ] Bug report  
</code></pre>

## The problem

A bulletin board for integration should be generated when a connector/extension is updated

## Expected behavior

Upgrading a connector/extensions should result in integrations using the connector/extension as having a message saying the integration should be updated.

lburgazzoli:
Fixes #2176

This PR is aimed to generate a bulletin board for integrations when a connector or an extension gets updated. 

The following endpoints returns a list of ConnectionOverview
- /api/v1/integration-support/overviews
- /api/v1/integrations

The following endpoints returns a single ConnectionOverview
- /api/v1/integrations/{id}
- /api/v1/integrations/{id}/overview

**NOTE**:
- I left `/integrations/{id}/overview` and `integration-support/overviews` to maybe ease the work on UI side but they can be removed as they are a mirror of `integrations` and `integrations/{id}`. 
- the generate board is a little more detailed that the one used for `Connection` as there are much more conditions that  have to be taken into account
-  the generate board include one or more message **per step**
- the new codes are:
    ```java
    enum Code {
        SYNDESIS000, // Generic message
        SYNDESIS001, // One or more properties have been updated
        SYNDESIS002, // One or more properties have been added or removed
        SYNDESIS003, // Connector has been deleted
        SYNDESIS004, // Extension has been deleted
        SYNDESIS005, // Action has been deleted
        SYNDESIS006, // One or more required properties is not set
        SYNDESIS007, // Secrets update needed
        SYNDESIS008, // Validation Error
    }
    ```

@gashcrumb do you mind checking it out and play a little with it ?


app/common/model/src/main/java/io/syndesis/common/model/WithVersion.java
```diff
@@ -22,4 +22,11 @@
     default int getVersion() {
         return 1;
     }
+
+    /**
+     * Marker interface to indicate that the revision should be automatically updated
+     * by the data manager
+     */
+    interface AutoUpdatable {
```
zregvart: Naming things is hard, this is a good name, might `Current` be a better one? So `WithVersion.Current`?

app/common/model/src/main/java/io/syndesis/common/model/bulletin/LeveledMessage.java
```diff
@@ -33,7 +33,14 @@
...
         SYNDESIS000, // Generic message
-        SYNDESIS001 // There are parameter updates for this connection
+        SYNDESIS001, // One or more properties have been updated
+        SYNDESIS002, // One or more properties have been added or removed
+        SYNDESIS003, // Connector has been deleted
+        SYNDESIS004, // Extension has been deleted
+        SYNDESIS005, // Action has been deleted
+        SYNDESIS006, // One or more required properties is not set
+        SYNDESIS007, // Secrets update needed
+        SYNDESIS008, // Validation Error
```
zregvart: Would it make sense to make the message also part of the `Code` enum?

lburgazzoli: It could but UI should not depend on the server message so they can have a much better message than the one we could think about (and maybe have a localized one)

app/common/model/src/main/java/io/syndesis/common/model/WithVersion.java
lburgazzoli: I can do that, but after PTO :)

zregvart: Or `Latest`, not insisting on this really...

app/common/model/src/main/java/io/syndesis/common/model/bulletin/LeveledMessage.java
zregvart: We already localize messages in `messages.properties` in `server-runtime`, it contains validation messages. We could have a joint configuration file with all messages used in the server or the UI at some point...

deeleman: And we will... I'm currently working on that and will come up with a desired config file format in a few days. That's why it is so important not to have hardcoded messages coming from the server within particular calls

gashcrumb:
> the generate board include one or more message per step

Hmm, per step?  Wonder if maybe step-related messages should just be output in the step?  As typically that's probably where they'd be displayed, at the relevant configuration screen.  Anyways I'll pull this branch down here and have a looksie, might be nice to rebase this PR once #2204 is merged as that contains some relevant work to this.

gashcrumb:
rebased anyways

gashcrumb:
Checking this out now that I've rebased it on master and it's looking good, the implementation I started to add for the connection detail page still works, and it still shows a connection updated message on the integration detail page.

I started pulling out the requests to `integration-support/overviews` in favor of just fetching `/integrations` however I don't see the state fields, also the ID of the current deployment etc.

For now I pushed those changes to this branch if you want to have a look when you get back -> https://github.com/gashcrumb/syndesis/tree/lburgazzoli-integration-overview

lburgazzoli:
@gashcrumb 

I'm working to fix your findings and yeah step-related messages should be part of the step but I didn't find a reasonable way to do so in the time-frame we have, can we do so after GA ?

gashcrumb:
Yep, absolutely...

lburgazzoli:
@gashcrumb the latest code produces something like:

```json
{
  "board": {
    "targetResourceId": "i-L9ACMCMNEHKkXEJ73fRz",
    "notices": 0,
    "warnings": 1,
    "errors": 0,
    "id": "i-L9ACMETNEHKkXEJ73fSz",
    "metadata": {
      "integration-id": "i-L9ACMCMNEHKkXEJ73fRz",
      "integration-version": "2"
    },
    "messages": [
      {
        "level": "WARN",
        "code": "SYNDESIS002",
        "metadata": {
          "step": "-L9ADHAX69BIU6RZJGf-"
        }
      }
    ],
    "createdAt": 1522753958878,
    "updatedAt": 1522754356965
  },
  "currentState": "Published",
  "targetState": "Published",
  "deploymentVersion": 2,
  "deployments": [
    {
      "userId": "developer",
      "currentState": "Published",
      "targetState": "Published",
      "stepsDone": {
        "build": "172.30.1.1:5000/syndesis/i-db2db:latest",
        "deploy": "5"
      },
      "id": "i-L9ACMCMNEHKkXEJ73fRz:2",
      "version": 2,
      "createdAt": 0,
      "updatedAt": 1522754327670
    },
    {
      "userId": "developer",
      "currentState": "Unpublished",
      "targetState": "Unpublished",
      "stepsDone": {
        "build": "172.30.1.1:5000/syndesis/i-db2db:latest"
      },
      "id": "i-L9ACMCMNEHKkXEJ73fRz:1",
      "version": 1,
      "createdAt": 0,
      "updatedAt": 1522754276559
    }
  ],
  "id": "i-L9ACMCMNEHKkXEJ73fRz",
  "steps": [ 
    ...
  ],
  "version": 2,
  "createdAt": 1522753958742,
  "updatedAt": 1522754211752,
  "tags": [
    "sql"
  ],
  "name": "db2db",
  "isDraft": false
}
```

Is there anything missing ?

paoloantinori:
guys can you ping be back when you are satisfied enough with what we have. I'd like to publish the feature in a tag as soon as we have anything consistent so that that the validation process can start without have to work on master.

gashcrumb:
Yeah, that looks good.  I'll give it a try here in a few to be sure, will
also have to adjust the UI to look at the 'board' attribute anyways.  If
you don't mind I'll push those changes to this branch then assuming it all
looks good.

lburgazzoli:
@gashcrumb go ahead

gashcrumb:
@lburgazzoli what command do you use to build/deploy the server locally?  I can't seem to get an updated build running in my minishift instance today via `tools/bin/syndesis`.

lburgazzoli:
something like:

  tools/bin/syndesis build -c -f -m server -d --openshift

gashcrumb:
K, pushed some updates, the UI no longer fetches overviews, just integrations, and I adjusted the code so that it gets integration messages from the `board` attribute.

I'd suggest at this point that we keep integration-related message IDs in their own range separate from connection messages, for example for whatever reason the integration I had deployed has a `SYNDESIS001` message, which means then my integration shows:

capture

gashcrumb:
I can do that, but I definitely want to try and stick to a simple 1:1 mapping of messages to status codes.  I would prefer to avoid having to interpret the status code + the context in which the related message would be displayed and whatever the contents of the metadata object is, that gets way too complicated.  Even here, the same status code is used for 2 different things really, on the connection detail page the metadata refers to a specific property, here on the integration detail page the metadata refers to a step ID.  That kind of inconsistency is just going to lead to issues.

Also what's odd here specifically is that I'm seeing this message but I haven't actually made any changes that should trigger it, I had created the integration I *think* before I managed to get the code on this branch deployed.  I hadn't done anything like updating the connection object, and actually re-publishing the integration hasn't made it go away either.  So not sure what's going on there, but in any case the UI is equipped to at least handle displaying error codes now :-)

lburgazzoli:
I definitively agree and some of the issue will be automatically solved when we move the messages at step level. However some code can be applied at connection and step levels i.e. when a new mandatory property is added so it could still make sense to have at least some generic codes for which the context is implicit. 

I think there was a bug in the code that generates the message even when it was not the case, ca you try with the latest code ?

lburgazzoli:
@gashcrumb do you still have the notification for newly created integrations ? if not I'd remove the wip

lburgazzoli:
@gashcrumb so we need someone to approve the PR and then break everything :)

PR #2181, opened 2018-03-28
bin/syndesis: create cookie for curl invocations

### LINKED ISSUE #2180: Get cookie to consume API from CLI
## This is a...

<pre><code>
[X] Feature request
[ ] Bug report  
</code></pre>

## The problem
Get a cookie to invoke API from `curl` without having to use a browser.

## Expected behavior
My current implementation:
```bash
16:27 $ /data/repositories/work/syndesis/syndesis/tools/bin/syndesis dev --cookie
Cookie created at /home/pantinor/syndesis_cookie.txt
Sample invocation: curl -k --cookie /home/pantinor/syndesis_cookie.txt https://myproject.192.168.42.202.nip.io/api/v1/connections

16:32 $ curl -k --cookie /home/pantinor/syndesis_cookie.txt https://myproject.192.168.42.202.nip.io/api/v1/connections
{"totalCount":1,"items":[{"connectorId":"sql","icon":"fa-database","description":"Connection to SampleDB","id":"5","tags":["sampledb"],"name":"PostgresDB","configuredProperties":{"password":"»ENC:adee434d4afeb6c39e280e074b4d396e32ddea59417cddfa732dcac2e13b774d5167551867251ef0fd3cfa53251a15fa","schema":"sampledb","url":"jdbc:postgresql://syndesis-db:5432/sampledb","user":"sampledb"},"isDerived":false}]}
```

paoloantinori:
fixes #2180 
Mind that I have tested it only on Fedora. I know that MacOSX `sed` and `grep` might support different flags


tools/bin/commands/util/openshift_funcs
```diff
@@ -222,3 +222,42 @@ guess_route() {
     fi
     echo "${project}.${domain#*.}"
 }
+
+create_api_cookie() {
```
rhuss: I wouldn't put it into openshift_funcs as it is minishift specific. This should be mentioned in the help message, too.

tools/bin/commands/util/openshift_funcs
```diff
@@ -222,3 +222,42 @@ guess_route() {
...
+
+    local COOKIE1="$HOME/syndesis_cookie.txt"
```
rhuss: Why hard code to a top-level file in the user's home directory ?

tools/bin/commands/util/openshift_funcs
rhuss: Also, I guess this is really only used for the `dev` command so happy to keep it there. Only when it needs to be shared it should be moved to the shared lib

tools/bin/commands/util/openshift_funcs
```diff
@@ -222,3 +222,42 @@ guess_route() {
...
+    rm -rf "$COOKIE1"
+    touch "$COOKIE1"
+
+    local SYNDESIS_URL="$(oc get route syndesis  --template={{.spec.host}})"
+    # first invocation is needed just to get the redierect url and to define a csrf token
+    local REDIRECTED_URL="$(curl --write-out "%{url_effective}\n" --location --silent --show-error --output /dev/null --insecure --cookie-jar  "$COOKIE1" -v "https://$SYNDESIS_URL" 2>/dev/null )"
+
+    # isolate required params
+    local LOGIN_URL="$(echo $REDIRECTED_URL | sed -e 's/?/\n/g' | head -n1)"
+    local THEN_URL="$(echo $REDIRECTED_URL | sed -e 's/=/\n/g' | tail -n1)"
```
rhuss: are you sure that the parameter value you want to pick up is always last ? Looks quite fragile to me, maybe better to use a regexp on the parameter key with a capturing group around the value ?

tools/bin/commands/util/openshift_funcs
```diff
@@ -222,3 +222,42 @@ guess_route() {
...
+    local CSRF=$(cat "$COOKIE1"  | grep -Po '(?<=\scsrf).*' | awk '{$1=$1;print}')
+
+    # login on OpenShift and save output to TEMP_STATE location
+    local TEMP_STATE="$(mktemp /tmp/syndesis_state.XXX.html)"
+    curl --referer "$REDIRECTED_URL" --insecure --location  -H "Cookie: csrf=$CSRF" --data-urlencode username="$USERNAME"  --data-urlencode password="developer" --data-ascii then="$THEN_URL" --data-urlencode csrf="$CSRF" --cookie-jar "$COOKIE2" -i -v -o "$TEMP_STATE" "https://$(minishift ip):8443/login"  2>/dev/null
```
rhuss: Not sure why use use `minishift ip` when you have $SYNDESIS_URL

tools/bin/commands/util/openshift_funcs
```diff
@@ -222,3 +222,42 @@ guess_route() {
...
+
+    # smoke test showing how to use the cookie
+    echo "Cookie created at $COOKIE1"
```
rhuss: Couldn't we just echo the COOKIE and let the user decide where to store it ? 

I mean using 

`curl -k --cookie $(syndesis dev --cookie) https://myproject.192.168.42.202.nip.io/api/v1/connections`

is much nicer and much more unixy than going over files. If you want to cache the cookie, then you could still do by letting syndesis dev pick up a cached copy (preferably a .dot file)

You could have

```
syndesis dev --cookie
```
to pick up an existing cookie under `~/.syndesis_api_cookie` or fetch the cookie

and a 

```
syndesis dev --cookie --refresh
```

to forcly fetch the cookie.

tools/bin/commands/util/openshift_funcs
```diff
@@ -222,3 +222,42 @@ guess_route() {
     fi
     echo "${project}.${domain#*.}"
 }
+
+create_api_cookie() {
+
+    local COOKIE1="$HOME/syndesis_cookie.txt"
+    # second cookie needed because curl flushes old values upon redirect. Might be a bug or some obscure flag missing
+    local COOKIE2="$(mktemp /tmp/syndesis_cookie.XXX.txt)"
```
rhuss: could we use lowercase vars for local variable and resever uppercase vars for exported env vars ? (its just that the other commands do it that way)

tools/bin/commands/util/openshift_funcs
```diff
@@ -222,3 +222,42 @@ guess_route() {
...
..." --cookie "$COOKIE1" -i -v --data "then=..%2Fauthorize%3Fapproval_prompt%3Dforce%26client_id%3Dsystem%253Aserviceaccount%253Amyproject%253Asyndesis-oauth-client%26redirect_uri%3Dhttps%253A%252F%252F$SYNDESIS_URL%252Foauth%252Fcallback%26response_type%3Dcode%26scope%3Duser%253Ainfo%2Buser%253Acheck-access%26state%3D$STATE%253A%252F&csrf=$CSRF&client_id=system%3Aserviceaccount%3Amyproject%3Asyndesis-oauth-client&user_name=$USERNAME&redirect_uri=https%3A%2F%2F$SYNDESIS_URL%2Foauth%2Fcallback&scope=user%3Ainfo&scope=user%3Acheck-access&approve=Allow+selected+permissions"  -H 'dnt: 1' -H "referer: https://$(minishift ip):8443/oauth/authorize/approve?client_id=system%3Aserviceaccount%3Amyproject%3Asyndesis-oauth-client&redirect_uri=https%3A%2F%2F$SYNDESIS_URL%2Foauth%2Fcallback&scope=user%3Ainfo+user%3Acheck-access&then=..%2Fauthorize%3Fapproval_prompt%3Dforce%26client_id%3Dsystem%253Aserviceaccount%253Amyproject%253Asyndesis-oauth-client%26redirect_uri%3Dhttps%253A%252F%252F$SYNDESIS_URL%252Foauth%252Fcallback%26response_type%3Dcode%26scope%3Duser%253Ainfo%2Buser%253Acheck-access%26state%3D$STATE%253A%252F" "https://$(minishift ip):8443/oauth/authorize/approve" -o /dev/null 2>/dev/null
```
rhuss: Also $(minishift ip) here.

rhuss: isn't `client_id=system` very specific to your setup ? Also, `serviceaccount=myproject` smells really like minishift only. And only when installed in project `myproject`. Many chose another project name, too (you might use $(oc project) at that spot.

The data looks very complex and not sure how robust this is.

tools/bin/commands/util/openshift_funcs
```diff
@@ -222,3 +222,42 @@ guess_route() {
...
+
+    # smoke test showing how to use the cookie
+    echo "Cookie created at $COOKIE1"
+    echo Sample invocation: curl -k --cookie "$COOKIE1" "https://$(oc get route syndesis --template={{.spec.host}})/api/v1/connections"
```
rhuss: That would need to go away if we are going to print out the cookie to stdout like suggested above.

tools/bin/commands/util/openshift_funcs
rhuss: That way the user also don't have to remember where the cookie file is created (its an implementation details of the command)

paoloantinori:
Hi @rhuss thanks for your feedback. I agree with many of your points, I will address them.

tools/bin/commands/util/openshift_funcs
paoloantinori: syndesis url is, syndesis url :wink:  

it includes the defined route.

`minishift ip` just gives you the ip. they are 2 different endpoints.

paoloantinori:
```bash
23:18 $ curl -k --cookie $(/data/repositories/work/syndesis/syndesis/tools/bin/syndesis dev --cookie --refresh )  "https://$(oc get route syndesis  --template={{.spec.host}})/api/v1/connections"
{"totalCount":1,"items":[{"connectorId":"sql","icon":"fa-database","description":"Connection to SampleDB","id":"5","tags":["sampledb"],"name":"PostgresDB","configuredProperties":{"password":"»ENC:adee434d4afeb6c39e280e074b4d396e32ddea59417cddfa732dcac2e13b774d5167551867251ef0fd3cfa53251a15fa","schema":"sampledb","url":"jdbc:postgresql://syndesis-db:5432/sampledb","user":"sampledb"},"isDerived":false}]}

23:18 $ curl -k --cookie $(/data/repositories/work/syndesis/syndesis/tools/bin/syndesis dev --cookie  )  "https://$(oc get route syndesis  --template={{.spec.host}})/api/v1/connections"
{"totalCount":1,"items":[{"connectorId":"sql","icon":"fa-database","description":"Connection to SampleDB","id":"5","tags":["sampledb"],"name":"PostgresDB","configuredProperties":{"password":"»ENC:adee434d4afeb6c39e280e074b4d396e32ddea59417cddfa732dcac2e13b774d5167551867251ef0fd3cfa53251a15fa","schema":"sampledb","url":"jdbc:postgresql://syndesis-db:5432/sampledb","user":"sampledb"},"isDerived":false}]}
```

currently working only on minishift and not on fuse-ignite, but I think we can find a way later on

tools/bin/commands/dev
```diff
@@ -1,15 +1,102 @@
 #!/bin/bash
 
+COOKIE1="$HOME/.syndesis_api_cookie"
+
+
+# helper python command to urlencode
+dev::urlencode() {
+
+local py_script="
+import sys, urllib as ul
```
rhuss: can we please avoid using Python ? I really would like to keep it a plain bash script for portability reasons. 

We deliberately choose bash as platform as the least common denominator which everyone can work on. Actually would have loved to implement it in Perl or Golang as these are much more powerful languages.

Shell is painful, true. But please let keep the pain pure.

tools/bin/commands/dev
```diff
@@ -1,15 +1,102 @@
...
+
+# helper python command to unescape html entities
+dev::xml_unescape() {
+
+local py_script="
+import sys, HTMLParser
+parser = HTMLParser.HTMLParser()
+print parser.unescape(sys.argv[1])
+"
+python -c "$py_script" $1
```
rhuss: Please no python.

tools/bin/commands/dev
```diff
@@ -1,15 +1,102 @@
...
+    variable="$(dev::xml_unescape $variable)"
+    variable="$(dev::urlencode $variable)"
+    echo "$variable"
+}
+
+dev::create_api_cookie() {
+
+    
+    # second cookie needed because curl flushes old values upon redirect. Might be a bug or some obscure flag missing
+    local cookie2="$(mktemp /tmp/syndesis_cookie.XXX.txt)"
```
rhuss: why do you use an absolute dir for mktemp ? Couldn't we just leave it to the OS where to put the tempdir ? (OS X its a different temp dir, although /tmp is alsow available)

tools/bin/commands/dev
```diff
@@ -1,15 +1,102 @@
...
+
+    # isolate required params
+    # split on question mark, so we are sure the first part is always the correct url
+    local login_url="$(echo $redirect_url | sed -e 's/?/\n/g' | head -n1)"
+    # split on question marks or ampersend, than filter on the specific param we are interested into
+    local then_url="$(echo $redirect_url | sed -e 's/[?&]/\n/g' | grep "then=" | sed -e 's/then=//' )"
+    local csrf=$(cat "$COOKIE1"  | grep -Po '(?<=\scsrf).*' | awk '{$1=$1;print}')
+
+    # login on OpenShift and save output to delegate_permissions_form location
+    local delegate_permissions_form="$(mktemp /tmp/syndesis_permissions.XXX.html)"
```
rhuss: dito wrt absolute path

tools/bin/commands/dev
```diff
@@ -24,4 +111,13 @@ dev::run() {
         local pod=$(oc get -o name pod -l component=syndesis-${name})
         oc port-forward ${pod//*\//} 5005:5005
     fi
+
+    if [ -n "$(hasflag --cookie)" ]; then
+        if [ -n "$(hasflag --refresh)" ]; then
+             dev::create_api_cookie
+        fi
+        echo $COOKIE1
```
rhuss: you should move this up and add an exit to in `--debug` as both options are orthogonal (as `oc portfoward` also spits on standardout and it doesn't make sense to allow both).

tools/bin/commands/dev
paoloantinori: Python is a pre-requisite on RHEL, since its package manager is written with that language, and is installed by default on MacOSX.
Not present on Windows, likewise bash.

An eventual additional package to implement urlencode from command line has way less chances to be installed on a platform than python.

golang is just not scripting, and I would never use it as a replacement for scripting

rhuss: I was afraid that you start arguing like that.

Ok, so I have to tell the long story. It's **not** about whether Python is installed or not. Perl is installed, too, everywhere. It's about respecting the team and allow as many people as possible to maintain this script.

When I started the script I seriously considered to write it in Perl because Perl is my mother language. I can do everything in Perl, super fast. "syndesis" would be much more powerful, we could do everything. 

But I would be the only one who could maintain it. So, after some discussion, e.g. with @iocanel, we bite in the the sour apple to implement everything in Shell. Not that I'm a big fan of shell scripting, not at all. But because *everyone* can understand it (or is supposed to at least). The challenge of course was to write sustainable shell script (where I learned quite some lessons, so many, that I'm even submitting CFP to conference for how to teach 'idiomatic shell' scripting).

For Python the same is true as for Perl. I don't know Python at all, at not many do. So you would be alone to maintain it.

So please, let's stick to plain shell. Don't open the door to Python or any other scripting language (why wouldn't we then have chosen Python from the beginning ?)

tl;dr - its not a technical reason, its a cultural and social reason to not introduce Python. I think we should be able to do it with plain UNIX commands.

(and what, btw, will happen if someone has Python 3 installed instead of Python 2 (e.g. with brew). Are you user urllib works here the same ? As they have removed urllib2 and are not really famous to be backwards comptabible ....)

tools/bin/commands/dev
rhuss: Does this work for Python 3, too ? Isn't it `import html.parser` here ?

tools/bin/commands/dev
zregvart: seems that urldecoding can be simply done in bash with `decoded=$(printf '%b' "${input//%/\\x}")` according to https://unix.stackexchange.com/questions/159253/decoding-url-encoding-percent-encoding

rhuss: * https://stackoverflow.com/a/10660730/207604 (not saying thats the best solution, just the first hit)

```
rawurlencode() {
  local string="${1}"
  local strlen=${#string}
  local encoded=""
  local pos c o

  for (( pos=0 ; pos<strlen ; pos++ )); do
     c=${string:$pos:1}
     case "$c" in
        [-_.~a-zA-Z0-9] ) o="${c}" ;;
        * )               printf -v o '%%%02x' "'$c"
     esac
     encoded+="${o}"
  done
  echo "${encoded}"    # You can either set a return variable (FASTER) 
  REPLY="${encoded}"   #+or echo the result (EASIER)... or both... :p
}
```

and now for the devils advocate argument, if not shell why not

```perl
value="$(perl -MURI::Escape -e 'print uri_escape($ARGV[0]);' "$1")"
```

rhuss: > seems that urldecoding can be simply done in bash with decoded=$(printf '%b' "${input//%/\\x}") according to unix.stackexchange.com/questions/159253/decoding-url-encoding-percent-encoding

I think its more about encoding then decoding. Though not sure why because you could always use `curl --data-urlencode`. @paoloantinori isn't that good enough when sending it back via curl ?

tools/bin/commands/dev
```diff
@@ -1,15 +1,102 @@
...
+    scopes=(${scopes// / })
+    for i in "${!scopes[@]}"
+    do
+        local encoded="$(dev::urlencode ${scopes[i]})"
+        joined_scopes="$joined_scopes --data scope=$encoded"
+    done
+
+    # approve permission delegation, weaving all the params in the GET parameters
+    
+    curl --insecure --location --cookie-jar "$COOKIE1" --cookie "$COOKIE1" -i -v --data "then=$then_url" --data "csrf=$csrf" --data "client_id=$client_id" --data "user_name=$user_name" --data "redirect_uri=$redirect_uri" $joined_scopes --data "approve=Allow+selected+permissions"  -H 'dnt: 1'  "https://$openshift_ip_and_port/oauth/authorize/approve" -o /dev/null 2>/dev/null
```
rhuss: @paoloantinori can't you use `--data-urlencode` here for the parameters and safe the custom encoding on your own ?

tools/bin/commands/dev
```diff
@@ -1,15 +1,102 @@
...
+    local redirect_uri=$(extract_and_escape "redirect_uri" $delegate_permissions_form )
+
+    local scopes=$(grep -Po '(?<=name="scope").*' $delegate_permissions_form | grep -Po '(?<=value=")[^"]+')
+    local joined_scopes=""
+    
+    scopes=(${scopes// / })
+    for i in "${!scopes[@]}"
+    do
+        local encoded="$(dev::urlencode ${scopes[i]})"
+        joined_scopes="$joined_scopes --data scope=$encoded"
```
rhuss: --data-urlencode instead of --data and avoiding custom encoding ?

rhuss:
I see. Interesting way to stop a review discussion.

PR #2190, opened 2018-03-29
New doc for HTTP/S and Slack connectors and some other updates

TovaCohen:
This is ready to merge. I need to bring it downstream so it will be in the TP4 doc. Thanks in advance for making sure this merges!


TovaCohen:
Oh wow. Someone gave me administrator privileges. Thank you!

Issue #2104, opened 2018-03-21
CSRF exploit using flash plugin


chirino:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
CSRF exploit using flash plugin.

## Expected behavior
3rd party pages not being able to use Syndesis APIs.

## Tasks involved / Steps to Reproduce

Reported by Andrej Vano.   He stated:

> "Basic" CSRF attempts are blocked by CORS, but this method is able to bypass the CORS settings using a remote server and one flash file.
> 
> Here is the example that can be used against Syndesis (case 2): https://www.geekboy.ninja/blog/exploiting-json-cross-site-request-forgery-csrf-using-flash/ .
> 
> For the demonstration I used only the json "{"name":"attacker"}" that was sent to https://app-proj232514.6a63.fuse-ignite.openshiftapps.com/api/v1/connections that ended up creating a new connector with name "attacker".
> see:  https://www.geekboy.ninja/blog/exploiting-json-cross-site-request-forgery-csrf-using-flash/

chirino:
We should use some specific header that will be sent with each [request](https://security.stackexchange.com/questions/23371/csrf-protection-with-custom-headers-and-without-validating-token).  Angular already [has support](https://angular.io/api/http/CookieXSRFStrategy) for this.

deeleman:
> Angular already [has support](https://angular.io/api/http/CookieXSRFStrategy) for this.

The linked resource was deprecated back in Angular 4.2, since all the valdiation is now performed in a generic way by the new Angular HttpClient for the entire application.  Requests make use of a custom `XSRF-TOKEN` cookie that needs to be set by the backend (at the OPTIONS handshake for instance) and will be returned in any call performed by the `HttpClient` class as a `X-XSRF-TOKEN`.

It will be up to the backend to define the cookie name and the header name where the reference will be attached on each request. The UI just requires those names, should they are customized.

Once done, this needs to be attached to the `AppModule` imports array:

```typescript
HttpClientXsrfModule.withOptions({
  cookieName: 'Syndesis-Xsrf-Cookie',
  headerName: 'Syndesis-Xsrf-Header',
})
```

Problem is that I have no idea how to apply this logic to Restangular ;P

deeleman:
[...]

PR #2196, opened 2018-03-29
chore (UI): LOOK MA! No Restangular! 

### LINKED ISSUE #2188: Remove our dependency on Restangular
## This is a...

<pre><code>
[x] MUST!
[ ] Feature request
[ ] Bug report  
</code></pre>

## The problem
Angular comes already with a powerful Http client that allows to model all verbs required for interacting with a REST API thru XHR. In fact, Restangular is built on top of such API with a little nuance though: Restangular makes use of an outdated (and VERY deprecated) version of the API.

Angular introduced a new version of the Http API back in v4.4.5, providing additional features that we ~will want to leverage in the near future~ need now. Restangular's _abstraction_ does not even provide such functionalities and  therefore is turning into a block more than a helper. 

A good example of this is that we cannot leverage the new Angular `HttpInterceptor` interface and related providers, which will allow for gaining a better control on custom headers, mock backends and stuff.

## Expected behavior
All XHR interactions should be handled natively with Angular and those specific to consume API endpoints should leverage our very own `ApiHttpClient` class and related types and helpers.

deeleman:
Fixes #2188 

Restangular is not actually needed in this project, adds a level of "abstraction" (sic) that only obfuscates the data logic of our services and clutters NgModules with unnecessary boilerplate.

**This PR entails:**

- Removing all instances of Restangular app-wide, replacing it by implementations of our very own `ApiHttpService`.
- `ApiHttpService` now implements a `.patch(body: any)` method to address the REST PATCH verb, (obviously).
- Adding a new `TestApiModule` stub module that we can use in our unit tests to wrap a mock implementation of `ApiModule`, providing builtin support for `HttpClientModule` so we will not need to import it in our tests anymore. Its use case allows for handling circular references when importing `ApiModule.forRoot()` in our tests leads to a circular reference.
- Tests decluttered of unused dependencies and useless tokens.
- Commented tests reinstated. Whoever did that (you know who you are) deserves to burn in hell for the rest of eternity. ;P

This PR also clears the way to use the new capabilities of `HttpClientModule` across the project, allowing to address low level issues such as #2104

### DIFF
--- a/app/ui/package.json
+++ b/app/ui/package.json
@@ -53,7 +53,6 @@
     "ng2-file-upload": "1.3.0",
     "ngx-bootstrap": "^2.0.2",
     "ngx-chips": "^1.6.2",
-    "ngx-restangular": "2.0.2",
     "patternfly": "3.26.1",
     "patternfly-ng": "^3.0.0",
     "pluralize": "5.0.0",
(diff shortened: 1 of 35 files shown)


kahboom:
Just looked this over now, great job @deeleman !! 👍

PR #2207, opened 2018-03-29
feature(ui): Full i18N support based on Redux and JSON dictionary files

### LINKED ISSUE #1726: Remove hard coded strings from UI and extract from doc source files instead
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## The problem

The problem is that currently all of the text used within the UI is hard coded in template/controller files. This blocks us from doing two important things

1) Driving tooltip text from the actual docs as made possible in https://github.com/syndesisio/syndesis/pull/1369

2) Supporting internationalization and localization

This issue is an evolution of https://github.com/syndesisio/syndesis/issues/295, and is meant to encompass all the site text, and not just tooltip content. Having all the language used in the site scattered across many files makes updating text time consuming and difficult.

## Expected behavior

The UI should never use literal text strings, but instead use a mechanism (probably a language service) that allows one to specify string **keys**, which in turn generate the appropriate text at runtime.

This will help improve consistency of the language used within the app and also speed up time required to update text on the site by moving power of control over to those who should have it in the first place.

## Related Issues
https://github.com/syndesisio/syndesis-documentation/issues/20
https://github.com/syndesisio/syndesis-documentation/issues/23
https://github.com/syndesisio/syndesis/issues/295

## API Endpoints and Schemas

`doc/tooltips/tooltips.json`

1. Move hard coded tooltip text into `doc/tooltips/tooltips.json` and negotiate key names with @TovaCohen 
2. Use mvn plugin tooling to generate tooltips.json so DOC folks then have finite control
3. Repeat this process for all site text, beyond just tooltips
4. Confirm what languages/locals we intend to support
5. Make a decision on whether or not we intend to support RTL languages
6. Address styling issues that come up, for example German translations are often much longer character strings than their English counterparts. The UI may not currently handle this flawlessly.

I can see number 3 above becoming quite a large work item, so it may make sense to turn this into an Epic and break the task down into more manageable chunks. I'll let someone else make the call as to how best we can divide up this work.

deeleman:
Fixes #1726 

For those who hate reading here you have a POC: https://app.hyfy.io/v/abZ6K38V1On/

(HYFY slows down my laptop quite a bit, that's why the reloading takes so long, but proves how I18N preferences are persisted)

This PR contains a fully-functional redux implementation for supporting I18N throughout the site. By means of our own custom I18N template bindings, we can centralize text management in ISO-locale JSON files so no hardcoded text will be found anymore in template views nor component controllers.

The case for this change is:

- Help content writers and managers edit text in a convenient fashion
- Set the ground for localizing the product in foreign languages
- Serve as a POC to showcase the power of Redux for handling and refreshing state in an entire application in a hi-performant way.

There is some side documentation in the works but this is pretty much the basics:

### Dictionary files live in `assets/dictionary`, named following the standard ISO locale convention. Eg: `en-GB.json`:

```json
{
  "locale": "en-GB",
  "dictionary": {
    "integrations": {
      "integration": "Integration",
      "integrations": "Integrations"
    },
    "dashboard": {
      "header": "Dashboard",
      "metrics": {
        "sysmetrics": "System Metrics",
        "uptime": "Uptime",
        "integrationsheader": "My {{0}} {{integrations.integrations}}"
      }
    },
    "shared": {
      "loading": "loading",
      "project": {
        "name": "Syndesis"
      }
    }
  }
}
````

### Values in the dictionary can be bound in templates like:

```html
<h1 class="col-xs-6 syndesis-dashboard__title">
    {{ 'dashboard.metrics.sysmetrics' | synI18n }}
</h1>
```

The engine will traverse the binding elements and will search for that value hierarchically in the JSON file: `dashboard > metrics > sysmetrics: "System Metrics"`.

With the data provided, the above will render:

```html
<h1 class="col-xs-6 syndesis-dashboard__title">
    System Metrics
</h1>
```

### Dictionary values can allocate references to other keys or indexed placeholders

```json
{
  "integrationsheader": "{{0}} {{integrations.integrations}}"
}
```

The above will expect a value to be passed when binding the dictionary key in the template and will also perform a lookup of the value corresponding to `integrations.integrations` like `integrations > integrations: "My Integrations"` and replace it on the output string. Hence:

```html
<h1>{{ 'dashboard.metrics.integrationsheader' | synI18n: 15 }}</h1>
```

will render:

```html
<h1>My 15 Integrations</h1>
```

This allows for creating complex translations with values generated on runtime.

### Date time configuration is localized too
A Redux side effect model take care of updating MomentJs locale preferences in a transparent fashion, being immediately available site-wide.

### Refreshing state is trivial

It just takes a call to `store.dispatch(new I18NActions.I18NFetch('it-IT'))` to refresh the state immediately, once the underlying JSOn file is loaded, with no document reload.

### State preferences are persisted
Once the state preference is refreshed, the chosen locale code is persisted, so user will find the product already localized next time he loads Syndesis.

## TODO:

- Finish documentation and a guide for UI developers (basically all the above) and content editors, with best practices for harnessing the power of the JSON dictionary while keeping it lean with a smart use of nested keys.
- Fix Angular locale registry, so we do not only localize the Redux state and MomentJs, but the Angular pipes setup.
- Strip all hardcoded texts from components and views, and also ensure we do not get hardcoded text from the backend. Move all these strings to JSON.
- Localize it in another language, if desired.


seanforyou23:
The DurationDiffPipe and I18NPipe are both exposed through the SyndesisCommonModule. But when its listed as an import in the DashboardMetricsComponent's test, we get a StaticInjectorError "No provider for Store!" thingie. I'll keep looking for fix, but seems like we've hit this one before.

deeleman:
This can be easily fixed @seanforyou23. Leave this up to me and I’ll submit a new commit today with an updated test implementation and the final documentation.

PR #2226, opened 2018-04-04
chore(build): upgrade PMD

### LINKED ISSUE #2209: `syndesis build` executed on *fuse ignite 1.3* release fails
## This is a...

<pre><code>
[ ] Feature request
[X ] Bug report  
</code></pre>

## The problem

`syndesis build` executed on *fuse ignite 1.3* release fails:

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-pmd-plugin:3.8:check (default) on project integration-project-generator: You have 1 PMD violation. For more details see: /hong/syndesis-fuse-ignite-1.3/app/integration/project-generator/target/pmd.xml

Contents on pmd.xml file --
https://gist.github.com/honghuac/eb14dfeb8601311c226323178eae28f6

zregvart:
This upgrades PMD Maven plugin to 3.9.0 and PMD version to 6.2.0 in an
attempt to try to fix a race issue when running PMD check from Maven
multiple times without performing a clean.

This unfortunately does not help to resolve the issue, and I had to
disable PMD cache.

This fixes PMD issues brought about with the PMD upgrade. Seems that
some of the issues were not reported with the previous version.

Removes the `break` that would prevent any additional integrations in
the submitted export ZIP file from being imported.

Fixes #2209, #2240

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/support/IntegrationSupportHandler.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/support/IntegrationSupportHandler.java
@@ -61,7 +61,6 @@ import io.syndesis.integration.api.IntegrationProjectGenerator;
 import io.syndesis.integration.api.IntegrationResourceManager;
 import io.syndesis.server.dao.file.FileDataManager;
 import io.syndesis.server.dao.manager.DataManager;
-import io.syndesis.server.endpoint.v1.handler.connection.ConnectionHandler;
 import io.syndesis.server.endpoint.v1.handler.integration.IntegrationHandler;
 import io.syndesis.server.jsondb.CloseableJsonDB;
 import io.syndesis.server.jsondb.JsonDB;
@@ -92,7 +91,6 @@ public class IntegrationSupportHandler {
     private final DataManager dataManager;
     private final IntegrationResourceManager resourceManager;
     private final IntegrationHandler integrationHandler;
-    private final ConnectionHandler connectionHandler;
     private final FileDataManager extensionDataManager;
 
     public IntegrationSupportHandler(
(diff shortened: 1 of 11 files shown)


ruleset.xml
```diff
@@ -386,7 +387,7 @@
       <property name="minimum" value="3" />
     </properties>
   </rule>
-  <rule ref="rulesets/java/naming.xml/MisleadingVariableName">
+  <rule ref="rulesets/java/naming.xml/MIsLeadingVariableName">
```
gashcrumb: Is this a typo or is this on purpose?

zregvart: Yeah, a typo, but on PMD's side I'm afraid :cry:

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/support/IntegrationSupportHandler.java
```diff
@@ -336,11 +332,10 @@ private  void importIntegrations(SecurityContext sec, JsonDbDao<Integration> exp
                 integrationHandler.update(id, builder.version(previous.getVersion()+1).build());
                 result.add(ChangeEvent.of("updated", integration.getKind().getModelName(), id));
             }
-            break;
```
zregvart: I think this `break` here is a bug that would prevent multiple integrations to load from an export. FYI @chirino @lburgazzoli

lburgazzoli: do we have multiple integration in an export ?

zregvart: This seems to add multiple to the exported ZIP file:

https://github.com/syndesisio/syndesis/blob/c9760574058f36a8ecd7c8f4d81c39b9f7e32a53/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/support/IntegrationSupportHandler.java#L155-L162

lburgazzoli: so maybe open a new issue :)

Issue #2249, opened 2018-04-05
Long connection name UI issues in integration details


michael-coker:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
* Long connection names cause issues in the integration details page

## Expected behavior
* Truncate long connection name in integration details at 1 line with title set for hover/accessibility to read the value.

## Screenshot

![screen shot 2018-04-05 at 9 30 27 am](https://user-images.githubusercontent.com/35148959/38373980-2eb4a7a4-38b8-11e8-80f9-71f6947c6039.png)

cc @dongniwang @sjcox-rh

PR #2258, opened 2018-04-05
truncate long connection names under integration details icons

michael-coker:
fixes https://github.com/syndesisio/syndesis/issues/2249

also snuck in a fix for a small bug where the empty state card for connections when you're creating an integration doesn't share the same bootstrap `col-*` classes as the rest of the cards so it was misaligned.

### DIFF
--- a/app/ui/src/app/integration/integration_detail/integration-description.component.html
+++ b/app/ui/src/app/integration/integration_detail/integration-description.component.html
@@ -1,15 +1,18 @@
 <section class="integration-description">
-  <p *ngIf="integration.steps">
+  <div *ngIf="integration.steps">
     <ng-template ngFor let-step [ngForOf]="integration.steps" let-index="index" let-last="last">
       <div *ngIf="step" class="step-block inline-block text-center">
         <ng-container [ngSwitch]="step.stepKind">
           <ng-container *ngSwitchCase="'endpoint'">
-            <div id="{{ step.connection.name | synSlugify }}" class="text-center connection" (click)="onViewDetails(step)" title="{{ step.connection.name }}&nbsp;{{ step.action.name }}">
+            <div id="{{ step.connection.name | synSlugify }}"
+              class="text-center connection" 
+              (click)="onViewDetails(step)" 
+              title="{{ step.connection.name }}&nbsp;{{ step.action.name }}">
               <div [class]="'step-line ' + getStepLineClass(index)"></div>
               <div class="icon">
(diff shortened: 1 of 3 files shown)


app/ui/src/app/integration/integration_detail/integration-description.component.html
```diff
@@ -25,7 +28,7 @@
         <span *ngIf="!last" class="step-sep fa fa-angle-right"></span>
       </div>
     </ng-template>
-  </p>
+  </div>
```
deeleman: You can make all this block leaner by removing redundant structural container directives:

```html
  <div *ngFor="let step of integration?.steps; let index=index, last=last" class="step-block inline-block text-center">
    <ng-container [ngSwitch]="step.stepKind">
      <div *ngSwitchCase="'endpoint'"
        id="{{ step.connection.name | synSlugify }}"
        class="text-center connection" 
        (click)="onViewDetails(step)" 
        title="{{ step.connection.name }}&nbsp;{{ step.action.name }}">
        <div [class]="'step-line ' + getStepLineClass(index)"></div>
        <div class="icon">
          <img class="syn-icon-small" [src]="step.connection | synIconPath">
        </div>
        <div class="syn-truncate__ellipsis">{{ step.connection.name | capitalize }}</div>
      </div>
      <div *ngSwitchDefault id="{{ (step.name || step.stepKind) | synSlugify }}" class="text-center step">
        <div class="step-line"></div>
        <div class="icon not-connection">
          <div class="icon-line"></div>
        </div>
        <div>{{ stepStore.getStepName(step) }}</div>
      </div>
    </ng-container>
    <span *ngIf="!last" class="step-sep fa fa-angle-right"></span>
  </div>
```

Please note how `*ngFor` or `*ngSwitchCase` have been applied directly on HTML elements, reducing the overall template boilerplate. ;)

michael-coker: Thanks! Would I still need `<div *ngIf="integration.steps">` wrapping it?

gashcrumb: I kinda feel like it's a safer bet but technically I think the `*ngFor` shouldn't execute if `steps` is undefined.  So you can probably leave it out.

But as you can see we like to make no assumptions as to what we get back for an API request :-)

michael-coker: thanks guys! updated the branch.

PR #2288, opened 2018-04-09
chore(ui): Customer portal user doc links updated and and managed thru i18n

### LINKED ISSUE #2278: Update URLs for links to customer portal user doc
## This is a...

<pre><code>
[ ] Feature request
[x ] Bug report  
</code></pre>

## The problem
The GA URLs for  Ignite user documentation on the Red Hat customer portal will be different than they were for the technology preview releases. The URLs no longer contain "jboss" and no longer contain "-tp". 

## Expected behavior
For GA, these are the correct URLs:

https://access.redhat.com/documentation/en-us/red_hat_fuse/7.0/html-single/integrating_applications_with_ignite/

https://access.redhat.com/documentation/en-us/red_hat_fuse/7.0/html-single/ignite_sample_integration_tutorials/

@fbolton could you please confirm that these URLs will be correct for GA? Thanks!

In the UI, the URLs need to be updated for
- Selecting Help > Sample Integration Tutorials 
- Selecting  Help > User Guide
- Settings  page text ".. See the documentation for help."  This URL will be:

https://access.redhat.com/documentation/en-us/red_hat_fuse/7.0/html-single/integrating_applications_with_ignite/#obtaining-authorization-to-access-applications

deeleman:
Fixes #2278 

@seanforyou23 Please note how the dictionary file is also a pretty convenient place to store information such as email and links, so those can be localized and embedded into some other more complex strings without having to pass those as pipe parameters.

### DIFF
--- a/app/ui/src/assets/dictionary/en-GB.json
+++ b/app/ui/src/assets/dictionary/en-GB.json
@@ -20,6 +20,10 @@
       "customizations": "Customizations",
       "settings": "Settings"
     },
+    "links": {
+      "userguide": "https://access.redhat.com/documentation/en-us/red_hat_fuse/7.0/html-single/integrating_applications_with_ignite/{{0}}",
+      "tutorial": "https://access.redhat.com/documentation/en-us/red_hat_fuse/7.0/html-single/ignite_sample_integration_tutorials/"
+    },
     "shared": {
       "loading": "loading",
       "project": {
--- a/app/ui/src/app/app.component.html
+++ b/app/ui/src/app/app.component.html
@@ -76,13 +76,13 @@
                 id="helpDropdownMenu"
                 aria-labelledby="helpDropdown">
               <li>
-                <a [href]="tutorialLink"
+                <a [attr.href]="'links.tutorial' | synI18n"
                     rel="nofollow"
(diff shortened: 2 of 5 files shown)


app/ui/src/assets/dictionary/en-GB.json
```diff
@@ -20,6 +20,10 @@
       "customizations": "Customizations",
       "settings": "Settings"
     },
+    "links": {
+      "userguide": "https://access.redhat.com/documentation/en-us/red_hat_fuse/7.0/html-single/integrating_applications_with_ignite/{{0}}",
+      "tutorial": "https://access.redhat.com/documentation/en-us/red_hat_fuse/7.0/html-single/ignite_sample_integration_tutorials/"
```
TovaCohen: Just a question. I see that the userguide link has a placeholder at the end of it. I assume that's for jumping to a particular location in the user guide. And it must be okay if this gets replaced with null since the Help > User Guide link jumps to the top of the document. But the tutorial link does not have a placeholder at the end of it. If we ever want to link to a particular tutorial, and not just the "book", would that {{0}} placeholder need to be added? Should it just be added now?

Issue #2331, opened 2018-04-10
Don't secure the TODO app by default, it breaks API examples when the SSL cert is not valid on the cluster.


chirino:
## This is a...

<pre><code>
[X] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
Don't secure the TODO app by default, it breaks API examples when the SSL cert is not valid on the cluster.

## Expected behavior
TODO app API example should still be useable even on clusters where the SSL cert is not valid.

PR #2332, opened 2018-04-10
Fixes: #2331 Don't secure the TODO app by default, it breaks API examples when the SSL cert is not valid on the cluster.

chirino:


dsimansk:
Would it be valuable to keep both http and https enabled? With `insecureEdgeTerminationPolicy: Allow`?

zregvart:
@chirino what do you think about @dsimansk suggestion to have both `http` and `https` enabled for the TODO app?

chirino:
@dsimansk good idea.  Just updated.

Issue #1641, opened 2018-02-22
Integration scaled down using OpenShift is not noticed by Syndesis


tplevko:
When I scale down integration pod on OpenShift (simulating sort of issue, where the pod is not available), integration on syndesis remains in "Published" state. Syndesis backend does not notice, the integration is down. There should probably be some sort of integration liveness check on syndesis side, which would notify the user, the integration is not available.

The described behavior: 
![rest_ignores_integ_delete](https://user-images.githubusercontent.com/4180208/36539807-5220e4f6-17d8-11e8-835f-f23caee11b0a.gif)

paoloantinori:
ping @kcbabo

kcbabo:
Is it Published or Active?  Strictly speaking, the integration is still published if it's scaled down, it's just not active, right?

zregvart:
> Is it Published or Active? Strictly speaking, the integration is still published if it's scaled down, it's just not active, right?

What I think is the questions are:
 - does the integration pod's status needs to be periodically checked so we can present the status in the UI?
 - do we need a start/stop functionality?

dsimansk:
This may also happen on delete of DC for example. Tbh web console might be the only "monitoring tool" for user to relay on. Therefor it should be able to provide update about status of integration runtime. E.g. pod maybe restarted due to OOM, but node has not enough memory to schedule it again, or any other kind of resource depletion. But from our UI everything seems in order, that integration is still running fine.

paoloantinori:
might be already fixed.

tplevko:
Tried it again and I unfortunately don't see any change in the behavior.

iocanel:
We do periodically check that the `desired state` matches the `current state`. Though the `current state` is read from the database, rather than determined based on the state of the deployment. This is something we can change quite easily. The side effect of such change though, will be that syndesis will try to redeploy integrations that have been manually deleted or scaled down.

iocanel:
Fix has been merged

iocanel:
The commit has been reverted so, lets reopen the issue.

dsimansk:
@iocanel FYI revert PR is still in queue https://github.com/syndesisio/syndesis/pull/2389

iocanel:
Pull request that fixes regression has been merged. So let's close this.

tplevko:
As mentioned by Ioanis, this is already fixed. I also verified it and works as expected. Closing.

PR #2338, opened 2018-04-11
This pull request is supposed to provide a fix for #1641.

iocanel:
What it actually does, is that it periodically polls for the actual state of all integrations (not just the ones triggered by a change event.

As a result, any integration externally modfied (dc deleted or scaled down) will be redeployed by the system.


zregvart:
> As a result, any integration externally modfied (dc deleted or scaled down) will be redeployed by the system.

I think this should be fine, it's Syndesis that should manage the integration's pod lifecycle. It should be easy enough for the user to unpublish an integration.

gashcrumb:
I wanna approve this PR so bad :-)  This also fixes integrations being stuck in unpublishing to doesn't it?

iocanel:
@gashcrumb: yes it does.

PR #2342, opened 2018-04-11
fix: Avoid parsing the error response, as it's an object

### LINKED ISSUE #2315: Incorrect error message displayed when uploading an invalid swagger file
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem

When uploading an invalid swagger file it appears the API can either return an HTTP 200 with an `error` attribute *or* return an HTTP 500 error response with the following:

```
{"errorCode":500,"userMsg":"Please contact the administrator and file a bug report","developerMsg":"Internal Server Exception. Swagger specification does not provide a `schemes` definition and the Swagger specification was uploaded so the originating URL is lost to determine the scheme to use"}
```

In the latter case where the above response is received the wrong error message is shown in the UI:

screenshot 2018-04-10 09 52 58

Probably caught by a generic error handler and not correctly propagated back to the component.

This can be reproduced by trying to create an API client connector with this swagger file:

[kie-server-swagger.json.txt](https://github.com/syndesisio/syndesis/files/1894902/kie-server-swagger.json.txt)

@chirino FYI

gashcrumb:
fixes #2315 

Though we should swap the messages in the returned error object so the right string is shown to the user.

### DIFF
--- a/app/ui/src/app/api/providers/api-http-provider.service.ts
+++ b/app/ui/src/app/api/providers/api-http-provider.service.ts
@@ -150,7 +150,7 @@ export class ApiHttpProviderService extends ApiHttpService {
       .filter(requestEvent => requestEvent.type === HttpEventType.Response)
       .map(requestEvent => requestEvent as HttpResponse<T>)
       .map(requestEvent => requestEvent.body)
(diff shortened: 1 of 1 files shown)


gashcrumb:
#2349 will also contain this change, so let's just go with that PR

PR #2349, opened 2018-04-11
Feat/218 better error handling

### LINKED ISSUE #218: UX: better handling of unavailable backend
|</img> @dsimansk | [2017-03-23](https://github.com/syndesisio/syndesis-ui/issues/307) | enhancement, Priority - Low, UX Design  |
|-|-|-|

Currently if ipaas-rest is not responding for some reason (misconfig of auth, 503 pod not available etc.) There's just spinner spinning endlessly. We should probably have timeout to request and provide laundered error message to end user, that something is terribly wrong.

### LINKED ISSUE #2315: Incorrect error message displayed when uploading an invalid swagger file

deeleman:
Fixes #218 
Fixes #2315 too

This PR encompass the following new features:

- Application wide errors are now funneled through a centralized Exception Handler Service, allowing for easier logging and reporting.
- Http errors (HTTP 500 errors, API calls yielding a 404 error, etc) when communicating with the backend API are now handled and not propagated, ensuring that components keep being responsive, and also reported through a toast notification. This will enhance the overall UX on those (many) components/flows where no proper Http exception handling had been implemented.
- Network errors (WiFi lags, Internet downtimes) are now reported on realtime with a toast notification informing the user that his Internet connection has been interrupted or is back.

FYI, the error strings configured for each toast notification are fully managed through the i18n dictionary.

### DIFF
--- a/app/ui/src/app/api/providers/api-http-provider.service.ts
+++ b/app/ui/src/app/api/providers/api-http-provider.service.ts
@@ -20,6 +20,7 @@ import {
 import { ConfigService } from '@syndesis/ui/config.service';
 
 const DEFAULT_ERROR_MSG = 'An unexpected HTTP error occured. Please check stack strace';
+const PROPAGATE_EXCEPTIONS = false;
 
 @Injectable()
 export class ApiHttpProviderService extends ApiHttpService {
@@ -69,7 +70,7 @@ export class ApiHttpProviderService extends ApiHttpService {
     const url = this.getEndpointUrl(endpointKey, ...endpointParams);
     return this.httpClient
       .get(url, options)
-      .catch(error => Observable.throw(this.catchError(error)));
+      .catch(error => this.handleError(error));
   }
 
   post<T>(endpoint: string | any[], body?: any, options?: ApiRequestOptions | any): Observable<T> {
@@ -77,7 +78,7 @@ export class ApiHttpProviderService extends ApiHttpService {
     const url = this.getEndpointUrl(endpointKey, ...endpointParams);
     return this.httpClient
       .post<T>(url, body, options)
-      .catch(error => Observable.throw(this.catchError(error)));
+      .catch(error => this.handleError(error));
   }
 
   put<T>(endpoint: string | any[], body: any, options?: ApiRequestOptions | any): Observable<T> {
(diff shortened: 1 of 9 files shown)


app/ui/src/app/api/providers/api-http-provider.service.ts
```diff
@@ -150,7 +151,7 @@ export class ApiHttpProviderService extends ApiHttpService {
       .filter(requestEvent => requestEvent.type === HttpEventType.Response)
       .map(requestEvent => requestEvent as HttpResponse<T>)
       .map(requestEvent => requestEvent.body)
-      .catch(error => Observable.throw(this.catchError(JSON.parse(error.error))));
+      .catch(error => this.handleError(error.error));
```
gashcrumb: I'd suggest passing the whole error object here too...

gashcrumb: This is basically the change I had in #2342 let's just go with this PR for that.

deeleman: Problem is that, contrary to the other `httpClient` handlers, the object returned through the RxJS `catch` handler here is not the actual HTTP error, but a native Angular wrapper around it, typed as a `HttpErrorResponse` object. Therefore the properties will not match with those expected by the logic in `handleError()`.

More info: https://github.com/angular/angular/blob/5.2.9/packages/common/http/src/response.ts#L288-L329

gashcrumb: ah, okay gotcha didn't realize that, thanks!  Well then I'm all +1 on this PR then :-)

Issue #2276, opened 2018-04-06
Breadcrumb not in full length on create API connector wizard pages 


dongniwang:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

In Create API Connector wizard, the breadcrumb on top of the page is not in full length and it looks larger than the usual ones. Also, the page is not displaying the correct breadcrumb.

## Expected behavior

It should look like this: 
![image](https://user-images.githubusercontent.com/24943812/38438972-12dd900c-39aa-11e8-8519-3abc82fbd1b2.png)

Also, the correct breadcrumb should be:
Home > Customizations > API Client Connector 

## Screenshot

![image](https://user-images.githubusercontent.com/24943812/38438852-bb7563b2-39a9-11e8-846c-83c12732f918.png)

## Tasks involved / Steps to Reproduce

1. Create a new integration (I use twitter mention -> salesforce update record)
1. Add a data mapper step (I map id -> id)
1. Click "Customizations"
1. Click "Create API Connector" button

michael-coker:
What browser are you using? I tried in chrome, firefox, and safari (on a mac) and am not seeing it.

dongniwang:
@michael-coker I was using chrome. If it's not showing up now maybe it was just something random...thanks for looking into this.

KurtStam:
This appears to have been fixed, likely by #2294.

michael-coker:
If you go through the steps to reproduce, can you see the bug? I just tried it and still can. And if you notice in your screenshot, there is no bottom border to the breadcrumbs, and there should be (it's there on initial page load).

![screen shot 2018-04-16 at 4 34 55 pm](https://user-images.githubusercontent.com/35148959/38836557-32528aae-4194-11e8-8d04-817ccec875b9.png)

PR #2402, opened 2018-04-16
Breadcrumb not in full length on create API connector wizard pages #2276

KurtStam:


KurtStam:
@michael-coker Well not sure how the UI style design is supposed to work but the 'toolbar' style is not loaded when things are styled correctly. So it seemed to me like it's not meant to be used. I can't really comment on the larger problem. Maybe @gashcrumb can shed some light on it?

michael-coker:
@KurtStam ah good catch! I'm guessing it was copied from another section that uses `.toolbar` (integrations or connections). Still not sure why those styles are being applied to this page though.

PR #2417, opened 2018-04-17
fix(connector): use proper twitter icon

zregvart:
Uses the SVG Twitter icon instead of fontawesome twitter glyph.

Now it's blue:

screenshot-2018-4-17 syndesis - development


gashcrumb:
whoops, test needs updating too I guess :-)

PR #2426, opened 2018-04-18
fix (#1641): Fix regression introduced by previous commit: isScaled()…

iocanel:
… now uses labels. IntegrationController now updates

deployment and removes scheduledChecks only when it needs to.


gashcrumb:
Giving this a test now...

PR #2429, opened 2018-04-18
feat: implement refresh of access token

### LINKED ISSUE #2408: Implement refresh of OAuth access token using the refresh token
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## The problem

When OAuth access tokens expire integrations fail to perform exchanges. If we receive refresh token in the OAuth authorization we should use it to get new access token.

## Expected behavior

Integration continues to run with one acceptable exchange failing if the access token is already expired when API is being invoked.

The nuance of access token refresh is:
 - we need to implement functionality of refreshing the OAuth access tokens without any user interaction in the integration pod as a part of the REST Swagger connector
 - this functionality will be shipped for but considered as tech preview
 - this functionality will be off by default and to enabled it users must specify a environment variable on the server pod
 - the logic of the functionality would be that we will track the expiry of the access token in the REST Swagger connector and use the refresh token to fetch a fresh access token before it expires
 - we will not retry the exchange if it fails caused by invalid access token, but refresh and use the new token on the next exchange

1. pass token endpoint to REST Swagger connector
2. pass refresh token to REST Swagger connector
3. implement access token refresh functionality

zregvart:
This implements additional logic and configuration options for the API
connector functionality.

OAuth2 access token refresh functionality is added to REST Swagger
connector. Circumstances under which this functionality is enabled is
the folowing:
 - API is using OAuth2 authentication type
 - refreshToken property was given, i.e. refreshToken was issued and
   received when the Syndesis connection was created in the OAuth2 flow
 - there is at least one HTTP status defined for which the refresh of
   the access token is performed

When the Camel exchange signals a HTTP error via
HttpOperationFailedException, this logic kicks in and if the list of
HTTP status codes contains the status code of the received response and
previous attempt (if any) to refresh the access token was not with the
same refresh token an attempt is made to to refresh the access token
with the current refresh token.

If we receive a new refresh token during the refresh of the access token
we will remember it for the next access token refresh in the integration
pod, the component property is mutated.

The first exchange that fails due to the expired access token will fail,
i.e. this will not retry that exchange if we receive a new valid access
token.

What the user needs to perform in order to trigger this functionality:
 - the user needs to add a vendor extension
   `x-refresh-token-retry-statuses` to the `securityDefinitions` of the
   OAuth2 security definition in the Swagger specification -- without it
   this functionality will not be performed

Example:
```yaml
securityDefinitions:
  oauth2:
    type: 'oauth2'
    flow: 'accessCode'
    authorizationUrl: '...'
    tokenUrl: '...'
    scopes:
      ...
    x-refresh-token-retry-statuses: 403
```

This will instruct the connector to refresh the access token for HTTP
403 status codes.

Fixes #2408


KurtStam:
nice! 👍

PR #2437, opened 2018-04-19
chore(connector): remove camel-connector-maven-plugin

### LINKED ISSUE #2436: Update API client connector to use the latest integration runtime
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## The problem

We don't require auto generated Spring (Boot) configuration classes. Let's remove those from the repository and disable their generation.

zregvart:
This removes `camel-connector-maven-plugin` Maven plugin usage and with
it Spring Boot auto configuration classes that were generated in
`src/main/java`

Fixes #2436


lburgazzoli:
I see that the spring stuffs generated by the camel connector plugin are now removed so I'm wondering how this should then work, I guess we only need to remove the plugin to avoid the regeneration of such files but left then in the repository as long as they are using the "old" connector stuff.

am I missing something ?

lburgazzoli:
Yes we should migrate swagger to the new style but I think by removing the camel plugin we can avoid updating the sources, we can maybe have a profile and activate it only when it is needed.

heiko-braun:
Closing due to inactivity

Issue #2155, opened 2018-03-26
SQL deadlock detected after a soak test of 2M messages


maschmid:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

Having a soak test run on OCP, using NFS as a persistent volume for the postgres
with 8 running mock integrations in total  (3x timer (1s) -> amq, 2x amq -> log, 3x timer (1s) -> log )

after a weekend run (Uptime 2 days 20 hours 40 minutes), there are 2,171,371 Total Messages

the UI is unresponsive, with most of the API calls taking more than 30 seconds to run (30 seconds is the OpenShift HAProxy default limit)

## Expected behavior

Syndesis should be usable with even after the DB is filled with lots of message logs

## Request and Response Data

(oc rsh to the syndesis-server pod)
time curl -H "X-Forwarded-User: foobar" -H "X-Forwarded-Access-Token: foobar" 127.0.0.1:8080/api/v1/integration-support/overviews
`
{
  "items": [
    {
      "tags": [
        "activemq",
        "ext-io-syndesis-extensions-syndesis-connector-timer"
      ],
      "version": 2,
      "deploymentVersion": 2,
      "steps": [
        {
          "connection": {
            "icon": "extension:icon.png",
            "connectorId": "ext-io-syndesis-extensions-syndesis-connector-timer",
            "name": "timer",
            "id": "i-L8Hz5LixhoStjWXXQZ1z"
          },
          "stepKind": "endpoint",
          "kind": "step",
          "action": {
            "name": "Simple Timer"
          },
          "id": "-L8HznxuOcxje_csrXF9"
        },
        {
          "connection": {
            "icon": "fa-puzzle-piece",
            "connectorId": "activemq",
            "name": "amq",
            "id": "i-L8HzkmIxhoStjWXXQZ2z"
          },
          "stepKind": "endpoint",
          "kind": "step",
          "action": {
            "name": "Publish messages"
          },
          "id": "-L8HzyniOcxje_csrXF9"
        }
      ],
      "targetState": "Published",
      "currentState": "Published",
      "draft": false,
      "name": "timer-amq",
      "id": "i-L8I-AibxhoStjWXXQZ3z"
    },
    {
      "tags": [
        "activemq",
        "ext-io-syndesis-extensions-syndesis-connector-log"
      ],
      "version": 1,
[...]

PR #2438, opened 2018-04-19
bug: #2155 SQL deadlock detected. Do a better job of Delete/insert records that are part of a transaction in path order to avoid deadlocks.

chirino:


zregvart:
errorprone complains:
```
[WARNING] /workspace/app/server/jsondb/src/main/java/io/syndesis/server/jsondb/impl/SqlJsonDB.java:[527,32] [JdkObsolete] It is very rare for LinkedList to out-perform ArrayList or ArrayDeque. Avoid it unless you're willing to invest a lot of time into benchmarking.
    (see http://errorprone.info/bugpattern/JdkObsolete)
```

PR #2440, opened 2018-04-19
Add Xsrf protection in syndesis backend

nicolaferraro:
Adding xsrf protection using spring security.
I changed the ui part to use the token on each request (also `GET` and `TRACE`) and to use the `awesome` token value by default unless the server asks to use a different one (the server will not, for now).
Multipart requests were ok, the problem was in the angular FileUploader that required a ad hoc config (fixed it).

Spring-security validates the token header twice under the hood, so I used a shared constant value for ui and server.

This can be merged. I put the wip flag until we decide if we want to protect Atlasmap endpoints too (cc: @igarashitm).


gashcrumb:
Cool, lemme give this a test locally too, but so far LGTM

gashcrumb:
Works for me here.  In the data mapper there's a POST to `inspect` that has a big JSON payload, the returned response does have strings that are shown in the UI so we might want to address this, question is if it's too late for GA or not.  But I'm definitely +1 to merging this PR.

PR #2446, opened 2018-04-19
chore(release): Updates to the release process

rhuss:
* --no-moving-tag-update to prevent updating the moving tag
* Adding syndesis-cli.jar to Dockerfile dir before building upgrade image
* Minor upgrade dockerfile tuning
* Disabling of 'warmup' maven build

### DIFF
--- a/tools/bin/commands/release
+++ b/tools/bin/commands/release
@@ -32,11 +32,27 @@ release::usage() {
     --docker-password <pwd>   Docker password for Docker Hub
     --product-templates       Create templates for the productised version only and tag
     --no-git-push             Don't push the release tag (and symbolic major.minor tag) at the end
+    --no-moving-tag-update    Don't move the minor tag (e.g. "1.3") to this release (e.g. "1.3.5-20180419"_
+                              This should be used for for daily releases only.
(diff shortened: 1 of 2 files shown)


tools/bin/commands/release
```diff
@@ -262,6 +272,9 @@ create_upgrade_docker_image() {
     # Copy syndesis.yml
     cp $topdir/install/syndesis.yml .
 
+    # Copy over syndesis-cli jar
+    cp $topdir/app/server/cli/target/syndesis-cli.jar .
+
     # Create the image
     docker build -t syndesis/syndesis-upgrade:${release_version} --build-arg version=${release_version} .
```
zregvart: Would it be better if the `cp` above is a part of the syndesis-upgrade Dockerfile?

rhuss: You can't add files in a Dockerfile which is outside the context directory. I.e. it must be in the same directory or below the Dockerfile.

Therefore we have to copy over it to the Dockerfile's dir.

rhuss: Same in the `build.sh`.

But shouldn't be a problem as the syndesis-cli created here during the release process it guaranteed to be created fresh and fitting to the version we are building.

PR #2449, opened 2018-04-20
chore(build): Allow git plugin in "flash" profile

rhuss:
This does not slow down the build considerably and is required
for getting git information into the version API endpoint.

Also, adapt the 'daily' release mode to run in flash, without tests,
to avoid sporadic test failures.
The rational is that its better to have daily releases which might have
some issues than having no such release because of test failures
(which might or might not be due to real errors)

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -322,16 +322,6 @@
             </executions>
           </plugin>
           <plugin>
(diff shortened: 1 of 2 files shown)


rhuss:
@zregvart hope this is ok to merge. Enjoy CITCON ! ;-)

PR #2451, opened 2018-04-20
fix(upgrade): Wait until all pods are gone when scaling down to 0

rhuss:
also, don't force replace deploymentconfigs anymore


paoloantinori:
jus tested and worked !

PR #2461, opened 2018-04-23
bug: Logout link was not logging user out.

chirino:


chirino:
Applying this PR will actually log you out be the next page is loaded with errors since the user is no longer logged in:

syndesis_-_development

Trying to do a page reload via:

    logout() {
      this.loggedIn = false;
      return this.userService.logout().toPromise().then(() => {
        location.reload();
      }).catch(() => {
        location.reload();
      });
    }

Does take you back to the login screen, but once you login, you get this error:

403_permission_denied

kahboom:
I think it would be good to do a redirect, as opposed to a refresh, and our authorization process should take care of the rest. Also, probably not necessary to convert to a promise here, but I'll try out a few things. Sorry for the delay.

gashcrumb:
Ideally we should have a redirect target in our `config.json` that the app would go to on logout, rather than trying to stay in the app at all.  We can merge this but we should create an issue to externalize that target URL the console should redirect to on logout.

app/ui/src/app/core/providers/user-provider.service.ts
```diff
@@ -26,6 +35,6 @@ export class UserProviderService extends UserService {
    * Log the user out
    */
   logout(): Observable<any> {
-    return this.apiHttpService.setEndpointUrl('/oauth/sign_out').get();
+    return this.httpClient.get('/oauth/sign_out');
```
gashcrumb: This change probably isn't necessary.  If the GET request isn't being triggered we probably just need to subscribe to this Observable, which we'd want to do anyways to trigger a redirect...  TBH I wouldn't inject httpClient directly in this class but instead minimize the changes and add the subscription...

chirino: this change is here to avoid the path prefix the apiHttpService adds to all endpoints.

gashcrumb: Ah of course, +1 then :-)

Issue #2472, opened 2018-04-25
API Connector - Refresh AuthTokens before they expire


KurtStam:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[*] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
Tokens can expire, and we won't notice till we fail a request.

## Expected behavior
If we have a RefreshToken we should refresh before it expires so that no requests fail.

PR #2473, opened 2018-04-25
Fix: API Connector - Refresh AuthTokens before they expire #2472

KurtStam:
Fixes: https://github.com/syndesisio/syndesis/issues/2472
Marking as WIP to prevent is from getting merged until we decide to put it in. Requesting a review from @zregvart before opening it up for wider review.


app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/OAuthRefreshTokenProcessor.java
```diff
@@ -126,6 +106,20 @@ void tryToRefreshAccessToken() {
...
+                    if (expiresInOverride != null) {
+                        expiresInSeconds = Long.valueOf(expiresInOverride);
+                    } else {
+                        final JsonNode expiresIn = body.get("expires_in");
+                        if (expiresIn != null && !expiresIn.isNull() && !expiresIn.asText().isEmpty()) {
+                            expiresInSeconds = expiresIn.asLong();
+                        }
+                    }
+                    if (expiresInSeconds != null) {
+                        component.setExpiresAt(new Date(new Date().getTime() + expiresInSeconds * 1000 - 60000l));
```
zregvart: I'm not 100% sure so I steered away from mutating the component state, component is shared among endpoints, so if we have two API actions using it there might be a race condition. Could be perfectly OK, but I was thinking it might be safer to store state mutations in the processors.
Again it might not be so good at all if we have competing processors refreshing the access token at the same time.
I think it would be good to make the state mutation logic the same though -- so either mutate the component state or mutate the processor state.

zregvart: Also wouldn't it be easier just to store, a `long` or `AtomicLong` instead of `java.util.Date`, less objects to instantiate. And `System::currentTimeMillis` is in UTC, so just as safe as `java.util.Date`.

app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/SwaggerConnectorComponent.java
```diff
@@ -46,6 +47,8 @@
 
     private String accessToken;
 
+    private Date expiresAt;
```
zregvart: Perhaps `long` instead?

KurtStam: Date and long are pretty much the same thing. I prefer Date since that's what it is.

zregvart:
@KurtStam I've added one more commit https://github.com/syndesisio/syndesis/pull/2473/commits/7d10d947bc3cb6b23055bfa20fcf4590c15fc163 can you have a quick look, and I think we can merge this.

KurtStam:
@zregvart lgtm

Issue #1512, opened 2018-02-09
syndesis-upgrade: "migrate_db"


rhuss:
#### Migrate Database

If the database schema has changed for the new version to apply, then migration is required.
Since our internal homegrown database JsonDB only supports a Java-based access, this migration needs to be performed with Java.

A Java CLI tool, which is stored in the `syndesis-rest` Docker image and which can be started with `/deployments/migrate-jsondb.sh` takes the following command line arguments:

* Connection parameters to the Postgresql database (URL, user, password)
* A directory holding the migration scripts written in JavaScript

This directory contains a migration script for every schema version:

.Example upgrade directory
```
/upgrade-jsondb/
    ...
    20.js
    21.js
    23.js
    ...
```

Each upgrade script can only upgrade from the prior version. In this example, if the DB is currently at schema `20` and the target schema is `23`, then the scripts `21.js` and `23.js` are executed.

These scripts contain a single javascript function:

.Simple JavaScript API
```javascript
function upgrade(jsondb) {
   // Perform migration by iterating of jsondb documents,
   // transforming them and then storing them back

}
```

with `jsondb` a still to defined context object for accessing, querying and updating JsonDB

These scripts can be part of the `syndesis-rest` Docker image so that an outside CLI tool just needs to call

.Starting the migration
```bash
oc port-forward $(pod syndesis-db) 5432:5432
docker run syndesis/syndesis-rest --net=host \
     /deployment/migrate-jsondb.sh \
         --url jdbc://localhost:5432 --user admin --password admin \
         --target-schema 23
```

NOTE: The upgrade script and mechanism could also be used internally by the syndesis-rest application to perform an upgrade during startup. However, this is recommended only for a development setup as there is no easy way to rollback if things go wrong.

#### Rollback
If any of the upgrade scripts fail with an error, a DB rollback needs to be performed.
For this, the backup created in the previous <<step-backup-db, step>> needs to be played back (on a fresh database).

rhuss:
@zregvart @paoloantinori I'm bit later especially with this story. So if there are some spare cycle who could help here, that would be awesome. 'happy to kick-off the story, but what needs to be done is do actually start the migration itself.

paoloantinori:
Ok, let me have a look! I'm expecting being able to pick this up, but I'm still going through the details to check if I can really help here.

paoloantinori:
It's been reported the current `upgrade db` part is breaking the upgrade process.

[...]

PR #2488, opened 2018-04-27
Upgrade, invoke db scripts

paoloantinori:


paoloantinori:
fixes #1512

avano:
hi @paoloantinori, @rhuss, @dsimansk 

how does the syndesis-cli access the syndesis database?

Doesn't the script miss the port forward to syndesis-db and parameters?

Because currently the script is trying to plain invoke java -jar syndesis-cli.jar that is trying to connect to localhost:5432 and it is causing the whole upgrade process to fail.

And second thing, if we want to externalize the syndesis-cli.jar to upgrade the DB, the syndesis-upgrade deployment will need to have the ENV vars for username and password for accessing the DB

rhuss:
@avano the idea is that the update typically should run from within the cluster, so there the name of the db host would be just 'syndesis-db'. And yes, username and password need to be picked up then from the proper configmap (we already know how to read it).

However it would be awesome also when the upgrade script would be able to run from the outside (even if only for testing), so we need to distinguish that case and then open a port-forward to the syndesis-db pod and connect to localhost. The DB user and password can be still read from the secrets.

rhuss:
Actually, setting up the connection is the real hard part of the db upgrade story ;)

PR #2491, opened 2018-04-27
fix(connector): proper data shape kind for mqtt

### LINKED ISSUE #2476: MQTT connector - should DataShape.kind be `any`?
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

## The problem
MQTT connector specifies `java.lang.String` as a `DataShape.type` with `DataShape.kind=java`
https://github.com/syndesisio/syndesis/blob/5e0192eb898232b487d767a04d1a59459ae2be94/app/connector/mqtt/src/main/resources/META-INF/syndesis/connector/mqtt.json#L86
https://github.com/syndesisio/syndesis/blob/5e0192eb898232b487d767a04d1a59459ae2be94/app/connector/mqtt/src/main/resources/META-INF/syndesis/connector/mqtt.json#L126

This causes an error with datamapper step as primitive types are not supported as a Document root (atlasmap/atlasmap#9).

## Expected behavior
Not sure, do we really want a plain single text as a DataShape for MQTT (which requires atlasmap/atlasmap#9 to be implemented ahead)? or want to have `DataShape.kind=any` so user can specify custom schema for JSON/XML via "Describe Data Type" scenario?

zregvart:
This set the input/output data shape kind to `any` for the mqtt
connector, previously it was erroneously set to `kind: 'java'` and
`type: 'java.lang.String'`.

Fixes #2476

### DIFF
--- a/app/connector/mqtt/src/main/resources/META-INF/syndesis/connector/mqtt.json
+++ b/app/connector/mqtt/src/main/resources/META-INF/syndesis/connector/mqtt.json
@@ -82,8 +82,7 @@
             "actionType": "connector",
             "descriptor": {
                 "inputDataShape": {
-                    "kind": "java",
-                    "type": "java.lang.String"
+                    "kind": "any"
                 },
                 "outputDataShape": {
                     "kind": "none"
@@ -122,8 +121,7 @@
                     "kind": "none"
                 },
                 "outputDataShape": {
-                    "kind": "java",
-                    "type": "java.lang.String"
+                    "kind": "any"
                 },
                 "configuredProperties": {
                 },


app/connector/mqtt/src/main/resources/META-INF/syndesis/connector/mqtt.json
```diff
@@ -82,8 +82,7 @@
             "actionType": "connector",
             "descriptor": {
                 "inputDataShape": {
-                    "kind": "java",
-                    "type": "java.lang.String"
+                    "kind": "any",
```
dsimansk: There's comma left at the end of line.

app/connector/mqtt/src/main/resources/META-INF/syndesis/connector/mqtt.json
```diff
@@ -122,8 +121,7 @@
                     "kind": "none"
                 },
                 "outputDataShape": {
-                    "kind": "java",
-                    "type": "java.lang.String"
+                    "kind": "any",
```
dsimansk: And same here ofc...

Issue #2489, opened 2018-04-27
Integration data buckets are not working


mcada:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

Integration with with more data buckets does not work and throws error:

```
2018-04-27 09:24:44.615  WARN 1 --- [r://integration] o.a.camel.component.timer.TimerConsumer  : Error processing exchange. Exchange[i-LB5PoTP8--C4zmSirFYz]. Caused by: [io.atlasmap.api.AtlasException - Errors: [Source document '-LB5Ox45jJWJ_mHtrd3E' doesn't exist: docId='-LB5Ox45jJWJ_mHtrd3E', path='/first_name'], ]

io.atlasmap.api.AtlasException: Errors: [Source document '-LB5Ox45jJWJ_mHtrd3E' doesn't exist: docId='-LB5Ox45jJWJ_mHtrd3E', path='/first_name'], 
	at org.apache.camel.component.atlasmap.AtlasEndpoint.onExchange(AtlasEndpoint.java:213) ~[camel-atlasmap-1.34.2.jar!/:na]
	at org.apache.camel.impl.ProcessorEndpoint$1.process(ProcessorEndpoint.java:71) ~[camel-core-2.21.0.jar!/:2.21.0]
	at org.apache.camel.util.AsyncProcessorConverterHelper$ProcessorToAsyncProcessorBridge.process(AsyncProcessorConverterHelper.java:61) ~[camel-core-2.21.0.jar!/:2.21.0]
	at org.apache.camel.processor.SendProcessor.process(SendProcessor.java:148) ~[camel-core-2.21.0.jar!/:2.21.0]
	at org.apache.camel.processor.DelegateAsyncProcessor.process(DelegateAsyncProcessor.java:110) ~[camel-core-2.21.0.jar!/:2.21.0]
	at org.apache.camel.processor.RedeliveryErrorHandler.process(RedeliveryErrorHandler.java:548) ~[camel-core-2.21.0.jar!/:2.21.0]
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201) [camel-core-2.21.0.jar!/:2.21.0]
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:138) ~[camel-core-2.21.0.jar!/:2.21.0]
[...]

PR #2502, opened 2018-04-27
bug: #2489 Integration data buckets are not working for scheduled connectors.

chirino:


app/integration/runtime/src/test/java/io/syndesis/integration/runtime/capture/OutMessageCaptureProcessorTest.java
```diff
@@ -213,6 +216,86 @@ public void testCaptureWithSplit() throws Exception {
         }
     }
 
+
+    @Test
+    public void testCaptureWithSplitAndSchedual() throws Exception {
```
igarashitm: while running a local build to try... typo? > Schedule

PR #2508, opened 2018-04-30
fix(upgrade): Fix rollback

### LINKED ISSUE #2499: [Upgrade] After performing a rollback, upgraded (and rolled back) deployments don't start
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

## The problem

I'm performing the rollback in upgrade_50_replace_template, where I'm appending `exit 1` after `update_version $tag` in `replace_template::run()`.

After everything is rolled back, the deployments are scaled back to 1, but only those, that are not affected by upgrade (prometheus, oauthproxy, db [i'm not doing db migration for now]) are started back.

This is how it looks like after rollback: 
uz3h

The completed deployments in my case can't be scaled up - when I scale them via oc or via web, it goes back to 0 after few seconds.

cc @rhuss @paoloantinori @dsimansk 

1.
2.
3.
4.

rhuss:
fixes #2499


rhuss:
@avano this should fix the rollback issue. Could you very this please ?

tools/upgrade/steps/upgrade_50_replace_template
```diff
@@ -29,17 +29,8 @@ replace_template::run() {
...
-extract_template_name() {
-    local file=$1
 
-    local ext="${file##*.}"
-    if [ $ext = "yml" ] || [ $ext = "yaml" ]; then
-        grep 'name:' $file | head -1 | sed -e 's/^[ ]*name:[ ]*//'
-    else
-        cat $file | jq -r .metadata.name
-    fi
+    exit 1
```
avano: this `exit 1` probably shouldn't be here

rhuss: oops. absolutely correct, left over from simulating a failure. 

Good find !

rhuss:
let's get it merged, seem to look good according to 2499

PR #2528, opened 2018-05-02
chore(server): set default max-age and expiry on client state cookies

### LINKED ISSUE #2443: Client side state cookies not removed
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem

Seems that the code to remove client side state cookies is not functioning properly. The cookies are still there after the connection is created. At that point those should be removed. Having them there we end up having really long `Cookie` headers and that ends up with `400 Bad Request: Request Header Or Cookie Too Large` on ngnix.

## Expected behavior

Remove the client state cookie when it's no longer needed.

1. Create a OAuth based connection several times, I've counted 9 client side cookies and 18 oauth-proxy cookies

zregvart:
If when creating OAuth connections the user cancels or gets an error the
client side state cookie might remain, this is an issue if many of these
cookies are present in the browser as there are limits on HTTP header
sizes in the browser and the server.

This adds default max-age and expiry to the cookie so it will be removed
from the browser at the same time as it times-out on the server.

Fixes #2443

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/state/ClientSideState.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/state/ClientSideState.java
@@ -22,10 +22,13 @@ import java.security.GeneralSecurityException;
 import java.security.MessageDigest;
 import java.security.SecureRandom;
 import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
 import java.util.Base64;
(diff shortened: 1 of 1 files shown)


paoloantinori:
hey @zregvart the issue related to this PR is not a blocker. can this wait  to after GA ?

paoloantinori:
Ah, I see, thanks. Anyhow "blockers only for this sprint" rule has higher priority over any other one.

PR #2534, opened 2018-05-02
fix(ui): remove the oauth settings by setting values to null

### LINKED ISSUE #2533: Can't see the connector config for an oauth enabled connector if the credentials are set and then deleted
## This is a...

<pre><code>
[ ] Feature request
[ ] Bug report  
</code></pre>

## The problem

If I set the client ID and secret for twitter and go into the create connection flow I see the 'Connect to Twitter' button on page 2.  If I then go back to 'Settings > OAuth Clients' and click 'Remove' on the 'Twitter' oauth client, and then go back to the create connection flow I *still* see the 'Connect to Twitter' button.

As a potential second issue (not sure if it's related to the above) if I create a connection using the oauth flow, and then *delete* the client id/secret from 'Settings > OAuth clients' and then go to the connection detail page, I still only see a 'Validate' button.

## Expected behavior

If no client ID/secret is set in the 'Settings' page, the connector form should be present on the create connection wizard and the connection detail page.

Marking this as a regression, as I seem to recall this working better in the past...

zregvart:
Syndesis server used to consider properties that are empty or contain
only whitespace as `null` (_trim to null_), this is no longer the case
and the property value needs to be specified as `null`.

Removing OAuth settings would only change the value to an empty string
(`""`) and thus would be still remain as configured.

Fixes #2533

### DIFF
--- a/app/ui/src/app/settings/oauth-apps/oauth-app-modal.component.ts
+++ b/app/ui/src/app/settings/oauth-apps/oauth-app-modal.component.ts
@@ -47,7 +47,7 @@ export class OAuthAppModalComponent {
 
   // Clear the store credentials for the selected oauth app
   removeCredentials() {
-    const app = { ...this.item.client, clientId: '', clientSecret: '' };
+    const app = { ...this.item.client, clientId: null, clientSecret: null };
     return this.store
       .update(app)
       .take(1)


zregvart:
@gashcrumb I've added toggling of `derived` flag of all Connections based on the state of OAuth settings of the Connector, this should fix the second part of #2533.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/setup/OAuthAppHandler.java
```diff
@@ -139,6 +140,16 @@ public void update(@NotNull @PathParam("id") String id, @NotNull @Valid OAuthApp
             .build();
 
         dataMgr.update(updated);
+
+        final boolean shouldBeDerived = app.clientId != null && app.clientSecret != null;
+
+        dataMgr.fetchAllByPropertyValue(Connection.class, "connectorId", id).forEach(connection -> toggleDerived(connection, shouldBeDerived));
```
rhuss: Is the change on the stream really persisted ?

zregvart: A bit further down in the `toggleDerived` method:

https://github.com/syndesisio/syndesis/pull/2534/files#diff-ed861ccdc3b0bd650c4032e9e633654cR152

PR #2535, opened 2018-05-02
Fix, upgradedb, uses correct user and password

### LINKED ISSUE #1512: syndesis-upgrade: "migrate_db"

paoloantinori:
fixes #1512 

Sample output:

```
...
syndesis-upgrade-1.3 upgrade === * Process new template to extract update resources (prep_40_process_template)
syndesis-upgrade-1.3 upgrade       - Current Syndesis version "latest"
syndesis-upgrade-1.3 upgrade       - Not updating syndesis-server-config
syndesis-upgrade-1.3 upgrade === * Migrate database (upgrade_10_migrate_db)
syndesis-upgrade-1.3 upgrade 16:09:04.677 Current database schema version: 26
syndesis-upgrade-1.3 upgrade 16:09:04.774 Migration done
syndesis-upgrade-1.3 upgrade === * Stop all deployments (upgrade_20_stop_all)
syndesis-upgrade-1.3 upgrade       - Stopping deployments syndesis-db
syndesis-upgrade-1.3 upgrade syndesis-meta
syndesis-upgrade-1.3 upgrade syndesis-oauthproxy
syndesis-upgrade-1.3 upgrade syndesis-prometheus
syndesis-upgrade-1.3 upgrade syndesis-server
syndesis-upgrade-1.3 upgrade syndesis-ui
...
```

### DIFF
--- a/tools/upgrade/steps/upgrade_10_migrate_db
+++ b/tools/upgrade/steps/upgrade_10_migrate_db
@@ -65,12 +65,26 @@ migrate_db::run() {
     local workdir=${2}
     local tag=${3}
 
-    # Ensure that DB users are really scaled down
-    scale_deployments 0 "syndesis-server syndesis-meta"
+    local host=${SYNDESIS_DB_SERVICE_HOST:-}
+    local port=${SYNDESIS_DB_SERVICE_PORT:-}
+    if [ -z "$host" ]; then
+        local db_pod=$(pod "syndesis-db")
+        check_error $db_pod
+        oc port-forward $db_pod 5432 &
+        local forward_pid=$!
+        trap "kill $forward_pid" EXIT
+        sleep 5
+        host="localhost"
+        port=5432
+    fi
 
-    echo "    - TODO: Call out to syndesis-server and start the migration CLI"
-    echo "      there with local miration scripts"
-    #java -jar syndesis-cli.jar migrate
+    local user=$(extract_param "POSTGRESQL_USER")
+    local password=$(extract_param "POSTGRESQL_PASSWORD")
+
+    java -jar $workdir/syndesis-cli.jar migrate \
+         --url="jdbc:postgresql://$host:$port/syndesis" \
+         --user="$user" \
+         --password="$password"
 }
 
 migrate_db::rollback() {
--- a/tools/upgrade/common.sh
+++ b/tools/upgrade/common.sh
@@ -98,6 +98,12 @@ read_global_config() {
     oc get secret syndesis-global-config -o jsonpath={.data.${key}} | base64 --decode
 }
 
+# Read a single parameter
+extract_param() {
+    local key=${1}
+    read_global_config params | grep "${key}" | sed -e "s/${key}=//"
+}
+
 # ===================================================================================================
 # Postgres funcs
 
--- a/tools/upgrade/upgrade.sh
+++ b/tools/upgrade/upgrade.sh
@@ -108,6 +108,9 @@ run() {
     local current_tag=$(read_global_config syndesis)
     preflight_version_check "$current_tag" "$tag"
 
+    # Copy the migration jar to the workdir for later reference
+    cp $basedir/syndesis-cli.jar $workdir/
+
     echo "============================================="
     echo "=== STARTING UPGRADE TO SYNDESIS $tag "
     echo "============================================="


tools/upgrade/steps/upgrade_10_migrate_db
```diff
@@ -65,9 +65,12 @@ migrate_db::run() {
...
     local tag=${3}
 
-    echo "    - TODO: Call out to syndesis-server and start the migration CLI"
-    echo "      there with local miration scripts"
-    #java -jar syndesis-cli.jar migrate
+    read_global_config params > ${workdir}/template_params.props
+    local POSTGRESQL_USER=$(grep -Po '(?<=POSTGRESQL_USER=)[^ ]+' ${workdir}/template_params.props)
+    local POSTGRESQL_PASSWORD=$(grep -Po '(?<=POSTGRESQL_PASSWORD=)[^ ]+' ${workdir}/template_params.props)
+    
+    java -jar syndesis-cli.jar migrate --url="jdbc:postgresql://$SYNDESIS_DB_SERVICE_HOST:$SYNDESIS_DB_SERVICE_PORT/syndesis" --user="$POSTGRESQL_USER" --password="$POSTGRESQL_PASSWORD"
```
rhuss: $SYNDESIS_DB_SERVICE_HOST is only set when the script is called from within Kubernetes. Its *not* available from the outside and if coming from the outside one would also open a port-forward first with `oc port-forward`.

Note also, that `grep -P` is not available on mac OS and is not Posix compliant and has to be avoided. I wonder why don't you just source template_params.props ? It's valid shell syntax.

paoloantinori: Is calling this script from outside the cluster a requirement? I thought we are using a container because we wanted to.
I remember also a discussion about releasing versioned upgrade containers, with all the information required.

In that same context
`grep -P` usage sounds like a no issue considering we are using a Centos based image.

By sourcing the properties file, I'm bringing in a lot of env vars I'm not interested into, and I should check if they can clash or override the context already set. I just need the 2 I have used in the script.

rhuss: Up to now everything works perfectly fine outside the container so that we can experiment easily with up/downgrades. It's really nice to keep, but if you insist on breaking that fine. I can send a PR to fix it later.

Please use `perl -n -e 'print $1 if /^POSTGRESQL_PASSWORD=(.*)$/'` instead of `grep -P` as it is more portable. We shouldn't compromise portability without good reasons (`grep -P` is not a good reason).

Also your regexp doesnt work if the password contains spaces, and no lookahead assertion is required here. 

What's about simply using this

```
grep POSTGRESQL_PASSWORD ${workdir}/template_params.props | head -1 | cut -d'=' -f2-
```

?

paoloantinori: I tend to favor simplicity over complexity. I do believe that 3 commands is more complex than 1, but other than that I don't have anything against that expression.
And, since we are already late with our planning, I'm happy to merge this, as I would have been with any working implementation.

I have added a commit with the alternative version, please check if I haven't introduced any error or typo.

rhuss: Sorry, my suggestion was not good enough, it will miss out subequent '=' in the password. Let me fix this and push it to your branch.

paoloantinori: Thanks, sorry, I didn't have a running test env at hand and it still takes me a lot to provision new ones, otherwise I would have done that myself.

avano: afaik currently you need to have the up-*.js scripts bundled in the syndesis-cli.jar as the DefaultMigrator uses "classpath:/migrations/up-" (as far as I remember) path prefix for the migrations scripts

paoloantinori: I have found that files are loaded from within the .jar classpath.
Currently no op script, that is executed each time lives here:

https://github.com/syndesisio/syndesis/blob/master/tools/upgrade/BOOT-INF/classes/migrations/up-30.js

I have tried to tweak the spring boot app to load `.js` files provided at CLI without luck. I can't tell if I haven't found the correct way to pass paths or if it cannot just be done due to the way the code is currently written or it's just me being dull and not able to tame the thing.

Anyhow, despite not ideal, we, as the engineering team, are in control of the script we need to publish to upgrade environments, and they can live in a .jar without specific drawback.

rhuss:
@paoloantinori I tested it locally (outside the container) and it works. Should work now also from within the cluster, but I still wonder how the migration scripts are handled.

PR #2542, opened 2018-05-03
fix(server): don't warn about missing properties on derived connections

### LINKED ISSUE #2532: Salesforce connection keep showing configuration issue message
## This is a...

<pre><code>
[ ] Feature request
[x ] Regression (a behavior that used to work and stopped working in a new release)
[x ] Bug report  
</code></pre>

## The problem

A valid salesforce connector configuration was set up and when the connection was created the following error message appear:

sfdc1
sfdc2

## Expected behavior

There should not be a warning, as the configuration is correct.

zregvart:
For Salesforce Connector Connections we support multiple ways of
configuring the properties, when the connection is configured without
OAuth settings set, i.e. manually, the username and password fields are
required; with OAuth settings, i.e. using OAuth flow, username and
password are not set and not required.

Given that the warning SYNDESIS006 is issued if a required property is
not configured, a warning will be shown on Salesforce connections
created from OAuth flow. We can distinguish those by having their
`derived` flag set and filter out this warning.

Fixes #2532


lburgazzoli:
lgtm but in the future we need a way to define groups or properties similar to what we did for the verifier, maybe this property validation should be delegated to the meta service.

Issue #2524, opened 2018-05-02
Deleting a connection which is used by an integration leads to inconsistent state


rhuss:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

After an import of integration with a broken connection (because of missing OAuth dance, see #2522), deleting the connection is possible without warning (e.g. "is used in an existing integration"). After the deletion, the following error happens on the integration overview:

![image](https://user-images.githubusercontent.com/99080/39515763-55e79360-4dfb-11e8-8a85-a67e92da40d7.png)

and in the server logs there's a NPE:

```
2018-05-02 09:24:07.765  INFO [-,,,] 1 --- [oller Scheduler] i.s.s.c.i.IntegrationController          : Checking integrations for their status.
2018-05-02 09:24:23.529 ERROR [-,bbc75ec8b624358d,bbc75ec8b624358d,false] 1 --- [  XNIO-3 task-8] .s.e.v.h.e.SyndesisServerExceptionMapper : Internal Server Exception. null

java.lang.NullPointerException: null
	at io.syndesis.server.endpoint.v1.handler.integration.IntegrationHandler.toCurrentConnection(IntegrationHandler.java:352) ~[server-endpoint-1.3.5-20180502.jar!/:1.3.5-20180502]
	at java.util.Optional.map(Optional.java:215) ~[na:1.8.0_151]
	at io.syndesis.server.endpoint.v1.handler.integration.IntegrationHandler.toCurrentSteps(IntegrationHandler.java:371) ~[server-endpoint-1.3.5-20180502.jar!/:1.3.5-20180502]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193) ~[na:1.8.0_151]
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1380) ~[na:1.8.0_151]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481) ~[na:1.8.0_151]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471) ~[na:1.8.0_151]
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708) ~[na:1.8.0_151]
[...]

PR #2545, opened 2018-05-03
Deleting a connection which is used by an integration leads to inconsistent state (backend)

lburgazzoli:
This PR is related to #2524 and:

- fixes the NPE thrown by syndesis-server when an integration with deleted connector is invoked
- introduces additional checks on integration level warn about inconsistent state of extensions or deleted connectors, @gashcrumb related codes are:

    ```java
        SYNDESIS009, // Connection has been deleted
        SYNDESIS010, // Multiple extension installed
    ```

The integration page now shows a warning on the impacted connection:

image

Going to integration summary, the following is shown:

image

There are some remaining issues I cannot fix on backend side

1. when an integration with deleted connections is edited, UI invokes the following endpoint

        /api/v1/connections/undefined/actions/${actionId}

    Here _undefined_ if obviously wrong and happens because the actualized connection does not have    the _connectorId_ set (as the connection does not exists any more).

2. New codes need to be integrated on UI side

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ hs_err_pid*
 .pmd
 .pmdruleset.xml
 .factorypath
+.sts4-cache/
 
 # System Files
 .DS_Store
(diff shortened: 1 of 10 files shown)


gashcrumb:
@lburgazzoli cool, lemme get those changes in place and update this PR.

gashcrumb:
Yeah, I'm -1 to this...  Like what am I supposed to do in the editor, delete the affected connection?  I think this PR is going in the wrong direction.

capture

lburgazzoli:
@gashcrumb rebased and squashed, now the step is re-setted but UI hangs now

gashcrumb:
@lburgazzoli think that last commit should do it, lemme know if you see anything else.

lburgazzoli:
@gashcrumb yes works for me, going to remove WIP

PR #2546, opened 2018-05-03
fix: correct the values being passed to patch and fix how it's being called

### LINKED ISSUE #2540: "Replace draft" when hit, it does basically nothing
## This is a...

<pre><code>
[ ] Feature request
[ x ] Bug report  
</code></pre>

## The problem

This is a small issue, however I don't know whether it somehow isn't related to this issue https://github.com/syndesisio/syndesis/issues/2539. Basically the "Replace draft" option for various integration versions does nothing. 

## Expected behavior

It could do something?

## Screenshot

screenshot from 2018-05-03 10-22-08

1. Create iintegration, edit it so there are at least v1 and v2
2. You can invoke dropdown menu with options "Replace draft" and "Publish" tight next to the integration version. 
3. The "Replace draft" does nothing

gashcrumb:
almost fixes #2540

The UI uses a PATCH request to update the `steps` attribute, however it also tries to update `isDraft` as well but that doesn't seem to be persisted.  So this halfway fixes the issue.

### DIFF
--- a/app/ui/src/app/integration/integration-actions-provider.service.ts
+++ b/app/ui/src/app/integration/integration-actions-provider.service.ts
@@ -222,10 +222,10 @@ export class IntegrationActionsProviderService extends IntegrationActionsService
 
   replaceDraftAction(integration: Integration | IntegrationOverview, deployment: IntegrationDeployment | DeploymentOverview): Promise<any> {
       return this.integrationSupportService.getDeployment(integration.id, deployment.version.toString())
-        .map(_deployment => {
-          this.store.patch(<any>integration, {
+        .switchMap(_deployment => {
+          return this.store.patch(_deployment.integrationId, {
             steps: _deployment.spec.steps,
-            draft: true
+            isDraft: true
           });
         }).toPromise();
   }
--- a/app/ui/src/app/integration/integration-actions-provider.service.ts
+++ b/app/ui/src/app/integration/integration-actions-provider.service.ts
@@ -224,8 +224,7 @@ export class IntegrationActionsProviderService extends IntegrationActionsService
       return this.integrationSupportService.getDeployment(integration.id, deployment.version.toString())
         .switchMap(_deployment => {
           return this.store.patch(_deployment.integrationId, {
-            steps: _deployment.spec.steps,
-            isDraft: true
+            steps: _deployment.spec.steps
           });
         }).toPromise();
   }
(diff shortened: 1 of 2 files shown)


app/ui/src/app/integration/integration-actions-provider.service.ts
```diff
@@ -222,10 +222,10 @@ export class IntegrationActionsProviderService extends IntegrationActionsService
 
   replaceDraftAction(integration: Integration | IntegrationOverview, deployment: IntegrationDeployment | DeploymentOverview): Promise<any> {
       return this.integrationSupportService.getDeployment(integration.id, deployment.version.toString())
-        .map(_deployment => {
-          this.store.patch(<any>integration, {
+        .switchMap(_deployment => {
+          return this.store.patch(_deployment.integrationId, {
             steps: _deployment.spec.steps,
-            draft: true
+            isDraft: true
```
zregvart: Turns out `isDraft` is not persisted, I think I have a fix for this on the server. @gashcrumb  do you mind if I push to this PR. I've tested a bit it seems to function. Though I'm a bit lost on the whole meaning of _Published_. Seems that if I create a new draft with this change then the status switches to _Unpublished_. Is this status meant to show only the latest versions state?

zregvart:
@gashcrumb, @paoloantinori this _seems_ to work, I think I wrapped my head around how the logic should be. Can you take a especially good look at this: https://github.com/syndesisio/syndesis/pull/2546/commits/0b2d8988302ca00eea2a53442d197088598df811 commit and see if that makes sense with the current world view on draft/current/target status

paoloantinori:
Looks reasonable to me!

zregvart:
For whatever reason the last deployment's Integration spec can differ from the Integration the deployment was created from (published). For SQL connector that I use to test with a empty Split definition is added to the ConnectorDescriptor. So comparing spec Integration and current Integration will result in a difference, and this makes poor choice for determining the `isDraft` status. Working on a version that compares select properties of steps.

paoloantinori:
@lburgazzoli  we have discussed something similar recently, despite it was for steps and actions. do you have any suggestion for @zregvart  issue?

lburgazzoli:
The split element should not be new, otherwise the whole integration would not work but what happen is that the sql connection is "actualized" and it will include the connector (note that the "dummy" connection we provide does not have an associated connector but a connectorId). 

Can we use IntegrationDeployment to discriminate between draft/non draft ? Don't know all the detail around that object but that's the object used to create a project so it should hold some state information.

zregvart:
> Can we use IntegrationDeployment to discriminate between draft/non draft ? Don't know all the detail around that object but that's the object used to create a project so it should hold some state information.

In a way we're using IntegrationDeployment, that is we're using the Integration contained in the `spec` within.

I think we missed a bit here:
 - we don't really know what IntegrationDeployment is the _published_ one -- given that the user can publish previous deployments; this PR contains logic that looks for the last non-_Unpublished_ IntegrationDeployment and considers it to be the one
 - with this Split popping up in ConnectorDescriptor comparing the IntegrationDeployment.spec with integration is non-deterministic; this PR compares configuredProperties of steps and stepKind

PR #2550, opened 2018-05-04
ENTESB-6910 Add licenses.xml to iPaaS

cunningt:
https://issues.jboss.org/browse/ENTESB-6910

My understanding is that this is needed as a gate check for release.


zregvart:
> I'm failing license-check on dummy readme.txt files. Any suggestions? Is there a better dummy file to use so that directory isn't empty by default, or is there something I can add to the text file?

Perhaps we could have just `.gitkeep` instead there?

paoloantinori:
fixes #2553

PR #2551, opened 2018-05-04
feat: add MAX_INTEGRATIONS_PER_USER parameter to syndesis-server configs

tplevko:
wdyt? If it's not against some of the conventions, I think this would be an easier way to specify the maximum number of integrations. 
I just didn't like that much the server config-map update, as described in [the documentation.](https://access.qa.redhat.com/documentation/en-us/red_hat_fuse/7.0/html-single/integrating_applications_with_ignite/#installing-on-ocp)

### DIFF
--- a/install/generator/01-header.yml.mustache
+++ b/install/generator/01-header.yml.mustache
@@ -129,6 +129,11 @@ parameters:
   name: SERVER_MEMORY_LIMIT
   value: 800Mi
   required: true
+- description: Maximum number of integrations single user can create
+  displayName: Maximum number of integrations
+  name: MAX_INTEGRATIONS_PER_USER
+  value: "1"
+  required: true
 - description: Key used to perform authentication of client side stored state.
   displayName: Client side state authentication key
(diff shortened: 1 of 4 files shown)


paoloantinori:
This pr is not entirely correct. It should have removed these lines as well:

https://github.com/syndesisio/syndesis/blame/master/install/generator/04-syndesis-server.yml.mustache#L211-L212

We now have a duplicated entry:

https://github.com/syndesisio/fuse-ignite-install/blob/1.3.6/resources/fuse-ignite-ocp.yml#L1103-L1105

and

https://github.com/syndesisio/fuse-ignite-install/blob/1.3.6/resources/fuse-ignite-ocp.yml#L1116-L1118

paoloantinori:
For the timing being I'm in hacking mode. We cannot fix it for GA at this layer, so I'm working on the ignite install repo to rectify this. We just need to keep track of this and put in place a maintainable strategy that divides the different responsibilities between upstream and product repo.

PR #2556, opened 2018-05-04
fix: trigger a window reload if we get a failed request to config.json due to browser CORS blocking

### LINKED ISSUE #2511: redirect to externalized URL when logging out
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## The problem

https://github.com/syndesisio/syndesis/pull/2461 fixes logging out, but when logging out the user is currently left within the app.  When a user logs out we should redirect to some page, and that page should be defined in the `config.json` file so it can be easily customized per environment.

gashcrumb:
fixes #2511 for firefox.


gashcrumb:
-1 for my own PR :-)

PR #2562, opened 2018-05-07
chore: Script for prefilling nodes on fuse-ignire cluster with the s2i image

rhuss:


rhuss:
The script is currently running for the 1.3.5 upstream image. Will check later whether it really improves deployment times.

rhuss:
I tested it now with 1.3.5 and the build time dropped down to 1.45s for me (after a republish). 40s of this is for the build, ~ 10s time for the OC deployment, the rest is the startup of the Spring Boot app.

PR #2593, opened 2018-05-11
Deleting API connector should delete it's connections

### LINKED ISSUE #2586: Custom API connector Delete behavior - new changes in CR1
## This is a...
<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem
There was a discussion about Deleting Custom API connector that haven't lead to any solution.
https://github.com/syndesisio/syndesis/issues/2124, https://github.com/syndesisio/syndesis/issues/1827

Meanwhile, a side effect of an other fix (probably https://github.com/syndesisio/syndesis/issues/2524)  caused change in that behavior, which is:

1. connector can be deleted if there is or isn't a connection created from the connector. THe connection is still editable. It wasn't before.
2. connector can't be deleted (Delete button inactive) if a connection created from the connector is being used in an integration
3. If the connector has been deleted and you want to use a connection created from the connector in a new integration, There is an error when selecting the connection in the integration.
connectionclicked
[server_pod.log](https://github.com/syndesisio/syndesis/files/1990798/server_pod.log)

There is also a warning when deleting a connector that says:

>You can no longer edit a connection that was created from this connector. **(not true)**
You can still add connections created from this connector to integrations. **(not true)** Deleting this connector does not affect running integrations. **(true, because you can't Delete it when a connection created from the connector is being used in an integration)**

>Are you sure you want to delete the "Todo App API" connector?

Please do something with that inconsistency. Whichever it is of an easiest way of changing the warning or getting things working properly. Thanks.

zregvart:
When a Connector is deleted it makes little sense in having Connections
of that Connector remaining. This removes all Connections of a Connector
when it's deleted.

Fixes #2586

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectorHandler.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectorHandler.java
@@ -48,6 +48,7 @@ import io.syndesis.server.inspector.Inspectors;
 import io.syndesis.common.model.Kind;
 import io.syndesis.common.model.ListResult;
 import io.syndesis.common.model.action.ConnectorAction;
+import io.syndesis.common.model.connection.Connection;
 import io.syndesis.common.model.connection.Connector;
 import io.syndesis.common.model.connection.ConnectorSummary;
 import io.syndesis.common.model.filter.FilterOptions;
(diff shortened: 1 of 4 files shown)


zregvart:
> hi zoran, we wanted to just update the description in the html part, to avoid further regression or changes. can you split the pr, so that we will merge only that part?

Not really, the text in the confirmation dialog now reflects the change of the functionality in the server. To me it makes little sense to just change the text and not address the underlying issue. Also, changing the text is a new issue as it differs from the plan I outlaid in https://github.com/syndesisio/syndesis/issues/2586#issuecomment-388189819. To change just the text we need a new issue and the exact wording for the change.

paoloantinori:
We don't want to risk to have another fix in that changes the behavior in such a way, that some possible workflow can break again. We don't have time for that.

Can you please than close this pr and open one that just fixes the warning message?

zregvart:
> We don't want to risk to have another fix in that changes the behavior in such a way, that some possible workflow can break again. We don't have time for that.

If you don't mind I'll keep the PR open, it can be merged at a later date. In my opinion the current workflow is broken and it will be remain broken until this PR is merged.

> Can you please than close this pr and open one that just fixes the warning message?

Sure, can you open a new issue with the exact wording of the warning message?

paoloantinori:
NP for the pr. the new issue is here: https://github.com/syndesisio/syndesis/issues/2594

PR #2609, opened 2018-05-14
[closes #2608] Get migrate DB scripts also from external folder

### LINKED ISSUE #2608: [Upgrade] Make syndesis-cli also take folder parameter
## This is a...

<pre><code>
[ ] Feature request
[ ] Bug report  
</code></pre>

## The problem

Currently you need to have up-*.js scripts for DB upgrade bundled inside syndesis-cli.jar

1.
2.
3.
4.

avano:
Original: up-50.js bundled in syndesis-cli.jar

```
java -jar syndesis-cli.jar migrate -t 99 --url="jdbc:postgresql://localhost:5432/syndesis" --user="syndesis" --password="FkGPHgdx8x5nvY0S"
10:44:14.708 Current database schema version: 26
10:44:14.714 Starting migration to version: 99
10:44:15.088 Migrating to schema: 50
Starting migration to schema 50
Migration to schema 50 completed
10:44:16.524 Migration done
```

up-50.js bundled in syndesis-cli.jar and up-98.js and up-99.js in external folder:

```
java -jar syndesis-cli.jar migrate -t 99 --url="jdbc:postgresql://localhost:5432/syndesis" --user="syndesis" --password="oB3IV5kakdyOUwyi" -f /tmp/upgrade
10:45:58.053 Current database schema version: 26
10:45:58.057 Starting migration to version: 99
10:45:58.235 Migrating to schema: 50
Starting migration to schema 50
Migration to schema 50 completed
10:45:59.267 Migrating to schema: 98
Starting migration to schema 98
Migration to schema 98 completed
10:45:59.310 Migrating to schema: 99
Starting migration to schema 99
Migration to schema 99 completed
10:45:59.357 Migration done

```

### DIFF
--- a/app/server/jsondb/src/main/java/io/syndesis/server/jsondb/dao/Migrator.java
+++ b/app/server/jsondb/src/main/java/io/syndesis/server/jsondb/dao/Migrator.java
@@ -21,5 +21,5 @@ import io.syndesis.server.jsondb.JsonDB;
  * Handles migrating the schema of a JsonDB store to a given schema of the model
  */
 public interface Migrator {
-    void migrate(JsonDB jsondb, int schema);
+    void migrate(JsonDB jsondb, int schema, String... resources);
 }
--- a/app/server/runtime/src/main/java/io/syndesis/server/runtime/DefaultMigrator.java
+++ b/app/server/runtime/src/main/java/io/syndesis/server/runtime/DefaultMigrator.java
@@ -15,18 +15,7 @@
  */
 package io.syndesis.server.runtime;
 
-import java.io.IOException;
-
-import javax.script.ScriptEngine;
-import javax.script.ScriptEngineManager;
-import javax.script.ScriptException;
-
-import io.syndesis.common.util.IOStreams;
-import io.syndesis.common.util.Resources;
-import io.syndesis.common.util.SyndesisServerException;
-import io.syndesis.server.jsondb.WithGlobalTransaction;
-import io.syndesis.server.jsondb.JsonDB;
-import io.syndesis.server.jsondb.dao.Migrator;
+import static io.syndesis.common.util.Json.map;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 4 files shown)


app/server/jsondb/src/main/java/io/syndesis/server/jsondb/dao/Migrator.java
```diff
@@ -22,4 +22,5 @@
  */
 public interface Migrator {
     void migrate(JsonDB jsondb, int schema);
+    void migrate(JsonDB jsondb, int schema, String migrationDir);
```
zregvart: Perhaps it would be better to have just the one method on the Migrator, something like:

`void migrate(JsonDB jsondb, int schema, String... baseResources)`

And the `resources` parameter would be in Spring [resource abstraction syntax](https://docs.spring.io/spring/docs/5.1.0.BUILD-SNAPSHOT/spring-framework-reference/core.html#resources-resource-strings).

So we could say here `classpath*:/migrations/` or `/path/to/dir/`.

app/server/runtime/src/main/java/io/syndesis/server/runtime/DefaultMigrator.java
```diff
@@ -15,56 +15,74 @@
...
-import org.springframework.core.io.ResourceLoader;
-import org.springframework.stereotype.Service;
-
-import static io.syndesis.common.util.Json.map;
-
 @Service
 public class DefaultMigrator implements Migrator {
     private static final Logger LOG = LoggerFactory.getLogger(DefaultMigrator.class);
-
+    private String migrationDir;
```
zregvart: Since spring gives us abstraction over resources, I think there is no need for `migrationDir`. We just specify the resource paths (in spring resource syntax) that the `ResourceLoader` needs to look into.

app/server/runtime/src/main/java/io/syndesis/server/runtime/DefaultMigrator.java
```diff
@@ -15,56 +15,74 @@
...
     private final ResourceLoader resourceLoader;
 
     public DefaultMigrator(final ResourceLoader resourceLoader) {
         this.resourceLoader = resourceLoader;
     }
 
     @Override
-    public void migrate(final JsonDB jsondb, final int toVersion) {
+    public void migrate(final JsonDB jsondb, final int schema) {
```
zregvart: Let's add `String... baseResources` as a parameter here. If there aren't any given we can default to `classpath:/migrations/up-`

app/server/runtime/src/main/java/io/syndesis/server/runtime/DefaultMigrator.java
```diff
@@ -15,56 +15,74 @@
...
         }
     }
 
+    @Override
+    public void migrate(JsonDB jsondb, int schema, String migrationDir) {
+        this.migrationDir = migrationDir;
+        migrate(jsondb, schema);
+    }
+
     private void performMigration(final JsonDB utilized, final int toVersion) {
```
zregvart: ...pass the `String... baseResources` parameter in here

app/server/runtime/src/main/java/io/syndesis/server/runtime/DefaultMigrator.java
```diff
@@ -15,56 +15,74 @@
...
         final String path = migrationsScriptPrefix() + toVersion + ".js";
```
zregvart: ...and loop over `baseResources[n]` so we can support from multiple base resources also, first one that locates the resource wins; I think that was the gist of the change...

avano:
@zregvart wdyt now? Currently you can specify only one additional path (-f file:/tmp/somedir), but I can change that to use multiple if we want to

Issue #2577, opened 2018-05-09
[Upgrade] Not possible to upgrade to prod version using docker img


avano:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

Currently it is not possible to upgrade to newer version using docker container, because of preflight check failing. The version defined the template in prod version upgrade container specifies syndesis: "1.3" version.

The upgrade script then fails here:

```
    if [ -n "${SYNDESIS_VERSION:-}" ]; then
        if [ "$SYNDESIS_VERSION" != "${target_version}" ]; then
            echo "Internal error: Container template's version is not the same as upgrade container tag"
            echo "- Container version:               $SYNDESIS_VERSION"
            echo "- Version extracted from template: $target_version"
            exit 1
        fi
    fi 
```

Because the container variable SYNDESIS_VERSION is set to "1.3.5.fuse-000002-redhat-1" and the "target_version" is parsed from the template and therefore resolves to "1.3" and the if shuts everything down

## Tasks involved / Steps to Reproduce

1.
2.
3.
4.

avano:
@paoloantinori @rhuss @dsimansk FYI

paoloantinori:
I had lost visibility of this one. Is this still  a blocker?

avano:
Hi,

Darren fixed this in the prod upgrade image, so not a blocker anymore, but still an issue. The question is if the syndesis template should specify syndesis version "1.3" instead of "1.3.6.xyz-redhat-1". If yes, then the preflight check should be reworked to take that into account.

rhuss:
Sorry, I'm still trying to catch up. Where does this 1.3.6.xyz-redhat-1 comes from ? This is definitely not an image tag which should be referenced in any template, but a Maven like artefact version.

avano:
Hi Roland,

the problem probably is that you can understand SYNDESIS_VERSION as two things in a productized build - "1.3" and "1.3.6.xyz-redhat-1". That is not a problem in upstream where tag and maven version are the same.

Currently Darren sets the SYNDESIS_VERSION in upgrade container to "1.3.6.xyz-redhat-1" and that causes this issue. The question is, what is the correct string there :-)

dsimansk:
@rhuss @avano @paoloantinori 
[...]

PR #2662, opened 2018-05-15
chore(upgrade): Relax preflight check

rhuss:
Only check up to the minor version number in the preflight check.


rhuss:
Also required to fix #2577 (beside using the proper SYNDESIS_VERSION)

PR #2665, opened 2018-05-16
Do batch inserts of the activity log records. Should incrementally improve performance

chirino:


chirino:
I don't think this change is required for GA, so lets avoid merging it until master is clear for post GA work.

rhuss:
added WIP label to avoid accidental merging

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/ActivityTrackingController.java
```diff
@@ -51,14 +52,15 @@
...
+import io.syndesis.server.jsondb.impl.JsonRecordSupport;
 import io.syndesis.server.openshift.OpenShiftService;
 
 /**
  * This class tracks pod controller and ingests them into our DB.
  */
 @Service
 @ConditionalOnProperty(value = "controllers.dblogging.enabled", havingValue = "true", matchIfMissing = true)
-@SuppressWarnings({"PMD.DoNotUseThreads", "PMD.ModifiedCyclomaticComplexity", "PMD.StdCyclomaticComplexity", "PMD.CyclomaticComplexity"})
+@SuppressWarnings({"PMD.DoNotUseThreads", "PMD.ModifiedCyclomaticComplexity", "PMD.StdCyclomaticComplexity", "PMD.CyclomaticComplexity", "PMD.GodClass"})
```
rhuss: Might be worth to split the class up at some point in time. At least this is commented ;-)

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/ActivityTrackingController.java
```diff
@@ -129,6 +131,34 @@ private int deleteFieldsLT(String path, String field) {
         });
     }
 
+    private void writeBatch(TreeMap<String, Object> batch) {
+        dbi.inTransaction((conn, status) -> {
+            PreparedBatch insert = conn.prepareBatch("INSERT into jsondb (path, value, ovalue) values (:path, :value, :ovalue)");
+            for (Map.Entry<String, Object> entry : batch.entrySet()) {
+                String key = "/activity" + entry.getKey() + "/";
```
rhuss: not sure but shouldn't this be `"/activity/" + entry.getKey() + "/"`;

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/ActivityTrackingController.java
```diff
@@ -288,7 +318,8 @@ private void processEventQueue() {
                     }
 
                     // Write the batch..
-                    jsonDB.update("/activity", Json.writer().writeValueAsBytes(batch));
+//                    jsonDB.update("/activity", Json.writer().writeValueAsBytes(batch));
+                    writeBatch(batch);
```
rhuss: so if I understand correctly, you are inserting directly into the DB here, bypassing jsondb. Wouldn't it make more sense to add a 'writeActivities()' method to jsondb and call it ? This would free this class from having internal knowledge of jsondb table structure and handling.

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/ActivityTrackingController.java
chirino: Yeah it's right.  entry.getKey() = "/somethign"

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/ActivityTrackingController.java
chirino: Not really.  The writeBatch method is doing stuff that is not generic.  It would be ugly to push higher 'app' level semantics into lower level generic apis.  I know its also not nice to break through the Json DB api abstractions here, but I'm only doing it to improve performance.

The reason that jsondb does not expose something similar right now is that this use pattern if used improperly can lead to inconsistent db state.  Example being inserting records that look like:
```
/a/b/ = "Hello"
/a/b/c/ ="World"
```

The exposed update method double checks those cases, and would delete the `/a/b/` record when the `/a/b/c/` record is added.  That double checking is partly what was slowing us down.

PR #2679, opened 2018-05-17
fix(install): Default value for max integrations

rhuss:


zregvart:
:+1: _(still not sure if we're merging to master)_

PR #2695, opened 2018-05-22
chore(feat): AWS S3 Component - set CamelAwsS3Key header

### LINKED ISSUE #2691: [ENH] AWS Component - set CamelAwsS3Key header
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## The problem

When AWS is used as a producer the write operation fails unless header `CamelAwsS3Key` is set. A workaround is to set it explicitly via a set header external extension or via a script.
We should provide a way to perform the operation in a way smoother way.

## Expected behavior

Allow this comoponent to be used without additional component required.

oscerd:
fixes #2691 

s3-fix

### DIFF
--- a/app/connector/aws-s3/src/main/java/io/syndesis/connector/aws/s3/AWSS3CopyObjectCustomizer.java
+++ b/app/connector/aws-s3/src/main/java/io/syndesis/connector/aws/s3/AWSS3CopyObjectCustomizer.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.connector.aws.s3;
+
+import java.io.IOException;
+import java.util.Map;
+
+import org.apache.camel.Exchange;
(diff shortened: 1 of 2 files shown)


paoloantinori:
:+1:

oscerd:
cc @TovaCohen 

The fileName in this case is not required: in case you're using the AWS S3 poll as starting connection you'll have the fileName already set, in that case you'll be able to choose if you want to maintain the name coming from the S3 start connection or if you want to override the name with the field in the copyObject.

TovaCohen:
@oscerd  Thanks for letting me know. 
Do I have this right:
In 1.0, when an S3 connection performs the Copy Object action, the user is not offered the option to specify a File Name. Now, in 1.1, a user can optionally specify a file name. 

Are you assuming that when the start connection's poll action obtains data that it always obtains exactly one file? I'm wondering if it is possible for the Poll action to obtain more than one file, and if it does obtain more than one file, then should we tell the user not to specify a file name for a Copy Object action?

Suppose an S3 start connection performs the Get Object action, can a user add a connection that performs Copy Object and specify a file name for the action?

Is there an expectation that when the user specifies a file name that the file name's extension (type) must match the file that the action is receiving? For example, if the Poll action obtained a MYFILE.doc file, does the user need to specify something like YOURFILE.doc?

oscerd:
> Are you assuming that when the start connection's poll action obtains data that it always obtains exactly one file? I'm wondering if it is possible for the Poll action to obtain more than one file, and if it does obtain more than one file, then should we tell the user not to specify a file name for a Copy Object action?

No, in that case you'd use the getObject. In case of poll action we'll need to document or specify to avoid setting the fileName in the copyObject, otherwise you'll get only the last file polled in your bucket.

> Suppose an S3 start connection performs the Get Object action, can a user add a connection that performs Copy Object and specify a file name for the action?

Yes, he can.

> Is there an expectation that when the user specifies a file name that the file name's extension (type) must match the file that the action is receiving? For example, if the Poll action obtained a MYFILE.doc file, does the user need to specify something like YOURFILE.doc?

If you want the to have the extension you can add the .doc but from the S3 perspective it doesn't have any effect, because S3 will be able to understand what kind of file is hosting.

PR #2696, opened 2018-05-22
chore: Moved label 'syndesis.io/app' back to 'app' as it has a meaning for OpenShift

rhuss:
This change reverts the change introduced in https://github.com/syndesisio/syndesis/commit/7027e646ac75af9548b34c557b0ade3d63f81fa6 where other labels has been moved to a 'syndesis.io' prefix.

The change for 'app' was by accident.

### DIFF
--- a/install/generator/01-header.yml.mustache
+++ b/install/generator/01-header.yml.mustache
@@ -13,7 +13,7 @@ metadata:
 # OAuthProxy Tag: {{Tags.OAuthProxy}}
 #
   labels:
-    syndesis.io/app: syndesis
+    app: syndesis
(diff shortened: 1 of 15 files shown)


rhuss:
@paoloantinori @dsimansk the 'app' label change was by accident. This PR should be applied, not sure when this should happen. We would need to either move 1.3.10 tag on syndesis (as it doesnt affect any image or Maven artefact I see no problem here) or release a 1.3.11. We could also release a 1.3.10.1 or so, and reference this tag in fuse-ignite-install to regen the product templates.

wdyt ?

PR #2710, opened 2018-05-25
chore(tools): add syndesis dev --cleanse tool

zregvart:
This adds `--cleanse` option to `syndesis dev` subcommand that deletes
pods that are in 'Completed' state.

### DIFF
--- a/tools/bin/commands/dev
+++ b/tools/bin/commands/dev
@@ -20,6 +20,7 @@ dev::usage() {
     cat <<"EOT"
     --debug <name>            Setup a port forwarding to <name> pod (default: server)
     --cookie                  Loads a local valid cookie to access Syndesis APIs when using Minishift
+    --cleanup                 Removes 'Completed' pods
     --refresh                 Used in conjuction with --cookie, forces a refresh of the cookie.
                               Ex: curl -k --cookie $(syndesis dev --cookie --refresh)  "https://$(oc get route syndesis  --template={{.spec.host}})/api/v1/connections"
     --version                 Show running version of cluster
@@ -128,5 +129,7 @@ dev::run() {
         echo $cookie_cached
     elif [ $(hasflag --version) ]; then
         oc get secret syndesis-global-config -o jsonpath={.data.syndesis} | base64 --decode
+    elif [ $(hasflag --cleanup ) ]; then
+        oc delete pods $(oc get pods --field-selector=status.phase=Succeeded -o=jsonpath='{range .items[*]}{.metadata.name} ')
     fi
 }


rhuss:
@zregvart sorry for my ignorance, but "clense" is not a word I really know, and I can't find it in a dictionary as well. To you mean "cleanse" ? But maybe even "--cleanup" would be more intuitive.

tools/bin/commands/dev
```diff
@@ -128,5 +129,7 @@ dev::run() {
         echo $cookie_cached
     elif [ $(hasflag --version) ]; then
         oc get secret syndesis-global-config -o jsonpath={.data.syndesis} | base64 --decode
+    elif [ $(hasflag --clense ) ]; then
+        oc delete pods $(oc get pods --field-selector=status.phase=Succeeded -o=jsonpath='{range .items[*]}{.metadata.name} ')
```
rhuss: would remove the extraneous space at the end of the jsonpath expression

rhuss: Alternative you could also just use

```
oc delete $(oc get pods --field-selector=status.phase=Succeeded -o name)
```

(though not tested, but should work)

zregvart:
> @zregvart sorry for my ignorance, but "clense" is not a word I really know, and I can't find it in a dictionary as well. To you mean "cleanse" ? But maybe even "--cleanup" would be more intuitive.

Like most things I do this has a inside joke attached to it. To cleanse is a very hipstery-new-age thing that's very trendy again :)

Though I thought about `--cleanup` and this seemed more appropriate as you would purge and remove things from within the namespace, not just scrub it of something on purpose.

Can change it, not that attached to it.

```
  cleanse
      v 1: clean one's body or parts thereof, as by washing; "clean up
           before you see your grandparents"; "clean your fingernails
           before dinner" [syn: cleanse, clean]
      2: purge of an ideology, bad thoughts, or sins; "Purgatory is
         supposed to cleanse you from your sins"
```

tools/bin/commands/dev
zregvart: Extra space is needed to separate pod names from each other, they get concatenated one to each other otherwise.

`-o name` would yield `pod/name` in the output where we need just the `name`, I actually quite dislike that it does that.

rhuss:
Yeah, would be nice to change to --cleanup. We are hipsters (or wannabes ;-), true. But not every one of our users are and the rest of our options are name quite old-school. So lets be boring and try to stay predictable.

Thanks !

zregvart:
Not an issue I've switched to `--cleanup`.

Issue #2761, opened 2018-06-05
Remove hover state from drag and drop uploader after file is uploaded


michael-coker:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
I just added the hover state when a file is uploaded on the extension import page https://github.com/syndesisio/syndesis/pull/2758 to match the integration import UI. However, after consulting the [design doc](https://github.com/syndesisio/syndesis/blob/master/ux/designs/importexport/importexport.md) the hover state should only be present on hover on those pages. 

## Expected behavior
Blue border around the file uploader should only be visible when dragging a file over the drag and drop zone.

## Tasks involved / Steps to Reproduce
1. Go to `/integrations/import` and drop an exported integration
2. Go to `/customizations/extensions/import` and drop an extension (you can use [this one](https://drive.google.com/file/d/1bwDvQfkjkJLO4HDbuA6O9sqFds7WxPHY/view?usp=sharing)
3. Note that the blue border is present after uploading a file - it should only present when dragging a file over the zone, and is not present in any other state.

PR #2762, opened 2018-06-05
Remove hover state from drag and drop uploader after file is uploaded

michael-coker:
Fixes https://github.com/syndesisio/syndesis/issues/2761


zregvart:
Seems that there is a issue compiling this:
```
[ERROR] Error: src/app/integration/import-export/import/integration-import.component.ts(75,10): error TS2339: Property 'checkIfDragAndDrop' does not exist on type 'IntegrationImportComponent'.
```

michael-coker:
@zregvart oops! Thanks, updated the branch.

PR #2763, opened 2018-06-05
Issue #636: Provided a web api to query used versions of Apache Camel…

### LINKED ISSUE #636: provide a web api to query used versions of Apache Camel, Spring Boot and Syndesis
in the eclipse tooling's Tech Extension wizard we currently let the user specify those versions manually. A more convenient way to do that would be to just ask for a URL of the syndesis instance and then trigger some web api to query the versions to be used.

valdar:
…, Spring Boot and Syndesis.

This fix #636

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/VersionService.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/VersionService.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.syndesis.server.endpoint;
+package io.syndesis.server.endpoint.v1;
 
 import java.util.Collections;
 import java.util.HashMap;
@@ -22,26 +22,30 @@ import java.util.Optional;
 
 import com.jcabi.manifests.Manifests;
 
-public final class Version {
+import org.springframework.stereotype.Service;
 
-    private Version() {
-    }
+@Service
+public final class VersionService {
 
-    public static Map<String, String> getDetailed() {
+    private final Map<String, String> detail;
+
+    public VersionService() {
         final Map<String, String> detail = new HashMap<>();
-        detail.put("version", getVersion());
+        putManifestValueTo(detail, "version", "Implementation-Version", "Project-Version");
         putManifestValueTo(detail, "commit-id", "Git-Commit-Id", "X-BasePOM-Git-Commit-Id");
         putManifestValueTo(detail, "branch", "Git-Branch");
         putManifestValueTo(detail, "build-time", "Build-Time");
         putManifestValueTo(detail, "build-id", "X-BasePOM-Build-Id");
 
-        return Collections.unmodifiableMap(detail);
+        this.detail = Collections.unmodifiableMap(detail);
     }
 
-    public static String getVersion() {
-        final String packageVersion = Version.class.getPackage().getImplementationVersion();
+    public Map<String, String> getDetailed() {
+        return detail;
+    }
 
-        return Optional.ofNullable(packageVersion).orElse("DEVELOPMENT");
+    public String getVersion() {
+        return Optional.ofNullable(detail.get("version")).orElse("DEVELOPMENT");
     }
 
     static void putManifestValueTo(final Map<String, String> detail, final String key, final String... attributes) {
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/VersionEndpoint.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/VersionEndpoint.java
@@ -20,7 +20,6 @@ import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.Response;
 
-import io.syndesis.server.endpoint.Version;
 import io.swagger.annotations.ApiOperation;
 import org.springframework.stereotype.Component;
 
@@ -31,17 +30,23 @@ import static java.util.concurrent.TimeUnit.HOURS;
 @Component
 public class VersionEndpoint {
 
-    @GET
-    @Produces("application/json")
-    @ApiOperation(value = "Get the version as JSON specification")
-    public Response doGetJson() {
-        return Response.ok(Version.getDetailed()).build();
+    private final VersionService versionService;
+
+    public VersionEndpoint(VersionService versionService) {
+        this.versionService = versionService;
     }
 
     @GET
     @Produces("text/plain")
     @ApiOperation(value = "Get the version")
     public Response doGetPlain() {
-        return Response.ok(Version.getVersion()).build();
+        return Response.ok(versionService.getVersion()).build();
+    }
+
+    @GET
+    @Produces("application/json")
+    @ApiOperation(value = "Get the version as JSON specification")
+    public Response doGetJson() {
+        return Response.ok(versionService.getDetailed()).build();
     }
 }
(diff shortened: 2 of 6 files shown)


zregvart:
I don't see the need for VersionService. I've commented on the https://github.com/syndesisio/syndesis/issues/636#issuecomment-394981866 that I think the better approach would be to define a Syndesis BOM rather than try to keep a list of relevant dependencies.

paoloantinori:
@zregvart your suggestion doesn't solve our use case. Our use case is: ask a runtime which version is using, so that a developer can configure his project accordingly (ideally automatically).

zregvart:
> @zregvart your suggestion doesn't solve our use case. Our use case is: ask a runtime which version is using, so that a developer can configure his project accordingly (ideally automatically).

@paoloantinori I'm not sure that you understand what I'm suggesting. I'm suggesting that rather than expanding the versions we provide in an endpoint we provide those versions via Syndesis BOM (that we need to create).

If I'm misunderstanding something please elaborate on how a BOM doesn't solve this use case?

paoloantinori:
this PR, as per title is about "Provided a web api".

the complexity of creating a BOM is outside the scope of it.
using a BOM or not using a BOM is not the critical aspect of the functionality and should not prevent a merge, in case anyone wants to refactor this later.

In fact this enhancement was actually requested to provide a single information, that is Syndesis version. So no risk of forgetting to add additional relevant information. The additional entries are a nice extra, that we can even throw away if we are afraid that they create an issue.

But again, please don't expand the scope of a specific feature. Open a new one if you have a suggestion  for further enhancements.

zregvart:
I'm sorry I do not agree with you and I fail to understand your argumentation, I find it quite dismissive and narrow sighted.

> this PR, as per title is about "Provided a web api".

This PR along with https://github.com/syndesisio/syndesis-extensions/pull/58 is an attempt to provide dependency information to the build tooling.

Providing a web API locks us to a certain way of doing things. I'm arguing that this way of doing things will cost us more to maintain.

> the complexity of creating a BOM is outside the scope of it.

Are you arguing that a solution should be considered regardless of consequences?

I do not agree that alternative solution to a problem is outside of the scope. Please provide arguments to why an alternative solution is not a better solution rather than arguing what is in scope or what isn't in scope.

> using a BOM or not using a BOM is not the critical aspect of the functionality and should not prevent a merge, in case anyone wants to refactor this later.

This refactoring *later* is something that never happens, will you be the one that refactors this? Can you consider not putting the burden on other team members in the first place?

> In fact this enhancement was actually requested to provide a single information, that is Syndesis version. So no risk of forgetting to add additional relevant information. The additional entries are a nice extra, that we can even throw away if we are afraid that they create an issue.

So in fact we provide the version of Syndesis without this PR, would that make this PR unnecessary?

> But again, please don't expand the scope of a specific feature. Open a new one if you have a suggestion for further enhancements.

I'm sorry I do not think I'm expanding the scope by arguing for an alternative solution to the issue this PR is trying to solve. And based on that why this PR should be reconsidered. The topic is this PR and the way we provide dependencies to tooling.

zregvart:
> @zregvart the service was in place because I understood, from some discussions, that the info might be needed in other components. If you think is overkilling I can move that in an utility class.

I don't mind it that much, but I would make that change if/when it's needed. Right now a static singleton class should suffice in my opinion.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/VersionService.java
```diff
@@ -22,29 +22,29 @@
...
 
-    public static Map<String, String> getDetailed() {
-        final Map<String, String> detail = new HashMap<>();
-        detail.put("version", getVersion());
+    private VersionService() {
         putManifestValueTo(detail, "commit-id", "Git-Commit-Id", "X-BasePOM-Git-Commit-Id");
         putManifestValueTo(detail, "branch", "Git-Branch");
-        putManifestValueTo(detail, "build-time", "Build-Time");
         putManifestValueTo(detail, "build-id", "X-BasePOM-Build-Id");
+        putManifestValueTo(detail, "version", "syndesys-version");
```
zregvart: typo: syndesys -> syndesis

app/server/runtime/pom.xml
```diff
@@ -101,6 +101,20 @@
...
+            <execution>
+            <id>default-jar</id>
+            <phase>package</phase>
+            <goals>
+                <goal>jar</goal>
+            </goals>
+            <configuration>
+              <archive>
+                <manifestEntries>
+                  <syndesys-version>${syndesis.version}</syndesys-version>
```
zregvart: typo: syndesys -> syndesis

This value already exists in the `Implementation-Version` manifest header, why do we need additional value?

valdar: from what I've understood this https://github.com/syndesisio/syndesis/blob/master/app/pom.xml#L36-L37 is the "official global syndesis version" and for whatever the case that might in future differ from the project.version of each component; if that is not the case, fine to read it from `Implementation-Version`.

valdar:
@zregvart if you say is fine to call this https://github.com/syndesisio/syndesis/blob/master/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/Version.java#L30-L39 each time the version endpoint is called, I'm ok with keeping just the changes to the templates to make the `/api/v1/version` endpoint.

zregvart:
@valdar this is the response we currently have for the `application/json` endpoint:

```json
{
  "build-id":"f7b3b47b-062e-42ed-8da8-8617dff3f65a",
  "commit-id":"f4c0f1e780268f32a8540e1e11e56807bdb6aea3",
  "version":"1.4-SNAPSHOT"
}
```

I see no reason why the `version` in that should diverge from $project.version in Maven.

valdar:
@zregvart that answer the first question, makes sense if they don't diverge. What about the second one?

paoloantinori:
> I'm sorry I do not agree with you and I fail to understand your argumentation, I find it quite dismissive and narrow sighted.

I'm saying that you can improve a project in infinite ways. But if you never make a single step, because you keep enlarging the scope you will never make any progress.

This feature was specific. You gave a suggestion. That's suggestion is valid, just outside the scope.

Merge this, make progress, move forward.

valdar:
@zregvart as per our irc discussion, I've updated the PR.
I've kept the @Service instead of the static singleton because since spring boot is used I thought was  better to leverage the framework.

As a side note, I've noticed some inconsistent behaviour on the produced output (i.e. sometime some less important attributes are missing) and I think that is due to https://github.com/jcabi/jcabi-manifests/issues/31 and in the light of that may be it wasn't a bad idea to add custom entries :)

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/VersionService.java
```diff
@@ -22,29 +22,33 @@
...
+        return detail;
+    }
 
-        return Optional.ofNullable(packageVersion).orElse("DEVELOPMENT");
+    public String getVersion() {
+        return Optional.ofNullable(detail.get("version")).orElse("DEVELOPMENT");
     }
 
-    static void putManifestValueTo(final Map<String, String> detail, final String key, final String... attributes) {
+    private void putManifestValueTo(final Map<String, String> detail, final String key, final String... attributes) {
```
zregvart: This method could remain `static`

zregvart:
> As a side note, I've noticed some inconsistent behaviour on the produced output (i.e. sometime some less important attributes are missing) and I think that is due to jcabi/jcabi-manifests#31 and in the light of that may be it wasn't a bad idea to add custom entries :)

Did you notice this in the test (running from the IDE?) or in the deployed? All those case have different classpaths and the `META-INF/MANIFEST.MF` content depends on the way the project is build, so the behaviour might change due to that. I would not worry about it if it behaves correctly when deployed.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/VersionService.java
valdar: Done.

Issue #1190, opened 2018-01-23
Create a simple ServiceNow Connector


rohanmars:
Implement ServiceNow connector component.

lburgazzoli:
We should provide a simple ServiceNow connector that provides basic functionalities , the user should provide data shape on its own.

PR #2770, opened 2018-06-06
chore: upgrade resteasy to 3.5.1.Final

lburgazzoli:
Background:

- while developing servicenow connectror #1190 I've experienced some issues because the of mismatch of javax.ws.rs-api which was forced to be 2.0 in syndesis but cxf uses 2.1
- we leverage [resteasy-spring-boot-starter](https://github.com/paypal/resteasy-spring-boot/tree/master/resteasy-spring-boot-starter) from payal which is not more maintained and has been donated to [resteasy](https://github.com/resteasy/resteasy-spring-boot)
- resteasy-spring-boot-starter from paypal is locked at v2.0 specs
- resteasy-spring-boot-starter from resteasy moved to specs v2.1 but is based on spring boot 2.x

So I've copied the resteasy-spring-boot-starter from paypal to our repository as app/commons/resteasy-spring-boot-starter and I've upgraded it to spec 2.1. I've fixed all the issues reported by our tools and removed the old dependency across the project. 

This is of course an interim solution till we migrate to spring boot 2.x and the official resteasy starter.

@syndesisio/backend I did run a quick test and things seems to work but I need a review as I'm not a super expert about ws-rs.

### DIFF
--- a/app/common/pom.xml
+++ b/app/common/pom.xml
@@ -74,6 +74,7 @@
   <modules>
 	  <module>util</module>
     <module>model</module>
+    <module>resteasy-spring-boot-starter</module>
   </modules>
 
   <dependencyManagement>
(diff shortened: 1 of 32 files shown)


zregvart:
The tests passed so I would be somewhat confident in merging this. @lburgazzoli do you know that the starter cannot be run in Spring Boot 1.5? We could try using the starter as is and excluding the Spring Boot 2.0 dependency/ies from it. Would that work?

lburgazzoli:
Don't know but as spring 2.x has changed the behavior of some annotation i.e. ConditionalOnXyz which is now an AND instead of a OR and some other stuffs, I'd say that it is better to stay with what we have. 

We may reach-out the resteasy people(in my todo list) to check if they can support both spring 1.x and 2.x.

lburgazzoli:
@zregvart removed the wip label so if you are confident, we can try to merge it :)

Issue #2095, opened 2018-03-21
Missing log entries


rhuss:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

For an periodic sql query integrations start point with a period of 5 secondes, there seems some entries missing at the start:

![image](https://user-images.githubusercontent.com/99080/37698704-4cd37b78-2ce4-11e8-9cf1-30f5d030124a.png)

Note that 10s distance between the first three entries. After this, period is always 5 seconds as it should be.

KurtStam:
@rhuss it is set to wait 10 seconds before the timer starts. Should I change it to 0 seconds?

KurtStam:
`docker log <your-integration-container-id> -f -t` shows the messages with a timestamp i.e:

> 2018-04-18T21:14:21.440079480Z {"exchange":"i-LAPav_x8Vo-tFZ3k3n7z","step":"-LAJKUU9aO0To1Em-4ir","id":"i-LAPav_y8Vo-tFZ3k3nAz","duration":797732}
> 2018-04-18T21:14:21.440111814Z {"exchange":"i-LAPav_x8Vo-tFZ3k3n7z","status":"done","failed":false}
> 2018-04-18T21:14:22.438620483Z {"exchange":"i-LAPavpa8Vo-tFZ3k3nCz","status":"begin"}
> 2018-04-18T21:14:22.442692899Z {"exchange":"i-LAPavpa8Vo-tFZ3k3nCz","step":"-LAJKSLtaO0To1Em-4ir","id":"i-LAPavpa8Vo-tFZ3k3nDz","duration":3995601}
> 2018-04-18T21:14:22.444198416Z {"exchange":"i-LAPavpa8Vo-tFZ3k3nCz","step":"-LAJKUU9aO0To1Em-4ir","id":"i-LAPavpe8Vo-tFZ3k3nFz","duration":1397562}
> 2018-04-18T21:14:22.444244181Z {"exchange":"i-LAPavpa8Vo-tFZ3k3nCz","status":"done","failed":false}
> 2018-04-18T21:14:23.439539739Z {"exchange":"i-LAPaw4D8Vo-tFZ3k3nHz","status":"begin"}
[...]

PR #2775, opened 2018-06-07
fixes: #2095 missing log entries.

chirino:
I just discovered that one of my log parsing assumptions was wrong. I assumed that all log entries from the API server started with a fix length time prefix that looked like:

    2018-01-12T21:22:02.068338027Z { ..... }

Turns out that that time prefix is actually variable length.


chirino:
With this change, if you "select * from todo" in the start connection, you should see an activity log that looks something like:

syndesis
...

syndesis

Where the integration steps for 1 activity repeat for each record in the database.

Issue #2303, opened 2018-04-10
Upgrade to Spring Boot 1.5.13


lburgazzoli:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
[X] Task
</code></pre>

## The problem

We are now using Spring Boot 1.5.8 but the latest release is 1.5.12 which among other things, fixes some CVE.

@paoloantinori @gashcrumb @chirino Should we upgrade before GA ?

aileenc:
We are currently using 1.5.10 in FIS to align with RHOAR

oscerd:
Also Camel 2.21.0 is based on 1.5.10, while 2.21.1 will be based on 1.5.11

lburgazzoli:
because of the CVEs does it make sense to upgrade the whole stack to 1.5.12 ?

chirino:
Yes.  If upstream does not do it, it will occur in the pipeline builds. I think it’s better if get ahead and upgrade your n the upstream.

chirino:
Lets go with latest SB 1.5.x version.

oscerd:
This should be closed, since we are using SB 1.5.13 now. 

https://github.com/syndesisio/syndesis/blob/master/app/pom.xml#L63

PR #2804, opened 2018-06-12
Bump Spring Boot to 1.5.13

chirino:


chirino:
This fixes #2303

zregvart:
Seems we're hitting https://github.com/google/error-prone/issues/780 I've triggered a rebuild, not sure if that'll help, if it doesn't we might need to turn off parameter naming check in errorprone by adding `-Xep:ParameterName:OFF` to compiler arguments.

zregvart:
Now we're getting 403 error when trying to download docker binary, hopefully this will be fixed at `get.docker.com` and we can retry the build later.

PR #2822, opened 2018-06-15
refactor: add concur icon and description

zregvart:
This adds the Concur icon and the updates the description.

screenshot-2018-6-15 syndesis 1

cc @syndesisio/uxd @syndesisio/support-docs @KurtStam @gaughan


KurtStam:
@zregvart let's commit. If UX comes up with something else we can always update.

gaughan:
Looks great!

zregvart:
> @zregvart let's commit. If UX comes up with something else we can always update.

Perhaps @gaughan and @TovaCohen can offer better name and description, the "Combined Concur 3.0 Swagger API" and "Use SAP Concur for travel and expense management" is probably not the best we can do. Super trivial to make changes to those.

TovaCohen:
How's this:

Concur
SAP expense, travel, and invoice management

gaughan:
How about? :

Concur (SAP)
Manage invoices, travel, and expenses.

sjcox-rh:
@zregvart Icon looks good! That was the latest Icon I found and since you were able to make it a vector, no need to change it.  Thanks!

Issue #2664, opened 2018-05-16
Clean up Maven build


zregvart:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>
We seem to duplicate plugin dependency configuration, we should try to manage this from a single Maven parent POM as much as possible.

## The problem

It's difficult to maintain multiple POM files in case of dependency or plugin configuration change.

## Expected behavior

Have a single place to maintain dependency and plugin configuration.

zregvart:
Resolved via #2838

PR #2838, opened 2018-06-16
Create a simple ServiceNow Connector

### LINKED ISSUE #1190: Create a simple ServiceNow Connector

lburgazzoli:
Fixes #1190

This PR implements a basic ServiceNow connector which lacks support for pagination for starting connectors.

### DIFF
--- a/app/connector/servicenow/pom.xml
+++ b/app/connector/servicenow/pom.xml
@@ -1,0 +1,100 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Copyright (C) 2016 Red Hat, Inc.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+<project
+  xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>io.syndesis.connector</groupId>
+    <artifactId>connector-parent</artifactId>
+    <version>1.4-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>connector-servicenow</artifactId>
+  <name>Connector :: ServiceNow</name>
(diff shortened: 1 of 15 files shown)


app/connector/servicenow/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+            <goals>
+              <goal>generate-connector-inspections</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+
+      <plugin>
+        <groupId>pl.project13.maven</groupId>
+        <artifactId>git-commit-id-plugin</artifactId>
```
zregvart: I never understood why are we skipping the `git-commit-id-plugin`, any idea, can this be removed? I plan to remove it for all connectors in #2664 anyhow.

app/connector/servicenow/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+    </plugins>
+  </build>
+
+  <profiles>
+    <profile>
+      <id>servicenow-skip-tests</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
+      <build>
```
zregvart: Perhaps it would make sense to just skip integration tests by default? i.e. define property `basepom.it.skip=true`

app/connector/servicenow/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+    </profile>
+
+    <profile>
+      <id>servicenow-tests</id>
+      <activation>
+        <property>
+          <name>env.SERVICENOW_INSTANCE</name>
+        </property>
+      </activation>
+      <build>
```
zregvart: Then here we could set `basepom.it.skip=false`.

app/connector/servicenow/src/main/java/io/syndesis/connector/servicenow/ServiceNowMetaDataExtension.java
```diff
@@ -0,0 +1,620 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * An implementation of the MetaData extension {@link MetaDataExtension} that
+ * retrieve information about ServiceNow objects as Json Schema as per draft-04
+ * specs.
+ *
+ * TODO: copied from upstream, push upstream and remove when swithing to 2.22.x
+ */
+@SuppressWarnings("PMD")
```
zregvart: :disappointed:

app/connector/servicenow/pom.xml
lburgazzoli: Don't know, I've used another connector's pom as template

app/connector/servicenow/src/main/java/io/syndesis/connector/servicenow/ServiceNowMetaDataExtension.java
lburgazzoli: I can disable individual checks if it sound better but:

- The class is complex

Yes it is complex as I need to generate the json schema by hand or other things.

- A method/constructor shouldnt explicitly throw java.lang.Exception 

Camel throws generic exceptions everywhere so I can eventually wrap them using runtime exception but as this class should be moved back to camel once done I do not see any reason to do it as that's the camel way.

lburgazzoli:
So i removed the bits from pom that are not needed and I've excluded only the PMD violations that can't be fixed

Issue #2452, opened 2018-04-20
Inline edit patternfly alignment


seanforyou23:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

There have been updates to the inline edit design experience in patternfly. Because of this, our implementation in Syndesis has fallen out of sync a bit, and some work is needed to bring this UI element back up to par.

## Expected behavior

Interacting with inline edit in Syndesis reflects the design specs noted in the pattern library.

http://www.patternfly.org/pattern-library/forms-and-controls/inline-edit/#design

michael-coker:
Here are the instances of inline editing that I've found so far. Does anyone know of any more?

cc @dongniwang @sjcox-rh 

## Integration editor - integration name
![screen shot 2018-06-13 at 3 08 31 pm](https://user-images.githubusercontent.com/35148959/41377141-cb68f30e-6f20-11e8-97e7-a0ccc234b079.png)

## API connector details
![screen shot 2018-06-13 at 3 04 54 pm](https://user-images.githubusercontent.com/35148959/41377142-cb7c0ffc-6f20-11e8-9719-703e8f86b1c2.png)

## Integration details - integration name and description 
![screen shot 2018-06-13 at 3 03 57 pm](https://user-images.githubusercontent.com/35148959/41377143-cb91ecc8-6f20-11e8-9116-b5561f8e2652.png)

## Connection details - connection name and description
![screen shot 2018-06-13 at 3 03 32 pm](https://user-images.githubusercontent.com/35148959/41377144-cba151fe-6f20-11e8-9d72-1fc9412887d8.png)

michael-coker:
On hold for the moment until UXD can decide on a good UI for both the editable textarea (integration/connection description) and text (integration/connection name) components.

PR #2873, opened 2018-06-21
update editable textarea component to the new patternfly ui

michael-coker:
part of https://github.com/syndesisio/syndesis/issues/2452

2 things I couldn't figure out...

* I can't seem to get the `blur` or `focusout` events to fire on the `textarea`, so that when you click outside of it, it runs `cancel()`
* If you have a value in the `textarea`, then edit the `textarea` and remove all of the contents and hit `cancel`, that works and the original value is shown. However, if you then go back in and edit the `textarea` again, the textarea is empty while you're editing it.

### DIFF
--- a/app/ui/src/app/common/editable/editable-textarea.component.ts
+++ b/app/ui/src/app/common/editable/editable-textarea.component.ts
@@ -24,8 +24,8 @@ import { EditableComponent } from './editable.component';
         autocomplete="off"
         aria-label="description"
         (keyup)="valueChanged($event)"></textarea>
+      <span class="help-block pull-left" *ngIf="errorMessage">{{ errorMessage }}</span>
       <div class="action-buttons">
-        <span class="help-block pull-left" *ngIf="errorMessage">{{ errorMessage }}</span>
         <button
           type="button"
           class="btn btn-primary form-control-pf-save"
--- a/app/ui/src/app/common/editable/editable-textarea.component.ts
+++ b/app/ui/src/app/common/editable/editable-textarea.component.ts
@@ -4,26 +4,45 @@ import { EditableComponent } from './editable.component';
 @Component({
   selector: 'syndesis-editable-textarea',
   template: `
-    <ng-template [ngIf]="!editing">
-      <em class="text-muted" *ngIf="!value">
-        {{ placeholder }}
-      </em>
-      <ng-container *ngIf="value">
-        {{ value }}
-      </ng-container>
-      <button type="button" class="btn btn-link" (click)="editing = true">
-        <i class="fa fa-pencil" aria-hidden="true" title="Click to edit"></i>
+    <div class="form-control-pf-editable form-control-pf-full-width"
+      [ngClass]="{'form-control-pf-edit': editing, 'has-error': errorMessage}">
+      <button
+        (click)="startEditing(textareaInput)"
+        type="button"
+        class="form-control-pf-value">
+        <em class="text-muted" *ngIf="!value">
+          {{ placeholder }}
+        </em>
+        <ng-container *ngIf="value">
+          {{ value }}
+        </ng-container>
+        <i class="glyphicon glyphicon-pencil" aria-hidden="true"></i>
       </button>
-    </ng-template>
-
-    <ng-template [ngIf]="editing">
-      <div class="form-group" [ngClass]="{'has-error': errorMessage}">
-        <textarea #textareaInput class="form-control" [ngModel]="value"></textarea>
-        <span class="help-block" *ngIf="errorMessage">{{ errorMessage }}</span>
+      <textarea #textareaInput
+        [ngModel]="value"
+        class="form-control form-control-pf-editor"
+        autocomplete="off"
+        aria-label="description"
+        (keyup)="valueChanged($event)"></textarea>
+      <div class="action-buttons">
+        <span class="help-block pull-left" *ngIf="errorMessage">{{ errorMessage }}</span>
+        <button
+          type="button"
+          class="btn btn-primary form-control-pf-save"
+          aria-label="Save"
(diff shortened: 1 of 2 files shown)


gashcrumb:
Hey!

So I pushed a commit that fixes those issues you mentioned.  I think part of the problem is that behind the scenes `NgModel` takes care of updating the DOM element for us in these controls.  I've added a bit to the template that gets a reference to the `NgModel` instance, from there in the controller we then have a lot more power over how the value gets updated from the form control etc.

I think this PR needs a bit more work, ideally I think we should avoid accessing the DOM value directly and instead use `NgModel`'s API -> https://angular.io/api/forms/NgModel#ngmodel

In `NgModel` there's a `model` and `viewModel` property, `model` refers to the value that's passed in via the template, and `viewModel` is the value you see on the screen but ironically a value that you don't easily get access to from your controller :-)  So that'd be a good option to use in `valueChanged`, compare the `model` and `viewModel` properties of the `NgModel` instance.  I think the `dirty` property could also be used, but that remains `true` if you say change the value and then change the value back to the original value in the control, like add a space and then delete that space.

We do still need a handle on the DOM element as well so we can focus it when you click the edit button, I put in some quick hacks here and there so the editable property base class will work with either the textarea or text input templates there, I'm guessing we'll need to update the text input component in a subsequent PR, right?

michael-coker:
@gashcrumb thanks! Looks good!

> I think the dirty property could also be used, but that remains true if you say change the value and then change the value back to the original value in the control, like add a space and then delete that space.

I'm OK with that, especially if it keeps the code cleaner. 

And yeah, at some point we will want another PR to update the editable-text component, but holding off for now since the text that's editable is in an `h1` and this new edit pattern doesn't work well with huge text. I think it might be weird to make the editable text small when editing, then large again when it's displayed, so we're just going to stick with the current UI for that component for now. When we redo the way the connection and integration details pages are displayed, we might be able to put this new pattern in place if we display the name somewhere on the page in normal text (similar to the editable "name" on the API connector details page).

Should we commit this as is or refactor using `NgModel` as you proposed? The `NgModel` properties are all new to me, but I can read up on them and take a stab, or whatever you'd suggest.

gashcrumb:
@michael-coker yeah if you want to spend more time on it that'd be awesome, otherwise I'll be happy to merge it tomorrow morning, leave a comment though if you want to work on it more but haven't had a chance to by then, sound good?

michael-coker:
@gashcrumb cool! I should be able to work on it this afternoon.

michael-coker:
@gashcrumb should the code you put in place to focus on the `textarea` in `startEditing()` focus that element? Doesn't appear to be working for me. However, this works...

`el.valueAccessor._elementRef.nativeElement.focus()`

Also, is there a way to keep `blur()` on the `textarea` from running `cancel()` if you click on the `disabled` check box?

gashcrumb:
> @gashcrumb should the code you put in place to focus on the textarea in startEditing() focus that element?

Strange it was working as pushed here this morning.  Though one thing about that `startEditing()` function, the text area control actually passes in an `NgModel` instance, the text area is available via the `control` attribute.  Maybe it depends on what page is used, I was testing it on the integration detail page.

> Also, is there a way to keep blur() on the textarea from running cancel() if you click on the disabled check box?

I'm not sure which check box you mean here?  But if it's the disabled property on the text area element you can check it in the controller and have `cancel()` return without doing anything in the controller.

gashcrumb:
> I'm not sure which check box you mean here? But if it's the disabled property on the text area element you can check it in the controller and have cancel() return without doing anything in the controller.

Wait, I get it, you mean the submit button.  Hmm, that's a tricky one that I'd probably have to play around with to figure out a good solution.

michael-coker:
> Wait, I get it, you mean the submit button. Hmm, that's a tricky one that I'd probably have to play around with to figure out a good solution.

Yep! Sorry, poor choice of words :)

gashcrumb:
So the main challenge I'm hitting this morning while trying to find solutions is that we've basically 1 component with 3 templates, we're trying to solve this issue specifically for the textarea, but we can't break the input/tag input variants of the control.  But now the textarea template is different than the text input template, so the component starts to get more complicated while dealing with these differences.  

I wonder, maybe we should take a step back, what are the exact gaps that we need to fix in these controls, specifically the textarea and text input?  I think we need to address them across all 3 templates, tbh I don't think the tag input is currently used but still, it's there and could be brought back at any point.

michael-coker:
Ah ok, I chatted with @dongniwang about it and we're going to meet at some point this or next week and go over how we can update both the name and description editor UI so they aren't out of sync. I'll go ahead and close this PR for now.

The changes to the UI are Patternfly has a bunch of classes/elements to use that control the way the inline editable area looks by default, on focus/hover, then in edit mode, with updates to spacing, buttons, etc.

The main functionality differences between what we have now and what we're looking to achieve are 

1. disable "save" button until content has changed
2. focus element when in edit mode
3. `blur` or whatever we use that indicates the element has lost focus (except when you click the disabled "save" button) cancels the edit
4. addition of a "clear value" button (in `<input type="text">` elements) that clears out the value of the element
5. a "spinner" in the element that shows while the value is updating. From a design perspective, not necessary if it's a quick action, but maybe it makes sense from an application standpoint because even quick actions could fail and the spinner is an indication that something isn't working?

Can you provide an example of what the tag input looked like and how it worked? Looking at the component code, maybe some sort of list of filteresque labels that you can add/remove/edit? 

And not sure if you saw it, but here's the [design](http://www.patternfly.org/pattern-library/forms-and-controls/inline-edit/#design) and [code sample](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/inline-edit-for-form.html) for the new inline edit pattern.

PR #2905, opened 2018-06-26
chore: replace publish/unpublish with start/stop and equiv verbs

### LINKED ISSUE #2896: Consider changing Publish/Unpublish to Start/Stop
## This is a...
[X] Feature request
</code></pre>

## The problem
The terms publish and unpublish are pretty severe from an end user POV.  If we want to call the initial button to deploy "Publish", I'm OK with that, but I think the state changes available for deployed integrations should be

- Start (was publish)
- Stop (was unpublish)
- Delete

Thoughts?

## Expected behavior
Action names meet user expectations.  FWIW, I got this feedback from a user. :-)

gashcrumb:
fixes #2896

'cause @kcbabo said so :-)

@TovaCohen FYI

quick example, also will show "Stopping" and "Starting" during state changes at the moment, dialogs and notifications are update as well:
capture

Also moved all of this text out of the app and into the dictionary file.  "Pending" shows up on the chart, previously the chart on the dashboard wasn't showing integrations that were in a transitional state, i.e. starting/stopping.  Which looks odd when you start/stop an integration and the total goes to 0 while the integration is transitioning.

I've kept "Pending" for this particular case for now, if there's a better term lemme know.  I didn't like the idea of having "Running", "Stopped", "Starting" and "Stopping" all in the legend there, seemed a bit much.


TovaCohen:
@gashcrumb  I think "Pending" is fine. I did consider "Transitioning", "Processing", "Spinning Up or Down", but I think "Pending" is just fine. Is "Pending"  a state, along with Draft, Running, Stopped, Error?
For a description of Pending - is this correct:
Pending indicates that this version of the integration is in the process of starting or stopping; it is not yet running or stopped.

gashcrumb:
@TovaCohen ok cool, thanks!

TovaCohen:
Stan, the image above, does the user see that when they click "Home", or "Integrations", in the navigation panel? I can't get the staging site to show me anything. :-(

gashcrumb:
@TovaCohen Yeah, that's a section of the 'Home' or Dashboard page.

PR #2924, opened 2018-06-28
feat: skip the action config if there's no properties

### LINKED ISSUE #2568: Skip Action configuration if there is no configuration
## This is a...

<pre><code>
[X] Feature request
[ ] Bug report  
</code></pre>

Not sure whether it already has been reported, but it would be awesome if we could avoid this empty unnecessary screen which forces the user to push a button even when an action doesn't need to be configured.

image

gashcrumb:
fixes #2568

### DIFF
--- a/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts
+++ b/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts
@@ -95,8 +95,15 @@ export class IntegrationConfigureActionComponent implements OnInit, OnDestroy {
     if (this.position > 0) {
       direction = 'input';
     }
-    this.router.navigate(['describe-data', this.position, direction], {
-      relativeTo: this.route.parent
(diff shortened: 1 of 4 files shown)


app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts
```diff
@@ -209,14 +216,18 @@ export class IntegrationConfigureActionComponent implements OnInit, OnDestroy {
...
       this.loading = false;
       // TODO figure out how to get a link in here that works
       this.error = {
         class: 'alert alert-info',
         icon: 'pficon pficon-info',
         message: 'There are no properties to configure for this action.'
       };
+      const metadata = this.step.metadata || {};
+      if (!metadata.configured) {
+        this.finishUp();
```
kahboom: We could probably do with a better name for this. 😂

Issue #2934, opened 2018-06-29
change API client connector details from inline edit to bulk edit


michael-coker:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
We should only be using inline edit when there is a single field to edit. And the icon you use/upload should be in the thumbnail beside the upload file input.

## Expected behavior
* This page should use a bulk edit form.
* When you're viewing connector details, the image should be in the upload preview box.
* When you upload a new image, it should be listed in the preview box, and only uploaded if you save the form.

PR #2935, opened 2018-06-29
change api connector details from inline editor

michael-coker:
fix https://github.com/syndesisio/syndesis/issues/2934

one thing I couldn't figure out was how to reset the form to the default/original values if you edit the form and click "cancel"

### DIFF
--- a/app/ui/src/app/customizations/api-connector/api-connector-info/api-connector-info.component.ts
+++ b/app/ui/src/app/customizations/api-connector/api-connector-info/api-connector-info.component.ts
@@ -1,12 +1,4 @@
-import {
-  Component,
-  OnInit,
-  Input,
-  Output,
-  EventEmitter,
-  Renderer2,
-  HostListener
-} from '@angular/core';
+import { Component, OnInit, Input, Output, EventEmitter, Renderer2, HostListener, ElementRef, ViewChild } from '@angular/core';
 import { FormBuilder, FormGroup, Validators } from '@angular/forms';
 
 import {
@@ -25,10 +17,16 @@ export class ApiConnectorInfoComponent implements OnInit {
   @Input() apiConnectorData: ApiConnectorData;
   @Output() update = new EventEmitter<CustomConnectorRequest>();
 
+  @ViewChild('connectorIconImg') connectorIconImg: ElementRef;
+  @ViewChild('connectorIconInput') connectorIconInput: ElementRef;
+  //@ViewChild('hostInput') hostInput: ElementRef;
+
   createMode: boolean;
+  editMode: boolean;
   apiConnectorDataForm: FormGroup;
   editControlKey: string;
   iconFile: File;
+
   private isDirty: boolean;
 
   constructor(private formBuilder: FormBuilder, private renderer: Renderer2) {}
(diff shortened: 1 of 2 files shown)


app/ui/src/app/customizations/api-connector/api-connector-info/api-connector-info.component.ts
```diff
@@ -52,63 +60,77 @@ export class ApiConnectorInfoComponent implements OnInit {
...
 
-  @HostListener('document:click', ['$event'])
-  private onDocumentClick(event: Event): void {
-    const element: any = event.target;
-    const elementTag = element.tagName.toLowerCase();
-    if (elementTag !== 'input' && elementTag !== 'textarea' && this.editControlKey) {
-      setTimeout(() => this.onSubmit(), 0);
+  onSubmit(): void {
+    if (this.apiConnectorDataForm.valid) {
+      if (this.isDirty) {
```
gashcrumb: `isDirty` can probably be replaced with looking at `this.apiConnectorDataForm.dirty`

app/ui/src/app/customizations/api-connector/api-connector-info/api-connector-info.component.ts
```diff
@@ -52,63 +60,77 @@ export class ApiConnectorInfoComponent implements OnInit {
...
+
+  clearValue(el) {
+    el.nativeElement.value = '';
+  }
+
+  isDisabled() {
+    return !this.editMode;
+  }
+
+  onChange() {
```
gashcrumb: Check if there's a `dirty` property on `apiConnectorDataForm`, if there is, then you don't need this.

app/ui/src/app/customizations/api-connector/api-connector-info/api-connector-info.component.ts
```diff
@@ -52,63 +60,77 @@ export class ApiConnectorInfoComponent implements OnInit {
...
+    this.enableEdit();
   }
 
-  onEditEnable(event: Event, key: string): void {
-    event.preventDefault();
-    event.stopPropagation();
-    this.editControlKey = key;
+  enableEdit() {
+    this.editMode = true;
+    this.apiConnectorDataForm.enable();
```
gashcrumb: Can probably grab the values from `apiConnectorDataForm` here like:

`this.oldValues = { ...apiConnectorDataForm.value };`

Then when the user clicks 'Cancel':

`this.apiConnectorDataForm.value = this.oldValues`

which I think *should* reset the values displayed in the form.

app/ui/src/app/customizations/api-connector/api-connector-info/api-connector-info.component.ts
michael-coker: `onChange()` only handles the `input type="file"` used to upload an image, and sets an `iconFile` var to populate the thumbnail `img`. I'm not sure how `dirty` could be used here?

app/ui/src/app/customizations/api-connector/api-connector-info/api-connector-info.component.ts
michael-coker: looks like `value` is a read only property, so I can't overwrite it, but I can use `setValue()`.

michael-coker:
@gashcrumb thanks for the feedback! updated with these changes.

PR #2944, opened 2018-07-02
chore: upgrade to patternfly-ng 4.x (partial)

### LINKED ISSUE #2941: Syndesis UI not displaying integrations/connections
## This is a...

<pre><code>
[ ] Feature request
[ x ] Bug report  
</code></pre>

## The problem

There seems to be an issue when I try to display "integrations" and "connections" page.

## Screenshot

syndesis_not_working

1. syndesis from master is running
2. try to access integrations/connections page

gashcrumb:
should hopefully fix #2941 as with patternfly 3.x we pull in an old version of ngx-bootstrap, which causes the runtime issue seen in #2941

### DIFF
--- a/app/ui/package.json
+++ b/app/ui/package.json
@@ -58,7 +58,7 @@
     "ngx-bootstrap": "^3.0.0",
     "ngx-chips": "^1.9.2",
     "patternfly": "3.51.2",
-    "patternfly-ng": "^3.0.0",
+    "patternfly-ng": "^4.0.0",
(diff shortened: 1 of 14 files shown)


seanforyou23:
I still see some bugginess around the forms (looking into that now) but doesn't impede the users workflow from what I've seen. This fixes the obvious and immediate issue and leaves it in a less broken state than it was, I say go for it

gashcrumb:
Cool, let's merge it and see what else shows up :-)

PR #2945, opened 2018-07-02
Change version to 1.3.11-SNAPSHOT - we need micro version for pipeline build.

cunningt:
Change version to 1.3.11-SNAPSHOT


cunningt:
Closing - Roland is doing a 1.3.11 tag which we can use instead.

rhuss:
Thanks @cunningt reason is that we try to keep the snapshot versions in the format 1.3-SNAPSHOT to avoid merge conflicts when keeping PR between releases.

PR #2951, opened 2018-07-03
chore: disable/remove package-lock.json in favor of yarn.lock

gashcrumb:
Given that package-lock.json is likely quite stale, yarn should be used to install dependencies.

@paoloantinori


paoloantinori:
should we document this somewhere? also, are we by any chance invoking npm to instlal dependencies in either our support scripts or in the Maven build?

gashcrumb:
I believe in the developer handbook for UI development we say to use yarn
when getting started.  Maven build uses yarn also.

paoloantinori:
Ah, perfect thanks!

Issue #2920, opened 2018-06-27
Update message dialog modals


michael-coker:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
This is a follow up from [updating the delete modal message dialogs](https://github.com/syndesisio/syndesis/issues/2879) to the new [PatternFly pattern for message dialogs](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/message-dialogs.html).

It looks like the wording between message dialogs is inconsistent (title, content, button text, etc), and we might be able to take advantage of the new styles for message dialogs for some of our modal actions.

### Delete OAuth
![screen shot 2018-06-27 at 3 49 15 pm](https://user-images.githubusercontent.com/35148959/41998979-bcc73c82-7a21-11e8-980c-1da4af4d7c05.png)

### Delete extension
![screen shot 2018-06-27 at 3 49 08 pm](https://user-images.githubusercontent.com/35148959/41998984-c34978e0-7a21-11e8-87a7-4d6561d6c94b.png)

### Delete extension in use with integration(s)
![screen shot 2018-06-27 at 3 53 16 pm](https://user-images.githubusercontent.com/35148959/41999137-4abdaeea-7a22-11e8-950f-9e5f40340843.png)

### Delete API client connector
![screen shot 2018-06-27 at 3 48 55 pm](https://user-images.githubusercontent.com/35148959/41998999-cc98f1d2-7a21-11e8-81ae-555e5093093c.png)

### Delete connection
![screen shot 2018-06-27 at 3 48 49 pm](https://user-images.githubusercontent.com/35148959/41999006-d3243cb4-7a21-11e8-8c99-69c38653b11b.png)

### Delete step from integration
![screen shot 2018-06-27 at 3 52 05 pm](https://user-images.githubusercontent.com/35148959/41999067-11a13df2-7a22-11e8-85b6-b69e1bf6e6d0.png)

### Delete integration
![screen shot 2018-06-27 at 3 48 42 pm](https://user-images.githubusercontent.com/35148959/41999038-f32b6c1c-7a21-11e8-9ad6-5d56b7ebe622.png)

### Start integration
![screen shot 2018-06-27 at 3 45 22 pm](https://user-images.githubusercontent.com/35148959/41998828-4fa39bfa-7a21-11e8-8abc-515fa51ac33a.png)

[...]

PR #2964, opened 2018-07-03
update modal message dialogs

michael-coker:
fix https://github.com/syndesisio/syndesis/issues/2920

### DIFF
--- a/app/ui/src/app/connections/list/list.component.html
+++ b/app/ui/src/app/connections/list/list.component.html
@@ -2,9 +2,9 @@
   <div class="container-cards-pf connections-list container-fluid">
     <div class="row row-cards-pf cards">
 
-      <syndesis-modal title="Confirm Delete"
+      <syndesis-modal title="Confirm Delete?"
                       type="delete"
-                      [message]="'Are you sure you would like to delete the \'' + selectedForDelete?.name + '\' connection?'">
+                      [message]="'Are you sure you want to delete the \'' + selectedForDelete?.name + '\' connection?'">
       </syndesis-modal>
 
       <ng-container *ngIf="connections && connections.length === 0">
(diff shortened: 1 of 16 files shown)


app/ui/src/app/connections/list/list.component.html
```diff
@@ -2,9 +2,9 @@
   <div class="container-cards-pf connections-list container-fluid">
     <div class="row row-cards-pf cards">
 
-      <syndesis-modal title="Confirm Delete"
+      <syndesis-modal title="Confirm Delete?"
                       type="delete"
-                      [message]="'Are you sure you would like to delete the \'' + selectedForDelete?.name + '\' connection?'">
+                      [message]="'Are you sure you want to delete the \'' + selectedForDelete?.name + '\' connection?'">
```
gashcrumb: It'd be cool if we could move this and other strings to the `en_GB.json` file for example like:

https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/connections/create-page/configure-fields/configure-fields.component.html#L59-L61

app/ui/src/app/customizations/api-connector/api-connector-list/api-connector-list.component.html
```diff
@@ -53,9 +53,8 @@ <h1>API Client Connectors</h1>
 </div>
 
 <ng-template #confirmDeleteModalTemplate>
-  <delete-confirmation-modal (delete)="onDeleteConfirm($event)">
-    <p class="lead">Removing a connector removes all of it's connections</p>
-    <p>This will remove all connections for this connector. Deleting this connector does not affect running integrations.</p>
-    <p>Are you sure you want to delete the "{{ deletableApiConnectorName }}" connector?</p>
+  <delete-confirmation-modal (delete)="onDeleteConfirm($event)" title="Confirm Delete?">
+    <p class="lead">Are you sure you want to delete the "{{ deletableApiConnectorName }}" connector?</p>
```
gashcrumb: You can also replace these lines, also you can have arguments in the string templates, for example:

https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/connections/create-page/configure-fields/configure-fields.component.html#L49

gashcrumb:
Lemme push a fix for those tests to your PR...

PR #2979, opened 2018-07-05
Allow to specify http headers for datamapper step in Webhooks

### LINKED ISSUE #2902: Allow to specify http headers for datamapper step in Webhooks
From the user story, when starting a integration from a Webhook, HTTP headers should be available in a datamapper step.

The user may choose which HTTP headers should be available for mapping.

lburgazzoli:
Fixes #2902 

This PR add initial support for webhooks headers averaging a specific json schema to define the headers that have to be included, the proposed schema template is:
 
```json
{
  "$schema": "http: //json-schema.org/schema#",
   "id": "io:syndesis:webhook",
   "type": "object",
   "properties": {
      "parameters": {
         "type": "object",
         "properties": {
         }
      },
      "body": {
         "type": "object",
         "properties": {
         }
      }
   }
}
```

The webhook connector wraps the payload only if:
- the data shape is of type `JSON Schema`
- the schema id is `io:syndesis:webhook`

### DIFF
--- a/app/connector/webhook/pom.xml
+++ b/app/connector/webhook/pom.xml
@@ -29,6 +29,16 @@
   <packaging>jar</packaging>
 
   <dependencies>
+
+    <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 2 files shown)


lburgazzoli:
@nicolaferraro done

Issue #2991, opened 2018-07-05
Create API connector form submit not working


michael-coker:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
When you create a new client connector, on the last step when you click "create api connector" it doesn't do anything.

## Expected behavior
That should create the connector.

PR #2992, opened 2018-07-05
fix submit handler for saving api connector details form when in create mode

michael-coker:
fix https://github.com/syndesisio/syndesis/issues/2991


gashcrumb:
good catch!

Issue #2988, opened 2018-07-05
Move message dialog text into dictionary file


michael-coker:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
From PR https://github.com/syndesisio/syndesis/pull/2964. The PR was merged before I finished.

Still need to dialog text from these modals to dictionary file
* delete api connector
* delete extension
* delete step (integration editor)
* delete/stop/start integration (add titles and button text)
* delete OAuth credentials (add title and button text)

gashcrumb:
oh sorry!  I didn't catch that...

michael-coker:
@gashcrumb no worries!

PR #2994, opened 2018-07-05
Move message dialog text into dictionary file

michael-coker:
fix https://github.com/syndesisio/syndesis/issues/2988


michael-coker:
@gashcrumb do you know why this test is failing? Looks like it's failing because of the translation service, but I looked at the test and can't figure out why.

gashcrumb:
@michael-coker yep, it's the same problem as the previous one, the test just needs to be updated to depend on the Core and Platform modules.  I can push a fix in a couple minutes.

michael-coker:
@gashcrumb thanks!

PR #2995, opened 2018-07-06
chore(ui): logout endpoint

### LINKED ISSUE #2561: Create LOGOUT Endpoint
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## The problem
Implement a "Logout" server-side endpoint, to take care of performing complex logic.
Different environments have a different specific configuration, and a server side endpoint is probably the ideal place to centralize the responsibility of taking care all the differences.

This is the continuation of this older enhancement: #2511

zregvart:
This adds a `/logout` endpoint backed by nginx and static logout.html
file.

When `/logout` endpoint is requested with the `SYNDESIS_XSRF_TOKEN` set
to `awesome` the response will expire the `_oauth_proxy` cookie. In any
case it will serve the `logout.html`.

The UI when performing the logout will replace the `window.document`
with the response of the `/logout` endpoint and will set the URL to
`/logout`.

This design protects the denial of service if the `/logout` endpoint is
invoked cross origin and allows for the served page to be reloaded
making it deterministic and thus safer.

The `logout.html` embeds the used CSS and just displays a simple _You
have been logged out_ message offering to Login again by linking to `/`.

Fixes #2561

### DIFF
--- a/app/ui/angular.json
+++ b/app/ui/angular.json
@@ -40,7 +40,8 @@
               "src/mstile-150x150.png",
               "src/safari-pinned-tab.svg",
               "src/manifest.webapp",
-              "src/browserconfig.xml"
+              "src/browserconfig.xml",
+              "src/logout.html"
             ],
             "styles": [
               "node_modules/patternfly/dist/css/patternfly.css",
--- a/app/ui/docker/nginx-syndesis.conf
+++ b/app/ui/docker/nginx-syndesis.conf
@@ -1,11 +1,22 @@
 server {
     listen       8080;
     server_name  localhost;
-    gzip on;
+    gzip         on;
+    root         /usr/share/nginx/html;
+
+    location = /logout {
+        set $cookie "";
+        if ($http_syndesis_xsrf_token = "awesome") {
+            set $cookie "_oauth_proxy=deleted; Expires=Thu, 01-Jan-1970 00:00:01 GMT; Domain=.$host; HttpOnly; Secure";
+        }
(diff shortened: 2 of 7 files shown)


gashcrumb:
Just a quick lint error looks like:

```
[INFO] /workspace/app/ui/src/app/core/providers/user-provider.service.ts
[INFO] ERROR: 48:9  semicolon  Missing semicolon
[INFO] 
```

zregvart:
> Should I wind up seeing the logout.html page? Testing here locally and even with an incognito window I seem to just wind up back at the openshift login page.

Kinda depends, it's served by nginx at `/logout` (without .html), and the cookie removing logic is also in the nginx configuration, so it won't actually work without nginx.

I can see logout.html if I run `yarn start` at `https://localhost:4200/logout.html`, so that helps with designing it. When developing I refrained from using `yarn start:minishift` as nginx is not used in that setup.

So with `yarn start:minishift` for `https://syndesis.$(minishift ip).nip.io/logout` we get 404, `https://syndesis.$(minishift ip).nip.io/logout.html` should show up though.

And for some background: I don't consider the way we currently perform authentication a API detail, it's handled by oauth-proxy at what I would call _platform_ level, so adding logic to the backend API didn't seem like a proper solution for this -- and that's why I added this to nginx configuration. The side effect of that is that the HTML of logout screen is also where it belongs in the UI project, easily discoverable when working on the front-end, so any design changes we would need can be worked on there.

gashcrumb:
Works for me then :-)

PR #3011, opened 2018-07-09
Enable state details monitoring by default and filtering to return pending state details only

dhirajsb:


dhirajsb:
@rhuss does the `app/server/runtime/src/main/resources/application.yml` file get pulled into `install/generator/04-syndesis-server.yml.mustache` somehow, or does that have to be edited separately?

rhuss:
@dhirajsb it has to be setup within the templates. Here's the entrypoint

https://github.com/syndesisio/syndesis/blob/de02e4c5f2fbe54cd75facb0a79361ff8bf20e13/install/generator/04-syndesis-server.yml.mustache#L176-L230

PR #3020, opened 2018-07-10
chore: OAuth app settings backed by configurati on from the backend

### LINKED ISSUE #2990: Fetch oauth app form configurations from the backend
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## The problem

Currently all the apps in the oauth settings page use the same form field configuration as it's defined client side, and the UI only fetches values from the backend and then populates those values in the form.  This leads to issues like #2921 and #2922 as not every setting in the generic form config would be relevant to a given oauth client, and also doesn't allow us to provide as nice of a UX, as we can't customize the oauth settings to match the target system's terminology as mentioned in https://github.com/syndesisio/syndesis/issues/240

## Expected behavior

The oauth settings form should really be defined by the connector rather than be hard-coded in the UI.

zregvart:
This changes the model of OAuthApp so it no longer has fixed properties.
Now the model is backed by properties/configuredProperties convention we
use for Connectors/Connections/..., allowing the properties as defined
by the Connector.

The properties used/exposed are determined by tag values, and those are
preexisting tags that are already used within the credentials module and
the Connector definitions.

Fixes #2990

### DIFF
--- a/app/ui/src/app/settings/oauth-apps/oauth-app-form.component.ts
+++ b/app/ui/src/app/settings/oauth-apps/oauth-app-form.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit, EventEmitter, Input, Output } from '@angular/core';
+import { Component, OnInit, EventEmitter, Input, Output, OnDestroy } from '@angular/core';
 import { Router } from '@angular/router';
 import { OAuthApp, OAuthApps } from '@syndesis/ui/settings';
 import { FormFactoryService } from '@syndesis/ui/platform';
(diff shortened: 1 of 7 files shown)


app/ui/src/app/settings/oauth-apps/oauth-app-form.component.ts
```diff
@@ -106,14 +72,13 @@ export class OAuthAppFormComponent implements OnInit {
...
   ngOnInit() {
-    const formConfig = (this.formConfig = JSON.parse(
-      JSON.stringify(OAUTH_APP_FORM_CONFIG)
-    ));
-    this.formModel = this.formFactory.createFormModel(
-      formConfig,
-      this.item.client,
-      ['*']
-    );
+    this.store.list.subscribe(oauthApps => {
```
gashcrumb: I would store the Subscription return value from this and have the component also implement `OnDestroy`, then add an `ngOnDestroy()` function and unsubscribe, like this -> https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.ts#L325-L327

zregvart:
@gashcrumb mind having a second look? thanks!

Issue #2993, opened 2018-07-05
Coding updates for simplified IVP design


amysueg:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
Implement the code to make the changes outlined in this UX Design: #2962.

## Expected behavior
Simplified, cleaner UI for the Integration Visualization Panel

## Screenshot
![screen shot 2018-07-05 at 4 14 03 pm](https://user-images.githubusercontent.com/31700119/42345780-79ee4ce0-806e-11e8-8940-1fddb1f8aa0b.png)

michael-coker:
@gashcrumb I think I can take care of all of this except for the connection & step details here: 

![screen shot 2018-07-05 at 3 17 45 pm](https://user-images.githubusercontent.com/35148959/42345958-a19f91ca-8066-11e8-8aa4-331e19b2368c.png)

Is that information currently available in the editor pages?

michael-coker:
@gashcrumb ah yes! Perfect, thanks. Though there is no info popup for the non-connection steps (filters, data mapper, log) - do those need any info below them?

gashcrumb:
We'll probably want this for steps too.  Steps don't have an 'action' like connections do, but there's config that we could figure out a way to show.

michael-coker:
@sjcox-rh what do you think about these states in the expanded version of the IVP?

![screen shot 2018-07-09 at 3 14 08 pm](https://user-images.githubusercontent.com/35148959/42473755-176fa008-838b-11e8-8dca-81d5146e06ba.png)
![screen shot 2018-07-09 at 3 14 11 pm](https://user-images.githubusercontent.com/35148959/42473757-18e2f1ec-838b-11e8-894b-46a301235b43.png)
![screen shot 2018-07-09 at 3 14 57 pm](https://user-images.githubusercontent.com/35148959/42473759-1a42290e-838b-11e8-9841-defa6d1270a6.png)

amysueg:
@michael-coker looking good.
I wonder why there is so much horizontal distance between the labels under each expanded section and the values. In other words, why so much white space between the ":" after "Name" for example, and the names? Seems like if that could be made tighter, then the expanded view would not need to take up quite as much horz space. @sjcox-rh WDYT?

michael-coker:
[...]

PR #3027, opened 2018-07-10
Simplified IVP

michael-coker:
fix https://github.com/syndesisio/syndesis/issues/2993

Outstanding issues I need code/design help with.

1. To get the "add step/connection" popover to display outside of the scrollable IVP (overflow hidden), I appended the popover to `body`. Then the `mouseover`/`mouseout` events on the "+" icon that would display/hide the popover no longer worked - as you hovered off the icon and into the popover, it would disappear. So I changed the popover to fire on click. I wonder if this is OK as is?
    * Another workaround this could be to display the popover below the "+" icon (`placement="bottom"`) when the IVP is collapsed, and to the right (`placement="right"`) when the IVP is expanded. Then the popover wouldn't overflow the IVP.
    * Or figure out how to keep the popover open when you hover over either the "+" icon or the popover itself.
1. ~~I don't know how to make tooltips work. @sjcox-rh wants a tooltip for the expand/collapse button that says "expand info", and a tooltip on the non-connection step icons that says "data mapper" or "log" or whatever that step is. For now, I just used the `title` attr instead. Maybe this is OK as is, too?~~ 
    * nevermind, figured it out

### DIFF
--- a/app/ui/src/app/integration/edit-page/edit-page.component.html
+++ b/app/ui/src/app/integration/edit-page/edit-page.component.html
@@ -3,7 +3,7 @@
     <div [class]="getPageRow()">
 
       <!-- Sidebar -->
-      <div [class]="getSidebarClass()">
+      <div class="wizard-sidebar">
         <syndesis-integration-flow-view></syndesis-integration-flow-view>
       </div>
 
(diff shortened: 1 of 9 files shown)


michael-coker:
cc @amysueg @sjcox-rh

gashcrumb:
Yeah, the popover has always been an issue, I think at least for a bit I actually had the mouseover/mouseout on the div *containing* the `+`, so the whole row, so that it kinda behaved properly.  All we really need is a stopgap until we can merge connections/steps into one page flow I guess.

PR #3056, opened 2018-07-12
Fix upgrade pod issues for operator

nicolaferraro:
@rhuss I've done some changes to fix the upgrade pod issues I've encountered:
- fixed script error
- upgraded `oc` client to 3.9 (requires EPEL repo)
- added rolebinding to "skip" list (until we figure out how to give proper permissions to the service account)

### DIFF
--- a/tools/upgrade/Dockerfile
+++ b/tools/upgrade/Dockerfile
@@ -1,7 +1,7 @@
 FROM centos:7
 
 ARG version=latest
-ARG openshift_yum_repo=centos-release-openshift-origin37
+ARG openshift_yum_repo=centos-release-openshift-origin39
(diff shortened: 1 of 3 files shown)


rhuss:
LGTM

@paoloantinori @heiko-braun as this is a bug fix for an upgrade issue, I tend to merge this. However, I wonder in general what the policy for merging PRs is as we are outside the feature window.

heiko-braun:
@rhuss I think the only policy is to apply common sense, i.e by looking at the justification and impact of a change. 

In this case a justification (bug fix) seems to be present. What about the risks involved? Do you see an issue updating `oc` to `3.9`?

rhuss:
Ok, there's no impact for upgrading to 3.9 as 3.9 is our baseline anyway.

Issue #2955, opened 2018-07-03
Basic filter step broken


tplevko:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

The basic filter step can't be used properly. I can't add any rules.

## Screenshot

![out](https://user-images.githubusercontent.com/4180208/42216649-de4fc392-7ec2-11e8-8a8d-eaa94951dea8.gif)

## Tasks involved / Steps to Reproduce

1. Syndesis from master running
2. Add new integration, specify some start/stop end connection
3. Specify the "Basic filter step"
4. Observe, that there is no possibility to specify options for basic filter step

gashcrumb:
Yep, think this was expected with the Angular 6 upgrade, @seanforyou23 I forget, had you raised an issue for this yet?  If not we can use this one :-)

PR #3059, opened 2018-07-12
fix(basic-filter-form): basic filter form rewritten with formbuilder

seanforyou23:
This PR fixes the basic filter step form which was broken as a result of the recent angular core upgrade (6.x). For this specific form we were bypassing our forms service and used the ng-dynamic-forms api directly. This stopped working and as a result we've rewritten this form using angular's FormBuilder.

Should help close https://github.com/syndesisio/syndesis/issues/2955

Remaining work items:
- Default value for predicate select control not correctly reflected in the UI
- Tie back in autocomplete for the property name field

I've marked this PR as WIP while I'm work through the previously mentioned issues. This work is generally stable enough, though, if it's desirable to go ahead and get it in considering the end of sprint is closing in.

ng-dynamic-forms => angular formbuilder
emit formattedProperties on form change
reload previously saved rule array groups on load
pass form validators as part of form construction
use button over link for in-view functionality
improve basic filter form keyboard experience
add missing accessible names


michael-coker:
Looks good! A few things:

1. When you choose something in the first select, the text beside it disappears.
1. The remove/trash icon shouldn't display if you only have one row.
1. Do we want a delete confirmation dialog to confirm deleting a row?

gashcrumb:
I'd be happy to merge this sooner rather than later, so worth taking off the WIP label, we can break out the remaining stuff as bugs to fix after the fact.

Did you nuke all the old ng-dynamic-form related code in this PR?  Worth doing if not, no reason to leave that stuff hanging around if it's not being used.

gashcrumb:
> Do we want a delete confirmation dialog to confirm deleting a row?

I don't recall the design having one, but yeah, this is worth at least considering, though an accidental delete in this case is less impactful from accidentally deleting a connection or integration.

app/ui/src/app/integration/edit-page/step-configure/filter-steps/basic-filter.component.ts
```diff
@@ -4,40 +4,35 @@ import {
...
-  rulesArrayModel: DynamicFormArrayModel;
+export class BasicFilterComponent implements OnChanges, OnDestroy {
+  form: FormGroup; // our model driven form
+  rulesArray: any = []; // what's a valid type to use here?
+  ruleOpsModel$: Observable<any>;
+  predicateOpsModel$: Observable<any>;
+  filterSettingsGroupObj;
+  pathsModel;
   loading = true;
+  formValudChangeSubscription: Subscription;
```
gashcrumb: nit: just a typo, Valud -> Valid

app/ui/src/app/integration/edit-page/step-configure/filter-steps/basic-filter.component.ts
```diff
@@ -4,40 +4,35 @@ import {
...
-export class BasicFilterComponent implements OnChanges {
-  basicFilterModel: DynamicFormControlModel[];
-  formGroup: FormGroup;
-  predicateControl: FormControl;
-  predicateModel: DynamicInputModel;
-  rulesArrayControl: FormArray;
-  rulesArrayModel: DynamicFormArrayModel;
+export class BasicFilterComponent implements OnChanges, OnDestroy {
+  form: FormGroup; // our model driven form
+  rulesArray: any = []; // what's a valid type to use here?
```
gashcrumb: Type here should be an array of these -> https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/edit-page/step-configure/filter-steps/filter.interface.ts#L8-L12

app/ui/src/app/integration/edit-page/step-configure/filter-steps/basic-filter.component.ts
```diff
@@ -4,40 +4,35 @@ import {
...
+  ruleOpsModel$: Observable<any>;
+  predicateOpsModel$: Observable<any>;
+  filterSettingsGroupObj;
+  pathsModel;
```
gashcrumb: Think the server gives us an array of strings for the `paths` in it's response

app/ui/src/app/integration/edit-page/step-configure/filter-steps/basic-filter.component.ts
```diff
@@ -66,53 +61,106 @@ export class BasicFilterComponent implements OnChanges {
...
     // Fetch our form data
     this.integrationSupportService
       .getFilterOptions(this.dataShape)
       .toPromise()
       .then((body: any) => {
         const ops = body.ops;
         const paths = body.paths;
-        initializeForm(ops, paths);
+        this.ruleOpsModel$ = of(body.ops);
+        this.pathsModel = body.ops;
```
gashcrumb: This should be 'paths' shouldn't it?  This would eventually hold an array of strings that would be used as a typeahead for the 'path' part of a given rule in the ruleset array.

gashcrumb:
Great stuff @seanforyou23 thanks!

PR #3061, opened 2018-07-13
Added publish monitoring service to syndesis server application.yml template

dhirajsb:


dhirajsb:
@rhuss a simple little change, just making sure it's in the right place

PR #3069, opened 2018-07-13
chore(db): add scheduled database maintenance

### LINKED ISSUE #2912: Optimize metrics retrieval
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem

When we present the dashboard (first screen), we have a number of request for `/api/v1/metrics/integrations`. I counted 6 on our staging environment, this uses connections to the database and prevents requests to `/api/v1/integrations` from completing: we get a 504.

This is then perceived as slowness of the UI.

I think we either need to reduce the rate of fetching metrics, prioritize fetching other data or push metrics from the server to the UI.

## Expected behavior

We need to serve the data for the most compelling parts of the UI first and reduce the load on the database.

## Screenshot

screenshot from 2018-06-27 15-14-31
screenshot from 2018-06-27 15-17-56

## API Endpoints and Schemas

`/api/v1/metrics/integrations`
`/api/v1/integrations`

1. Reload the dashboard on staging site
2. Observe the numerous requests to metrics and errors while fetching integrations in the UI and in the network inspector
3.
4.

zregvart:
This adds a CronJob backed by ConfigMap that contains a shell script to
run periodical database maintenance.

Currently it runs `VACUUM ANALYSE jsondb` and `REINDEX TABLE jsondb` at
04:15 each day.

Fixes #2912

### DIFF
--- a/install/generator/04-syndesis-db.yml.mustache
+++ b/install/generator/04-syndesis-db.yml.mustache
@@ -65,6 +65,71 @@
       /var/lib/pgsql/sampledb/add-sample-db.sh &>  /proc/1/fd/1
 
 - apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    labels:
+      app: syndesis
+      syndesis.io/app: syndesis
+      syndesis.io/type: infrastructure
+      syndesis.io/component: syndesis-db
+    name: syndesis-db-maintenance-script
+  data:
+    maintenance.sh: |
+      #!/bin/bash
+      cat << EOF| psql -h syndesis-db -U $POSTGRESQL_USER -d $POSTGRESQL_DATABASE
+      VACUUM FULL ANALYSE jsondb;
+      REINDEX TABLE jsondb;
+      SELECT COUNT(*) FROM jsondb;
+      EOF
+
+- apiVersion: batch/v1beta1
+  kind: CronJob
+  metadata:
+    name: syndesis-db-maintenance
+    labels:
+      app: syndesis
+      syndesis.io/app: syndesis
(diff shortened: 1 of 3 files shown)


install/generator/04-syndesis-db.yml.mustache
```diff
@@ -64,6 +64,71 @@
...
+            containers:
+            - name: syndesis-db-maintenance
+              env:
+              - name: POSTGRESQL_USER
+                value: ${POSTGRESQL_USER}
+              - name: PGPASSWORD
+                value: ${POSTGRESQL_PASSWORD}
+              - name: POSTGRESQL_DATABASE
+                value: ${POSTGRESQL_DATABASE}
+              image: centos/postgresql-95-centos7
```
zregvart: I could not find a way to reference the image stream here. Any ideas?

install/generator/04-syndesis-db.yml.mustache
```diff
@@ -64,6 +64,71 @@
...
+                - -c
+                - /data/maintenance.sh
+              volumeMounts:
+              - mountPath: /data
+                name: syndesis-db-maintenance-script
+            restartPolicy: Never
+            volumes:
+            - name: syndesis-db-data
+              persistentVolumeClaim:
+                claimName: syndesis-db
```
rhuss: @zregvart In certain setups (e.g. OpenShift Starter), it's forbidden to use the same PVC in different pods. 

This might be an issue e.g. on our dedicated cluster.

Actually we do this for our update pod, too, so that might be issue over there as well.

PR #3079, opened 2018-07-14
Added deployer pod monitoring to state details

dhirajsb:


dhirajsb:
@gashcrumb I went ahead and added deployer pod monitoring as well, as it could theoretically have issues like running out of resources when deploying a new integration. 
No UI changes should be required, as the backend will just return the appropriate pod name and  `linkType` for the deployer pod.

Issue #2693, opened 2018-05-22
Integration with same name - cumbersome change of specified name


tplevko:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

I noticed, that on creation of the integration, the name of the integration is saved in UI even in case, it is not unique. This causes quite unpleasant user experience, when the user has to change the integration name in the upper left corner, where the integration name, which can't be used is displayed. 
His other option is to cancel the whole not yet saved integration and go through the creation process again.
I think this could be prevented, if the integration name, which can't be used was removed and the user would be able to start with the naming process again from scratch.  

## Expected behavior

This naming process in case the first name of the integration was not unique should be improved.

## Screenshot

![out](https://user-images.githubusercontent.com/4180208/40346828-4efce58e-5d9e-11e8-9976-a7fba25cd7ea.gif)

## Tasks involved / Steps to Reproduce

1. Create integration with "name"
2. Create second integration, when naming it, select the same "name"
3. Notice the error, the "Publish" button spins without stopping
4. The name can't be easily changed through the name input field

phantomjinx:
Requires the Publish button be re-enabled and the spinner css classes removed.

So, when the 'not unique' error is encountered:
* User changes the name in the original input text box (no need to use the top-left text component);
* User clicks Publish a second time;
* The integration begins publishing.

This works because the underlying model of the name text box is still tied name() setter function so still emits the change property event to the currentFlowService to update the name.

To remove the spin classes and re-enable the Publish button will require handling the http callback error that causes the display of the original 'not unique' error.

PR #3084, opened 2018-07-16
Issue #2693: resuming after publish of integration violates same name

phantomjinx:
* When the publish operation fails, the publish button is left disabled and
  the spinner is displayed. By setting publishInProgress to false in the
  event of an error will reset both of these and allow the name to be
  re-edited and the publish be re-attempted with the new name.


oscerd:
@paoloantinori  this looks good but we are in code freeze, what do you think?

gashcrumb:
It's a bug fix and not new functionality, so I think it's fine.

PR #3086, opened 2018-07-16
fix: don't try to mount syndesis-db-data volume

zregvart:
Removes the mount of `syndesis-db-data` volume, it's not required and
prevents the creation of the pod as the `syndesis-db` maintains a lock
on `syndesis-db-data` volume.

### DIFF
--- a/install/syndesis.yml
+++ b/install/syndesis.yml
@@ -570,9 +570,6 @@ objects:
                 name: syndesis-db-maintenance-script
             restartPolicy: Never
             volumes:
-            - name: syndesis-db-data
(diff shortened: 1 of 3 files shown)


install/syndesis.yml
```diff
@@ -1212,6 +1209,11 @@ objects:
           sensitive: false
         jsondb:
           enabled: true
+      monitoring:
```
rhuss: @zregvart any idea where this comes from ?

rhuss: I guess it comes from a PR where the generator has not been run.

PR #3097, opened 2018-07-17
Unecessary requests to `/api/v1/integrations`

### LINKED ISSUE #3071: Unecessary requests to `/api/v1/integrations`
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem

It seems that we're hitting the `/api/v1/integrations` endpoint multiple times. This really hurts the backend (~10sec responses).

## Expected behavior

Load only when needed, probably only when update is received via socket?

## Screenshot

On staging:
peek 2018-07-13 16-54

## API Endpoints and Schemas

`GET /api/v1/integrations`

1. Open Syndesis
2. Wait on the dashboard screen doing nothing
3. Notice the requests going to `/api/v1/integrations`

zregvart:
Copy paste from the commits:

**fix(server): don't perform updates without changes**

When we compare periodically state of objects in controllers we run for integration deployments, integrations, connections and connectors we should not generate updates of the JsonDB or emit changes via WebSockets if the state did not change.

This checks if the previous state is equal to the current state and only performs the database update end emits the WebSocket event if the change has occurred. Thusly it increases our performance by reducing the number of refreshes in the UI and decreasing the database load for updates and fetches initiated by the emitted change to the UI.

**fix(ui): consider refreshing the store only for events of the stores kind**

Events that are not related to the store should not be handled by the store.

With this we check for the `kind` of the store and if it matches the `kind` of the event then it will be processed by the store, i.e. reload the data from the server.

This decreases the load on the server by not requesting data for unrelated update events.

Fixes #3071

### DIFF
--- a/app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/ConnectionUpdateHandler.java
+++ b/app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/ConnectionUpdateHandler.java
@@ -20,6 +20,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import javax.validation.Validator;
 
@@ -64,6 +65,7 @@ public class ConnectionUpdateHandler extends AbstractResourceUpdateHandler<Conne
             dataManager.fetchAllByPropertyValue(Connection.class, "connectorId", id)
                 .filter(connection -> connection.getConnector().isPresent())
                 .map(connection -> computeBoard(connection, connection.getConnector().get(), connector))
+                .filter(Objects::nonNull)
                 .forEach(boards::add);
         }
 
@@ -78,8 +80,7 @@ public class ConnectionUpdateHandler extends AbstractResourceUpdateHandler<Conne
 
         if (board != null) {
             builder = new ConnectionBulletinBoard.Builder()
-                .createFrom(board)
-                .updatedAt(System.currentTimeMillis());
+                .createFrom(board);
         } else {
             builder = new ConnectionBulletinBoard.Builder()
                 .id(KeyGenerator.createKey())
@@ -104,7 +105,13 @@ public class ConnectionUpdateHandler extends AbstractResourceUpdateHandler<Conne
         builder.putMetadata("connector-version-connection", Integer.toString(oldConnector.getVersion()));
         builder.messages(messages);
 
-        return builder.build();
+        ConnectionBulletinBoard updated = builder.build();
+
+        if (!updated.equals(board)) {
+            return builder.updatedAt(System.currentTimeMillis()).build();
+        }
+
+        return null;
     }
 
     private static Map<String, String> merge(final Map<String, String> one, final Map<String, String> two) {
--- a/app/server/controller/src/main/java/io/syndesis/server/controller/integration/IntegrationController.java
+++ b/app/server/controller/src/main/java/io/syndesis/server/controller/integration/IntegrationController.java
@@ -204,12 +204,14 @@ public class IntegrationController {
                     // handler.execute might block for while so refresh our copy of the integration
                     // data before we update the current status
                     IntegrationDeployment current = dataManager.fetch(IntegrationDeployment.class, integrationDeploymentId);
-                    dataManager.update(current.builder()
+                    final IntegrationDeployment updated = current.builder()
                         .statusMessage(Optional.ofNullable(update.getStatusMessage()))
                         .currentState(update.getState())
                         .stepsDone(update.getStepsPerformed())
-                        .updatedAt(System.currentTimeMillis())
-                        .build());
+                        .build();
+                    if (!(updated.equals(current))) {
+                        dataManager.update(updated.builder().updatedAt(System.currentTimeMillis()).build());
+                    }
                 });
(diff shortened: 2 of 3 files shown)


app/ui/src/app/store/entity/entity.store.ts
```diff
@@ -90,7 +94,9 @@ export abstract class AbstractStore<
       this._current,
       this.changeEvents.pipe(
         filter(event => {
-          return event.id.startsWith(this.currentId);
+          log.debug(`got change event for kind: ${event.kind}, id: ${event.id}, `
+                 + `this store is for kind: ${this.service.kind}, id: ${this.currentId}`);
+          return event.kind === this.service.kind && event.id.startsWith(this.currentId);
```
gashcrumb: The only concern here I'd have is if an integration-bulletin-board event comes in the old code would reload the associated integration, which would then ensure that any new messages in the integration's bulletin board attribute would be picked up and displayed.  Same goes for integration deployments.  If updating these ultimately results in the associated integration having a change event triggered, then I'm all for this.  And then TBH we may as well suppress change events for deployments/bulletin boards, the UI doesn't maintain any local store of these objects.

app/ui/src/app/store/entity/entity.store.ts
```diff
@@ -72,7 +72,11 @@ export abstract class AbstractStore<
     return observableMerge(
       this._list,
       this.changeEvents.pipe(
-      debounceTime(500),
+        filter(event => {
+          log.debug(`got change event for kind: ${event.kind}, id: ${event.id}, this store is for kind: ${this.service.kind}`);
+          return event.kind === this.service.kind;
```
gashcrumb: Same comment for this line as below :-)

app/ui/src/app/store/entity/entity.store.ts
gashcrumb: Easy way to test BTW is to create an integration with an extension, then update the extension.  That should trigger a 'configuration required' type of message on the integration.

app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/ConnectionUpdateHandler.java
```diff
@@ -78,8 +80,7 @@ ConnectionBulletinBoard computeBoard(Connection connection, Connector oldConnect
 
         if (board != null) {
             builder = new ConnectionBulletinBoard.Builder()
-                .createFrom(board)
-                .updatedAt(System.currentTimeMillis());
```
lburgazzoli: why is "updatedAt" removed ?

app/ui/src/app/store/entity/entity.store.ts
zregvart: @gashcrumb how we have it now is that every event triggers `list` of every resource, perhaps we should remove this subscriber all together and listen only to individual resource changes. Too drastic?

app/ui/src/app/store/entity/entity.store.ts
zregvart: Yeah I don't think change of `integration-bulletin-board` will trigger a `integration` event.

What alternatives do we have?
 - add a new endpoint for the `integration-bulletin-board` and update integration state based on that, it might be difficult/touches a lot of places?
 - change the condition from `event.kind === this.service.kind` to `event.kind.startsWith(this.service.kind)`, seems a bit smelly, but could work
 - also trigger `integration` event on `integration-bulletin-board` change on the backend, not too difficult, but bit smelly -- I mean why do we have `integration-bulletin-board` event?

I think same arguments that can be made for `integration-bulletin-board` events can be made for `integration-deployment` events.

zregvart: Hmm, found out about `setChangeEventsFilter` that uses `event.kind.startsWith('integration')` in `IntegrationStore`.

So perhaps this filter can be removed as there is already one added from the `IntegrationStore`.

app/ui/src/app/store/entity/entity.store.ts
gashcrumb: no, this isn't shared for all resource, this is a base class for each resource, and we filter by kind in the constructor of this base class.  So this doesn't mean an event triggers fetching all resources, just for a given kind.

app/ui/src/app/store/entity/entity.store.ts
gashcrumb: Right, the existing code basically did that, it triggered an update when events came in for integrations and related types.

Why the UI has to deal with watching all these types is another question for another time perhaps :-)

zregvart:
@gashcrumb I've removed the UI bits from this, we should probably rethink about how we emit and handle events (created #3105). At least with the server bits in we won't get superfluous updates when no state change has occurred.

@lburgazzoli can you take a look at https://github.com/syndesisio/syndesis/pull/3097#discussion_r203044466 to see if that answers your question?

app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/ConnectionUpdateHandler.java
lburgazzoli: oh yeah, didn't saw that updatedAt was moved so fine with me.

PR #3102, opened 2018-07-17
Running yarn with Openshift using oc cluster up

phantomjinx:
* Can use the same scripts and configuration as running yarn with
  minishift. Couple of small modifications to the pertinent scripts.

* minishift-setup.sh
 * Provides an openshift-implementation independent mechanism to return the
   openshift console url

* minishift-restore.sh
 * Fixes the restoration of the syndesis-ui service and endpoint including
   extra labels and ensuring both are added back separately.
 * Effectively reverses/undoes the effect of minishift-setup.sh

### DIFF
--- a/app/ui/scripts/minishift-restore.sh
+++ b/app/ui/scripts/minishift-restore.sh
@@ -2,31 +2,85 @@
 
 set -euo pipefail
 
-# Backup the existing service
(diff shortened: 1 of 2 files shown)


phantomjinx:
Could have changed the script names to something more generic but figured there is a lot of documentation involving these names so not exactly worth it yet.

PR #3104, opened 2018-07-17
fix: ensure the integration and connection names are displayed properly

### LINKED ISSUE #3101: Integration import results UI
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## problem1
When I import an integration, the results seems to list all of my integrations in the import results area. Shouldn't that just list the integration(s) that were imported? I only imported one integration in the screenshot below.

To reproduce:

1. Create more than one integration.
2. Export one of your integrations.
3. Import it.
4. You should see all of your integrations listed in the import results area.

screen shot 2018-07-17 at 9 24 56 am

## problem2

If I remove the connections that an exported integration used, then import that integration, the import results list and buttons render but with no content for a moment. After a second or so, those areas populate with content. This seems to be dependent on the system speed and is much more apparent on staging.

1. Create an integration with 2 connections.
2. Export the integration.
3. Delete the integration and the 2 connections it used.
4. Import the integration.
5. Observe an empty results list and tiny buttons with no text.

screen shot 2018-07-17 at 9 49 18 am

cc @dongniwang

gashcrumb:
fixes #3101

Also removes the 'Cancel' button, as there's no 'Cancel' the way the API is currently designed.

### DIFF
--- a/app/ui/src/app/integration/import-export/integration-import-export.component.html
+++ b/app/ui/src/app/integration/import-export/integration-import-export.component.html
@@ -18,9 +18,9 @@
         </ol>
       </div>
       <div class="toolbar-pf-action-right">
-        <button class="btn btn-default"
-                (click)="showCancelModal()">
-          Cancel
+        <button class="btn btn-primary"
+                (click)="redirectBack()">
+          Done
         </button>
       </div>
     </div>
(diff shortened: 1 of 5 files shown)


app/ui/src/app/integration/import-export/integration-import-export.component.html
```diff
@@ -18,26 +18,13 @@
         </ol>
       </div>
       <div class="toolbar-pf-action-right">
-        <button class="btn btn-default"
-                (click)="showCancelModal()">
-          Cancel
+        <button class="btn btn-primary"
+                (click)="redirectBack()">
+          Done
```
kahboom: Shouldn't we have i18n for this?

gashcrumb: I'll get it in the next pass through :-)

PR #3112, opened 2018-07-18
fix: fetch more things from the backend, add a config parameter

### LINKED ISSUE #2925: Only 20 connectors are returned at most when creating a connection
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem

When you install a few extensions that add connectors you'll quickly find that some connectors vanish from the list of available connectors when you go to create a connection.

With a bit of testing by slowly adding connector extensions I found that the API response for /connectors appears to have a maximum limit of 20.

## Expected behavior

All connectors should be returned as the 1st page of the create connection wizard doesn't do pagination.

1. Open up the browser developer console and look at the network tab, check the "Preview" tab for the HTTP request to /connectors
2. Add a connector extension, note how the `totalCount` in the response increases by 1
3. Add more connector extensions so that you have more than 20 connectors
4. Note that `totalCount` stops at 20, and some connectors that were available are now missing.

gashcrumb:
fixes #2925

This just bumps up the item limit to 50 by default on the UI side, this can be controlled in `config.json` by adding a `perPage` attribute to the root object with a numeric value.

@rhuss FYI

Should probably create an epic for proper pagination as a future enhancement for all pages, what's there now I don't think will be suitable long term.

### DIFF
--- a/app/ui/src/app/store/action/action.service.ts
+++ b/app/ui/src/app/store/action/action.service.ts
@@ -2,10 +2,12 @@ import { Injectable } from '@angular/core';
 
 import { ApiHttpService, Action, Actions } from '@syndesis/ui/platform';
 import { RESTService } from '../entity';
+import { ConfigService } from '@syndesis/ui/config.service';
 
 @Injectable()
 export class ActionService extends RESTService<Action, Actions> {
-  constructor(apiHttpService: ApiHttpService) {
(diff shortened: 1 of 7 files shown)


gashcrumb:
Created https://github.com/syndesisio/syndesis/issues/3113 to hopefully track improved pagination, think it'll require work across the stack.

Issue #3041, opened 2018-07-11
Concur connector has incorrect data shapes


zregvart:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

Swagger file we created Concur connector from has little to do with reality. I just checked with [List Items API Swagger](https://developer.concur.com/api-explorer/v3-0/ListItems.swagger2.json) and the `GET /common/listitems` operation's response is defined as:

```json
    "ListItemGetCollection": {
      "properties": {
        "Items": {
          "$ref": "#/definitions/ListItemGet"
        },
        "NextPage": {
          "type": "string",
          "description": "The URI of the next page of results, if any."
        }
      }
```

And based on the response we receive it should be defined as:

```json
    "ListItemGetCollection": {
      "properties": {
        "Items": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/ListItemGet"
          }
        },
        "NextPage": {
          "type": "string",
          "description": "The URI of the next page of results, if any."
        }
      }
```

One can only safely assume that this mistake was made on all of 30 other Swagger specifications provided and merged into the Concur connector.

## Expected behavior

Data shapes of the connector should converge with reality.

## Screenshot

![ezgif-4-914a9cb720](https://user-images.githubusercontent.com/1306050/42581578-5e5a18d8-852d-11e8-893c-10ca7a5525c9.gif)

heiko-braun:
@zregvart It's not obvious from the description where the differences to the current impl are. Can you provide an example if that?

heiko-braun:
//cc @KurtStam

zregvart:
> @zregvart It's not obvious from the description where the differences to the current impl are. Can you provide an example if that?

Description has JSON schema snippet as defined by the Swagger Concur publishes and what it needs to be.

To make it more plastic, this could be the JSON as defined by the published Swagger:
```json
{
  "Items": { ... }
}
```

Whereas what's actually returned is:

```json
{
  "Items": [
    { ... },
    { ... }
  ]
}
```

heiko-braun:
@KurtStam Is this something you could look into?

zregvart:
This was addressed in #3139

PR #3139, opened 2018-07-20
Fix for the concur swagger definitions #3041

KurtStam:


zregvart:
@KurtStam the `specification` property seems too big, are you building this with the latest `connector-generator` that includes the `specification` trimming from ?
The size of the `connector-concur.json` on my end is 799K, this one is 2.0M.

Issue #3060, opened 2018-07-12
Cancel modal dialogs


michael-coker:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

Looks like we need to update the layout and title, body, and button wording of these dialogs

* Cancel during in create API Connector workflow 
* Cancel during integration import

This is potentially part of a larger discussion - should these buttons even have modal dialog confirmations?

The following workflow cancel buttons do not have a confirmation dialog - they just exit the workflow

* extension import
* create integration
* create connection

And assuming any of these workflows (including the first 2 mentioned in the issue) should have cancel confirmation modal dialogs, should the dialog only appear when you click the "cancel" button (current behavior), or should the modal appear when you navigate/cancel out of the workflow some other way? (click a link in the breadcrumbs, navigation, etc)

## Screenshot
![screen shot 2018-07-12 at 3 11 55 pm](https://user-images.githubusercontent.com/35148959/42657096-e3b86dae-85e6-11e8-92b4-f0282281a2db.png)
![screen shot 2018-07-12 at 2 57 01 pm](https://user-images.githubusercontent.com/35148959/42657097-e3ce6b22-85e6-11e8-9c65-3bc4dbddcfcb.png)

thoughts? @TovaCohen @amysueg @sjcox-rh @dongniwang @seanforyou23

gashcrumb:
> It's probably worth noting that we'll have the capbability of autosave in the near future.

Right, true, though I wonder if this is maybe something to consider for 7.1 or as a stopgap until we've implemented proper saving of not finished stuff.

dongniwang:
> Right, true, though I wonder if this is maybe something to consider for 7.1 or as a stopgap until we've implemented proper saving of not finished stuff.

Well, if that's the case, then I vote for having cancel confirmation dialogs once users have actually started a flow - e.g. after user uploaded a file, put in an URL, or any other situations where new changes/data has been introduced, etc.

michael-coker:
@TovaCohen thanks! Updated the API Connector cancel dialog, and the integration import cancel button was removed (https://github.com/syndesisio/syndesis/issues/2587), so we're good there.

[...]

PR #3143, opened 2018-07-20
add cancel modal dialog to integration-basics step in integration editor

michael-coker:
### Don't merge this please!

@gashcrumb just taking a stab at adding a modal by using the API client connector as a reference. Seems to work :) This is re: https://github.com/syndesisio/syndesis/issues/3060

### DIFF
--- a/app/ui/src/app/integration/edit-page/integration-basics/integration-basics.component.html
+++ b/app/ui/src/app/integration/edit-page/integration-basics/integration-basics.component.html
@@ -23,7 +23,7 @@
             </div>
             <div class="toolbar-pf-action-right">
               <button class="btn btn-default"
-                      (click)="cancel()">Cancel
+                      (click)="showCancelModal()">Cancel
               </button> &nbsp;
               <button class="btn btn-default"
                       (click)="save()"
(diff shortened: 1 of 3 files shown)


michael-coker:
@gashcrumb cool, I looked at the API Connector guard before and wasn't sure if it was something I could do, but I'll take another stab at it :) thanks!

michael-coker:
@gashcrumb just an aside, but would a better way to trigger the modal when you click the cancel button when creating a connection to be to remove the [cancel target](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/connections/create-page/configure-fields/configure-fields.component.ts#L78) from the guard?

gashcrumb:
yeah, think so.

gashcrumb:
Sorry just double-checking, is this good to go?  On second look I think it's fine but now I read the old conversation and I can't tell if there's remaining items :-)

michael-coker:
@gashcrumb yo! This only introduces a cancel confirmation dialog on the "connection-basics" step. My intent was just to see if that's the correct way to implement it, and if so, I was going to introduce it on the rest of the editor steps with a cancel button. However, since we were going to introduce the route guard in the integration editor, we decided not to integrate this on each step, and to use the route guard to trigger a modal instead. And since the route guard work was shelved in lieu of auto-save, this was never implemented.

I don't think this should be merged since it only adds a confirmation dialog to one step. I suppose our options are to either 1) Pick this PR back up and implement the cancel confirmation dialog in the other editor steps, or 2) close it for now and revisit after we implement autosave. 

FWIW, per our discussions about autosave, I _think_ we discussed not having a route guard or cancel modal dialog at all once autosave is enabled. I brought up still having the route guard in place, in case you accidentally navigate away from the editor workflow, but I think the group consensus was  to not have this confirmation dialog if you click "cancel," as that's an intentional click and you won't lose any work, so no need to introduce another thing for people to click on.

What do you think?

gashcrumb:
Ah of course, I remember now!

Cool, let's revisit when we implement autosave/route guards.  But yeah, for a single page you're implementation looks good to me :-)

PR #3146, opened 2018-07-22
Enforce the minimum version of oc

phantomjinx:
* Since 3.9.0 version of oc is required, check the version during the setup
  of oc stage of running syndesis commands.

### DIFF
--- a/tools/bin/commands/util/openshift_funcs
+++ b/tools/bin/commands/util/openshift_funcs
@@ -1,5 +1,80 @@
 #!/bin/bash
 
+OC_MIN_VERSION=3.9.0
+
+compare_version_part() {
+    local test=$1
+    local min=$2
+
+    test=`expr $test`
+    min=`expr $min`
+
+    if [ $test -eq $min ]; then
+        echo 0;
+    elif [ $test -gt $min ]; then
+        echo 1;
+    else
+        # $test -lt $min
+        echo -1
+    fi
+}
+
+compare_oc_version() {
+    local test=$1
+    local min=$2
+
+    echo -n "Testing oc version '$test' against required minimum '$min' ... "
+
+    testparts=( ${test//./ } )
+    minparts=( ${min//./ } )
+
+    local i=0
+    while [ $i -lt ${#minparts[@]} ]
+    do
+        local testpart=${testparts[$i]}
+        local minpart=${minparts[$i]}
+
+        if [ -z "$testpart" ]; then
+            # test version does not extend as far as minimum
+            # in parts so append a 0
+            testpart=0
+        fi
+
+        ret=$(compare_version_part $testpart $minpart)
+        if [ $ret == -1 ]; then
+            #
+            # version part is less than minimum while all preceding
+            # parts were equal so version does not meet minimum
+            #
+            echo "ERROR: oc version ($test) should be at least $min"
+            return
+        elif [ $ret == 1 ]; then
+            #
+            # version part is greater than minimum so no need to test
+            # any further parts as version is greater than minimum
+            #
+            echo "OK"
+            return
+        fi
+
+        #
+        # Only if the version part is equal will the loop continue
+        # with further parts.
+        #
+        i=`expr $i + 1`
+    done
+
+    echo "OK"
+}
+
+check_oc_version()
+{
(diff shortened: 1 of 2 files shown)


tools/bin/commands/util/openshift_funcs
```diff
@@ -1,11 +1,66 @@
...
+
+compare_version_part() {
+    local test=$1
+    local min=$2
+
+    test=$(($test + 0))
+    min=$(($min + 0))
+
+    if [ $test -eq $min ]; then
+        return 0;
```
rhuss: Actually we are using a different semantics for returning values from functions, which can be used also for string return values.

It works with eval and echo:

```
a() {
   ...
   echo "myretval"
}

x=$(a)
```

Also we have golang like error handling:

```
a() {
    ...
    echo "ERROR: Can't do this"
}

x=$(a)
check_error $x
```

See also other spots in the scripts how this is used.
It would be awesome if we could stick to this convention, also for pure numeric return values.

tools/bin/commands/util/openshift_funcs
```diff
@@ -1,11 +1,66 @@
 #!/bin/bash
 
+OC_MIN_VERSION=3.9.0
+
+compare_version_part() {
+    local test=$1
+    local min=$2
+
+    test=$(($test + 0))
+    min=$(($min + 0))
```
rhuss: If we could avoid bashism, that would be awesome. Actually we try to avoid shell specific constructs as much ass possible.

tools/bin/commands/util/openshift_funcs
```diff
@@ -1,11 +1,66 @@
...
+        echo "Error: oc version should be at least $min"
+        exit 1
+    fi
+
+    ret=$(compare_version_part ${testparts[2]} ${minparts[2]})
+    ret=$?
+    if [ $ret -gt 1 ]; then
+        echo "Error: oc version should be at least $min"
+        exit 1
+    fi
```
rhuss: Coudld we make this into a loop please ?

```bash
for i in 0 1 2
do
    ret="$(compare_version_part ${testparts[$i]} ${minparts[$i]})"
    if [ $ret -gt 1 ]; then
        echo "ERROR: oc version should be at least $min"
        return
    fi  
done
```

Don't exit from within a helper function. Instead propagate the error up and the let the calling method decided (therefore that 'ERROR:' prefix)

tools/bin/commands/util/openshift_funcs
```diff
@@ -1,11 +1,66 @@
...
+
+    echo "OK"
+    set -e
```
rhuss: If we are using the eval convention for return valus, there is no need to switch on/off -e

tools/bin/commands/util/openshift_funcs
```diff
@@ -1,11 +1,66 @@
...
+    elif [ $test -gt $min ]; then
+        return 1;
+    else
+        return 2;
+    fi
+}
+
+check_oc_version()
+{
+    set +e
```
rhuss: Please don't do this over the scope of a whole function, only in rare circumstances.

tools/bin/commands/util/openshift_funcs
```diff
@@ -1,11 +1,66 @@
...
+    set -e
+}
+
 setup_oc() {
 
     # Check path first if it already exists
     set +e
     which oc &>/dev/null
     if [ $? -eq 0 ]; then
+      check_oc_version
```
rhuss:
```bash
err=$(check_oc_version)
check_error $err
```

phantomjinx:
Changes made ... WDYT?

tools/bin/commands/util/openshift_funcs
```diff
@@ -1,12 +1,86 @@
...
+            echo "OK"
+            return
+        fi
+
+        #
+        # Only if the version part is equal will the loop continue
+        # with further parts.
+        #
+        i=`expr $i + 1`
+    done
```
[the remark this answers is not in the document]
phantomjinx: Done

phantomjinx:
PR for test suite create [here](https://github.com/syndesisio/syndesis/issues/3161).

Issue #2706, opened 2018-05-24
SQL Connector should not parse body when no parameters are set


nicolaferraro:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

When inserting data into a DB, if the insert query has no parameters, the SQL connector should not parse the body and not expect it to be present or to be JSON data.

I've a integration from webhook to SQL with query: `INSERT INTO TODO (TASK) VALUES (1)`.
On invocation I get:
```
java.lang.IllegalArgumentException: Unable to parse given JSON
	at io.syndesis.connector.sql.common.JSONBeanUtil.parseSqlParametersFromJSONBean(JSONBeanUtil.java:80)
	at io.syndesis.connector.sql.customizer.SqlConnectorCustomizer.doBeforeProducer(SqlConnectorCustomizer.java:56)
	at org.apache.camel.util.AsyncProcessorConverterHelper$ProcessorToAsyncProcessorBridge.process(AsyncProcessorConverterHelper.java:61)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:138)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:101)
	at io.syndesis.integration.component.proxy.ComponentProxyProducer.process(ComponentProxyProducer.java:44)
	at org.apache.camel.processor.SendProcessor.process(SendProcessor.java:148)
	at org.apache.camel.processor.DelegateAsyncProcessor.process(DelegateAsyncProcessor.java:110)
	at org.apache.camel.processor.RedeliveryErrorHandler.process(RedeliveryErrorHandler.java:548)
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:138)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:101)
	at org.apache.camel.processor.DelegateAsyncProcessor.process(DelegateAsyncProcessor.java:110)
	at io.syndesis.integration.runtime.logging.ActivityTrackingInterceptStrategy$EventProcessor.process(ActivityTrackingInterceptStrategy.java:77)
	at org.apache.camel.processor.DelegateAsyncProcessor.process(DelegateAsyncProcessor.java:110)
[...]

PR #3153, opened 2018-07-23
Issue #2706 - sql connector should not parse body when no parameters

phantomjinx:
* JSONBeanUtil
 * Checks the json parameter and returns if empty
 * Check the jdbcTypeMap and assigns VARCHAR if null

* JSONBeanUtilTest
 * Test to check what happens if parameters are null

### DIFF
--- a/app/connector/sql/src/main/java/io/syndesis/connector/sql/common/JSONBeanUtil.java
+++ b/app/connector/sql/src/main/java/io/syndesis/connector/sql/common/JSONBeanUtil.java
@@ -17,6 +17,7 @@ package io.syndesis.connector.sql.common;
 
 import java.io.IOException;
 import java.sql.Types;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


app/connector/sql/src/main/java/io/syndesis/connector/sql/common/JSONBeanUtil.java
```diff
@@ -73,16 +73,30 @@ public static Properties parsePropertiesFromJSONBean(final String json) {
     }
 
     public static Map<String,SqlParameterValue> parseSqlParametersFromJSONBean(final String json, final Map<String, Integer> jdbcTypeMap) {
+        final Map<String,SqlParameterValue> ret = new HashMap<>();
+
+        if (json == null || json.length() == 0) {
```
zregvart: nitpick: json == null || json.isEmpty()

app/connector/sql/src/main/java/io/syndesis/connector/sql/common/JSONBeanUtil.java
```diff
@@ -73,16 +73,30 @@ public static Properties parsePropertiesFromJSONBean(final String json) {
...
         } catch (final IOException e) {
             throw new IllegalArgumentException("Unable to parse given JSON", e);
         }
-        final Map<String,SqlParameterValue> ret = new HashMap<>();
         for (String key : parsed.keySet()) {
             Object value = parsed.get(key);
-            Integer jdbcType = (jdbcType = jdbcTypeMap.get(key)) != null ? jdbcType : Types.VARCHAR;
+
+            Integer jdbcType = null;
+            if (jdbcTypeMap != null) {
```
zregvart: looking at:

https://github.com/syndesisio/syndesis/blob/93d1784d2de1ee354e651325bace008cb27d0ba3/app/connector/sql/src/main/java/io/syndesis/connector/sql/customizer/SqlConnectorCustomizer.java#L82-L98

it doesn't seem that `jdbcTypeMap` can be `null`

zregvart:
@phantomjinx perhaps in the future you could use conventional commits and reference the issue from the commit message body.

You can see it in this commit e06646e07c32a0d0649a8198dea5be74c9445d30

See https://syndesis.io/community/engineering_guidelines/#commit-messages-should-follow-the-conventional-commits-standardized-commit-message-format

You can also install pre-commit hooks https://pre-commit.com/#install that have some of these checks there. You need to install it to git repository with `pre-commit install`

app/connector/sql/src/main/java/io/syndesis/connector/sql/common/JSONBeanUtil.java
phantomjinx: @zregvart 
You are right and did consider this. However, upon writing the unit test and the static nature of the function, it seemed prudent that since I was handling the possibility of 1 parameter being null, I might as well check the other. Probably unnecessary but at least consistent.

phantomjinx:
@zregvart 
Fixed the changes and removed the label so will merge when the checks have completed.
Thx!

PR #3156, opened 2018-07-23
chore: Update to AtlasMap 1.35.5 in Integration BOM

zregvart:
We need to update AtlasMap to 1.35.5 in Integration BOM as well,
otherwise our integrations run with older version.

cc @igarashitm

### DIFF
--- a/app/integration/bom/pom.xml
+++ b/app/integration/bom/pom.xml
@@ -29,7 +29,7 @@
   <properties>
     <spring-boot.version>1.5.13.RELEASE</spring-boot.version>
     <camel.version>2.21.0.fuse-000094</camel.version>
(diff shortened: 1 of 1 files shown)


zregvart:
> hi guys, which bug is this PR fixing?

None, I didn't think we need a bug for bumping versions.

paoloantinori:
normally we don't need it. 

but since we are in code freeze, it means that only bug related commits can get in.

zregvart:
@paoloantinori this is an additional task that we need to do for Integration BOM, a follow-on on the AtlasMap upgrade we did in #3127.

paoloantinori:
thanks, the correlation with that activity was what was needed.

Issue #3096, opened 2018-07-17
API connector - Next button does nothing after specifying a swagger file


mastepan:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

API Client Connector - After specifying a swagger file to load, clicking **Next** button does nothing. I'm unable to proceed further.

## Screenshot

![apiconnnext](https://user-images.githubusercontent.com/8707251/42819367-a183012e-89d3-11e8-9085-b24835a66f42.png)

## Tasks involved / Steps to Reproduce

1.
2.
3.
4.

heiko-braun:
@mastepan What's in the browser/javascript logs? Any errors?

gashcrumb:
I think this is because there's still work ongoing with the apicurio integration, this is the line that's bombing:

![capture](https://user-images.githubusercontent.com/351660/42824695-5708b780-89ae-11e8-8ff9-79d20e8a571b.PNG)

Might be a problem actually, as to be able to edit a swagger file the UI needs the actual file, and I think the API response back is just the summary information.

@kahboom thoughts?  I think initially we had just tackled uploading a local file vs. editing a file downloaded from a URL.  Maybe we should just disable the editor for this case in the short term.

mastepan:
@gashcrumb, @kahboom it's same for a local file or a one get from an URL.

kahboom:
@mastepan - It might behoove you to put that in a gist so that I can parse that more easily (mentally 😂 ). I'll test this issue with the swagger files I have as soon as I sort out this Minishift issue. @____@

@gashcrumb you'll be getting a ping from me in the morning if none of this works out haha. I knew that summary response would eventually become an issue. Are you suggesting we disable the editor for what exactly? It sounds like it's for both file uploads and URLs. I can take a look and see if there's a potential workaround.

Also, on a similar but unrelated note, while I know we are not wanting to store OpenAPI specifications on the backend, I think we will reach a point where we will want to return the new specification, whether to export it for use elsewhere (perhaps to actually update their API), or maybe for situations like this. Another use case would be allowing us to continuously save on changes, which would sort out that issue we discussed in the F2F where users can take a lunch, come back, and lose all their changes if someone turned the laptop off, etc. I know some people may have long specifications, but I think it's worth looking into further.

cc @zregvart

zregvart:
[...]

PR #3173, opened 2018-07-24
API connector - Clicking Next button now takes you to the Review Actions page whe either file or URL is used

elvisisking:
- See #3096
- FileReader instance is now only used for when file is used
- When URL is used, there is still a problem when Review/Edit is clicked to try and get to the ApiCurioComponent. This is because the ApiDefinition.spec is not being set correctly. ApiDefinition.spec should be set by using a File or a JSON object which I'm not sure we have one at this point. This will be worked in a different issue per @kahboom.


kahboom:
Here is the other issue #3174

PR #3193, opened 2018-07-26
refactor(operator): Used meta direct as key, simplified a bit

rhuss:


rhuss:
Thats the last pr from syndesis-operator ported to here --> https://github.com/syndesisio/syndesis-operator/pull/16

PR #3198, opened 2018-07-26
Fix(install): Using oc determine try and determine the console url

### LINKED ISSUE #3076: Console URL doesn't seem to be correctly set when installed via 'syndesis install'
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem

Seems like the 'open log in openshift' link and the 'logs' link don't work properly out of the box on openshift, with a little digging we found that the `consoleUrl` setting in the UI's `config.json` file seems to be empty or can often be the wrong URL.

## Expected behavior

Ideally this URL should be set correctly when running `syndesis install`

@dhirajsb FYI

As a workaround this can be fixed by editing the `syndesis-ui-config` ConfigMap.  Also with #3072 `yarn start:minishift` will set up the dev instance with the right console URL at least via `minishift`.

phantomjinx:
* install
 * Moves the --console option reading to openshift_funcs

* openshift_funcs
 * Dedicated determine_console() function for working out the right url
 * Extract the url from oc status output
 * If url does not already have '/console' on the end, append it
 * Handle if an url cannot be extract and has not been specified

Fixes #3076

### DIFF
--- a/tools/bin/commands/util/openshift_funcs
+++ b/tools/bin/commands/util/openshift_funcs
@@ -104,6 +104,31 @@ setup_oc() {
     exit 1
 }
 
+determine_console() {
+    local console=$(readopt --console)
+
+    if [ -z "${console}" ]; then
+        #
+        # 1. Show the status of oc
+        # 2. Cut the output to only the first line
+        # 3. Reduce the line to only the https ... url
+        #
+        # If logged out then console should be empty
+        #
+        console=$(oc status | head -n 1 | sed s/.*https/https/)
+    fi
+
+    if [[ ! ${console} =~ .*\/console ]]; then
+        #
+        # Ensure that console is appended on the end of the url
+        #
+        console="${console}/console"
+    fi
+
+    # user specified console url so prefer that
+    echo ${console}
+}
+
 recreate_project() {
     local project=$1
     local dont_ask=${2:-false}
@@ -163,7 +188,6 @@ create_oauthclient() {
 
 create_and_apply_template() {
     local route=$1
-    local console=${2:-}
     local template=${3:-syndesis}
     local tag="$(readopt --tag)"
     local use_local_resource="$(hasflag --local)"
@@ -171,6 +195,13 @@ create_and_apply_template() {
 
     if [ -z "$route" ]; then
         echo "No route given"
+        exit 1
+    fi
+
+    # Determine the url of the console
+    local console=$(determine_console)
+    if [ -x "${console}" ]; then
+        echo "No openshift console url could be determined. Try specifying it using --console <console>."
         exit 1
     fi
 
--- a/tools/bin/commands/install
+++ b/tools/bin/commands/install
@@ -44,8 +44,6 @@ install::run() {
         check_error $route
     fi
 
-    local console=$(readopt --console)
-
     # Setup oc
     setup_oc
 
@@ -53,7 +51,7 @@ install::run() {
     create_oauthclient "$(readopt --tag)" "$(hasflag --local)"
 
     # Apply a template, based on the given arguments
-    create_and_apply_template "$route" "$console" "$(select_template $(hasflag --dev))"
+    create_and_apply_template "$route" "$(select_template $(hasflag --dev))"
 
     if [ $(hasflag --watch -w) ] || [ $(hasflag --dev) ] || [ $(hasflag --open -o) ]; then
         wait_for_deployments 1 syndesis-server syndesis-ui syndesis-meta


tools/bin/commands/util/openshift_funcs
```diff
@@ -104,6 +104,31 @@ setup_oc() {
...
+        # If logged out then console should be empty
+        #
+        console=$(oc status | head -n 1 | sed s/.*https/https/)
+    fi
+
+    if [[ ! ${console} =~ .*\/console ]]; then
+        #
+        # Ensure that console is appended on the end of the url
+        #
+        console="${console}/console"
```
rhuss: Shouldn't we always take over the console URL literally when provided from the outside ?

tools/bin/commands/util/openshift_funcs
```diff
@@ -104,6 +104,31 @@ setup_oc() {
...
+        # 1. Show the status of oc
+        # 2. Cut the output to only the first line
+        # 3. Reduce the line to only the https ... url
+        #
+        # If logged out then console should be empty
+        #
+        console=$(oc status | head -n 1 | sed s/.*https/https/)
+    fi
+
+    if [[ ! ${console} =~ .*\/console ]]; then
```
rhuss: Please try to avoid bash constructs like `[[ ... ]]`. You can test this easily with parameter substitutions, too (something similiar like `${console//\/console$} != ${console}` or so ...

tools/bin/commands/util/openshift_funcs
phantomjinx: Probably but then I never trust users to get it right first time ;-)
So checking for the console suffix seemed prudent.

tools/bin/commands/util/openshift_funcs
phantomjinx: I was going to but checked the rest of the script and noted that L234 uses the same kind of syntax hence thought this would suffice. Would you like L234 changed too?

phantomjinx:
@rhuss 
Well that's a PITA!!

That implies that in the case of Openshift Online etc... users require the --console switch. This does override so is the preferred option in that context.

Any way to detect the type of cluster automatically? Otherwise, it comes back to documentation I would have thought.
This is still a valid approach from a single user / developmental use-case so would this still fly as an 80/20 solution?

heiko-braun:
@rhuss 

> Actually, we already tried hard to find out this URL in a generic fashion, but it is somewhere burried in the configuration of each cluster, which is not exposed over the API (at least not that we know this).

Time for a change request in openshift upstream?

tools/bin/commands/util/openshift_funcs
rhuss: yeah, sorry. Sometimes I get a bit too dogmatic ;-) no problem, just wanted to mentionedn there are easier ways like e.g `[ ${console#/console} != $console ]` (the first expr removes the a trailing "/console"). That is POSIX compliant afaik.

sorr for the nit pick.

rhuss:
> That implies that in the case of Openshift Online etc... users require the --console switch. This does override so is the preferred option in that context.
> 
> Any way to detect the type of cluster automatically? Otherwise, it comes back to documentation I would have thought.
> This is still a valid approach from a single user / developmental use-case so would this still fly as an 80/20 solution?

Actually, if we cannot for sure detect the URL (and we can't) we should not go for a default. In the case of a missing --console-url Syndesis is still working, just with not showing a link to the console. We agreed some time ago on this behaviour as this is supposed to be better than having a link which leads to nirvana.

rhuss:
@heiko-braun good idea, but not sure that this a very commmon use case.

Nevertheless, here we go --> https://github.com/openshift/origin/issues/20447

phantomjinx:
@rhuss 
<blockquote> We agreed some time ago on this behaviour as this is supposed to be better than having a link which leads to nirvana.</blockquote>

Should point out that during my testing with this commit, hence the correct console url specified, Syndesis provides the link to the openshift log during the build phase. Sure enough I can see the log.
However, when I clicked on it during the following phases:
1) assembling: The link went to no where since there was nothing yet for it to point to;
2) deploying: The link went to no where since the deploy pod went by so quickly that the link was almost instantly invalid.

rhuss:
@phantomjinx thanks for the feedback, and yes. the link can be invalid depending on the state of pods and deployment. However as discussed in length with @kcbabo and @chirino this was acceptable (i.e. that this link is invalid from time to time).

phantomjinx:
@rhuss 

Due to the operator script changes it no longer merges so would have to be rebased. What would you recommend to do with this now?

rhuss:
Let me revisit when 1.4.2 has been released and we branched off a 1.4.x branch. Sorry, busy times ...

rhuss:
Sorry, I don't think we can reliably do the kind of automatism for every cluster. There is some ongoing discussion to make the console API discoverable. I think we have to wait for this to make it truely automatically available.

So I'm going to close this PR for now (thanks nevertheless !), we can reopen it if required.

PR #3206, opened 2018-07-27
feat: mark Connectors provided as tech preview

### LINKED ISSUE #3201: Visual Distinction between Preview Connector types
## This is a...

<pre><code>
[ X] Feature request
[ ] Bug report  
</code></pre>

## The problem

There are essentially two types of connectors:

**Full Support connectors**
- have user stories
- are tested
- are documented
- are fully supported by Red Hat

**Preview connectors** 
- do not require user stories
- may or may not be tested
- have upstream documentation at best
- are not supported by Red Hat, but we do take feedback from customers and plan to include in product at some point

Need to visually differentiate the two groups. Preview connectors would be labeled "Preview"

## Expected behavior

Preview connectors would be indicated as being a "Preview". Determine best visual indicator.

## Screenshot

screen shot 2018-07-26 at 1 18 32 pm

1. Determine visual Design
2. Review
3. Implement

zregvart:
Adds a new tag `tech-preview` that can be added to connectors that are provided for early access technology preview.

First such Connector is the IRC Connector.

Visually when Connection is displayed the top of the Connection card will contain `Technology preview` text, which is also displayed during connection creation.

Fixes #3201

### DIFF
--- a/app/connector/irc/src/main/resources/META-INF/syndesis/connector/irc.json
+++ b/app/connector/irc/src/main/resources/META-INF/syndesis/connector/irc.json
@@ -9,9 +9,9 @@
            "id": "@project.groupId@:@project.artifactId@:@project.version@"
       }
   ],
-  "tags": [
-  	"tech-preview"
-  ],
+  "metadata": {
+  	"tech-preview": true
+  },
     "actions": [
         {
             "name": "IRC PRIVMSG",
--- a/app/connector/irc/src/main/resources/META-INF/syndesis/connector/irc.json
+++ b/app/connector/irc/src/main/resources/META-INF/syndesis/connector/irc.json
@@ -9,6 +9,9 @@
            "id": "@project.groupId@:@project.artifactId@:@project.version@"
       }
   ],
+  "tags": [
+  	"tech-preview"
+  ],
     "actions": [
         {
             "name": "IRC PRIVMSG",
--- a/app/ui/src/app/common/card-tech-preview.component.scss
+++ b/app/ui/src/app/common/card-tech-preview.component.scss
@@ -1,0 +1,26 @@
+@import 'syndesis-sass';
+
+.syn-card-info {
+  text-align: right;
+
+  &:after {
+    /* taken verbatim from patternfly/_icons.scss */
+    content: $pficon-var-info;
+    display: inline-block;
+    font-family: "#{$icon-font-name-pf}";
+    font-style: normal;
+    font-variant: normal;
+    font-weight: normal;
+    line-height: 1;
+    speak: none;
+    text-transform: none;
+    /* Better Font Rendering =========== */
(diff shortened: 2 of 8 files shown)


gashcrumb:
Looks awesome!  I personally would have suggested not using tags for this but instead setting a metadata property, but this works :-)

>    is the loading from store needed -- seems that Connections have connector property set

I don't think so.  When we get connections back the connector is a property that's always present.

The only other page where we want to be sure this shows up is the first page of the create connection wizard, the same list/card view is used, looks like you've handled that already though :-)

>    breaking encapsulation -- I found no other way of styling card-pf-body, otherwise it would be slightly higher then other cards for tech preview connectors and that would look really ugly

I'd ask @michael-coker but yeah, normally if you need to affect a child element that's nested in some sub-component you can use `::ng-deep` these days, for example -> https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/connections/list/list.component.scss#L5

zregvart:
> I'd ask @michael-coker but yeah, normally if you need to affect a child element that's nested in some sub-component you can use ::ng-deep these days, for example -> /app/ui/src/app/connections/list/list.component.scss@master#L5

Yeah, this case is a bit more complicated than that, we need to change style of a sibling of the parent of the host component in case it is in tech preview. Took a while to get that working, would appreciate a cleaner solution if any. Perhaps with different markup to start with, not sure.

zregvart:
> > is the loading from store needed -- seems that Connections have connector property set
> 
> I don't think so. When we get connections back the connector is a property that's always present.

Let me remove those parts before we merge and keep it as simple as it can be.

zregvart:
> Looks awesome! I personally would have suggested not using tags for this but instead setting a metadata property, but this works :-)

Oh, and I kinda like tags. Having this on a small percentage of Connectors kinda eliminates the need for a property (IMHO).

That said we can add another property if that would make it nicer?

gashcrumb:
>Oh, and I kinda like tags. Having this on a small percentage of Connectors kinda eliminates the need for a property (IMHO).

>That said we can add another property if that would make it nicer?

I forget if connectors can have metadata like integrations can or not, but this feels more like an item that'd be a metadata thing.  Tags work but they're really for grouping and can be user-edited, at least when we exposed them.  If we decide to re-expose them at some point, then using tags to also toggle this tech-preview indicator I think won't be ideal.

gashcrumb:
> Do we need to revisit the tag usage for properties (OAuth) and actions (marked for meta, validation)?

I've no idea, the UI AFAIK doesn't use tags for these types for anything.

I only mention using metadata as I recall from @lburgazzoli it's a map and it's contents are pretty open so seems a good spot for this kind of arbitrary data.

TBH I think tags/name/description should all eventually wind up in the metadata map too for all types, but you know, later :-)

Issue #2888, opened 2018-06-25
Setup and integrate "fuse-online-operator" into regular build


rhuss:

rhuss:
@nicolaferraro if you have some cycles to spend and since you already played around with the Operator SDK, would it be possible for you to help in adding an "fuse-online-operator" skeleton to our mono-repo ?

I would take over the part to include the build into the "syndesis" script.

nicolaferraro:
I've some work to do on Webhooks, but it's definitely something I can start to do right after (this week).

rhuss:
This integration has not been started but is crucial for a coherent productisation process.
For Sprint 32 this should be one of the highest priority and should be started asap.

The idea is to introduce https://github.com/syndesisio/syndesis-operator into https://github.com/syndesisio/syndesis as a submodule, so that the operator and syndesis can be tagged together so that we also publish a new operator version per release.

This story also requires to update `syndesis` to include the building and releasing the operator.

PR #3213, opened 2018-07-27
feature(operator): Initial build and install support for syndesis-operator

rhuss:
This PR will change the way how "syndesis" works, and it is expected,
as it is not a small change, that there might be some kinks.

Not yet finished, but good enough to give it a try.


rhuss:
Fixes https://github.com/syndesisio/syndesis/issues/2888

rhuss:
I think I'm ready for a review and possible merge. Its quite a mouthful, but I had to completely adapt the way how we are performing the build with `syndesis`

I updated the following parts

* `syndesis build` creates the operator (`-m operator`) as well as the Docker images. Build can be done locally with a local go installation (faster) or via a Docker image (more robust). For a local build to work you have to checkout syndesis at $GOPATH/src/github.com/syndesisio/syndesis. The image build either works via a Docker daemon (`--docker`, required by the CI) or a S2I Dockerfile build (useful to autmatically update the operator in minishift like the other images). Also the Docker build for the Java Docker images has been deprecated and the options have been slightly changed

* `syndesis minishift` now does the following (instead of installing the oauthclient and applying the template):
  - Switch to system:admin to register the Syndesis CRD an create a role with some missing perms for `developer` user. Add a rolebdinding to `developer` for this role
  - As `developer`: Install `syndesis-operator` as a regular DeploymentConfig
  - Create a Syndesis resource on the fly reflecting the parameter given

* For `syndesis install` the same is performed, except that there is no automatic change to admin as this is not possible in general. So you can use multiple steps:
  `syndesis install --setup-crd` (as admin)
  `syndesis install --fix-permission <user>` add permissions to user (to be run as admin)
  `syndesis install` as normal user to install the operator and the custom resource

* The operator now reads the template from the filesystem. The proper template is packaged in to the operator's Docker image.

* `syndesis release` has been updated to release the operator with the proper template

* CircleCI job now also builds the Operator Images and pushes it to Docker Hub

-----------

There are for sure still some kinks, i.e. the update needs to be tested.

nicolaferraro:
Looks very good :smile:

rhuss:
> Let's merge and fix any issues that come up, if they come up.

well, seems that this didn't really worked out.

It's all my fault that I haven't mailed the notes from https://github.com/syndesisio/syndesis/pull/3213#issuecomment-408915960 and haven't updated the user manual yet.

PR #3220, opened 2018-07-29
bug fix: Replaced ignite with fuse_online in user doc URLs

TovaCohen:
@zregvart - Thanks for remembering to update the version number in the doc URLs in the dictionary file. It reminded me to replace "ignite" with "fuse_online" in the 7.1 URLs. In the future, I will make sure to check the dictionary file for every release. 

While I was in there, I removed the instances of "please", except for one, since Red Hat style says this is not appropriate in UIs.I also removed some unnecessary words and  updated the description of an extension.

### DIFF
--- a/app/ui/src/assets/dictionary/en-GB.json
+++ b/app/ui/src/assets/dictionary/en-GB.json
@@ -4,7 +4,7 @@
     "customizations": {
       "extensions": {
         "heading": "{{extensions}}",
-        "description": "An extension defines one or more custom steps for use in integrations.  Find out more at <a href=\"{{links.userguide}}#adding-extensions\" target=\"_blank\">{{shared.project.name}} Help</a>",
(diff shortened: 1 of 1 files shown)


fbolton:
@zregvart - sounds like a good idea to use a variable for the latest version, instead of 7.1. But I was not involved in the creation of this `.json` file format, so I do not know how to do that here.

zregvart:
@fbolton @TovaCohen with #3222 merged we would have a single version in the JSON file to maintain.

@fbolton I was thinking if the published site could contain `latest` in the URL so we wouldn't need to maintain a version here. Come to think of it that might make it easier for us but difficult for users as prior versions would point to latest docs.

fbolton:
@zregvart, oh, I see. Yes, that would be cool. Unfortunately, it's currently not supported on the customer portal.

fbolton:
And thanks very much for merging #3222 @zregvart !

PR #3222, opened 2018-07-30
refactor(ui): variable for project version

### LINKED ISSUE #3221: Introduce variable for referencing the version of Fuse Online
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## The problem

As discovered in #3220 we could introduce a single variable holding the version of Fuse Online. This can be helpful so we keep our versions up to date and don't repeat the version string in multiple places.

Perhaps we could also use this on the about screen?

## Expected behavior

Have a single variable holding the productized version.

zregvart:
Introduces a single variable to hold the project version in order to
make maintenance easier.

Fixes #3221

### DIFF
--- a/app/ui/src/assets/dictionary/en-GB.json
+++ b/app/ui/src/assets/dictionary/en-GB.json
@@ -120,10 +120,10 @@
     },
     "links": {
       "contactus": "//access.redhat.com/support/cases/#/case/new",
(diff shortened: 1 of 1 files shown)


TovaCohen:
Thanks for making the version number a variable Zoran!
I'll make sure it gets updated for each release.

PR #3229, opened 2018-07-30
Integration edit guard

michael-coker:
fix https://github.com/syndesisio/syndesis/issues/3060

Some things I couldn't figure out

* Clicking "save as draft" from save-or-add-step when editing an integration takes you back to integration details.
* Clicking "cancel" from save-or-add-step when editing gives 2 modals.
* I'm not sure what routes to test for from the integration-basics step. It takes me back to /integrations/[unique string for the integration]. If I use that value, that would disable the route guard from that link in the breadcrumbs.
* If you trigger the route guard from the "connection-select" step, the wrong modal comes up.
* Tova would like different wording for the "cancel" modal for the integration create and edit paths.

### DIFF
--- a/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
+++ b/app/ui/src/app/integration/edit-page/action-configure/action-configure.component.html
@@ -104,8 +104,8 @@
           </div>
         </syndesis-loading>
       </div>
-      <syndesis-modal title="{{ 'connections.leave-page-title' | synI18n }}"
-        message="{{ 'connections.leave-page-message' | synI18n }}">
+      <syndesis-modal title="{{ 'integrations.leave-page-modal-title' | synI18n }}"
+        message="{{ 'integrations.leave-page-modal' | synI18n }}">
       </syndesis-modal>
     </div>
   </div>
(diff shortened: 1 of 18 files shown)


michael-coker:
@gashcrumb just checking in since we're nearing the end of the sprint. Let me know have any questions or want me to do any more work with this one.

gashcrumb:
Yeah, lemme give this a test locally...

zregvart:
@gashcrumb should we merge this in the upcoming release :steam_locomotive::railway_car::railway_car:

gashcrumb:
I'd like to get this merged, but I need to look into these outstanding items.  Will try and get this tested today.

rhuss:
@gashcrumb any chance to get this in within the next hour ?

gashcrumb:
@rhuss no

rhuss:
ok, lets leave it then for next version.

gashcrumb:
@michael-coker I wonder if maybe we should defer these guards in favor of developing autosave and implementing the improved state diagram?  I feel like the implementation of those two features is going to have a huge impact on this...

michael-coker:
@gashcrumb agreed, that sounds good to me. And from the discussion this morning, we might choose not to have a guard at all.

gashcrumb:
Ok, let's close this for now then, we might still want to steal some of this later on :-)

PR #3250, opened 2018-08-01
Fixes #2611, Added check for duplicate names across integration deployments, fixed integration delete to delete all deployments

### LINKED ISSUE #2611: Syndesis lifecycle - issue with renaming of integrations and older versions
## This is a...

<pre><code>
[ ] Feature request
[ x ] Bug report  
</code></pre>

## The problem

It seems like there is an issue in current lifecycle design: 
I created an integration with name "test". I then published the integration and then renamed it to "test_updated". Then I edited the integration and deployed the newer version of this integration (this starts the pod with the name "i-test-updated"). Then I revert the integration to the older "version 1" - this will start a pod with name "i-test", the integration name stays however "test_updated".
The issue is now, when I start a new integration "test", this will recreate the running pod "i-test", which is now dedicated to the "test_updated" integration and this will break both my old and new integration, because the "i-test" pod will try to recreate endlessly.

## Expected behavior

There could be some mechanism, which would prevent from creating the integration, which starts the pod with the same name as the running one, or some other solution. 

1. Create integration, run it
2. Rename the integration 
3. Edit the integration - run the new version
4. Revert to older version 1 of the integration
5. Create a new integration, with the original name of the integration, that is already running
6. Observe in OpenShift the incorrect behavior

dhirajsb:


dhirajsb:
@gashcrumb this fix throws an exception when duplicate names are detected. The HTTP status message and code (CONFLICT) are:
```
409 Name already used by integration<i>, version <v>
```
The UI isn't handling this for some reason, so a UI fix might be needed.

dhirajsb:
Working on moving the duplicate check into a proper bean validator.

dhirajsb:
@gashcrumb UI handles bean validation, so it should work after updating to use a bean validator.

dhirajsb:
@gashcrumb ok, so it turns out, the integration details page where the name can be edited on the top left hand side, ignroes errors and assumes it worked. Although, if I refresh the page it shows the original unchanged value. 
image

dhirajsb:
@heiko-braun although it looks like a lot of files were changed, adding a new bean validator didn't change much of existing behavior. So it's a fairly low risk and permanent fix.

heiko-braun:
Sounds good @dhirajsb . Thanks for taking care of this.

tools/bin/commands/build
```diff
@@ -21,6 +21,7 @@ build::usage() {
 -e  --ensure                  If building a Go module, run 'dep ensure' before building
 -l  --local                   For Go builds: Use a locally installed go, otherwise run the Go build from a container
 -c  --clean                   Run clean builds (mvn clean)
+    --resume <m1>             Resume build from maven module (mvn -rf m1)
```
rhuss: @dhirajsb could you please add this to the documentation, too ?

dhirajsb: @rhuss will do

dhirajsb: Looks like it's been updated already.

PR #3251, opened 2018-08-01
Fixes #2694, rename conflicting integration, connections, and extensions during import

### LINKED ISSUE #2694: Import and publish of integration with same name is possible
## This is a...

<pre><code>
[ ] Feature request
[ x ] Bug report  
</code></pre>

## The problem

It is possible to import integration with the same name as has the already running integration in syndesis.
If the imported integration is published, the running pod in syndesis is rewritten with the newly published integration, both then end up in inconsistent "Published" state.

## Expected behavior

The user should not be allowed to import integration with the same name, or the one integration should be rewritten (the user should be notified about this).

## Screenshot

out

same_name_integration

1. Create integration with "name"
2. Import an integration with the same "name" - previously created and exported 
3. Publish the integration with the same name
4. Observe the build and deployment on openshift

dhirajsb:


heiko-braun:
@chirino It would be good to get a review today, because @dhirajsb will be on PTO the next to days. Same for #3250

dhirajsb:
@chirino, @heiko-braun this is a slightly risky change, but I've tested it locally and it works a treat. I would like it to be tested more thoroughly by QE to make sure there are no edge conditions I may have missed.

heiko-braun:
@rhuss Can we get a second pair of eyes? From my POV this should be included, it's relevant to #2694, but it's not my call to make.

dhirajsb:
The aesthetic of `-imported-n` is bad by design. 
It's a naming conflict that needs user attention, and is self explanatory.

heiko-braun:
@rhuss @zregvart would be good to have this in 7.1. Can you review and come to a conclusion?

rhuss:
when should I do the review ? If we can delay the release until tomorrow or the day after, I can potentially do it.

PR #3255, opened 2018-08-01
fix(server): use timeout of 35mins

### LINKED ISSUE #2598: activity view does not show recent activity after ~day of uptime
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

## The problem

The activity log seems to stop at some point and does not show any of the recent entries. I have an integration that starts with a 1s timer, and after a day of runtime, the last activity shown is May 10, 2018, 7:01:44 PM

, while the current time is    May 11, 2018 17:52:13 PM

watching the actual pod log shows that the integration is still running correctly and reports logs every seconds as expected.

## Expected behavior

activity log should show recent entries even after days of uptime

## Screenshot

Activity log view stopped at 7:01:44 PM

1.
2.
3.
4.

zregvart:
This adds a read timeout that should provoke `SocketTimeoutException` if no data was received for 35mins. This is to fix an issue whereby the client keeps trying to read from a half-open connection and never receiving any data or any notice that the server has stopped sending the data.

Fixes #2598 (perhaps)


zregvart:
So this is my third attempt of getting #2598 fixed, first two included using TCP keep-alive (didn't work), or switching to websockets (too big of an investment).
This _should_ fix if the assumptions in the test hold, that is, if the test is testing what can be observed on staging and QE environments.
This deliberately tries not to make huge changes to the way logs are collected, but targets to fix just this one issue. Further evolution of this architecture should be considered in the future, possibly including pushing of activities from the integration pod to a dedicated storage for activities, i.e. getting this out of the server pod and not using logs for activity delivery.

heiko-braun:
@rhuss Would be good to have a fix for the issue in 7.1. Can you review it and take a decision together with @zregvart on how to proceed with this?

zregvart:
This I would consider a band-aid, not a proper fix, proper fix would include re-architecting activity tracking functionality. For now, I think, it's the best we can do.

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/KubernetesSupport.java
```diff
@@ -42,10 +44,12 @@
 
     private final KubernetesClient client;
     private final OkHttpClient okHttpClient;
+    /** Read timeout for the HTTP client, we expect to receive at least one log line in 30 mins. */
+    private Duration readTimeout = Duration.ofMinutes(35);
```
rhuss: Is this 30 mins for getting a logline configured somewhere ?

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/KubernetesSupport.java
```diff
@@ -79,22 +88,35 @@ public void onFailure(Call call, IOException e) {
...
+                            if (response.code() == 200) {
+                                try (InputStream responseStream = response.body().byteStream()) {
+                                    handler.accept(responseStream);
+                                }
                             } else {
                                 LOG.info("Failure occurred while processing controller for pod: {}, http status: {}, details: {}", podName, response.code(), response.body().string());
                                 handler.accept(null);
                             }
+                        } catch (SocketTimeoutException timeout) {
+                            LOG.warn("Timed out reading the log stream");
```
rhuss: I would add the timeout to the warn message and omit the debug statement.

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/PodLogMonitor.java
```diff
@@ -46,9 +48,7 @@
 import io.syndesis.server.jsondb.JsonDBException;
 import io.syndesis.server.openshift.OpenShiftService;
 
-/**
- *
- */
+@SuppressWarnings("PMD.GodClass")
```
rhuss: ;-)

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/PodLogMonitor.java
```diff
@@ -198,7 +198,7 @@ InflightData getInflightData(String exchangeId, String logts) throws IOException
     }
 
     @SuppressWarnings({"PMD.CyclomaticComplexity", "PMD.ExcessiveMethodLength", "PMD.NPathComplexity", ""})
-    private void processLine(String line) throws IOException {
+    void processLine(String line) throws IOException {
```
rhuss: where else is processLine used ?

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/PodLogMonitor.java
```diff
@@ -116,22 +116,22 @@ public void run() {
...
+            } finally {
                 logsController.schedule(this::run, 5, TimeUnit.SECONDS);
             }
         } else {
             logsController.schedule(this::run, 5, TimeUnit.SECONDS);
         }
     }
 
-    private void processLogStream(final InputStream is) throws IOException, InterruptedException {
+    void processLogStream(final InputStream is) throws IOException, InterruptedException {
```
rhuss: where else is processLogStream used ?

app/server/logging/jsondb/src/test/java/io/syndesis/server/logging/jsondb/controller/KubernetesSupportIT.java
```diff
@@ -0,0 +1,157 @@
...
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.skife.jdbi.v2.DBI;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+public class KubernetesSupportIT {
```
rhuss: Love it ! Thanks for adding tests even in stressful times ...

app/server/logging/jsondb/src/test/java/io/syndesis/server/logging/jsondb/controller/KubernetesSupportIT.java
```diff
@@ -0,0 +1,157 @@
...
+            } catch (final InterruptedException ignored) {
+                return;
+            } finally {
+                exchange.close();
+            }
+        });
+    }
+
+    @Test
+    public void shouldTolerateNetworkTimeouts() throws IOException, InterruptedException, ExecutionException {
```
rhuss: Any feeling how much extra time this tests add to the overall run ? looks like +5 sec or so ?

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/KubernetesSupport.java
zregvart: No and it is just an assumption that every integration should perform some activity within 30min, so after 30min + 5min of grace period (all ball parked) if the client hasn't received any log lines we consider the connection to be half-open and we need to timeout and retry.

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/PodLogMonitor.java
zregvart: I think that's a leftover from the test I deleted where I needed to override the method.

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/PodLogMonitor.java
zregvart: ditto

app/server/logging/jsondb/src/test/java/io/syndesis/server/logging/jsondb/controller/KubernetesSupportIT.java
zregvart: About 30sec, it won't be run on CI as we don't have failsafe configured in this module though.

zregvart:
I really would not want to invest or waste any more time into this, let's merge it and plan on re-architecting activity logging.

zregvart:
Thanks @rhuss

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/KubernetesSupport.java
rhuss: I can imagine cron-job like integration which fire up only once a day or so for (batch processing), so while this is a good rule of thumb for now, I don't think that this assumption holds in general.

PR #3256, opened 2018-08-01
feat(server): Provides uses element to Connections API

### LINKED ISSUE #3098: Add a field to the connection object that tracks how many integrations refer to it
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## The problem

To implement preventing the user from deleting connections for #2582 the UI needs to know if the connection has references to it from any integrations.  That currently isn't tracked in the API response, and I'd rather not have to also fetch integrations also and correlate all this in the UI.

Instead, I'd rather have a `uses` field or whatever like we have for API client connectors and extensions that the UI can then use to determine if a connection can be deleted or not.

phantomjinx:
* Adds an augment mechanism to the ConnectionHandler which finds the
  integrations that utilise the connections and logs their total in a
  'uses' property. Works in the same manner as used by the Connectors.

Fixes #3098

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/EmptyListResult.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/EmptyListResult.java
@@ -1,0 +1,23 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


paoloantinori:
@zregvart guys we are in code freeze. We are still in code freeze. This seems to be related to a feature. Can you give more context? otherwise this is candidate to be reverted and closed for the time being

PR #3263, opened 2018-08-01
fix(install): use image without VOLUME for CronJob

### LINKED ISSUE #3252: Errors in db-maintenance events log
## This is a...

<pre><code>
[ ] Feature request
[x ] Bug report  
</code></pre>

## The problem
I'm running soak test on OSO Pro and I'm seeing following errors in db-maintenance pod events after couple of hours: 
`
Error: Error response from daemon: create 29a8108185585a4b7ce779cf359083624733438f85a9fc634f32966784d64d1f: error while creating volume path '/var/lib/docker/volumes/29a8108185585a4b7ce779cf359083624733438f85a9fc634f32966784d64d1f/_data': mkdir /var/lib/docker/volumes
/29a8108185585a4b7ce779cf359083624733438f85a9fc634f32966784d64d1f: permission denied
`

screen shot 2018-08-01 at 10 06 09

I believe due to this I cannot see activities after couple of hours (~200 000 total messages):
screen shot 2018-08-01 at 10 12 04

zregvart:
Fixes #3252

### DIFF
--- a/install/generator/04-syndesis-db.yml.mustache
+++ b/install/generator/04-syndesis-db.yml.mustache
@@ -111,7 +111,7 @@
                 value: ${POSTGRESQL_PASSWORD}
               - name: POSTGRESQL_DATABASE
                 value: ${POSTGRESQL_DATABASE}
-              image: centos/postgresql-95-centos7
+              image: openshift/base-centos7
               command:
                 - /bin/sh
                 - -c
--- a/install/syndesis-dev.yml
+++ b/install/syndesis-dev.yml
@@ -560,7 +560,7 @@ objects:
                 value: ${POSTGRESQL_PASSWORD}
               - name: POSTGRESQL_DATABASE
                 value: ${POSTGRESQL_DATABASE}
-              image: centos/postgresql-95-centos7
+              image: openshift/base-centos7
               command:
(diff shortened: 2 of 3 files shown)


zregvart:
Thinking of another approach, with supported (this one hasn't been updated in 2yrs) image and perhaps rsh-ing into the db pod.

heiko-braun:
awesome, good work!

rhuss:
Actually I have concerns not only with this fix but with any new image we are introducing. 
This has to do with productisation as we need to have find a productised image having the same functionality.

Also I doubt that we should add an image which has not been updated since two years.

As discussed on IRC we should avoid referencing other images as dependencies as much as possible.

Still my offer stands to do this kind of cleanup from the syndesis-operator, which gets productised anyway.

zregvart:
Seems to me that the least complicated way of having this functionality would be to tune autovacuum so we remove the CronJob entirely.

I think adding this to syndesis-operator will needlessly tie syndesis-operator to syndesis-db (and vice versa) and put additional burden on the syndesis-operator.

I'll close pull request this but I'll keep the branch open so we can revisit if autovacuum fails, as this, I think, is quick and simple way of fixing the issue.

PR #3275, opened 2018-08-02
refactor(build): when installing with --local use provide the operator with local template

zregvart:
Now when we build with `--local` operator uses the template it was built with and it is contained in the image of the operator.

This introduces a `syndesis-template` ConfigMap that contains the content of the local `syndesis.yml` OpenShift template. The `syndesis-operator` DeploymentConfig is patched to mount this template from the ConfigMap into the `/conf` directory and to use that template when creating Syndesis resources.

### DIFF
--- a/tools/bin/commands/util/openshift_funcs
+++ b/tools/bin/commands/util/openshift_funcs
@@ -104,6 +104,14 @@ deploy_syndesis_operator() {
     fi
 
     create_openshift_resource "install/operator/deploy/syndesis-operator.yml"
+    if [ $(hasflag --local) ]; then
+        # we need to include the `syndesis-template.yml` akka `syndesis.yml` in
+        # the operator image, this adds a ConfigMap that we mount into the
+        # operator container
+        oc delete configmap syndesis-template >/dev/null 2>&1
+        oc create configmap syndesis-template --from-file=template=$(appdir)/../install/syndesis.yml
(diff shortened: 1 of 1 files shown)


rhuss:
Cool, thanks ! Going to give a review soonish.

Not sure whether you mean "building" or "installing" in the title ?

rhuss:
@zregvart could we briefly discuss the use case ? When you want to include the local template in the operator it could be already done with "syndesis build -m operator --image" as this will _always_ pick up the local template to include and as the build is done via s2i and if the operator would be installed with an image trigger (not sure that it does), then you could just a build to the the local template.

But its very likely I'm missing the context.

Technically the PR looks ok. Cool path btw ;-)
Hhowever, it might be that the operator cannnot deal with the local template as they should always "fit" together. The assumption that they are always on the same version.

zregvart:
I don't think we can have developing changes to the template tied to `syndesis build`.

For example, say you want to change the template and run it on a new minishift instance or in a new namespace via `syndesis minishift --install` or `syndesis install`. In that case the operator would pick up the template from the image it was built with not the one modified locally.

Don't know if that makes sense, but it was my workflow...

rhuss:
> For example, say you want to change the template and run it on a new minishift instance or in a new namespace via syndesis minishift --install or syndesis install. In that case the operator would pick up the template from the image it was built with not the one modified locally.

I think it can work when the operator image pulled from Docker Hub and the locally modified template doesn't too much in version. As mentioned the operator itself relies on certain aspects of the template but for minor tuning it might work. The good thing is, if you find it doesn't you will know that you have to adapt the operator ;-)

So, when you change the template you do a full reinstall ?

Let me give it a quick test drive and then lets merge it ...

zregvart:
> So, when you change the template you do a full reinstall ?

Yeah, kinda feel that this makes it more stable and that everything went in... I might be foolish about this, but I feared dangling resources, like I had some experiments with removing the maintenance CronJob and without full reinstall it would remain...

rhuss:
I had no chance yet to try out (sorry) but as it enabled only with an option I guess its safe to include for now.

PR #3284, opened 2018-08-03
Hide Review or Edit btn for URL-fetched specifications

kahboom:


gashcrumb:
I like @riccardo-forina's solution in #3288

PR #3288, opened 2018-08-03
Hide the Review/edit button if the provided URL to the spec file is unfetchable

### LINKED ISSUE #3174: Apicurio Editor: URL-fetched specification not supported
## This is a...

<pre><code>
[ ] Feature request
[ x ] Bug report  
</code></pre>

## The problem
Specifications should be treated differently based on how they are obtained (e.g. URL fetch or file upload), due to the data structure. See [here](https://gist.github.com/kahboom/9f9e606df271b49973bf7337de3ed0d1) for reference. This is related to #3096.

When a URL is specified, the user can then proceed to Review/Edit from the Review Actions step (2), but the editor is blank, as if a blank specification had been provided.

## Expected behavior
User should be able to either upload a file or specify a URL for the Swagger/OpenAPI spec and see the details of it in the Review/Edit Apicurio editor.

## Screenshot
screenshot 2018-08-01 12 38 31

## Request and Response Data
See [here](https://gist.github.com/kahboom/9f9e606df271b49973bf7337de3ed0d1) for reference.

1. Navigate to Customizations.
2. Proceed as if you were creating a custom API connector, except select the URL option.
3. Test with any valid Swagger/OpenAPI URL.
4. You should see an error.

riccardo-forina:
Fixes #3174 

# Changes

## Fetchable url (same protocol and CORS enabled)

from-fetchable-url

## Unfetchable url (different protocol and/or CORS disabled)

from-non-fetchable-url-tooltip

## Todo

~Maybe it would be better to disable the Review/edit button instead of hiding it, showing the reason why it has been disabled.~ Done this, the reason for the disabled state is shown in a tooltip when hovering the button.

But I think it would be even better if we could change the backend to return the spec after the validation.

### DIFF
--- a/app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-create.component.html
+++ b/app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-create.component.html
@@ -47,7 +47,7 @@
             [apiConnectorTemplateName]="'OpenAPI'"
             [apiConnectorData]="(apiConnectorState$ | async)?.createRequest"
             [showNextButton]="true"
-            [showEditButton]="enableEditor"
+            [enableEditButton]="enableEditor"
             (reviewComplete)="onReviewComplete($event)">
           </syndesis-api-connector-review>
           <syndesis-api-connector-auth *ngIf="currentActiveStep == 3"
--- a/app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-create.component.html
+++ b/app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-create.component.html
@@ -47,6 +47,7 @@
             [apiConnectorTemplateName]="'OpenAPI'"
             [apiConnectorData]="(apiConnectorState$ | async)?.createRequest"
             [showNextButton]="true"
+            [showEditButton]="enableEditor"
             (reviewComplete)="onReviewComplete($event)">
           </syndesis-api-connector-review>
           <syndesis-api-connector-auth *ngIf="currentActiveStep == 3"
--- a/app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-create.component.ts
+++ b/app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-create.component.ts
@@ -45,6 +45,7 @@ export class ApiConnectorCreateComponent implements OnInit, OnDestroy {
   displayDefinitionEditor = false;
   editorHasChanges = false;
   validationResponse: CustomConnectorRequest;
+  enableEditor = true;
 
   @ViewChild('_apiEditor') _apiEditor: ApiEditorComponent;
   apiDef: ApiDefinition;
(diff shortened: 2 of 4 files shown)


EricWittmann:
This is a really nice compromise.  Good work. :)

dongniwang:
@riccardo-forina - nice work! 

Did you already ping @TovaCohen about the content we provided in the tooltip? If I remember correctly, it's suggested that we don't use "please" in sentences.

riccardo-forina:
Ah, no sorry, I didn't know about that, Paolo and I wrote something down very quickly to avoid a lorem ipsum :P

@TovaCohen this is the content we placed in the tooltip:

> The platform has not been able to download the linked specification. Please try manually downloading it an providing the downloaded file to enable the review/edit feature.

Could you please provide a better one?

TovaCohen:
Hi @riccardo-forina  - In your demos above, you enter a URL for an OpenAPI spec, right? So Syndesis is able to see and read that because the user gets the Review Actions page. Are you saying that while Syndesis can read the spec at that URL and can create an API connector (assuming the spec is valid), Syndesis cannot obtain a file that contains the spec? If that's the situation, how about this for the tooltip:

Unable to download a file that contains the OpenAPI specification. Try manually downloading the specification, save a file that contains the specification, and create the API connector by uploading this file. 

Is there a variable for the product name? If so, it you be great to use it:

Syndesis/Fuse Online is unable to download a file that contains the OpenAPI specification. Try ...

riccardo-forina:
> In your demos above, you enter a URL for an OpenAPI spec, right? 
Right

> So Syndesis is able to see and read that because the user gets the Review Actions page. 
Exactly

> Are you saying that while Syndesis can read the spec at that URL and can create an API connector (assuming the spec is valid), Syndesis cannot obtain a file that contains the spec?

Ah, no. It's a bit more subtle. While Syndesis can validate the provided OpenAPI spec URL because it's something that happens on a server done by Syndesis's backend, the same URL _could_ not be usable for Reviewing/editing purposes because of some security restriction imposed to Syndesis UI by the browser itself.

So yes, we can read the spec file from an URL, but sometimes we can't use the same URL to feed the editor.

> Unable to download a file that contains the OpenAPI specification. Try manually downloading the specification, save a file that contains the specification, and create the API connector by uploading this file.

I think this message would be great in general in case of wrong/unreachable URLs, but not for the case described above.

Did this help?

TovaCohen:
That definitely helps!
How's this:  (I am assuming that there is a variable for the product name.)

Syndesis/Fuse Online cannot make this specification available for review/edit. Something in the specification that prevents this. If you want to review or edit the API before Syndesis/Fuse Online creates the connector, try manually downloading the specification and then upload that file.

riccardo-forina:
What if we remove the hint about something wrong in the specification? Like this:

> Syndesis/Fuse Online cannot make this specification available for review/edit. If you want to review or edit the API before Syndesis/Fuse Online creates the connector, try manually downloading the specification and then upload that file.

TovaCohen:
Sure!

I put that in because if it is something that is in the specification that prevents making it available for review/edit, then it is nice of us to let the user know in case the user wants to change something to make it possible in the future.   Up to you. Fine to leave that sentence out.

riccardo-forina:
The kind of problems that can block us from enabling the edit/review button will almost surely not be about the specification file itself, but about _where_ the specification file is hosted. 

Maybe we can rephrase `Something in the specification prevents this.` to refer to the kind of issues that could cause the problem. From the top of my mind, I can think of 2 major reasons:
- URL is not https:// (http:// URLs will not work)
- the server hosting the specification file doesn't have CORS enabled

The reasons are fairly technical I'm afraid :(

TovaCohen:
Let's leave the sentence out. And I'll add something to the documentation to indicate why Syndesis might not be able to make an API available for review/edit. Okay? I'll ask you to review it when I add it. So the tooltip is:

Syndesis/Fuse Online cannot make this specification available for review/edit. If you want to review or edit the API before Syndesis/Fuse Online creates the connector, try manually downloading the specification and then upload that file.

Agreed?  :-)

riccardo-forina:
Agreed 👍

PR #3299, opened 2018-08-03
refactor(db): replace database maintenance CronJob with aggressive autovacuum configuration

### LINKED ISSUE #3252: Errors in db-maintenance events log

zregvart:
It's difficult for us to use the CronJob as the container image used would need to be productized. Seems that as CronJob is a non-OpenShift resource referencing the same PostgreSQL image we use for the database pod -- referencing via lookup image stream from other namespace (in this case openshift) is currently not possible.

With this we loose `FULL ANALYSE` from `VACUUM` and `REINDEX TABLE` for `jsondb` table, the hope is that a more aggressive autovacuum configuration will be good enough.

See https://dba.stackexchange.com/questions/21068/aggressive-autovacuum-on-postgresql

Fixes #3252

### DIFF
--- a/install/generator/04-syndesis-db.yml.mustache
+++ b/install/generator/04-syndesis-db.yml.mustache
@@ -72,59 +72,18 @@
       syndesis.io/app: syndesis
       syndesis.io/type: infrastructure
       syndesis.io/component: syndesis-db
-    name: syndesis-db-maintenance-script
+    name: syndesis-db-conf
   data:
-    maintenance.sh: |
-      #!/bin/bash
-      cat << EOF| psql -h syndesis-db -U $POSTGRESQL_USER -d $POSTGRESQL_DATABASE
-      VACUUM FULL ANALYSE jsondb;
-      REINDEX TABLE jsondb;
-      SELECT COUNT(*) FROM jsondb;
-      EOF
-
-- apiVersion: batch/v1beta1
-  kind: CronJob
-  metadata:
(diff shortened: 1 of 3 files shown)


rhuss:
@chirino as the jsondb expert, do you think these settings are fine ? The discussion in the SO answer was a bit controversal, wasn't it ?

rhuss:
@zregvart Do you feel comfortable to merge ? Then I would say, lets do it.

zregvart:
We should test this on staging by doing a number of busy integrations and see over few days if the performance is tolerable...

PR #3307, opened 2018-08-05
chore(install): Added better error handling and split up admin setup

rhuss:


zregvart:
@rhuss seems that this needs to be rebased. Do we need to update the docs?

rhuss:
Yes, doc update is part of the "wip". Will come in today, as well as the update to the standalone installation script "install-syndesis" (which is the basis of the OCP installation script). THanks for the heads up !

rhuss:
@zregvart I'm good to go for a review now. Rearranged also a bit your --local addition.

Thanks !

doc/sdh/cli/cmd_install.adoc
```diff
@@ -102,22 +108,22 @@ syndesis install
...
+If not given, no link appears.
 
 By default, this commands installs Syndesis in the currently connected project, but you can specify an alternative project with `--project <project>`.
 If this project already exists, it gets deleted unconditionally before the deployment, so be careful when using this option.
-By default you are asked whether you really want to delete the project for recreation.
+By default, you are asked whether you want to delete the project for recreation.
 You can switch off the security question with the option `--yes` (short: `-y`).
 
-WARNING: Don't use `syndesis install --project $(oc project -q) --yes`. You will shoot yourself into the foot. Ask the author if you want to know more details.
+WARNING: Don't use `syndesis install --project $(oc project -q) --yes`. You shoot yourself into the foot. Ask the author if you want to know more details.
```
zregvart: nit, grammar _You'll_ (I think)

tools/bin/commands/install
```diff
@@ -6,9 +6,12 @@ install::description() {
 
 install::usage() {
     cat <<EOT
--s  --setup-as-admin <user>   Install CRDs clusterwide and optionally add permission to the given user so that
-                              user can install the operator in his projects. Othwerwise only admin can install the
-                              operator. You have to run this once as cluster admin
+-s  --setup-as-admin          Install CRDs clusterwide. Use --prepare-for-user if you want a specific user to be
```
zregvart: would it make sense to call this just `--setup`, `install_syndesis_crd` fails if the user isn't admin and prints a helpful message

rhuss:
@zregvart chance for quick review ? If ok, I will prepare a mail to send out about the change in the setup to be performed as admin, but it should be already documented.

'will give some examples, too.

tools/bin/commands/install
```diff
@@ -53,7 +67,16 @@ install::run() {
     # Setup oc
     setup_oc
 
+    # Check for the proper setup
+    set +e
+    oc get syndesis >/dev/null 2>&1
```
zregvart: We loose output of the command here, I wonder if we could have it captured and displayed also to help troubleshooting

tools/bin/commands/install
```diff
@@ -6,9 +6,12 @@ install::description() {
 
 install::usage() {
     cat <<EOT
--s  --setup-as-admin <user>   Install CRDs clusterwide and optionally add permission to the given user so that
-                              user can install the operator in his projects. Othwerwise only admin can install the
-                              operator. You have to run this once as cluster admin
+-s  --setup-as-admin          Install CRDs clusterwide. Use --prepare-for-user if you want a specific user to be
+                              able to install Syndesis. You have to run this option once as cluster admin.
+-u  --prepare-for-user <user> Add permissions for the given user so that user can install the operator
```
zregvart: `--setup-user`?

doc/sdh/cli/cmd_install.adoc
```diff
@@ -10,73 +10,79 @@ If you want to install to https://www.openshift.org/minishift/[Minishift] the <<
...
 syndesis install --setup-as-admin
 ----
 
-This will install only the CRD. In addition to grant a user _developer_ the proper permission to create a `Syndesis` resource, you should add the user name:
+This will install only the CRD. In addition to grant a user _developer_ the proper permission to create a `Syndesis` resource, you should add `--prepare-for-user <user>`:
 
 [source,bash,indent=0,subs="verbatim,quotes"]
 ----
-syndesis install --setup-as-admin developer
+syndesis install --setup-as-admin --prepare-for-user <user>
```
zregvart: Perhaps we should note that this command also needs to be run as admin user

rhuss:
have to add some last minute changes ...

tools/bin/commands/install
```diff
@@ -6,7 +6,7 @@ install::description() {
 
 install::usage() {
     cat <<EOT
--s  --setup-as-admin          Install CRDs clusterwide. Use --prepare-for-user if you want a specific user to be
+-s  --setup-crd               Install CRDs clusterwide. Use --prepare-for-user if you want a specific user to be
```
zregvart: much better :star2:

doc/sdh/cli/cmd_install.adoc
rhuss: My grammarly says, that I should all write in present tense. But agreed, future makes more sense here ;-)

rhuss:
> So my question is, could we have this even simpler? Something like:
> 
> $ syndesis install --setup
> $ syndesis install --grant developer

Yes, make probably more sense, but somehow I would also make it clear from the option name, that you have to run this as cluster admin. Of course, the error message is much better now (took me a bit to tune this) so maybe that's not so much of an issue.

You still need `--cluster` as you have to distinguish whether you want grant the perms to read the CRD only for the current project (I should probaby better also check --project here), or for all projects (now and in the future).

For running "sy install --project myproject" which replaces or creates a new project you would have had to setup the permission via "--cluster" otherwise you get the error you get.

Ok, let me change the option names a last time. I think, the shorter the better anyway.

doc/sdh/cli/cmd_install.adoc
zregvart: I was always hesitant to install something that looks at all my text with no compensation from me (and it wasn't open source), that could be just my paranoia surfacing...

rhuss:
@zregvart I'm good now. If you don't mind I would merge it with a single squash merge.

PR #3315, opened 2018-08-06
fix: hide the log/event link if the console URL isn't set

### LINKED ISSUE #3301: If consoleUrl is not defined and invalid URL to the logs is rendered
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>
In 7.0 the behaviour was that if the consoleUrl is not defined no link to the console was rendered. This was the compromise we came to after long discussion since it is not always easily possible to provide this information and we wanted to have a zero config setup

With 7.1 it seems that the URL to the pod event and pod logs into the console is unconditionally rendered, with an invalid URL like `/project/myproject/browse/pods/i-logtest-1-vcv7l?tab=logs` when not `consoleUrl` was configured.

gashcrumb:
fixes #3301

### DIFF
--- a/app/ui/src/app/integration/list/status-detail.component.ts
+++ b/app/ui/src/app/integration/list/status-detail.component.ts
@@ -85,6 +85,11 @@ export class IntegrationStatusDetailComponent implements OnChanges {
                                   current,
                                   total]);
       this.barWidth.width = ((current / total) * 100) + '%';
+      const consoleUrl = <string> this.configService.getSettings('consoleUrl');
+      // Check if the console URL is set properly, if not don't set up the log/event link
+      if (!consoleUrl || !consoleUrl.length) {
+        return;
+      }
       this.logUrl = undefined;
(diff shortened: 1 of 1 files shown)


app/ui/src/app/integration/list/status-detail.component.ts
```diff
@@ -85,6 +85,11 @@ export class IntegrationStatusDetailComponent implements OnChanges {
                                   current,
                                   total]);
       this.barWidth.width = ((current / total) * 100) + '%';
+      const consoleUrl = <string> this.configService.getSettings('consoleUrl');
+      // Check if the console URL is set properly, if not don't set up the log/event link
+      if (!consoleUrl || !consoleUrl.length) {
+        return;
+      }
       this.logUrl = undefined;
```
rhuss: Shouldn't this go up before the if ? Actually, it its unitialized it doesn't matter, but we should erase it if no consoleUrl is given.

PR #3325, opened 2018-08-07
fix(syndesis): Fix broken --cluster option

rhuss:
This is a backport of #3323


rhuss:
@zregvart could I please have a quick review here ? We need to respin the release anyway, would be good to have this in.

Issue #3341, opened 2018-08-07
UI updates to required form fields.


mcoker:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
When all fields in a form are required
* remove asterisks
* helper text should read "All fields are required."

When some fields are required
* helper text should read "The fields marked with * are required."

When all fields are optional
* helper text should read "All fields are optional."

related - https://github.com/syndesisio/syndesis/issues/3319

PatternFly code for reference - https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/forms.html

#### All fields required
<img width="1410" alt="screen shot 2018-08-07 at 12 19 20 pm" src="https://user-images.githubusercontent.com/35148959/43792623-ded7423a-9a3e-11e8-8193-0188ef0337e1.png">

#### Some fields required
<img width="1410" alt="screen shot 2018-08-07 at 12 29 36 pm" src="https://user-images.githubusercontent.com/35148959/43792617-dccd4c96-9a3e-11e8-86f1-634130db8598.png">

#### All fields optional
<img width="1418" alt="screen shot 2018-08-07 at 12 38 00 pm" src="https://user-images.githubusercontent.com/35148959/43792613-da86f270-9a3e-11e8-9151-1ace6d16fb4e.png">

cc @dongniwang

dongniwang:
cc: @elvisisking since he's working on #3319

gashcrumb:
So all of these pages are ng-dynamic-forms unlike #3319 

So most likely this will need to be implemented in our form service that creates the ng-dynamic-forms objects and probably some work will be needed in our ng-dynamic-forms UI component.  I think it should be fairly straightforward, we should probably move the current text that shows `the fields marked * are required` which is currently in the connection configuration template to the UI component, then we can easily display the correct hint text.

dongniwang:
@gashcrumb @michael-coker - can we close this issue now since PR #3345 is in place?

michael-coker:
@dongniwang the original issue was about the app as a whole. I'm also seeing the original issue still.

<img width="1593" alt="screen shot 2018-10-29 at 4 01 18 pm" src="https://user-images.githubusercontent.com/35148959/47680098-e7c62c80-db93-11e8-91c5-c5ad86f6e974.png">

And here are some others at a quick glance
[...]

Issue #3319, opened 2018-08-06
API client connector workflow - required fields should be indicated with a red asterisk (*)


dongniwang:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

Required fields are not indicated on step 3 of API Client Connector creation workflow. 

<img width="806" alt="screen shot 2018-08-06 at 11 36 58 am" src="https://user-images.githubusercontent.com/24943812/43727148-2cd10f78-996f-11e8-8f7b-8766d3932a3b.png">

## Expected behavior

Indicate required fields with a red asterisk (*) to the left of the filed label. Should also include the sentence "The fields marked with * are required." 

See PatternFly field labeling for detailed instruction https://www.patternfly.org/pattern-library/forms-and-controls/field-labeling/

(The spacing in the following mock-up is not correct, please reference PF site for correct implmentation - FYI @michael-coker )

![required_fields](https://user-images.githubusercontent.com/24943812/43727743-cd84b3f6-9970-11e8-8144-d3ad1257925c.png)

cc: @amysueg @michael-coker

elvisisking:
Work was done in #3345 and conforms to PatternFly guidelines.

PR #3345, opened 2018-08-07
Fixes UX issues with API client connector workflow required auth fields

elvisisking:
- See Issues #3319 and #3341
- i18n'd ApiConnectorAuthComponent html file
- Added recommended PatternFly message at top of wizard page indicating all fields are required
- Added message under each field that displays an error message when field has been touched and is empty or invalid (attn: @TovaCohen)
- Put focus on first field when form is displayed


elvisisking:
Can someone look at the change I made at line 50 of `api-connector-auth.component.ts` to see if it makes sense? Before making this change, if there were more than one authentication type on the "Specify Security" step (more than one radio button), the validation logic I added in this PR was not getting executed unless I changed the radio button selection. I was using the `invalid-type.petstore.swagger.json` found in the server test resources to test this.

app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-auth/api-connector-auth.component.ts
```diff
@@ -47,9 +47,15 @@ export class ApiConnectorAuthComponent implements OnInit, OnDestroy {
       .get('authenticationType')
       .valueChanges.subscribe(value => this.setOAuthFormValidation(value));
 
-    this.setOAuthFormValidation(
-      authenticationType ? authenticationType.defaultValue : ''
-    );
+    this.setOAuthFormValidation( defaultAuthenticationType );
```
kahboom: Wonder if `this.setOAuthFormValidation( defaultAuthenticationType ? defaultAuthenticationType : '' );` might be good. In reality, this code doesn't make sense (the `setOAuthFormValidation` method), not referring to your fix, considering that it gets passed a boolean, expects a string, and doesn't fail. We really should start setting return types, otherwise defeats the purpose of using TS.

PR #3349, opened 2018-08-08
chore(release): Fix quoting for operator version number when creating moving tag

rhuss:


rhuss:
The fix is included in #3348

PR #3379, opened 2018-08-14
Add a dependency-management version for hibernate-validator.   

cunningt:
syndesis-parent currently sets a hibernate.validator.version but does not respect it because there is no dependency-management entry.


cunningt:
@zregvart I'm patching 1.4 already for this in the dev build- I guess it wouldn't hurt to put cherry-pick into 1.4.x though.

Issue #3200, opened 2018-07-26
UI - Refactor business code to be framework agnostic


paoloantinori:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
We might need to change UI framework, moving away from Angular to something else.
To ease the activity we can start refactoring our current code so that all business code moves away from templates, if any, and is collected as much as possible in the business layer.

This is probably an ongoing long-term activity. Tracking it here so that the whole UI team is aware of this.

Updated based on @kahboom suggestions:

- take a stance on i18N solution to normalize our code base or officially deprecate the approach
- take a stance on Redux architecture to normalize our code base or officially deprecate the approach
- implement new tslint rules as necessary to comply with requirements around jsx/webcomponents, example tslint attached

Preview of tslint errors we'll encounter https://gist.github.com/seanforyou23/6da78befadfd75175e7c2e51e5081a00

Here attached is an example of the forthcoming tslint rules we can prepare for. We're using these rules with TypeScript 3.x so some of the properties may not be directly translatable with our current implementation - skip any of those that we encounter.
[tslint.txt](https://github.com/syndesisio/syndesis/files/2340164/tslint.txt)

paoloantinori:
Hi, yes, thank you Heiko.

This activity is a prerequisite we need to ease the migration to different frameworks in the future

kahboom:
I think @gashcrumb and I had mentioned in #syndesis-ui that the easiest way to approach this would be to first complete the Redux architecture, or the majority of it, over time, and then it'd be much easier to add in React or whichever framework or library we want. But please correct me if I'm wrong!

kahboom:
So, I think the plan for this we discussed was to maybe move to a Redux architecture (which was started but not completed), and I think stripping text into a JSON file (like the i18n library we have in place) will also help, but to be 100% framework agnostic would be quite difficult I think. Minimizing the framework boilerplate though is always a good idea and can help make the code base easier to work with, should we decide to move from one framework to another.

paoloantinori:
i've update the issue description based on @kahboom heuristics. Anything else that we should point out explicitly?

kahboom:
Think @elvisisking has been working a bit on i18n for the React initiative and can perhaps chime in on whether or not this task is still worth it, with respect to i18n?

[...]

Issue #2500, opened 2018-04-27
Wrong time format in metrics view


rhuss:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

![image](https://user-images.githubusercontent.com/99080/39368012-c19aa44c-4a38-11e8-891e-9884b3b6b9ae.png)

Should be "16:32" or "4:32 pm" ....

rhuss:
The other time (last processed) is in "4:32 pm" format ....

rhuss:
but it's cool that its not shown with the server's timezone but with the one of the user ;)

rhuss:
I still think this is relevant.

tplevko:
Seems to be still issue at least in the integration version history view:  ![screenshot from 2018-10-05 13-07-02](https://user-images.githubusercontent.com/4180208/46532026-a184f280-c89f-11e8-911d-32f82f506d35.png)

tplevko:
Even the Uptime detail in integration is the same, reopening 
![screenshot from 2018-10-05 13-12-52](https://user-images.githubusercontent.com/4180208/46532994-5cfb5600-c8a3-11e8-8003-748fda3ec6d4.png)

tplevko:
Verified, is OK now!

Issue #2329, opened 2018-04-10
Import/Export: Move strings into i18n dictionary


kahboom:
## This is a...

<pre><code>
[ x ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
Related to #2232. We need to move strings for the import/export-related components into the i18n dictionary.

## Expected behavior
We should be able to edit strings from a single dictionary. This work will be broken up into separate areas, and this specific issue is for the import/export module. See [this](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/platform/types/i18n/README.md) README for additional information.

PR #3381, opened 2018-08-14
[NOMERGE] RFC - Switch to Angular's own i18n tools

riccardo-forina:
From issue #3200:
> take a stance on i18N solution to normalize our code base or officially deprecate the approach

I decided to try porting the current in-house backed i18n solution to [Angular's default one](https://angular.io/guide/i18n) and share a list of pros and cons of this solution as I understand it.

Feel free to update the list with your feedback! There is some amount of labor involved in switching i18n module, so before proceeding further I'd like to know if we all agree to do it or not.

### Pros
* the whole toolchain is provided by Angular, so we can drop the in-house backed i18n module and reduce the LOC count (and the surface area for bugs)
* translations are handled through XLIFF (XML Localization Interchange File Format) files, which is an open standard. [This is how it looks like for this POC](https://github.com/riccardo-forina/syndesis/blob/i18n/app/ui/src/locale/messages.en.xlf). It supports all the things an i18n module should support, like pluralization.
I'd like to hear @TovaCohen about this since she is the stakeholder for translations right now.
* I think the HTML code looks cleaner this way, and it's surely easier for the developers since it reduces the amount of steps to 1 (which is, mark the text to be translated with the `i18n` attribute), versus the actual process where you have 1 + N steps (1 is to write the `syni18n` pipe in the HTML, N is the number of dictionaries to update - one per supported language).
* the toolchain can be configured to mark as warnings the missing translations while developing, and stop the build with errors when building for production, thus ensuring that no "broken" app is served to the end users.
* Angular handles also date/numbers/currency localization, which is a good thing. Especially for dates, it removes the current ambiguity with date formatting, for which we sometimes use Angular's pipes and sometimes use moment.js format. By sticking to Angular, we make it simpler to drop moment.js completely. Relevant issue #2500. 

### Cons

* the app must be built for every single language we want to support, and (continued to the next point).
* the server must be instructed to understand which language should be delivered to the user. This can be done by looking at the [Accept-Language](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language) header for example, but should also allow the user to manually change it if desired.
* this solution doesn't automatically make the move to a new framework (eg. React) trivial. But being the XLIFF a standard, there are tools to migrate it to other formats if needed.
* strings defined in TS files can't be translated [yet](https://github.com/angular/angular/issues/11405).

# Screenshot

syndesis-it

Here is a sample of the app running against the Italian messages that I translated to test the app serving.

### DIFF
--- a/app/ui/angular.json
+++ b/app/ui/angular.json
@@ -61,6 +61,12 @@
             ]
           },
           "configurations": {
+            "it": {
+              "aot": true,
+              "i18nFile": "src/locale/messages.it.xlf",
+              "i18nFormat": "xlf",
+              "i18nLocale": "it"
+            },
             "production": {
               "optimization": true,
               "outputHashing": "all",
@@ -76,7 +82,10 @@
                   "replace": "src/environments/environment.ts",
                   "with": "src/environments/environment.prod.ts"
                 }
-              ]
+              ],
+              "i18nFile": "src/locale/messages.en.xlf",
+              "i18nFormat": "xlf",
+              "i18nLocale": "en"
             }
           }
         },
@@ -86,6 +95,9 @@
             "browserTarget": "syndesis-ui:build"
           },
           "configurations": {
+            "it": {
+              "browserTarget": "syndesis-ui:build:it"
+            },
             "production": {
               "browserTarget": "syndesis-ui:build:production"
             }
--- a/app/ui/package.json
+++ b/app/ui/package.json
@@ -14,6 +14,7 @@
     "start": "yarn ng serve --ssl true --ssl-key localhost.key --ssl-cert localhost.crt --proxy-config proxy.conf.js",
     "start:prod": "yarn ng serve --ssl true --ssl-key localhost.key --ssl-cert localhost.crt --proxy-config proxy.conf.js --aot --prod",
     "start:minishift": "./scripts/minishift-setup.sh && yarn ng serve --host 0.0.0.0 --disable-host-check --public-host $(oc get routes syndesis --template '{{.spec.host}}'); ./scripts/minishift-restore.sh",
+    "start:minishift:it": "./scripts/minishift-setup.sh && yarn ng serve --configuration=it --host 0.0.0.0 --disable-host-check --public-host $(oc get routes syndesis --template '{{.spec.host}}'); ./scripts/minishift-restore.sh",
     "start:minishift-aot": "./scripts/minishift-setup.sh && yarn ng serve --aot --host 0.0.0.0 --disable-host-check --public-host $(oc get routes syndesis --template '{{.spec.host}}'); ./scripts/minishift-restore.sh",
     "restore:minishift": "./scripts/minishift-restore.sh",
     "build:ci": "yarn ng build --aot --prod --progress=false",
--- a/app/ui/src/app/app.component.html
+++ b/app/ui/src/app/app.component.html
@@ -30,7 +30,7 @@
           aria-label="Toggle Main Naivgation"
           type="button"
           class="navbar-toggle">
-          <span class="sr-only">{{ 'menu.navbartoggle' | synI18n }}</span>
+          <span class="sr-only" i18n="The sidebar toggle button, also known as hamburger icon@@menu.navbartoggle">Toggle navigation</span>
           <span class="icon-bar"></span>
           <span class="icon-bar"></span>
           <span class="icon-bar"></span>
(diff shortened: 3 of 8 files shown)


TovaCohen:
I don't completely understand what this approach would mean for me. How would this change what I do?
Looking at the messages.en.xlf file, it's not clear to me if I would ever edit that file. 
If this approach is adopted, does that mean that we no longer use the i18n dictionary, referenced in this issue: https://github.com/syndesisio/syndesis/issues/2329

I like the ability to update text strings that appear in the UI. For connectors, I can easily update the JSON files. As long as I can continue to contribute by updating text strings myself, any approach that you all deem better is fine with me. I'm sure someone will point me in the right direction and I can learn how to update text strings with the new approach.

riccardo-forina:
@TovaCohen you will still be able to update the text strings in the UI, the only difference for you would be working with these new xlf files instead of the json ones.

Going into more details, the two formats are different in the approach. Xlf files will contain every single string used in the UI, with the context of where it's been used (and possibly why, if the developer will provide such info). There are also tools that should ease working with the translations, eg. https://omegat.org/
The biggest difference though is that with the proposed system translation messages can't be joined together. Something like [this](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/assets/dictionary/en-GB.json#L50) for example would not be possible, and the full translation must be provided even if it means repeating the same translation many times.

elvisisking:
Does this approach provide a way to translate strings found in TypeScript (*.ts) files @riccardo-forina? Seems like I looked at this awhile back and at that time it didn't (or couldn't figure out how it did). Thanks.

TovaCohen:
This change is okay with me. :-)

elvisisking:
This method @riccardo-forina has a couple examples:
https://github.com/syndesisio/syndesis/blob/cd92d3f9bb36423664e12cfd8a6b196cfedf49a2/app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-create.component.ts#L201

riccardo-forina:
In that case, moving the translation to the HTML as 2 different `ng-container` blocks with the proper `*ngIf` should work.

But looking at how we use `.localize` across the app, I understand that for some cases there is no easy fix and we should probably come up with something different, like for the translation of error messages.

kahboom:
@riccardo-forina - Thanks for the contribution, really appreciate it. Build is failing due to some lint issues, but I wanted to talk a bit more about the i18n changes here. I don't think this is the direction we should be going in because of a few things:

- Well, I don't think we really need it when there are so many other parts of the app that need attention.
- We decided as a team to follow the current implementation because of the flexibility to use a single deployment for multiple languages. With this approach, you would need to build and deploy the app in each language you wanted, separately. If we used this a proper Redux fashion we should be able to switch the language easily at runtime. It would also make testing a bit trickier and unwieldy I think.
-  As you and @elvisisking mentioned, there is currently no way to translate strings in TS yet.
- JSON is just overall easier to work with and better adopted regardless of what framework we work with. I personally think it wouldn't be as easy to add keys, but that might just be personal preference.
- Current solution is pretty lean without much bloat, so even though it is rolled out, it is easy to work with and maintain, or to port over for whatever we want to do, including using it in the Redux setup.
- We don't know for certain that we are going the React route; there is actually an issue open to evaluate how framework-agnostic the UI can become. Even if we did go the React route, we don't know that there is some type of translation library that would be compatible with this, and if it'd still be maintained by the time we did that type of migration (who knows how long a whole app migration to React would take, and is not within the scope of the next Fuse release).

It's not a complete solution, as we don't have a way to load different locales yet, and that kind of defeats the purpose of having a flexible solution like we do now. With the current implementation, that would be just one issue to sort out.

Of course, would like for everyone else to chime in though, as this type of change should be a group decision.

seanforyou23:
@riccardo-forina nice investigation and pioneering!! I do agree with @kahboom's points though in that It's quite a bit of work to move to this native approach and in the end we would lose several important capabilities we have with the existing implementation. I know it was pretty involved and strategic to allow for single deployments and updating of locale during runtime so I see these as the two biggest issues.

All that said, I think it would be good to forward this investigation work over into the POC work we're doing with React. If we decide to go that route, whatever solution we have for i18N needs to be compatible with the new architecture.

elvisisking:
I agree with @kahboom that the current approach seems a lot easier when it comes to maintaining the i18n keys and strings. The XLIFF file seems a bit *ugly* to me and contains lots of information that *might* not really be necessary; and possibly makes it harder for a translator to work with. And trying to come up with a temporary solution for translating strings in TS files until the framework adds this feature seems like a lot work that might have to be undone later.

gashcrumb:
Yeah, I definitely think whatever solution we use needs to allow for:

1)  Building/combining strings using keys like [this kind of thing](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/assets/dictionary/en-GB.json#L189)
2)  The ability to insert text into a string template programatically
3)  Should be able to build strings either in a template or component code.

riccardo-forina:
Thank you all for the feedback, I guess this can be closed now :)

Issue #2850, opened 2018-06-18
Enable drag/drop UI for API connector upload


michael-coker:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
API client connector upload should have the same drag/drop interface as the tech extension and integration upload UI's.

ping @gashcrumb @kahboom

PR #3383, opened 2018-08-14
Enables drag/drop UI for API connector upload to be similar as rest of app

elvisisking:
- See issue #2850 
- ApiConnectorSwaggerUploadComponent is now using the same file upload component as in other places in app that allow DND
- Added code in api-connector-create.component.html that does not allow a file to be DND'd in the ApiCurio wizard unless in a drop zone.
- i18n'd files that were touched

@michael-coker please take a look at CSS change to see if what I did is OK
@TovaCohen please take a look at the i18n file for any changes you might want to make

### DIFF
--- a/app/ui/src/assets/dictionary/en-GB.json
+++ b/app/ui/src/assets/dictionary/en-GB.json
@@ -45,6 +45,7 @@
         "url-upload-instructions": "Upload the OpenAPI 2.0 file for your custom API client connector. Custom APIs are RESTful APIs and can be hosted anywhere, as long as a well-documented API specification is available and conforms to the OpenAPI standard.",
         "api-upload-valid-file": "Successfully uploaded '{{0}}'.",
         "api-upload-invalid-file": "'{{0}}' is not a valid file. Only files that end in '.json' can be uploaded.",
+        "api-upload-helper-text": "Accepted file type: .json",
         "api-url-upload": "Use a URL",
         "api-url-upload-note": "* Note: After uploading this specification, Syndesis/Fuse Online does not automatically obtain any updates to it. You would have to upload an updated specification and create a new API client connector that incorporates the updates.",
         "url-validation-error": "You must enter a valid URL before submitting the form."
--- a/app/ui/src/assets/dictionary/en-GB.json
+++ b/app/ui/src/assets/dictionary/en-GB.json
@@ -40,10 +40,11 @@
         "create-details-step-title": "Review/Edit Connector Details",
         "edit-tooltip": "{{project.name}} cannot make this specification available for review/edit. If you want to review or edit the API before {{project.name}} creates the connector, try manually downloading the specification and then upload that file.",
         "api-file-upload": "Upload an OpenAPI file",
-        "api-file-upload-dnd": "Drag and drop your OpenAPI file here or click the 'Choose File' button to use the file chooser.",
+        "api-file-upload-dnd": "Drag and drop your OpenAPI file here or",
         "api-upload-choose-method": "Choose how you want to create your connector:",
         "url-upload-instructions": "Upload the OpenAPI 2.0 file for your custom API client connector. Custom APIs are RESTful APIs and can be hosted anywhere, as long as a well-documented API specification is available and conforms to the OpenAPI standard.",
-        "api-upload-invalid-file": "'{{0}}' is not a valid file. Only files that end in '.json' can be imported.",
+        "api-upload-valid-file": "Successfully uploaded '{{0}}'.",
+        "api-upload-invalid-file": "'{{0}}' is not a valid file. Only files that end in '.json' can be uploaded.",
         "api-url-upload": "Use a URL",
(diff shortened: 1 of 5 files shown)


kahboom:
This is purposely not allowing multiple, correct?

elvisisking:
Yes @kahboom. Since we are only uploading one API definition, if there are multiple files being DND'd, I ignore all but the last one.

kahboom:
Looks good to me, thanks so much @elvisisking . Waiting on @TovaCohen and @michael-coker 's blessing before approving.

app/ui/src/assets/dictionary/en-GB.json
```diff
@@ -38,7 +38,15 @@
...
+        "edit-tooltip": "{{project.name}} cannot make this specification available for review/edit. If you want to review or edit the API before {{project.name}} creates the connector, try manually downloading the specification and then upload that file.",
+        "api-file-upload": "Upload an OpenAPI file",
+        "api-file-upload-dnd": "Drag and drop your OpenAPI file here or click the 'Choose File' button to use the file chooser",
+        "api-upload-choose-method": "Choose how you want to create your connector:",
+        "url-upload-instructions": "Upload the OpenAPI 2.0 file for your custom API client connector. Custom APIs are RESTful APIs and can be hosted anywhere, as long as a well-documented API specification is available and conforms to the OpenAPI standard.",
+        "api-upload-invalid-file": "'{{0}}' is not a valid file. Only files that end in '.json' can be imported.",
+        "api-url-upload": "Use a URL",
+        "api-url-upload-note": "* Note: we will download this for you but any updates to the online version will not be updated in our app.",
```
TovaCohen: Can we change this note to:
"After uploading this specification, Syndesis/Fuse Online does not automatically obtain any updates to it. You would have to upload an updated specification and create a new API client connector that incorporates the updates."
Too long? Just use the first sentence.

app/ui/src/assets/dictionary/en-GB.json
```diff
@@ -38,7 +38,15 @@
...
+        "url-validation-error": "You must enter a valid URL before submitting the form"
```
TovaCohen: There should be a period at the end of that message since it is a complete sentence.

app/ui/src/assets/dictionary/en-GB.json
```diff
@@ -38,7 +38,15 @@
         "create-review-step-title": "Review Actions",
         "create-security-step-title": "Specify Security",
         "create-details-step-title": "Review/Edit Connector Details",
-        "edit-tooltip": "{{project.name}} cannot make this specification available for review/edit. If you want to review or edit the API before {{project.name}} creates the connector, try manually downloading the specification and then upload that file."
+        "edit-tooltip": "{{project.name}} cannot make this specification available for review/edit. If you want to review or edit the API before {{project.name}} creates the connector, try manually downloading the specification and then upload that file.",
+        "api-file-upload": "Upload an OpenAPI file",
+        "api-file-upload-dnd": "Drag and drop your OpenAPI file here or click the 'Choose File' button to use the file chooser",
```
TovaCohen: There should be a period at the end of that message since it is a complete sentence.

TovaCohen:
I added three comments. The rest looks good.  Thanks Daniel!

elvisisking:
Thanks for your review and comments @TovaCohen. I made the changes you recommended.

michael-coker:
@elvisisking looks pretty good, but it does need a few changes. If you want to go ahead and get this merged, I'll create a follow up issue and update the CSS/layout.

@TovaCohen if you look in the DnD import box for importing an extension or integration, we have helper text like "Accepted file type: .jar" and "Note: The imported integration will be in the draft state. If you previously imported and this environment has a draft version of the integration, then that draft is lost."

Should we have some helper text in this import box, too?

TovaCohen:
@michael-coker Do we require a .json file to create an API client connector? If yes, then sure, it would good to have helper text as for import. 
If the user specifies a URL, does it have to end with .json? Again, if there are requirements and we won't upload it if the ending or the file extension is wrong, then yes, helper text would be good.

michael-coker:
@TovaCohen looks like there is validation on the file uploader that requires the file have a `.json` extension. 

I'm not sure about validation on the URL field.

TovaCohen:
Thanks for checking, Michael. 
I think for the file, "Accepted file type: .jar" just like the import operation would be great.

app/ui/src/assets/dictionary/en-GB.json
michael-coker: @TovaCohen just checking to see if you saw this comment? Sorry if I missed your reply :)

TovaCohen: Thanks Michael, I did miss your comment. I think that the text here should match the text in the integration and extension importers.

elvisisking:
Thanks @michael-coker and @TovaCohen. I will make the changes in your comments here and also the changes mentioned in the email (always clear choose file input, move selected file message to the right of the choose file input).

michael-coker:
@elvisisking @TovaCohen got sign off from the designers on a few updates and updated the PR if you want to take a look.

TovaCohen:
LGTM!

elvisisking:
Looks good @michael-coker !

PR #3393, opened 2018-08-16
fix(server): don't destroy Connector's configuredProperties on second startup

### LINKED ISSUE #3386: OAuth credentials disappearing
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem

Syndesis version 1.4.5
While playing with concur using OAuth credentials from settings page, it already happened 3 times in last 24 hours that those credentials disappeared and connector stopped working (it worked before the credentials loss). I created an integration with concur connection and then edit the integration and publish over and over again, last time the credentials disappeared on fourth edit - publish. 

## Expected behavior

Credentials in settings page must be permanent and not disappear when they want.

zregvart:
On startup `io.syndesis.server.runtime.Migrations::run` is invoking `DataManager::resetDeploymentData`, we need to make sure that the `configuredProperties` of the Connector are not overwriten by the loaded Connector definition.

Fixes #3386

### DIFF
--- a/app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
+++ b/app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
@@ -150,6 +150,14 @@ public class DataManager implements DataAccessObjectRegistry {
 
                         if (connector != null) {
                             LOGGER.info("Load connector: {} from resource: {}", connector.getId().orElse(""), resource.getURI());
+                            final String id = connector.getId().get();
+                            final Connector existing = fetch(Connector.class, id);
+                            if (existing != null) {
+                                // the only mutable part of the Connector
+                                final Map<String, String> configuredProperties = existing.getConfiguredProperties();
+
+                                connector = connector.builder().configuredProperties(configuredProperties).build();
+                            }
                             store(connector, Connector.class);
                         }
                     }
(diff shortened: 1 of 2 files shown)


zregvart:
Headsup @syndesisio/qe if this is merged it will affect the `/api/v1/test-support` API, the `configuredProperties` of Connectors won't get reset to empty/default values.

lburgazzoli:
is that the right thing ? a connector definition should be immutable imho

zregvart:
> is that the right thing ? a connector definition should be immutable imho

@lburgazzoli OAuth settings are part of the Connector (in the `configuredProperties`), there's no other place to store them as the Connection doesn't exist yet, so until we have a ConnectorSettings (or similar) model this is required...

lburgazzoli:
I understand but it smells. I think we need a follow up issue.

zregvart:
Created #3394

mcada:
@zregvart we need a reliable way how to reset everything to default state in-between tests and I don't think deleting connections/credentials by ourselves is the correct way, WDYT?

zregvart:
> @zregvart we need a reliable way how to reset everything to default state in-between tests and I don't think deleting connections/credentials by ourselves is the correct way, WDYT?

I can add reset of connector configuration for the `/api/v1/test-support` endpoint, won't be that nice, we can revisit this in #3394, does that make sense?

avano:
@zregvart actually there is nothing to do for the support, as the test support first deletes all from DB (therefore also the connections) and then calling the resetDeploymentData

https://github.com/syndesisio/syndesis/blob/ae68817f8e158fe9968d7e492bd436fcf4e8efac/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/tests/TestSupportHandler.java#L69-L71

so we have the clean state because "existing" in your PR is always null :-)

zregvart:
@avano my bad, perfect, less work :smile:

PR #3441, opened 2018-08-24
fix(slack connector): Slack Connector: when used with username as act…

zregvart:
…ion the action will fail

(cherry picked from commit 8c587d3abaa715d08e2fa4c2378db2263ab7def1)


zregvart:
:robot:

PR #3448, opened 2018-08-24
feature(upgrade): Adapte templates and operator for referencing a registry

### LINKED ISSUE #3440: Template for upgrade pod for OCP misses registry reference.
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

The template for the upgrade pod in the main template for the productised image references the image without registry. I don't think this work, I think it should point to image in the registry (now its looked up in docker hub)

```yaml
- apiVersion: template.openshift.io/v1
  kind: Template
  ....
  objects:
  - apiVersion: v1
    kind: Pod
    metadata:
      name: syndesis-upgrade-${SYNDESIS_VERSION}
    spec:
      containers:
        image: fuse7/fuse-ignite-upgrade:1.1-2
      ....
```

cc @heiko-braun

rhuss:
Now the operator can be called with a `-registry <registry>` argument for using that registry when instantiation the upgrade pod

A productised Dockerfile should use this option to point the target registry.

Also, the template generator has an options `--upgrade-tag` to directly specify the upgrade tag.

Fixes #3440

### DIFF
--- a/install/syndesis.yml
+++ b/install/syndesis.yml
@@ -1617,9 +1617,9 @@ objects:
       openshift.io/documentation-url: "https://syndesis.io"
       openshift.io/support-url: "https://access.redhat.com"
   parameters:
-  - name: SYNDESIS_VERSION
-    description: 'The version to upgrade to or "latest" for the newest version'
-    value: latest
+  - name: UPGRADE_REGISTRY
+    description: 'Registry from where to pick up the upgrade pod'
+    value: docker.io
     required: true
   message: |-
     Syndesis upgrade process started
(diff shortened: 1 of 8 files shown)


rhuss:
@djcoleman @heiko-braun this should fix #3440.

The productisation process should issue the --upgrade-tag of the template generator to directly feed in the taget tag incuding build number and the option -registry of the syndesis-operator when creating the productised docker image.

install/syndesis.yml
```diff
@@ -1617,22 +1617,22 @@ objects:
...
+    description: 'Registry from where to pick up the upgrade pod'
+    value: docker.io
     required: true
   message: |-
     Syndesis upgrade process started
   objects:
   - apiVersion: v1
     kind: Pod
     metadata:
-      name: syndesis-upgrade-${SYNDESIS_VERSION}
```
heiko-braun: Why has the version been hardcoded?

heiko-braun: @rhuss Thanks, I understand. Please make sure to let @djcoleman know what he needs to change on his end

rhuss:
@djcoleman

IMO you could leverage that change by:

* Using --upgrade-tag 1.1-5 or so when generating the template
* User in you Dockerfile for the operator something like "CMD -registry brew-pulp...." (you get the idea)

Issue #3459, opened 2018-08-27
apicurio - warnings are not updated when changed via gui and saved


mcada:
See also **https://issues.jboss.org/browse/ENTESB-11443**

## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

When adding custom API connector and editing it in apicurio, operation numbers are refreshed, but warnings are not.

## Expected behavior

Warnings from apicur gui should be propagated out when saved

## Screenshot

![screenshot_20180827_125204](https://user-images.githubusercontent.com/14313995/44656331-bc3e8e80-a9f8-11e8-97f5-3d0800931ad6.png)

## Tasks involved / Steps to Reproduce

1. From customization page update an OpenAPI file with warnings
2. Go next and click review/edit button to open apicurio gui
3. Solve at least one warning and save, you will be returned to overview page
4. Note that warnings number was not updated

tplevko:
@mcada seems like this should be fixed by now in https://github.com/syndesisio/syndesis/pull/3483 could you please verify?

mcada:
@elvisisking I tried it with syndesis master branch --local installation and it is not fixed.

elvisisking:
Not sure what could be going on @mcada. I just got latest from master and did this:
1. From the `Customization` page click the `Create API Connector` button to start the upload wizard.
1. On the `Upload OpenAPI Specification` step of the wizard choose the `beer-api_2.0.json` file.
1. Click the `Next` button to advance to the `Review Actions` step.
1. On the `Review Actions` step notice there are 1 error and 8 warnings.
1. Click the `Review/Edit` button to open the *ApiCurio* editor.
1. In the *ApiCurio* editor, delete the `Delete` method at path `/beers/{beerId}`.
1. Click `Save` button to take you back to the `Review Actions` step.
1. Notice that the number of warnings has decreased by one and is now seven.

Maybe let me know what steps and file you are using and I will give that a try. Thanks!

mcada:
I use this file in my test: https://github.com/syndesisio/syndesis-qe/blob/master/ui-tests/src/test/resources/swagger/connectors/invalid/kie-server-swagger.json

I do not delete whole operation, I fix one warning:
![out](https://user-images.githubusercontent.com/14313995/45410844-0a979280-b673-11e8-89fd-b9ba55e59ff8.gif)

[...]

PR #3483, opened 2018-08-28
Fix for Issue 3459 apicurio - warnings are not updated when changed via gui and saved

elvisisking:
- See #3459 
- the problem was that the `CustomConnectorRequest` instance is being reused and all the validation properties were not being cleared before re-validation.
- now clearing out these properties before requesting validation: actionsSummary, errors, and warnings.
- fixed an issue where the `Save` button would be enabled even though there was no change made in the editor. This would happen when the API was changed previously using the editor.


gashcrumb:
Build seems to keep failing, I've kicked it off from the beginning, if if fails again can you maybe rebase this PR in case there's been some other update that might fix whatever seems to be failing in the extensions modules?

gashcrumb:
@syndesisio/backend any idea on the build failure here?

@elvisisking you can try running `syndesis build` locally, but this looks like it's failing due to maven deps vs any change that this PR is related to.

elvisisking:
As mentioned to you in IRC, the local build works fine @gashcrumb

gashcrumb:
Maybe try just one more rebase and push and see if this builds?

PR #3485, opened 2018-08-28
add user namespace to use for the subjecAccessReview in oauthproxy

JameelB:
Allow the custom resource to accept a `USER_NAMESPACE` env var. This allows us to specify the user's namespace where the Syndesis service instance is provisioned in and use it for the `subjectAccessReview` to check for the user's authorization.

**Use Case:**
To provide a managed Syndesis installation, we would like to deploy the syndesis operator and deployment configs into a non-user controlled namespace, separate from the user's namespace. The user's namespace will only have the Syndesis service instance with the Syndesis dashboard URL.

At the moment, the `subjectAccessReview` is set to check against the OpenShift namespace where the operator and the deployment configs are deployed in. Due to this, the user who provisioned the Syndesis service will not have access to the Syndesis dashboard. 

Setting the `subjectAccessReview` to check against the user's namespace (`USER_NAMESPACE`) will allow the user to be authorized to access the Syndesis dashboard. If this variable was not defined, it will then default to the OpenShift namespace where the deployment config and operator are deployed in.

### DIFF
--- a/install/generator/01-header.yml.mustache
+++ b/install/generator/01-header.yml.mustache
@@ -30,6 +30,10 @@ parameters:
   description: The name of the OpenShift project Syndesis is being deployed into.
   displayName: OpenShift project to deploy into
   required: true
+- name: SAR_PROJECT
+  description: The user needs to have permissions to at least get a list of pods in the given project in order to be granted access to the Syndesis installation in the $OPENSHIFT_PROJECT
+  displayName: OpenShift project to be used to authenticate the user against
+  required: true
 - name: OPENSHIFT_OAUTH_CLIENT_SECRET
   description: OpenShift OAuth client secret
   generate: expression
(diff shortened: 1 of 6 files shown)


chirino:
Cool.  I hope the user will have view access into what's running in that admin controlled namespace.  Otherwise it will be very hard for him to troubleshoot thing when they go wrong.

JameelB:
@chirino From my understanding, the admin user should be responsible for troubleshooting the pods. I'm not sure if there's any logs available on the dashboard, but if its one user testing this out, they should have access to that admin user. Can you confirm this? @david-martin @maleck13

maleck13:
As mentioned in the google doc, I think we could use RBAC when creating the namespace to allow the provisioning user view access to pods in the namespace

Issue #3531, opened 2018-08-31
Error using API Connectors 


KurtStam:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[*] Bug report  
[ ] Documentation issue or request
</code></pre>

## scope and impact of the issues
**You cannot work with data returned from API's**. This issue affects all connectors created from a swagger upload. 

## description of the behaviour
Lets take concur as an example: "Create a new listitem" action does not return anything in syndesis so I can not save newly created listitem's ID.
My integration created new listitem and then it failed in data mapper where I mapped ID from response it said in logs that /ID does not exist.

## description of the behaviour you expect
I expect that I can use the data which I get from API connector action later in the integration - available data mapper with visible data from previous api connector action. 

## Reproducer A
[Integration_with_concur-export.zip](https://github.com/syndesisio/syndesis/files/2345063/Integration_with_concur-export.zip)
To reproduce, you can import this integration. (correct concur OAuth credentials required) 
Due to constants not working in data mapper, I add those values into our database which runs with syndesis:

insert into CONTACT values ('Anivia' , 'Bird', 'gWnytPb$pHxNJMPz4yL0nosnCQ4r30gRWs4w' , 'some lead', '1999-01-01');

You can see how are those values used from attached integration.
Run the integration and check integration logs - there is an error that /ID does not exist.

## priority
Concur is just technical preview, I don't mind marking it as p1. But as it affects all connectors created via swagger upload, I think this should be marked as blocker as it has no workaround. I (mcada) only pointed on this error that it does not work, @KurtStam investigated it deeper and should state the priority. 

## Reproducer B:

[db_custom_api_db-export.zip](https://github.com/syndesisio/syndesis/files/2349271/db_custom_api_db-export.zip)

Set up:
Add following values into database:
```
insert into CONTACT values ('evelyn' , 'Queen', '1' , 'some lead', '1999-01-01')
insert into TODO values (1, 'malphite' , 0)
```
After the import, perform these steps **before publishing the integration**:

- change the todo connection host name to reflect your minishift host name
- edit the integration name, save it
- delete and recreate the todo connection step in the flow (otherwise the new host name is not pick up)

Swagger for the custom connector I used:
Note that I added .txt to be able to upload it here.
[todo.swagger.yaml.txt](https://github.com/syndesisio/syndesis/files/2349279/todo.swagger.yaml.txt)

[...]

Issue #1993, opened 2018-03-14
AMQ and AMQP connections incorrectly offer RequestResponse as a finish connection action


TovaCohen:
## This is a...

<pre><code>
[ ] Feature request
[x ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
Dhiraj confirmed for me that AMQ and AMQP connections that are used as the finish connection in an integration should not perform the Request Response action. That was the behavior in TP3. But now, the staging site displays the Request Response action as a choice when you add an AMQ or AMQP connection as a finish connection. 

## Expected behavior

When you add an AMQ or AMQP connection as a finish connection, the only action choice should be Publish Messages. 

## Screenshot
This appears when I am creating a new integration. I selected an AMQ connection as the finish connection, and these are the actions that I can select from. Request Response should NOT appear. 
![image](https://user-images.githubusercontent.com/25067106/37419998-2384254a-278c-11e8-8e06-46a750d2106a.png)

## Tasks involved / Steps to Reproduce

1. Create an AMQ connection.
2. Create a new integration and add the start connection.
3. For the finish connection, select an AMQ connection.
4. Two actions are available here, publish and request/response. Only publish should be available.

gashcrumb:
Currently the UI filters the available actions based on the `pattern` attribute for an action, and that only has `From` and `To`, where `To` can be anywhere and `From` can only appear as a start connection.  Obviously we need a third option or better define this particular attribute.

iocanel:
It might make sense to also use output datashape kind, so that we can distinguish between things that need to be terminal actions (and thus has output datashape kind equal to 'none').

Not sure if this approach is correct. I am experimenting it with it anyways. @rhuss, @lburgazzoli thoughts?

gashcrumb:
Seems legit :-)

Though yeah, I'd maybe caution against being too hasty, it could be that we don't have any actions that have a `none` output data shape...

gashcrumb:
And also the semantics of that increase complexity, it's certainly a lot easier to have one field that we can definitively check vs. 2 fields that require some interpretation.

iocanel:
FWIW, this approach does work. So I removed the WIP and its up for review.

gashcrumb:
K!  Note that if there's a bug raised for xyz action isn't available as a finish step you're gonna get a grumpy cat meme e-mailed to you :-)

[...]

PR #3532, opened 2018-08-31
[fix] Generated actions should have default pattern of 'Pipe' #3531

KurtStam:
This issue (#3531) was found by @mcada during testing the Concur connector. The fix allows generated action to allow a datamapper both before and  _after_ the connection executes. I think this is the most common use case so therefor the default action pattern should be 'Pipe' and not 'To'. 

This PR contains an updates Concur connector, as well as a four character code fix in the code the generates connectors from Swagger in general, along with updates to unit tests to satisfy the code change.


KurtStam:
Hi @lburgazzoli , this issue was found by @mcada. It should allow generated action to allow a datamapper after the connection executes. I think this is the most common use case so therefor the default action pattern should be 'Pipe' and not 'To'. Please take a look. I think this fix should be applied to the 1.4 branch as well.

lburgazzoli:
I think we had a similar discussion quite a few time ago but I do not remember what `Pipe` was supposed to mean, I guess it means that the input and output data shape are the same.

I'm not very sure about this solution and to be honest I think that:

- data mapper should not care about the step pattern, the only things that matters are the in/out data shapes (unless of course I'm missing something) otherwise the definition becomes a little confusing (i.e. I can define a step whit patter pipe with output data shape 'none'). I remember this was needed in the early days we we assumed that only start/end components had a shape. @gashcrumb can you clarify the behaviour ?
 
- I'm not very sure if the result of a REST operation is of the same type of the request so setting 'pipe' as pattern does not feel right.

heiko-braun:
@oscerd , @nicolaferraro Mind a review?

nicolaferraro:
The `Pipe` pattern is a surprise to me. Syndesis allows actions with `To` to be used in the middle of a integration. A Syndesis action may contain a input and output datashape, and I expect that the datamapper is enabled between two connections if they respectively have a output datashape and a input datashape defined.

I see the pattern discussion started from here: https://github.com/syndesisio/syndesis/issues/1993

Maybe @gashcrumb can shed light on it.

nicolaferraro:
From a API consumer point of view, if `Pipe` enables the datamapper, we should set `Pipe` if the swagger action has a response type.

If we set it unconditionally.. bringing this forward, I'd say that we can replace `To` with `Pipe` everywhere (thus, removing `Pipe` is a better option).

heiko-braun:
@nicolaferraro I don't fully understand this:

> If we set it unconditionally.. bringing this forward, I'd say that we can replace To with Pipe everywhere (thus, removing Pipe is a better option).

Can you elaborate?

nicolaferraro:
I think that the presence of the datamapper should be driven by the presence of the input/output datashapes combination, not by the kind of action. We should remove 'Pipe' and apply the corresponding UI logic also to 'To' actions in my opinion.

heiko-braun:
@igarashitm @gashcrumb Can we get your feedback on @nicolaferraro suggestion to drive "the presence of the datamapper should be driven by the presence of the input/output datashapes combination, not by the kind of action"

gashcrumb:
I think that's fine, the one issue I would see is the case where the
datatshape is based on some configuration property, the UI still needs a
hint in that case that the action should be available in a given position,
as in some cases no datatshape is set on the action until the user starts
to configure the action in a flow.

heiko-braun:
@chirino Can we get your comments on the issue (#3531) and the proposed solution(s) as well? We need to take a decision today.

KurtStam:
So can I conclude that I can delete this PR and we are going to change how to the UI works? FYI the Concur connector has both in and output shapes defined. 

Can we get rid of `Pipe` or is there still a use for it?

heiko-braun:
@gashcrumb I don't fully understand which of the proposal you support: The one in the PR or the approach @nicolaferraro did suggest? If you support the later, what changes would it require on the UI side or in AtlasMap (cc @igarashitm)?

igarashitm:
datamapper step is only looking at datashapes
https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/edit-page/step-configure/data-mapper/data-mapper-host.component.ts#L199-L214

gashcrumb:
@heiko-braun If I recall correctly the UI only uses these fields in one spot, which is to determine what actions/connections should be available at the start, middle and end of a flow.  We could change this to rely on the presence and kind of datashapes for a given connection or action, however in cases where the datashape is based dynamically on the how an action is configured the UI needs a fallback mechanism to know if a connection or action should be available at a given spot in the flow.  So I'm fine with getting rid of this pattern field, but this will potentially be an issue that we'll need to solve as a result.

Note that this field doesn't determine whether or not the data mapper is available, that's calculated via the available data shapes, and I think is a different thing altogether.

gashcrumb:
Note, I don't recall at all where `Pipe` came from exactly :-)

nicolaferraro:
I've not fully understood what is the issue here.

The concur action mentioned in the issue (`create a new listitem`) has a output datashape and I'm able to map it.

screenshot from 2018-09-04 16-42-08

So, is this issue appearing only when the integration is running @mcada? It seems something related to the backend runtime or datamapper and not related to the `Pipeline` pattern, that is something UI only.

heiko-braun:
Here's the error @mcada's reproducer reveals:

Error from my reproducer (I will add the reproducer in the description later):

```
io.atlasmap.api.AtlasException: Errors: [Source document '-LL_2gxr3Jq4DC6RiYJQ' doesn't exist: docId='-LL_2gxr3Jq4DC6RiYJQ', path='/task'], 
	at org.apache.camel.component.atlasmap.AtlasEndpoint.onExchange(AtlasEndpoint.java:213) ~[camel-atlasmap-1.35.7.jar!/:na]
	at org.apache.camel.impl.ProcessorEndpoint$1.process(ProcessorEndpoint.java:71) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.util.AsyncProcessorConverterHelper$ProcessorToAsyncProcessorBridge.process(AsyncProcessorConverterHelper.java:61) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.SendProcessor.process(SendProcessor.java:148) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.DelegateAsyncProcessor.process(DelegateAsyncProcessor.java:110) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.RedeliveryErrorHandler.process(RedeliveryErrorHandler.java:548) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201) [camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:138) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:101) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.DelegateAsyncProcessor.process(DelegateAsyncProcessor.java:110) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at io.syndesis.integration.runtime.logging.ActivityTrackingInterceptStrategy$EventProcessor.process(ActivityTrackingInterceptStrategy.java:79) ~[integration-runtime-1.4.7.fuse-710001-redhat-00001.jar!/:1.4.7.fuse-710001-redhat-00001]
	at org.apache.camel.processor.DelegateAsyncProcessor.process(DelegateAsyncProcessor.java:110) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.RedeliveryErrorHandler.process(RedeliveryErrorHandler.java:548) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201) [camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:138) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:101) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.RedeliveryErrorHandler.process(RedeliveryErrorHandler.java:548) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201) [camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.MulticastProcessor.doProcessSequential(MulticastProcessor.java:715) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.MulticastProcessor.doProcessSequential(MulticastProcessor.java:638) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.MulticastProcessor.process(MulticastProcessor.java:248) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.Splitter.process(Splitter.java:122) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.DelegateAsyncProcessor.process(DelegateAsyncProcessor.java:110) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.RedeliveryErrorHandler.process(RedeliveryErrorHandler.java:548) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201) [camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:138) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:101) ~[camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201) [camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.component.timer.TimerConsumer.sendTimerExchange(TimerConsumer.java:197) [camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at org.apache.camel.component.timer.TimerConsumer$1.run(TimerConsumer.java:79) [camel-core-2.21.0.fuse-710018-redhat-00001.jar!/:2.21.0.fuse-710018-redhat-00001]
	at java.util.TimerThread.mainLoop(Timer.java:555) [na:1.8.0_181]
	at java.util.TimerThread.run(Timer.java:505) [na:1.8.0_181]

{"exchange":"i-LL_3bXudlluvFmkLmvwz","status":"begin"}
{"exchange":"i-LL_3bXudlluvFmkLmvwz","step":"-LL_2hHU3Jq4DC6RiYJR","id":"i-LL_3bXzdlluvFmkLmw-z","duration":1215591}
{"exchange":"i-LL_3bXudlluvFmkLmvwz","step":"-LL_2gxr3Jq4DC6RiYJQ","id":"i-LL_3bY0dlluvFmkLmw1z","duration":14538442}
2018-09-04 14:52:55.888  WARN 1 --- [r://integration] o.a.c.component.atlasmap.AtlasEndpoint   : Null or non-String source document: docId='-LL_2gxr3Jq4DC6RiYJQ': docId='-LL_2gxr3Jq4DC6RiYJQ', path='null'
2018-09-04 14:52:55.890 ERROR 1 --- [r://integration] o.a.camel.processor.DefaultErrorHandler  : Failed delivery for (MessageId: i-LL_3bYEdlluvFmkLmw2z on ExchangeId: i-LL_3bXydlluvFmkLmvyz). Exhausted after delivery attempt: 1 caught: io.atlasmap.api.AtlasException: Errors: [Source document '-LL_2gxr3Jq4DC6RiYJQ' doesn't exist: docId='-LL_2gxr3Jq4DC6RiYJQ', path='/task'],
```

heiko-braun:
@mcada Let's keep the info related this issue here, rather than on the PR

heiko-braun:
@nicolaferraro I think the reproducer @mcada showed had a second mapping step in the flow. And the second one failed with the exception above. @mcada Can you confirm this or explain the differences between your case setup and one @nicolaferraro is showing above?

A screenshot like the one above might be useful.

KurtStam:
@nicolaferraro ok so we have some pilot error on my behalf here. If the final step is a log step then it won't let you pick a datamapper, as the data shape of the logger is `any`. When I change the final connector to something with an inputDatashape I in fact do get a mapper. So now onto the stack trace listed above. But this PR will get deleted for sure, but not until we solve the remaining issue.

nicolaferraro:
Yes @KurtStam, the log action at the end of the flow confused also me at first. But mapping seems to work as expected from the UI side.

KurtStam:
Closing this PR however opening https://github.com/syndesisio/syndesis/issues/3543 since any api-connector does now seem broken with the latest from master.

PR #3537, opened 2018-09-04
Allow multiple issue templates

rhuss:


heiko-braun:
One for user stories would be good as well

PR #3553, opened 2018-09-05
chore: Add tech preview label to concur connector

### LINKED ISSUE #3457: SAP Concur connector has no technology preview label
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem

Concur is marked as technology preview in documentation and release notes, we should add this label into UI.
Its not really a blocker, but it is very easy to fix and has no regression threats.

rhuss:
Fixes #3457.

Not 100% sure whether this is the right spot. Please verify @KurtStam


KurtStam:
We should add it to the project (that we are trying to opensource) that generates the concur-connector.json or else we will loose this change next time we regenerate.

rhuss:
@KurtStam so how do you currently generate this descriptor file ? Do you have a pointer for me ?

PR #3554, opened 2018-09-05
chore: Move concur-connector out of "dao" to "connectors" in a separate module.

### LINKED ISSUE #3292: Create a new home for the Concur connector, and perhaps other API connector based connectors in the future
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## The problem

We bundle the Concur connector by loading the `concur-connector.json` from `/META-INF/syndesis/api-connectors/` in the DAO module.

The way we build the `concur-connector.json` is to merge several Swagger specifications we download from Concur, patch them to be correct and run a Groovy script to artisanally tune the resulting connector metadata.

Let's create a Maven module in under `app/connector` tree that produces artefact containing the connector descriptor. We could also produce multiple artefacts (I think this is better) or bundle multiple connectors in the same artefact.

## Expected behavior

Make the world a better place.

## Screenshot

giphy

1. Create a new Maven module under `app/connector`
2. Create Maven build that merges Concur Swaggers, and creates the resulting connector JSON with proper metadata (names/descriptions/icons/tags...)
3. The result should be a JAR file with only the connector JSON
4. Profit

rhuss:
This fixes #3292 (and also #3457 on master) but needs to be verified that it works as expected.


rhuss:
Motivation of this move is layed out in #3292 

And as I _think_ from inspecting the connector usage in the DataManager, the usage for a plain connector and an api connector is the same (at least it is treated the same in the DataManager), so I don't think it is required to distinguish them on the user level. At the end for a user its all "connectors" (regardless how they are build internally). 

Could be that I missed something, but for me it feels natural to put the concur API connector to the "connectors" group, regardless whether its a standard or API connector.

KurtStam:
We generate the concur-connector.json from the concur swagger. It is this project that should live in 
connectors/api-connectors/concur. And that should create the jar that contains the concur-connector.json. I've asked concur if we can publish their swagger.json in our repository 

https://forum.developer.concur.com/t/what-is-the-license-of-the-swagger-api-documents/1839

and if it's ok to derive code from their swagger.  But I have not yet heard back. 

My preference is to keep loading the api-connectors from a different path until we can do this migration and it becomes clear we are loading an api-connector.

rhuss:
> connectors/api-connectors/concur

why connectors/api-connectors/concur and not just connectors/concur ? That it's an API connector is just an implementation details. From a usage point its still can be used just as any other connector.

rhuss:
@KurtStam please give me a compelling argument why it is important for the user of a connector which is based on the api connector and predefined for Syndesis to know its an "API Connector" if it used like any other connector like ServiceNow, Salesforce, ....

For the user it doesn't matter.

What technical reason do you have to make this distinction ? Isn't it just part of the build how we create the connector's descriptor ? And that can be different of any connector ?

So, what real benefit would we have to make it super explicit within the build that this is an "API connector" ? Where do we technically require this distinction ?

KurtStam:
@rhuss It's a maintenance argument. Not a technical one.

rhuss:
What's actually the maintenance argument if everything, including creating of the descriptor files would be included in the module ? What maintenance benefit would you have if you put in into an completely different group (which you have to add to the build scripts) ?

KurtStam:
If there is a bug fix to the api-generator as we are calling it now, all connectors that are based on this should get regenerated. Our build structure should reflect this. We need a subdirectory under connectors that houses these projects as well as the common code. Until we have this we some way to tell the developer that the concur connector is an instance of api-connector.

rhuss:
@KurtStam how is the script actually created ? 


Couldn't we make the recreation part of the regular build, so when the way how it is created changes and because that mechanism is included as a e.g. a plugin dependencies it automatically gets the update fitting to the version. Recreating the file is part of the build of this specific module, and each module's build should be isolated from other modules build, so the location of the module doesn't matter. This doesnt mean the modules might depend on each other, but this is declared in the pom.xml (not via directory relation ships).

If you look into the build configuration of the module, a developer will immediately see how the connector descriptor is created.

rhuss:
@lburgazzoli could you give this smallish PR please a quick review ? Do you think it would be possible to add the generation of the concur-connector.json to the build of this module ? (actually I'm not sure how this file is created as it is currently stored in the DAO module and comes out of the blue)

lburgazzoli:
the pr looks good to me but I'm afraid, I do not have any clue about how this file is generated

rhuss:
@KurtStam so the update of the file is currently a manual process or can it be automated ? It it can be automated, we could put it directly as part of the build in this module, which would be awesome.

What is the status quo for an update ? You update the json file and copy it over to Syndesis, is it like this ? I think in a first step it would be cool if the README of this module would explain how the json descriptor could be regenerated.

KurtStam:
@rhuss yes it can be automated, and this is what I've been saying the whole time. We are trying to sort of licensing issues.

rhuss:
cool, awesome. Then this new module is the right place if the issues are sorted out. I just updated the json with your latest version from master.

KurtStam:
@rhuss actually as requested above, let's put it in `connectors/api-connectors/concur`, so that we can put common code in `connectors/api-connectors`.

rhuss:
Ok, a possible compromise:

* Let's merge this PR now as its not clear how the generation of the concur-connector.json could be included here.
* When possible, you can move the generation of this json into this module, ideally with a build integration, so that it might get regenerated during every build (so concur-connector.json would be checked in, but generated into `target/` which then ends up in the jar.
* When we get to a second API connector which shares builder code or something else, we can decided to proceed: A dedicacted Maven plugin, shared script code in some location, collection all API connectors in one place, ...

@KurtStam Would this something you could live with ?

Issue #3436, opened 2018-08-23
Can't edit API connector definition added using the URL


tplevko:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

When creating new API connector, I can't edit the swagger.json file in the apicurio editor, if it was added using the "URL". The "review/edit" button is not clickable in this case, as can be seen in the screenshot. In case the same swagger.json was added using locally downloaded file, it can be edited.
The adding using URL is also mentioned in the related "swagger editing user story": https://github.com/syndesisio/syndesis/issues/2750 so it should be possible.

## Expected behavior

It should be possible to edit this API connector definition.

## Screenshot

![screenshot from 2018-08-23 09-45-37](https://user-images.githubusercontent.com/4180208/44511940-5335de80-a6b9-11e8-9896-1c1a8afc4a78.png)

## Tasks involved / Steps to Reproduce

1. Go to "API connectors"
2. Add one using URL, you can use http://petstore.swagger.io/v2/swagger.json
3. The "review/edit"

zregvart:
This is a consequence of our current approach, see #3174.

@kahboom with #3317 server now echoes the resolved specification in the `specification` `configuredProperty`.

tplevko:
@riccardo-forina seems like this was updated in https://github.com/syndesisio/syndesis/pull/3562 Can I close this?

riccardo-forina:
Totally, thanks!

PR #3562, opened 2018-09-06
fix: remove the workaround to fetch a remote OpenAPI file

riccardo-forina:
Remove the workaround to fetch a remote OpenAPI file and use the spec as returned by the backend instead.


riccardo-forina:
Closes #3436

PR #3569, opened 2018-09-07
Fix for Issue 3495 API connector wizzard has no back buttons

elvisisking:
- added back button to the ApiConnectorAuthComponent
- ApiConnectorAuthComponent now initializes state to values in the CustomConnectorRequest if they exist or to the default values
- ApiConnectorCreateComponent handles all back pressed events emitted by the steps of the wizard
- added click listeners to ApiConnectorSwaggerUploadComponent for the upload method radio buttons
- ApiConnectorSwaggerUploadComponent now initializes state to values in the CustomConnectorRequest if they exist
- added back button to the ApiConnectorInfoComponent
- added back button to the ApiConnectorReviewComponent
- added i18n where needed

### DIFF
--- a/app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-auth/api-connector-auth.component.html
+++ b/app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-auth/api-connector-auth.component.html
@@ -71,9 +71,14 @@
     </fieldset>
 
     <fieldset>
-      <syndesis-button type="submit" [disabled]="authSetupForm.invalid">
-        Next
-      </syndesis-button>
+      <div class="list-pf-actions">
+        <syndesis-button theme="default" type="button" (click)="onBackPressed()">
+          &lt; {{ 'shared.back' | synI18n }}
+        </syndesis-button>
+        <syndesis-button type="submit" [disabled]="authSetupForm.invalid">
+          {{ 'shared.next' | synI18n }}
+        </syndesis-button>
+      </div>
     </fieldset>
   </form>
 </div>
--- a/app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-auth/api-connector-auth.component.ts
+++ b/app/ui/src/app/customizations/api-connector/api-connector-create/api-connector-auth/api-connector-auth.component.ts
@@ -21,6 +21,7 @@ export class ApiConnectorAuthComponent implements OnInit, OnDestroy {
   authSetupFormValueSubscription: Subscription;
   @Input() customConnectorRequest: CustomConnectorRequest;
   @Output() authSetup = new EventEmitter();
+  @Output() backPressed = new EventEmitter();
 
   constructor(private formBuilder: FormBuilder) {}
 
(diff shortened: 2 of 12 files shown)


elvisisking:
@dongniwang Please take a look at the button positioning. I did add the back button to another step, the auth step (not pictured below), but it only had the back and next buttons. Thanks.

@michael-coker Can you please look at the way I used CSS to position the "Review/Edit" and "Create API Connector" buttons? I'm sure they can be done a better way. Those can be found in the `api-connector-review.component.scss` and the `api-connector-info.component.scss` files. Let me know and thanks. Also could you and/or @seanforyou23 take a look at how I ended up changing the html for those buttons in that bottom button bar on those 3 steps? Some files were using `syndesis-button` and `fieldset` while other files weren't. Just not sure. Thanks.

review-actions-step

review-details-step

elvisisking:
Another thing of note on how things are coded: If a user uploads a file, or uses a URL import, then edits that API spec, the hits next once or twice, then goes back to the first step, I am throwing out the changes that were made in the editor. The changes are only lost if the user goes back to the first step (the upload step). Keeping the changes going back to the first step maybe nice but seems to cause some UX issues like: how do we annotate that the spec has been edited and how does the user throw out the changes if they want to start over with the same spec.

dongniwang:
One thing I noticed in the screenshot. I think the "Next" button should have an arrow pointing to the right. "Next >"

I'm not sure I fully understand what's the issue here. I'm happy to set up a meeting to discuss further. 
>Another thing of note on how things are coded: If a user uploads a file, or uses a URL import, then edits that API spec, the hits next once or twice, then goes back to the first step, I am throwing out the changes that were made in the editor. The changes are only lost if the user goes back to the first step (the upload step). Keeping the changes going back to the first step maybe nice but seems to cause some UX issues like: how do we annotate that the spec has been edited and how does the user throw out the changes if they want to start over with the same spec.

michael-coker:
@elvisisking gave you a PR with some updates to the button layout.

elvisisking:
@dongniwang I've updated PR so that all the next buttons now have the '>' symbol.

dongniwang:
> I've updated PR so that all the next buttons now have the '>' symbol.

Cool thank you!

>Another thing of note on how things are coded: If a user uploads a file, or uses a URL import, then edits that API spec, the hits next once or twice, then goes back to the first step, I am throwing out the changes that were made in the editor. The changes are only lost if the user goes back to the first step (the upload step). Keeping the changes going back to the first step maybe nice but seems to cause some UX issues like: how do we annotate that the spec has been edited and how does the user throw out the changes if they want to start over with the same spec.

@elvisisking About this issue ^^, do you have a recording of the flow? If not, we can set up a meeting to discuss further.

gashcrumb:
lint error:

```
[INFO] $ yarn ng lint --format stylish
[INFO] /workspace/app/ui/src/polyfills.ts
[INFO] ERROR: 74:8  rxjs-add  Unused patched operator in polyfills.ts: first
[INFO] 
[ERROR] Lint errors found in the listed files.
[ERROR] error Command failed with exit code 1.
```

dongniwang:
Hi @elvisisking - thanks for the walkthrough, it helped me understand the issue much better. Can you please open a separate issue for the back button behavior in api client connector flow so that @syndesisio/uxd can investigate that. Thanks!

Issue #3564, opened 2018-09-06
Query parameters are not set in the API-Connector


KurtStam:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[*] Bug report  
[ ] Documentation issue or request
</code></pre>

Found by @mcada 

## Description
The DELETE action for listitems has a path of /common/listitems/{id}, 

https://developer.concur.com/api-explorer/v3-0/ListItems.html#/

however it also has a required query parameter called listId. The id and the listId are both available as parameters. The id is set in the RestProducer by substituting the {id} with the value of the id, however the listId query parameters is never set on the request as listId is never identified as a queryParameter.

This makes the request to concur fail. I'mn still trying to figure out what code is supposed to identify it as a queryParameter so that it can be set. Maybe it's a bug in camel-core altogether.

So to be clear; this happens with all type of HTTP requests GET/UPDATE/DELETE with query parameters like

/common/listitems/{id}?listId={listId}

mcada:
@KurtStam it still does not work for concur delete :/

KurtStam:
@mcada you tried on master?

mcada:
Yes I tried it on up-to-date master.

KurtStam:
ok will double check, and a GET + query params worked?

KurtStam:
@mcada - I think the fix is working but I think there was a bad merge. Here is a screenshot of the proof that the params are in fact set on a DELETE request

<img width="894" alt="screen shot 2018-09-13 at 12 53 02 pm" src="https://user-images.githubusercontent.com/35576/45503339-92a19900-b754-11e8-81b2-2dda268eef96.png">

The concur swagger is loaded in into the DB though. So if it's not a fresh install then you will need to run "delete from jsondb" in the syndesis db and then kill the server pod, so it repopulates the db. You will have to rebuild your concur connection and integrations.

KurtStam:
@mcada I reopened until you can verify this is working now.

mcada:
@KurtStam I did test it with fresh master as I always do, but I forgot to update my test to also set listId when deleting the item :( After I fixed that it works. My bad, thanks for your screen shot, I immediately knew what was my mistake after seeing it. Closing.

KurtStam:
Great! Thx Michael.

On Sep 14, 2018, at 04:40, mcada <notifications@github.com> wrote:

[...]

PR #3572, opened 2018-09-09
[fix] Query parameters are not set in the API-Connector #3564

KurtStam:
Fixes query parameters not being set on the HTTP request due to the fact that the swagger definition was pruned too much. This fixes the issue for any new swagger document being uploaded to be use to generate an api-connector.

There is only a small code change, and some updates test resources with updated swaggers.

https://github.com/syndesisio/syndesis/issues/3564


KurtStam:
retest please

PR #3574, opened 2018-09-10
Various api provider endpoint improvements

### LINKED ISSUE #3547: Adapt API provider endpoints to UI requirements
I've created a set of initial REST endpoints for the API provider. I've used them to test the backend and, since they are able to create a multi-flow integration from a swagger spec, they can be used to design the frontend too.

Multiflow integrations currently look this way:
screenshot from 2018-09-05 08-44-46

This will be handled as part of #3482.

Here's a description of current endpoints. We can discuss here what needs to be changed to support UI requirements.

**1) Validation**
Path: `POST /api/v1/apis/info`
Consumes: a multipart form with the OpenApi spec in the `specification` field
Produces: a "APISummary" object, that is the same object returned when validating a spec in the API consumer context, so that the UI components can be reused

Validation returns a summary of the operations, warnings and errors, as in the case of the API consumer. Validation rules are slightly different at backend side.

**2) Create initial integration**
Path: `POST /api/v1/apis/generator`
Consumes: a multipart form with the OpenApi spec in the `specification` field
Produces: a multi-flow Integration object with a flow for each REST operation in the spec

Each flow is configured with a start and a end action. The end action allows to customize the return code, that is set to 501 by default (select box of available HTTP return code is included).
This integration can be used as initial template by the UI.

**3) Create initial integration and save**
Path: `POST /api/v1/apis/generator/save`
Same input/output data as in *2*.

The difference is that the integration is also stored in Syndesis and appears as "Stopped" and "Draft".
This has been useful for testing the backend. I don't know which strategy (2 or 3) will be better from a UI point of view.

@gashcrumb , @riccardo-forina .

nicolaferraro:
Fix #3547.

This addresses the issues outlined in #3547.

- *Ensure the original schema is present in the response of the validation endpoint (needed if the user wants to go back and edit the integration)*

**Solution:** The implementation follows what is implemented in API consumer and the OpenAPI spec is already included.

- *Provide a endpoint to upload the schema via URL, to overcome browser restrictions*

**Solution:** like for the API consumer, just upload a URL instead of a openapi spec in the "specification" field of the form and it will be downloaded by the backend (and returned back from the call to /apis/info)

- *Add a field to the Flow object that contains the last column seen in the list of flows when editing a integration (the one reporting the current HTTP return code)*

**Solution:** I've added a "excerpt" field in the Flow object that will contain a text like "501 Not Implemented" (and other return codes)

- *Mark actions belonging to a API flow with a special "read-only" tag, so the UI does not allow to replace them*

**Solution:** I've added a "locked-action" tag to each action in the start and finish step of a generated integration. If present, the UI should prevent the user to change them (don't actually know if that was possible).

cc: @riccardo-forina , @gashcrumb

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java
@@ -15,24 +15,25 @@
  */
 package io.syndesis.common.model.integration;
 
-import java.io.Serializable;
-import java.util.Collections;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
-
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import io.syndesis.common.model.WithId;
+import io.syndesis.common.model.WithMetadata;
 import io.syndesis.common.model.WithName;
 import io.syndesis.common.model.WithTags;
 import io.syndesis.common.model.connection.Connection;
 import io.syndesis.common.util.json.OptionalStringTrimmingConverter;
 import org.immutables.value.Value;
 
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+
 @Value.Immutable
 @JsonDeserialize(builder = Flow.Builder.class)
 @SuppressWarnings("immutables")
-public interface Flow extends WithName, WithId<Flow>, WithTags, WithSteps, Serializable {
+public interface Flow extends WithName, WithId<Flow>, WithTags, WithSteps, WithMetadata, Serializable {
 
     class Builder extends ImmutableFlow.Builder {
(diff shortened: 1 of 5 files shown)


app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java
```diff
@@ -48,6 +48,8 @@
     @JsonDeserialize(converter = OptionalStringTrimmingConverter.class)
     Optional<String> getDescription();
 
+    Optional<String> getExcerpt();
```
lburgazzoli: Can we use attributes for that ?

nicolaferraro: Yes, better using attributes...

nicolaferraro:
@lburgazzoli I've moved the field to the "metadata" collection.

Issue #3482, opened 2018-08-28
Update integration detail page to add support for API provider multi flow information


gashcrumb:
Design -> https://redhat.invisionapp.com/share/4JNR7J5DMNY#/screens/316053627

Issue #3476, opened 2018-08-28
Update "save or add step" page to implement "add flow/integration" page based on design


gashcrumb:
Update [this](https://github.com/syndesisio/syndesis/tree/master/app/ui/src/app/integration/edit-page/save-or-add-step) page to suit [this](https://redhat.invisionapp.com/share/3CNPQLOKAEW#/screens/316356339) design.  This page in general is where you wind up after completing a sequence of pages in the editor.

riccardo-forina:
The updated design is here https://redhat.invisionapp.com/share/CBNWDY58AHG#/screens/318306622

We just need a link back to the Operations page, the inner content is the same as with a normal flow.

PR #3585, opened 2018-09-13
feat: ui changes to support API provider integrations

### LINKED ISSUE #3475: Create API provider choose operation page
Implement [this page](https://redhat.invisionapp.com/share/3CNPQLOKAEW#/screens/316356335).

We may have to muck with the routing for the editor in general, as we'll have an integration ID and then ultimately a flow ID when going into an individual flow.  Will have to work with @lburgazzoli  and @nicolaferraro  to be sure we understand what model changes are involved.

Routing for the editor is defined [here](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/integration.module.ts#L47-L85) and then added to the integration module's routes [here](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/integration.module.ts#L97-L101) and [here](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/integration.module.ts#L106-L110).  I think I'd probably add the flow ID to these existing routes, and put this new page as a parent page kinda, mostly to maintain some state when refreshing (particularly when developing).

I'd probably add a new directory [here](https://github.com/syndesisio/syndesis/tree/master/app/ui/src/app/integration/edit-page) for this new page, this directory is the root of the integration editor.

### LINKED ISSUE #3480: Implement multi-flow integration visualization in integration list view per designs

riccardo-forina:
Closes #3480
Closes #3475
Partially implements #3482
Partially implements #3476

# Todo

* [ ] Santa Hat Operations page's sidebar
* [ ] Make the Save or Add step's sidebar API Provider aware, to make the first node read-only

# Changes

api-provider


app/ui/src/app/integration/integration_detail/integration-detail.component.html
```diff
@@ -1,7 +1,13 @@
 <syndesis-modal [title]="modalTitle" [message]="modalMessage" [type]="modalType" [primaryText]="modalPrimaryText">
 </syndesis-modal>
-<syndesis-loading [loading]="loading">
-  <div *ngIf="integration" class="integration-detail syn-scrollable">
+
+<div *ngIf="loading; else content">
```
gashcrumb: The `syndesis-loading` component should be kinda doing this already, was it not working for this case?

app/ui/src/app/integration/integration_detail/integration-description.component.ts
```diff
@@ -1,10 +1,10 @@
-import { Integration, Step } from '@syndesis/ui/platform';
+import { Flow, Integration, IntegrationType, Step } from '@syndesis/ui/platform';
 import {
   Component,
   Input,
   Output,
   EventEmitter,
-  ViewEncapsulation
+  ViewEncapsulation, OnChanges
```
gashcrumb: I'm surprised lint isn't complaining about this, could just be we've the line length restriction but no lint rules on import formatting perhaps.

app/ui/src/app/integration/integration_detail/integration-detail.component.html
gashcrumb: Gotcha!  Didn't realize that, can you raise a bug against the loading component?  We should try and improve it so that it's template isn't evaluated until the loading input is false, that's the intent anyways.

gashcrumb:
Other than the two newlines in the imports LGTM!

app/ui/src/app/integration/integration_detail/integration-detail.component.html
riccardo-forina: Sure! I tried to use the same logic inside the loading component itself, but as I expected it doesn't solve the issue.

If we were using React, I'd know how to fix this, but with Angular, I'm a bit at a loss.

PR #3592, opened 2018-09-13
fix: update syndesis-loading to accept a TemplateRef too

### LINKED ISSUE #3587: [UI] syndesis-loading children is evaluated regardless of its loading state
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description

Children elements of the `syndesis-loading` component are parsed regardless of the loading state of the component.

A simple example to try it out is the following:

```
// example.component.ts
export class Example {
   alert = alert;
}

// examle.component.html
<syndesis-loading [loading]="true">
  {{ alert("sorry, still evaluated!") }}
</syndesis-loading>
```

## Workaround

The following code does what you'd expect it to do.

```
// example.component.ts
export class Example {
   alert = alert;
}

// examle.component.html
<ng-container *ngIf="true; else content">
  <syndesis-loading loading="true"></syndesis-loading>
</ng-container>

<ng-template #content>
  {{ alert("this will not be called!") }}
</ng-template>
```

gashcrumb:
and update all the code using it.

fixes #3587 now code can pass a template ref to `syndesis-loading` using the `content` input:

```
<syndesis-loading [loading]="whatevs" [content]="ref">
  <ng-template #ref>
    <span>Hello! {{ stuff("world!") }}</span>
  </ng-template>
</syndesis-loading>
```

In this case `stuff()` doesn't get called until `whatevs` is false.

### DIFF
--- a/app/ui/src/app/common/loading/loading.component.html
+++ b/app/ui/src/app/common/loading/loading.component.html
@@ -1,6 +1,11 @@
-<ng-container *ngIf="loading">
+<ng-container *ngIf="loading; else body">
   <div class="syn-loading">
       <div class="spinner spinner-lg loading"></div>
   </div>
 </ng-container>
-<ng-content *ngIf="!loading"></ng-content>
+<ng-template #body>
+  <ng-container *ngIf="content" [ngTemplateOutlet]="content"></ng-container>
+  <ng-container *ngIf="!content">
+    <ng-content></ng-content>
+  </ng-container>
+</ng-template>
--- a/app/ui/src/app/common/loading/loading.component.ts
+++ b/app/ui/src/app/common/loading/loading.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input } from '@angular/core';
+import { Component, Input, TemplateRef } from '@angular/core';
 
 @Component({
   selector: 'syndesis-loading',
@@ -7,4 +7,5 @@ import { Component, Input } from '@angular/core';
 })
 export class LoadingComponent {
   @Input() loading: boolean;
+  @Input() content: TemplateRef<any>;
 }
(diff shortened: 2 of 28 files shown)


app/ui/src/app/common/loading/loading.component.html
```diff
@@ -1,6 +1,11 @@
-<ng-container *ngIf="loading">
+<ng-container *ngIf="loading; else body">
```
riccardo-forina: I think you can get away with just 

```
<ng-container *ngIf="loading; else content">
  <div class="syn-loading">
      <div class="spinner spinner-lg loading"></div>
  </div>
</ng-container>
```

app/ui/src/app/common/loading/loading.component.ts
```diff
@@ -7,4 +7,5 @@ import { Component, Input } from '@angular/core';
 })
 export class LoadingComponent {
   @Input() loading: boolean;
+  @Input() content;
```
riccardo-forina: `@Input() content: TemplateRef<any>;`

app/ui/src/app/common/loading/loading.component.html
gashcrumb: I think I could but I kinda wanted to keep support for *not* using an *ng-template* still also.

app/ui/src/app/common/loading/loading.component.ts
gashcrumb: Yep, good call

app/ui/src/app/common/loading/loading.component.html
riccardo-forina: Oh, I didn't think about that, you're right!

gashcrumb:
Updated, but we can hold off merging until *after* the sprint demo just to be on the safe side :-)

PR #3593, opened 2018-09-14
chore: turn on Typescript's noUnusedLocals

riccardo-forina:
I noticed quite a bit of unused code around the app. I enabled the relative tsconfig option to get an error when this happens, and fixed what I found.

### DIFF
--- a/app/ui/src/app/connections/common/configuration/configuration.service.ts
+++ b/app/ui/src/app/connections/common/configuration/configuration.service.ts
@@ -45,7 +45,7 @@ export class ConnectionConfigurationService {
         break;
       default:
     }
-    const formModel = this.formFactory.createFormModel(
+    this.formFactory.createFormModel(
       config,
       values,
       controls
--- a/app/ui/src/app/connections/common/configuration/configuration.service.ts
+++ b/app/ui/src/app/connections/common/configuration/configuration.service.ts
@@ -45,11 +45,6 @@ export class ConnectionConfigurationService {
         break;
       default:
     }
-    this.formFactory.createFormModel(
-      config,
-      values,
-      controls
-    );
     return this.formFactory.createFormModel(config, values, controls);
   }
 
--- a/app/ui/src/app/core/providers/integration-support-provider.service.ts
+++ b/app/ui/src/app/core/providers/integration-support-provider.service.ts
@@ -9,8 +9,6 @@ import {
   Integration,
   IntegrationDeployment,
   IntegrationDeployments,
-  IntegrationOverview,
-  IntegrationOverviews,
   IntegrationStatus,
   IntegrationSupportService,
   ApiHttpService,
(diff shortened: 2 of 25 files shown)


app/ui/src/app/connections/common/configuration/configuration.service.ts
```diff
@@ -45,7 +45,7 @@ export class ConnectionConfigurationService {
         break;
       default:
     }
-    const formModel = this.formFactory.createFormModel(
+    this.formFactory.createFormModel(
```
gashcrumb: You can actually delete this block I think, the next line is the same thing but with a return

app/ui/src/app/core/providers/integration-support-provider.service.ts
```diff
@@ -183,6 +183,7 @@ export class IntegrationSupportProviderService extends IntegrationSupportService
       .get();
   }
 
+  // @ts-ignore
```
gashcrumb: Maybe on these can you stick a `TODO` line on?  I think actually we could remove these two functions

app/ui/src/app/platform/types/integration/integration.reducer.ts
```diff
@@ -210,7 +210,7 @@ export function integrationReducer(
     case IntegrationActions.DELETE_INTEGRATION_FAIL: {
       const error = (action as IntegrationActions.IntegrationDeleteFail)
         .payload;
-      const restCollection = state.collection.filter(
+      state.collection.filter(
```
gashcrumb: I think this block could be deleted altogether.

app/ui/src/app/settings/settings-root.component.ts
```diff
@@ -1,11 +1,11 @@
 import { Component } from '@angular/core';
-import { ActivatedRoute, Router } from '@angular/router';
+// import { ActivatedRoute, Router } from '@angular/router';
```
gashcrumb: Should we just delete these lines?

app/ui/src/app/store/connector/connector.service.ts
```diff
@@ -43,7 +43,7 @@ export class ConnectorService extends RESTService<Connector, Connectors> {
         // Try and clear any stale cookies, though we can't touch HttpOnly ones
         document.cookie.split(';').forEach(function(c) {
           if (c.startsWith('cred-')) {
-            const newCookie = c
+            c
```
gashcrumb: I think we can delete this whole split/forEach block starting with the comment on line 43 there

app/ui/src/app/store/integration/integration.service.ts
```diff
@@ -91,6 +91,7 @@ export class IntegrationService extends RESTService<Integration, Integrations> {
   }
 
   // design time function
+  // @ts-ignore
```
gashcrumb: Good call, I like having an example response customizer function :-)

app/ui/src/app/connections/common/configuration/configuration.service.ts
riccardo-forina: 🤦‍♂️ I didn't notice that!

PR #3600, opened 2018-09-17
fix(s2i): Different ordering of maven modules during packaging phase

phantomjinx:
* Both the maven-dependency-plugin and the exec-maven-plugin are set to
  be enacted during the package phase. It is expected that the dependency
  gets executed first since the exec requires it. However, sometimes they
  are reversed by maven and the build fails.

* Shifting the dependency execution to the validate phase ensure the
  sequence occurs in the correct order.

### DIFF
--- a/app/s2i/pom.xml
+++ b/app/s2i/pom.xml
@@ -65,7 +65,7 @@
         <executions>
           <execution>
             <id>copy</id>
-            <phase>package</phase>
+            <phase>validate</phase>
             <goals>
               <goal>copy</goal>
             </goals>


lburgazzoli:
@zregvart I think you are more knowledgeable about all the maven things we have set-up so do you mind having a look ?

zregvart:
I wonder when it's so, the order of the execution should be the order in which the plugins are listed in the POM. Also putting stuff in `validate` is a heavyweight approach to ordering, it gets executed far too frequently that way.

Is this a problem when only `exec:exec` goal is invoked?

phantomjinx:
Well, another day and another reboot of the laptop. Everything works fine with no requirement for this PR. Executed 10 or so builds with everything in the s2i module building without a hitch. So, will close this.

Sorry for the noise!

@lburgazzoli @zregvart

zregvart:
@phantomjinx glad it started working for you :fireworks:

PR #3604, opened 2018-09-17
feat: #3568 Migrate from nginx to a simple go static asset server: Simble

### LINKED ISSUE #3568: Migrate from nginx to some simple go static asset server
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description

image

chirino:
fixes #3568


gashcrumb:
Awesome!  Will give this a test this afternoon...

gashcrumb:
Hmm, when the operator stuff was added to the build I didn't find I needed to explicitly install `go`, but on this PR it seems like I need to as the build fails at the UI module with an error because `go` can't be found.  Also the CI build fails as well.  @rhuss how did the addition of the operator tackle fetching `go` on the fly and is it something we could re-use here?

gashcrumb:
Hmm, different error actually on CI:

```
java.io.IOException: Cannot run program "make" (in directory "/workspace/app/ui/cmd"): error=2, No such file or directory
```

lemme just recreate my error and paste it here too to be on the safe side.

gashcrumb:
Yep, so get:

```
[INFO] --- exec-maven-plugin:1.6.0:exec (default) @ ui ---
go get github.com/go-bindata/go-bindata/go-bindata@v1.0.0
make: go: Command not found
Makefile:9: recipe for target '../target/go/go-bindata' failed
make: *** [../target/go/go-bindata] Error 127
[ERROR] Command execution failed.
```

rhuss:
> @rhuss how did the addition of the operator tackle fetching `go` on the fly and is it something we could re-use here?

This was a tricky beast which is resolved now but caused a lot of frustration.

The build is done via a go compiler from a docker image, when using a remote Docker daemon, then an extra rsync step is required. I think we should integrate this (as well as maybe the whole UI build) in the "syndesis" script, too. There an issue open, but don't know it by heart right now.

Apologises for my ignorance but what actually is the reason for writing an own HTTP server binary instead of relying on an existing one ? What is the concrete issue with nginx ?

gashcrumb:
> Apologises for my ignorance but what actually is the reason for writing an own HTTP server binary instead of relying on an existing one ? What is the concrete issue with nginx ?

https://github.com/syndesisio/syndesis/issues/3568#issuecomment-419886469 ?

gashcrumb:
> This was a tricky beast which is resolved now but caused a lot of frustration.

Would it be easier to make `go` and `make` prerequisites for building syndesis I wonder?

app/ui/cmd/Makefile
```diff
@@ -0,0 +1,25 @@
...
+#BUILD_FLAGS=-gcflags "all=-N -l"
+
+all: linux
+
+../target/go/go-bindata:
+	go get github.com/go-bindata/go-bindata/go-bindata@v1.0.0
+	go build -o $@ github.com/go-bindata/go-bindata/go-bindata
+
+bindata_assetfs.go: ../target/go/go-bindata $(ASSET_FILES)
+	../target/go/go-bindata -prefix ../dist -o $@ ../dist/...
```
rhuss: Why you won't do put the static assets into the binary ? Would the standalone binary ever be used outside of a container, as its hard to adapt and to develop against it ?

As there is also data served from `/html` wouldn't it be easier and more straightforward to server everything from /html ?

app/ui/docker/Dockerfile
```diff
@@ -1,60 +1,27 @@
-FROM docker.io/centos:7
-
-ENV NGINX_VERSION 1.13.4-1.el7
+FROM scratch
```
rhuss: really ? I always find it super hard to debug, but you might have other experiences here.

app/ui/cmd/go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/syndesisio/syndesis
```
rhuss: Could we agree to stay with go 1.10 and eg dep as long the go 1.11 module support is marked as experimental ? Happy to switch if the module support has been proven to be stable in the wild, though, as it will be the future of dependency management. Also it will free us from staying in the GOPATH (this issue is solved in "syndesis" for go 1.10 a bit differently)

chirino:
Yeah. I could convert this to using a Docker Builder.

app/ui/docker/Dockerfile
chirino: I find debugging go binaries hard in a container even with a centos base layer. By avoiding the base layer you avoid having to rebuild due to base layer CVEs.

app/ui/cmd/go.mod
chirino: I’d rather stick with 1.11. Modules are just easier to develop with.

app/ui/cmd/Makefile
chirino: I could make that change easily.  Was thinking that by embedding them into the binary, might be less at risk of something not being put into the docker image properly during productization.

rhuss: Actually, I think it would be easier to avoid adding the assets within the binary for the following reasons:

* Less heap memory usage (not really sure how large the assets are)
* We could use a generic image which is built outside the UI build, freeing UI developers from running a go compiler each time
* Savin an additional pre-processor step saves some build times.

FYI for the operator, we also added the OpenShift template with gobin first, too but soon moved it out in the filesystem again. Also that the assets are more accessible in case some debug is required.

app/ui/docker/Dockerfile
rhuss: I'm fine with this but we should double check with Darren that scratch images are explicitly allowed as I remember darkly that there were some restrictions (but can be wrong here). What was Darrens GitHub handle again ?

app/ui/cmd/go.mod
rhuss: Ok, but then we should upgrade the operator also to 1.11 and use the same builder image.

rhuss:
The PR is really good, but there is one important piece missing: A proper integration into "syndesis build", not only indirectly via Maven calling "make".

My suggestion is to follow the same lines as for the operator build:

* If the assets are not included in the go binary (which I would recommend, see above), then we can split this up into two steps:
 - One build of the base image in a dedicated step
 - Adding the assets in a second layer

This would speed up the build time (as the base image needs not to be build everytime as I suspect it changes quite infrequently) and we could avoid the requirement on "make".

* The go build of the binary should follow the same lines as the operator build and support three ways how to build the go binary
  * Directly with a log go environment when the --local option is given
  * Indirectly via a gobuilder image that we define at https://github.com/syndesisio/syndesis/tree/master/tools/images/godev When we switch the operator to use 1.11 modules, then we can get rid of this (intermediate) builder image and use an official golang image directly.
  * When the Docker daemon is not a local one so that volume mounts of the source code into the builder image is not possible, and the remote Docker daemon is from minishift (eval $(minishift docker-env)), then the source needs to be rcp-ed into the remote VM before the builder image is started and the resulting binary needs to be copied back again.

Astonishingly the latest option seems to be the most popular one amongst backend developers. 
Unfortunately, Minishift's Docker daemon is too old and does not support multi-stage builds so this rcp copy trick is required.

All of this code can be found in https://github.com/syndesisio/syndesis/blob/master/tools/bin/commands/build and https://github.com/syndesisio/syndesis/blob/master/tools/bin/commands/util/operator_funcs

IMO its really important if we could keep a consistent build experience across building go based apps and that we don't increase the build requirements (local go env) if not necessary (it's not).

gashcrumb:
Awesome!  Build completed successfully for me but the UI pod is failing to start.  Output is:

```
standard_init_linux.go:178: exec user process caused "exec format error"
```

Building the go part doesn't look wrong as far as I can see:

```
[INFO] --- exec-maven-plugin:1.6.0:exec (default) @ ui ---
===========================================
Running go build via a docker build
===========================================
Sending build context to Docker daemon  8.687MB
Step 1/4 : FROM golang:1.11.0
Trying to pull repository docker.io/library/golang ...
1.11.0: Pulling from docker.io/library/golang
05d1a5232b46: Pulling fs layer
5cee356eda6b: Pulling fs layer
89d3385f0fd3: Pulling fs layer
80ae6b477848: Pulling fs layer
94ebfeaaddf3: Pulling fs layer
dd697213ec59: Pulling fs layer
715d1281dfe7: Pulling fs layer
80ae6b477848: Waiting
dd697213ec59: Waiting
715d1281dfe7: Waiting
94ebfeaaddf3: Waiting
89d3385f0fd3: Verifying Checksum
89d3385f0fd3: Download complete
5cee356eda6b: Verifying Checksum
5cee356eda6b: Download complete
05d1a5232b46: Verifying Checksum
05d1a5232b46: Download complete
80ae6b477848: Verifying Checksum
80ae6b477848: Download complete
715d1281dfe7: Verifying Checksum
715d1281dfe7: Download complete
94ebfeaaddf3: Verifying Checksum
94ebfeaaddf3: Download complete
dd697213ec59: Verifying Checksum
dd697213ec59: Download complete
05d1a5232b46: Pull complete
5cee356eda6b: Pull complete
89d3385f0fd3: Pull complete
80ae6b477848: Pull complete
94ebfeaaddf3: Pull complete
dd697213ec59: Pull complete
715d1281dfe7: Pull complete
Digest: sha256:e8e4c4406217b415c506815d38e3f8ac6e05d0121b19f686c5af7eaadf96f081
Status: Downloaded newer image for docker.io/golang:1.11.0
 ---> fb7a47d8605b
Step 2/4 : WORKDIR /src
 ---> 353e4a31922a
Removing intermediate container 855d40c5c519
Step 3/4 : COPY . /src
 ---> bb145a889294
Removing intermediate container fbb369b0079d
Step 4/4 : RUN CGO_ENABLED=0 go build -mod=vendor -o /syndesis-ui *.go
 ---> Running in 03185363d305

 ---> b7ef021962c4
Removing intermediate container 03185363d305
Successfully built b7ef021962c4
===========================================
Extracting go binary from the Docker image
===========================================
base64: invalid option -- 'D'
Try 'base64 --help' for more information.
[INFO]
```

To be on the safe side lemme test again with a full re-install on this branch.

app/ui/cmd/build.sh
```diff
@@ -0,0 +1,13 @@
...
+echo ===========================================
+echo Running go build via a docker build
+echo ===========================================
+docker build -t syndesis-ui-builder .
+mkdir -p ../target/go/linux-amd64/
+
+echo ===========================================
+echo Extracting go binary from the Docker image
+echo ===========================================
+docker run -t syndesis-ui-builder bash -c "cat /syndesis-ui | base64" | base64 -D >  ../target/go/linux-amd64/syndesis-ui
```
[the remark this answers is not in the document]
gashcrumb: > Building the go part doesn't look wrong as far as I can see

I take that back :-)

rhuss:
When running "syndesis build --image --module ui" the image is build twice, once via maven exec:exec and once via OpenShift s2i build. I would expect only one Docker build.

When called "syndesis build --module ui" no Docker image is created which is as expected.

app/ui/cmd/build.sh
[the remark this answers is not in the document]
chirino: Doh. Back to the drawing board.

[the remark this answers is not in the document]
chirino: Doh. Back to the drawing board.

app/ui/cmd/go.mod
heiko-braun: However we approach this, please use one baseline for docker based projects that need to productised. This means, same docker version, dependency management, base images, etc

chirino:
Rebased.  This PR now serves up the static content via https to avoid MiM attacks.

chirino:
@rhuss could you re-review, the vendor dir is no longer checked in.

gashcrumb:
cool, will give it a retest here too...

gashcrumb:
Awesome, UI pod seems to start alright now for me but it doesn't seem like the config.json file is being picked up properly.  Looking at the pod does this seem right?

image

is it meant to reference that nginx path still?  Anyways, gonna try one more rebuild just to be on the safe side.

gashcrumb:
Hmm, if I do `syndesis minishift --full-reset --install --local` the UI pod doesn't seem to come up but this bit in the deployment seems better than the last screenshot:

image

chirino:
@gashcrumb I think you just need to build the ui module now, so that the image stream gets updated with a local copy.

rhuss:
@chirino have you implemented `syndesis --local` for the ui build to use a local go compiler (like for the operator) ? If not we could implement it later.

Have you checked https://github.com/syndesisio/syndesis/pull/3604#issuecomment-422405894 for the double build of the image ?

If this is fixed (have not much more time to review, sorry) its fine for me technically wise. 

Not sure whether it should be merged as long as the mail discussion is ongoing because of the more non-technical, long term issues. wdyt, @heiko-braun ?

rhuss:
I mean, whether we should wait until we agree on the common direction.

gashcrumb:
@chirino Yep, that did the trick, app worked fine, thanks!

I did `yarn start:minishift` though to test out that workflow as well and just get a 504 gateway timeout page after refreshing.  Here's the output of the `minishift-setup.sh` script:

```
] yarn start:minishift
yarn run v1.9.4
$ ./scripts/minishift-setup.sh && yarn ng serve --host 0.0.0.0 --disable-host-check --public-host $(oc get routes syndesis --template '{{.spec.host}}'); ./scripts/minishift-restore.sh
+(./scripts/minishift-setup.sh:6): main(): set -euo pipefail
++(./scripts/minishift-setup.sh:7): main(): dirname ./scripts/minishift-setup.sh
+(./scripts/minishift-setup.sh:7): main(): cd ./scripts
+(./scripts/minishift-setup.sh:8): main(): cd ..
++(./scripts/minishift-setup.sh:11): main(): ./node_modules/.bin/my-local-ip
+(./scripts/minishift-setup.sh:11): main(): LOCAL_IP=10.0.0.86
+(./scripts/minishift-setup.sh:12): main(): echo 'Local IP is: 10.0.0.86'
Local IP is: 10.0.0.86
+(./scripts/minishift-setup.sh:15): main(): mv -f src/config/config.json src/config/config.json.bak
+(./scripts/minishift-setup.sh:16): main(): cp src/config/config.json.minishift src/config/config.json
++(./scripts/minishift-setup.sh:18): main(): oc status
++(./scripts/minishift-setup.sh:18): main(): head -n 1
++(./scripts/minishift-setup.sh:18): main(): sed 's/.*https/https/'
+(./scripts/minishift-setup.sh:18): main(): OS_HOST=https://192.168.42.124:8443
++(./scripts/minishift-setup.sh:20): main(): oc get route syndesis '--template={{.spec.host}}'
+(./scripts/minishift-setup.sh:20): main(): sed -i.bu s#syndesis.192.168.64.2.nip.io#syndesis.192.168.42.124.nip.io# src/config/config.json
+(./scripts/minishift-setup.sh:21): main(): sed -i.bu 's#https://192\.168\.64\.2:8443#https://192.168.42.124:8443#' src/config/config.json
+(./scripts/minishift-setup.sh:22): main(): sed -i.bu 's/Syndesis/Syndesis - DEVELOPMENT/' src/config/config.json
+(./scripts/minishift-setup.sh:23): main(): rm src/config/config.json.bu
+(./scripts/minishift-setup.sh:25): main(): oc replace --force -f -
service "syndesis-ui" deleted
service "syndesis-ui" replaced
endpoints "syndesis-ui" replaced
$ /home/gashcrumb/Source/syndesis/app/ui/node_modules/.bin/ng serve --host 0.0.0.0 --disable-host-check --public-host syndesis.192.168.42.124.nip.io
WARNING: Running a server with --disable-host-check is a security risk. See https://medium.com/webpack/webpack-dev-server-middleware-security-issues-1489d950874a for more information.
** Angular Live Development Server is listening on 0.0.0.0:4200, open your browser on http://localhost:4200/ **
```

The endpoint looks like I'd expect, and the above looks okay:

image

But seems there's a disconnect somewhere.

Also `yarn restore:minishift` doesn't seem to work any longer on this branch:

```
] yarn restore:minishift
yarn run v1.9.4
$ ./scripts/minishift-restore.sh
syndesis-ui current at revision 4
Currently running syndesis-ui pod: syndesis-ui-4-887cj
syndesis-ui-4-deploy
service "syndesis-ui" deleted
service "syndesis-ui" replaced
error: json: line 20: invalid character '\n' in string literal
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Which means I can't get at the UI served out from minishift anymore.

> I mean, whether we should wait until we agree on the common direction.

@rhuss for sure, also want to be sure we test this to ensure there isn't an impact on our workflow, which I'd like to ensure isn't impacted when this PR as merged.  Like for example right now we wouldn't be able to use the development server looks like, which would definitely be an impact :-)

chirino:
@gashcrumb that’s likely due to the recent switch to https. Let me see what I can figure out for that.

chirino:
Guys thank for the reviews.. Going the close this PR down for now since it was decided to hold of on switching to simble until it get more mileage being used by other projects.

rhuss:
@chirino Thanks a lot for all your efforts ! This PR contains a bunch of very good ideas, which we should consider to take over to our build (like using COPY and stdout for compiling with a builder image without bind mounts).

I also opened an issue for the SSL certs, which we still need to configure for all of our services (not only syndesis-ui).

PR #3605, opened 2018-09-17
Feature: Template step

phantomjinx:
Note: This has 2 explicit commits where 1 addresses the changes for the backend while the other deals with the UI.
@lburgazzoli: if you would mind having a look at the backend or defer to whoever you think best
@gashcrumb likewise, if you could take the UI elements or defer to whoever you think best

Many thanks.


app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/TemplateStepHandler.java
```diff
@@ -0,0 +1,225 @@
...
+        //
+        try {
+            String openDelimiter = URLEncoder.encode(MUSTACHE_OPEN_DELIMITER, "UTF-8");
+            String closeDelimiter = URLEncoder.encode(MUSTACHE_CLOSE_DELIMITER, "UTF-8");
+
+            StringBuilder uri = new StringBuilder()
+                       .append("mustache")
+                       .append(COLON).append("dummy")
+                       .append(QUESTION_MARK).append("startDelimiter").append(EQUALS).append(openDelimiter)
+                       .append(AMPERSAND).append("endDelimiter").append(EQUALS).append(closeDelimiter);
```
lburgazzoli: This may be problematic as if you add two templates in the same context, they will class, you should name the endpoint after the flow and step index.

I guess you can also use Camel's URISupport to create an uri.

app/integration/runtime/src/test/java/io/syndesis/integration/runtime/IntegrationTestSupport.java
```diff
@@ -19,21 +19,37 @@
...
+import io.syndesis.common.model.integration.Step;
+import io.syndesis.common.util.StringConstants;
 
-public class IntegrationTestSupport {
+public class IntegrationTestSupport implements StringConstants {
     private static final Logger LOGGER = LoggerFactory.getLogger(IntegrationTestSupport.class);
 
+    protected ObjectMapper mapper = new ObjectMapper();
+
+    protected static class MsgSnippet {
```
lburgazzoli: Not very sure to understand the need of this msg snippet thing, isn't a map and/or Jackson enough ?

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/TemplateStepHandler.java
phantomjinx: @lburgazzoli 
Yeah, I'll update the "dummy" literal to something unique as you suggest.
These template camel modules do support an URI but are limited to looking for the source of the URI as a file on the classpath (unless its an absolute file). Seeing as I already have the content and they support placing it in the header, it's become more straightforward to leave it that way.

app/integration/runtime/src/test/java/io/syndesis/integration/runtime/IntegrationTestSupport.java
phantomjinx: A MsgSnippet is simply a key/value pair object rather than a map so more akin to a Map.Entry. In the tests it allows to generate snippets and adds snippets as the value of snippets (so recursively). Basically, just reducing complexity and verbosity of test setup (started with maps and it was getting rather cluttered).

app/ui/src/app/integration/edit-page/step-configure/templater/templater.component.ts
```diff
@@ -0,0 +1,358 @@
...
+    };
+  }
+
+  private initCodeMirror() {
+    /**
+     * Defines an overlay mode for mustache,
+     * which does not yet have its own mode
+     * defined in CodeMirror
+     */
+    CodeMirror.defineMode('mustache', function(config, parserConfig) {
```
gashcrumb: Is this just a one time thing like at app bootup?  Wonder if it'd be better doing this initialization in some service in the vendor module instead.  Unfortunately none of that is set up currently, but maybe it's worth identifying this as a potential refactoring task (i.e. just add a TODO comment maybe?) when we need to add more non-standard highlighting modes etc.

gashcrumb:
+1 from me on the UI code, LGTM.  I won't mark it approved yet as it'll get merged possibly before you've handled Luca's feedback.

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/TemplateStepHandler.java
lburgazzoli: What I mean here is that instead of using StringBuilder, you can leverage Camel's `URISupport.appendParametersToURI(String, Map)` to generate the final uri. Not a blocking issue of course :)

app/integration/runtime/src/test/java/io/syndesis/integration/runtime/IntegrationTestSupport.java
lburgazzoli: Os so I'd try to find a better name as `message` has a very well defined meaning in Camel and it may confuse if used with different meaning (same for the utility methods like `jsonMessage(MsgSnippet...)` and `message(MsgSnippet...)`

app/integration/runtime/src/test/java/io/syndesis/integration/runtime/handlers/TemplateStepHandlerTest.java
```diff
@@ -0,0 +1,363 @@
...
+                                                             ),
+                                                             snippet("text", "Will be going to the University of Southampton.")));
+
+        testMessages.add(message(snippet("name", "Susan"),
+                                                             snippet("course",
+                                                                   snippet("name", "Physics"),
+                                                                   snippet("name", "German"),
+                                                                   snippet("name", "Art")
+                                                             ),
+                                                             snippet("text", "Will be joining Evans, Richards and Dean.")));
```
lburgazzoli: can we reduce a little this indentation ? it makes it hard to review as the line goes beyond github limits. Of course not a blocking issue :)

oscerd: +1

app/integration/runtime/pom.xml
```diff
@@ -84,6 +84,10 @@
       <groupId>org.apache.camel</groupId>
       <artifactId>camel-spring</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.camel</groupId>
+      <artifactId>camel-mustache</artifactId>
+    </dependency>
```
lburgazzoli: it is fine for now but we should include this dependency only if the integration has a template step. We may need to do the same for atlasmap.

app/integration/runtime/src/test/java/io/syndesis/integration/runtime/IntegrationTestSupport.java
phantomjinx: ok... renamed.

app/ui/src/app/integration/edit-page/step-configure/templater/templater.component.ts
phantomjinx: Sure. Will need to revisit when adding support for velocity and framemaker templates. So leaving a TODO note.

app/integration/runtime/pom.xml
phantomjinx: Absolutely. So I left this in simply for 7.2. After that, we are intending to add both velocity and framemaker support in these template. At that point, the library dependency would be dynamic based upon the user's choice of syntax.

phantomjinx:
@lburgazzoli @gashcrumb 
Ok. Updated the PR with the latest changes so if you wouldn't mind taking another look or approving. Thx.

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/TemplateStepHandler.java
```diff
@@ -0,0 +1,225 @@
...
+        // Apply the template to the mustache header property
+        // Then add to the route path
+        //
+        route.setHeader(MustacheConstants.MUSTACHE_TEMPLATE).constant(template);
+
+        //
+        // Encode the delimiters since they are applied as URI query parameters
+        //
+        try {
+            String id = step.getId().orElse(UUID.randomUUID().toString()) + UNDERSCORE + UNDERSCORE + stepIndex;
```
lburgazzoli: better not to use the step id as it may contain not allowed chars, a simple strategy would be to use something like:

    step-${flowIndex}-${stepIndex}

lburgazzoli:
@phantomjinx looks good so far, only a minor thing about the the way the endpoint uri is computed.

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/TemplateStepHandler.java
phantomjinx: done.

lburgazzoli:
@phantomjinx LGTM

zregvart:
@phantomjinx one unused import raises a PMD issue that [fails the build](https://circleci.com/gh/syndesisio/syndesis/43239):

```
[INFO] --- maven-pmd-plugin:3.8:check (default) @ integration-runtime ---
[INFO] PMD Failure: io.syndesis.integration.runtime.handlers.TemplateStepHandler:24 Rule:UnusedImports Priority:4 Avoid unused imports such as 'java.util.UUID'.
```

phantomjinx:
@zregvart 
"Once more unto the breach dear friend..."

phantomjinx:
come again?
circleci fails due to lack of npm??

```
[INFO] [4/5] Rebuilding all packages...
[INFO] [5/5] Cleaning modules...
[INFO] success Saved lockfile.
[INFO] $ npm run snyk-protect; yarn snyk-protect
[ERROR] /bin/sh: 1: npm: not found
[INFO] yarn run v1.2.1
[INFO] $ snyk protect
[INFO] An unknown error occurred. Please include the trace below when reporting to Snyk:
[INFO] 
[INFO] Error: 500
[INFO]     at /workspace/app/ui/node_modules/snyk/lib/snyk-test/npm/index.js:127:23
[INFO]     at /workspace/app/ui/node_modules/snyk/lib/request/index.js:12:7
[INFO]     at <anonymous>
[INFO]     at process._tickCallback (internal/process/next_tick.js:188:7)
[INFO] 
[ERROR] error Command failed with exit code 1.
[INFO] info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
[ERROR] error Command failed with exit code 1.
[INFO] info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
[INFO] ------------------------------------------------------------------------
```

@zregvart 
Can you re-trigger please?

Issue #3454, opened 2018-08-24
Google Calendar Connector


gaughan:
Google Calendar
As a citizen Integrator I want to be able to use the google calendar Camel component http://camel.apache.org/googlecalendar.html in Fuse Online:

I would like to be able to use the same authentication details if I have already provided for google. 
I would like to be able to configure both:
**start connector** to accept events from my google calendar 
**finish connector** to insert or move event in a calendar

To accept Google Calendar **Poll** events for a start connector, I will configure a google watch url for my calendar https://www.googleapis.com/apiName/apiVersion/resourcePath/watch
I expect  a post message to be sent to my url conforming to google api standard (including the syndesis webhook the event can be sent to) 
I want to monitor a channel (calendar) for new events created and consume these via my fuse online integration.

DoD

- I can create events in google calendar
- I can subscribe to google calendar resources and in turn be notified if an event is created or changed.
- The feature owner has confirmed one of the following:
--There is an initial draft of the user doc for this feature. 
--The writer acknowledges receiving enough information from engineering to write the initial draft of the user doc for this feature.

oscerd:
I believe using Push events is overkilling since we already have a Java API for this https://developers.google.com/calendar/v3/reference/events/watch and we can use the Events Watch 
https://developers.google.com/calendar/quickstart/java

Let me know what do you think, the result should be the same.

oscerd:
Using push notification will be a problem. We'll need to have a domain where we can send the notifications, with a SSL certificate and we'll need to demonstrate we are the owner of that domain, by giving informations or adding html tag in our code (different for each domain). So I don't think is feasible to use push notification at this point, my suggestion is to list the event and filter them by update time. @paoloantinori @gaughan @heiko-braun @kcbabo

oscerd:
More information here: https://developers.google.com/calendar/v3/push

My suggestion is to use a polling approach.

We can think about a general solution, but this would be part of a different issue, google-calendar should avoid using push notifications.

oscerd:
I'll wait for your feedback before I start working on this.

heiko-braun:
I think polling is fine if users have some degree of control how often we poll

paoloantinori:
My suggestion would be to continue with a polling based logic and to start setting up the discussion for an eventually support at Syndesis level per Push based apis.

A quick conversation with engineering pointed out that there are some hard requirements, dictated by the remote service provider, to comply with the service terms.
[...]

PR #3621, opened 2018-09-19
feat(connectors): New Google-Calendar connector

oscerd:
Related to #3454 

The connector works fine, we still need to add additional options to the create and update operation, but the basic stuff is done.

I noticed that Atlasmap repors warning while parsing the Event model class from Google API, this is something to check for @igarashitm , a part the warnings and some options the connector is done.

### DIFF
--- a/app/connector/google-calendar/pom.xml
+++ b/app/connector/google-calendar/pom.xml
@@ -1,0 +1,190 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Copyright (C) 2016 Red Hat, Inc.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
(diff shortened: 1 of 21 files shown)


oscerd:
Still have to fix a couple of tests. Updating.

zregvart:
I'm cc-ing @TovaCohen to have a look on the text in [Connector JSON](https://github.com/syndesisio/syndesis/blob/760096f17e5459eb5a5361f0c82bb7ce5943b6d0/app/connector/google-calendar/src/main/resources/META-INF/syndesis/connector/calendar.json) and to be aware that this needs to be documented.
There's a user story on the Google Calendar epic in #3454.

oscerd:
Sure, this is just the first version, but there is still place to have it improved and with different option.

TovaCohen:
Thanks Zoran! I'll take a look.

PR #3628, opened 2018-09-19
ENTESB-9436: Added an artifact for the s2i m2 repository zip

djcoleman:
A zip of the s2i m2 directory is required by the prod build process as the maven and docker builds are done in different build systems, so an artifact can be downloaded from the internal prod maven repository and extracted into the container image.

### DIFF
--- a/app/s2i/pom.xml
+++ b/app/s2i/pom.xml
@@ -65,7 +65,7 @@
         <executions>
           <execution>
             <id>copy</id>
-            <phase>package</phase>
+            <phase>prepare-package</phase>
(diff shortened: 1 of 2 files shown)


app/s2i/pom.xml
```diff
@@ -65,7 +65,7 @@
         <executions>
           <execution>
             <id>copy</id>
-            <phase>package</phase>
+            <phase>prepare-package</phase>
```
zregvart: cc @phantomjinx :smile_cat:

PR #3631, opened 2018-09-20
Fix #3460: co-locate upgrade pod with database and add service account

### LINKED ISSUE #3460: 'Unable to mount volumes for pod "syndesis-upgrade-1.4.7...' during upgrade from 1.4.6 to 1.4.7 on OSO pro
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem
I'm unable to upgrade syndesis on openshift pro:

3:31:18 PM | Warning | Failed Mount | Unable to mount volumes for pod "syndesis-upgrade-1.4.7_upgrade-project(ee1e559b-a9fb-11e8-937e-125b034d2f46)": timeout expired waiting for volumes to attach/mount for pod "upgrade-project"/"syndesis-upgrade-1.4.7". list of unattached/unmounted volumes=[backup-dir]5 times in the last 10 minutes
-- | -- | -- | --
3:20:13 PM | Normal | Successful Mount Volume | MountVolume.SetUp succeeded for volume "syndesis-operator-token-jwf4x"
3:20:13 PM | Warning | Failed Attach Volume | Multi-Attach error for volume "pvc-655fe43a-a9fb-11e8-ba69-12b5519f9b58" Volume is already used by pod(s) syndesis-db-1-2m8n5
3:20:13 PM | Normal | Scheduled | Successfully assigned syndesis-upgrade-1.4.7 to ip-172-31-58-128.ec2.internal

screen shot 2018-08-27 at 15 26 39

1. deploy syndesis 1.4.6 on OSO Pro using template
2. manually create operator service account 
3. try to upgrade to 1.4.7 via template ("old way")

nicolaferraro:
This is to co-locate the upgrade pod with the database so that the upgrade process can mount the same (RWO) volume to perform backup in a real cluster.

@rhuss can you have a look? If it is ok, I'll apply it also to 1.4.x and fuse-online-install.
Discussion in #3460.

### DIFF
--- a/install/syndesis.yml
+++ b/install/syndesis.yml
@@ -1634,6 +1634,14 @@ objects:
     metadata:
       name: syndesis-upgrade-latest
     spec:
+      affinity:
+        podAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchLabels:
+                "syndesis.io/app": "syndesis"
+                "syndesis.io/component": "syndesis-db"
+            topologyKey: "kubernetes.io/hostname"
       serviceAccountName: syndesis-operator
       containers:
       - name: upgrade


install/syndesis.yml
```diff
@@ -1634,6 +1634,14 @@ objects:
...
       name: syndesis-upgrade-latest
     spec:
+      affinity:
+        podAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchLabels:
+                "syndesis.io/app": "syndesis"
+                "syndesis.io/component": "syndesis-db"
+            topologyKey: "kubernetes.io/hostname"
```
rhuss: Hmm, don't really understand the documentation for this field. Do you have any idea what the topologyKeu is for ?

nicolaferraro: The whole thing means: find a pod with labels matching the ones I provide, then make sure this new pod has the same topology key as the one you've found (if any), i.e. "kubernetes.io/hostname" means same node. See the "redis" example: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#always-co-located-in-the-same-node

Btw I'm going to spin a new PVC for the upgrade pod :)

nicolaferraro:
Going to create a new PV for the upgrade pod.

install/syndesis.yml
rhuss: ok. sounds good. But we have not to forget to negotiate this with the OSO team to increase the PV quota for Fuse Online.

Issue #2880, opened 2018-06-25
Publishing integration fails as k8s secret cannot be created


zregvart:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

Creating a k8s secret can fail if the secret is larger in file size than 1MB. This can happen if the `specification` property of the API connector containing the Swagger specification is oversized. This can be noticed when using the new Concur connector.

## Expected behavior

Secret can be created, integration is published.

PR #3633, opened 2018-09-20
[fix] Query parameters are not set in the API-Connector #3564

zregvart:
(cherry picked from commit 61ab88570c43e296b06c8fe4ecc39bedb51a7133)

@KurtStam I saw that the PR #3572 had `pr/needs-backport` label but it was subsequently removed. I think we should backport this fix, but I could be missing some discussion around this and we could have decided not to backport the fix.

cc @paoloantinori @heiko-braun


heiko-braun:
Thanks @zregvart . We had discussed it and decided to keep the changes in 7.1.1 to a minimum. I'll loop in @gaughan for feedback. 

Gary, do you think https://github.com/syndesisio/syndesis/issues/3564 should be included in 7.1.1?

zregvart:
Without this integrations using custom API client connectors can't use HTTP query parameters.

They get mapped and received, but the minimization of the Swagger specification removes the query parameters. So the `camel-rest-swagger` Camel component doesn't propagate the query parameter to the HTTP request as the parameter is not found in the Swagger specification that the component is configured with.

And further background on why we minimize the Swagger specification can be found in #2880.

gaughan:
@heiko-braun Agree on including in 7.1.1

heiko-braun:
@zregvart See above. We agreed to include it in 7.1.1

PR #3657, opened 2018-09-24
Carry over 1.4.x docs from master

heiko-braun:


heiko-braun:
@TovaCohen Can you double check if this covers it?

TovaCohen:
This looks good to me. I assume this captures every commit to the doc directory starting after the branch on August 6th and until early September. I see that it does not, as it should not, carry over Paul's commit of doc/integrating_applications/topics/add_template_step.adoc. So it looks like this captures everything and only what it should. Thanks!

PR #3668, opened 2018-09-25
Long term OAuth support for GMail

### LINKED ISSUE #3608: gmail connection is limited by expiration of oauth token
Creating a gmail connection requires adding an OAuth Client ID to the gmail API and registering syndesis accordingly (documented in the user guide). Publishing an integration involving gmail works as expected. However, subsequent republishings of the same integration or using the connection for other integrations frequently results in the published integration failing due to '401 unauthorized errors'.

The workaround is to remove the OAuth Client ID in the gmail API, recreate it and re-register the integration. The gmail documentation actually [states](https://developers.google.com/gmail/api/auth/web-server):

```This access token is passed to the Gmail API to grant your application access to user data for a limited time.```

Is it possible that the token is quickly expiring and requires refreshing? If so, Is it possible to refresh the token automatically or lengthen the expiry time?

zregvart:
This adds:
 - support for additional query parameters in the (initial) authorization URL, so we can specify `access_type=offline` needed for Google OAuth to issue a refresh token
 - adds the `access_type=offline` to GMail connector metadata
 - modifies the OAuth token URL to the newer version

Issues:
 - if I understand correctly this does not require db migration, connectors are overwritten on startup (correct?)
 - will not work with existing Google API project, I needed to create new project and new credentials in order to get it to work; alternatively perhaps going to https://myaccount.google.com/permissions and revoking existing permission might work (didn't for me)
 - the end result should be that the created connection has a `refreshToken` property, and that the Camel GMail component is configured with it (did not test that part)

Fixes #3608


oscerd:
LGTM. I'll test tomorrow and report on the issue. Thanks for your work @zregvart

oscerd:
If it will work I'll add the same option to google-calendar connector too.

oscerd:
[ERROR] /workspace/app/server/runtime/src/test/java/io/syndesis/server/runtime/credential/TestCredentialProviderFactory.java:[53,55] cannot infer type arguments for io.syndesis.server.credential.OAuth2CredentialProvider<>
  reason: cannot infer type-variable(s) S

PR #3678, opened 2018-09-26
fix(server): don't consider hidden OAuth properties

### LINKED ISSUE #3674: gmail and google calender oauth form config missing defaultValue parameters
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Both the gmail and google calendar connectors have oauth form config JSON that have "hidden" fields but don't provide values for those fields, for example [see here](https://github.com/syndesisio/syndesis/issues/3672#issuecomment-424743192).

Ideally it should look like this:

image

when `type` is set to `hidden`, the UI will post these values then when saving the oauth config.

Otherwise I'd say just remove these form fields and set them at the backend if there's no reason to expose them to the user.

zregvart:
We should not send `type=hidden` properties to the UI for the OAuth
setup, it makes little sense as the UI will not be able to anything
meaningful with them.

Fixes #3674


gashcrumb:
That works too :-)

zregvart:
[gifbot:robot]

PR #3679, opened 2018-09-26
[CI] Automatic pull request opened by the Fuse 7 dev pipeline to synchronize versions

cunningt:
@janstey @zregvart

I'd like to start to automatically open PRs into syndesis master after a Fuse 7 pipeline build.    Here's the initial run, I think there are some issues here that we probably need to be squared away  ...

- we're building jboss-fuse/f-m-p 3.5.x.redhat-7-x, which is based off of 3.5.33, but syndesis is using 3.5.38 - do we need to update the product branch of f-m-p?

- what activemq version should we be using in the Fuse 7 build?   Right now we're set to 5.11.0.redhat-630343, and syndesis is using 5.15

Everything else looks pretty reasonable, I think (correct me if I'm wrong)

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -45,26 +45,26 @@
     <image.meta>syndesis/syndesis-meta:%l</image.meta>
     <image.ui>syndesis/syndesis-ui:%l</image.ui>
 
-    <fabric8.maven.plugin.version>3.5.38</fabric8.maven.plugin.version>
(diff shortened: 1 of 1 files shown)


oscerd:
I wouldn't change the ActiveMQ version here.

cunningt:
Closing, going to resubmit without redhat artifacts.

PR #3706, opened 2018-10-01
Upgraded to latest Apicurio release (0.2.15).  Fixes #3647.

### LINKED ISSUE #3647: Update the Apicurio Editor to the latest version (0.2.15)
## Description
I just did a new release of Apicurio (0.2.15.Final) which includes a new release of the editor component.  You can find it here:

https://www.npmjs.com/package/apicurio-design-studio

It is now version 0.2.15 and Syndesis should upgrade when possible.

When upgrading, you'll need to also upgrade the following dependencies in `package.json`:

* oai-ts-core (should be upgraded to version **0.2.23**
* oai-ts-commands (should be upgraded to version **0.2.41**

Finally, you'll need to add two more library dependencies (also to `package.json`) that the Apicurio editor now requires:

* marked (should be version 0.5.0)
* pluralize (should be version 7.0.0)

I've made these same changes to Apicurito already.  You can find the changeset here:  https://github.com/Apicurio/apicurito/commit/e3b77626c3499333693ba2ad31e5bd6ff99a309a

Just make the same changes in Syndesis and I think you should be good.

EricWittmann:
Upgraded Apicurio to the latest version.  I don't have a good way to test these changes locally - so would appreciate it if someone could do a quick sanity check to make sure that the API Connector flow still works with Apicurio in the mix.


riccardo-forina:
When modifying the `package.json` we must be sure to run a `yarn install`  to generate an updated `yarn.lock` file, which is required to have deterministic installs of the dependencies in CI and locally.

I think #3714 was caused by missing this step here.

It's something to should remember to check when reviewing, also.

PR #3716, opened 2018-10-02
chore: update yarn.lock

zregvart:
Updates `yarn.lock`, this one is generated with yarn 1.10.1.


zregvart:
We updated `yarn.lock` in #3722, so closing this.

Issue #3651, opened 2018-09-21
Increase default size of step icon circle in IVP


amysueg:
## This is a...

<pre><code>
[ x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
First raised in #3630 and referenced in a new request for step icons in #3650, increase the size of the icon container/circle in the IVP for steps to accommodate the new icons.

Assigned to @michael-coker. FYI @sjcox-rh @dongniwang @seanforyou23 @jameslau

amysueg:
Just a reminder that we'll need to determine what impact this will have on the Integration Details page, since it includes a horizontal representation of the Integration flow and currently uses the smaller circles to represent steps.

amysueg:
Thanks, @michael-coker for updating this to include the tooltip. We're planning to update the tooltip on the "Collapse" arrows (<<) to say "Collapse" (currently it says "Expand info"). Also, we agreed that the terms "Expand" and "Collapse" are more precise and we should remove the abbreviated "info" from the tooltip.
@dongniwang @sjcox-rh

PR #3748, opened 2018-10-04
refactor IVP code 

### LINKED ISSUE #3641: Decrease space between connection/step icons in collapsed IVP (in the integration editor), fix icon placement
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

From https://github.com/syndesisio/syndesis/issues/3630#issuecomment-423187172:

> We seem to have a regression in the IVP where there's a massive amount of space between the icons again. It looks like they are being sized based on the details text - when the details panel is opened then the spacing seems right. Can we just squash the steps together when the details pane is not open?

## The problem
* There is too much space between connections/steps when the IVP is collapsed.
* The warning icon covers the step icon.

## Expected behavior
* There should be less space there so the icons are closer together.
* The warning icon should be positioned like a badge, per the [UXD design](https://redhat.invisionapp.com/share/UYNXZ30MQ34#/screens/318752832)

1. Edit an integration
2. See icon spacing in collapsed IVP
3. ☹️

jameslau:
fixes #3641 #3651.

### DIFF
--- a/app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.scss
+++ b/app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.scss
@@ -49,7 +49,7 @@
   .icon {
     border-radius: 50%;
     border: 3px solid $color-pf-black-300;
-    padding: 11px;
+    padding: 12px;
     width: 100%;
     height: 100%;
     background-color: #ffffff;
@@ -60,8 +60,6 @@
 
     &.not-connection {
       background-color: #ffffff;
-      height: 100%;
-      width: 100%;
     }
 
     .connection-icon {
@@ -100,7 +98,9 @@
       border-radius: 50%;
       padding: 5px;
       cursor: pointer;
-      transform: translate(100%, -100%);
+      top: 0;
+      right: 0;
+      transform: translate(25%, -25%);
 
       &::before {
         color: $color-pf-white;
--- a/app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.scss
+++ b/app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.scss
@@ -60,12 +60,8 @@
 
     &.not-connection {
       background-color: #ffffff;
-      border-width: 10px;
-      padding: 5px 0px;
-      border-radius: 100px;
-      height: 33px;
-      margin: 11px;
-      width: 32px;
+      height: 100%;
+      width: 100%;
     }
 
     .connection-icon {
@@ -103,9 +99,8 @@
       background: #ec7a08;
       border-radius: 50%;
       padding: 5px;
-      top: -6px;
-      right: -6px;
       cursor: pointer;
+      transform: translate(100%, -100%);
 
       &::before {
         color: $color-pf-white;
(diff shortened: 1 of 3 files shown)


app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.scss
```diff
@@ -60,12 +60,8 @@
...
     &.not-connection {
       background-color: #ffffff;
-      border-width: 10px;
-      padding: 5px 0px;
-      border-radius: 100px;
-      height: 33px;
-      margin: 11px;
-      width: 32px;
+      height: 100%;
+      width: 100%;
```
michael-coker: I think the height and width here are unnecessary. This icon should get width/height 100% from https://github.com/syndesisio/syndesis/blob/cd59a1b99ebcce04f66290f9d559084adfcf947f/app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.scss#L53

app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.scss
```diff
@@ -103,9 +99,8 @@
       background: #ec7a08;
       border-radius: 50%;
       padding: 5px;
-      top: -6px;
-      right: -6px;
       cursor: pointer;
+      transform: translate(100%, -100%);
```
michael-coker: This positions the icon from the center of the connection icon. That's one way of doing it, but it's dependent upon the size of the connection icon and warning icon. A better way would do it is to assign `top: 0; right: 0; transform: translate(25%,-25%);`. It's always a good idea to add a vertical (`top`, `bottom`) and horizontal (`left`, `right`) positioning to non-`static` positioned elements.

app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.scss
jameslau: - property and values width/height has been removed. selector inherits width: 100% from parent (`&`).

app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.scss
jameslau: Thanks for the tip @michael-coker! Correct settings applied to icons now.

app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.scss
```diff
@@ -49,7 +49,7 @@
   .icon {
     border-radius: 50%;
     border: 3px solid $color-pf-black-300;
-    padding: 11px;
+    padding: 12px;
```
michael-coker: what's the reasoning behind changing the padding?

michael-coker:
Looks like this is still an issue.

> Also the last connection icon in the IVP is oddly shaped when you're creating an integration, and slightly non-circular (55x52) when editing an integration.

app/ui/src/app/integration/edit-page/flow-view/flow-view-step.component.scss
jameslau: Nice find. Padding and height has been added to the `.icon-container`. Now all step icons are 55x55 pixels.

jameslau:
Hey @michael-coker, the last bit of change for this issue has been pushed up and is ready for a review.

jameslau:
Adjusted the sizing of the last step icon circle. Please verify code @michael-coker. Thanks!

Issue #3405, opened 2018-08-20
Failed to stop integration Error stopping integration: [object Object] during publishing 


mmelko:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
Currently it's not possible to stop integration during publishing,  _Failed to stop integration Error stopping integration: [object Object]_ is displayed. I can delete integration though. 

## Expected behavior
I should be able to stop integration during publishing. When I make some mistake I can't edit either so for fixing it I have to delete it and create new one. 

## Tasks involved / Steps to Reproduce

1.
2.
3.
4.

heiko-braun:
@lburgazzoli @nicolaferraro Any thoughts on this?

KurtStam:
When an integration is started/deployed the POST request returns an IntegrationDeployment data structure with the new version that is being started/deployed. When the user clicks on Stop the url that is called is: 

POST: /api/v1/integrations/i-LO2eKU4E1UksX4jFH5Rz/deployments/undefined/targetState

where 'undefined' is supposed to be the version. More over the backend, the backend sends a 404 bc the versionId is defined as an int. So

/api/v1/integrations/i-LO2eKU4E1UksX4jFH5Rz/deployments/{versionId}/targetState

does NOT match the request made.

tplevko:
@mmelko please verify the fix

PR #3762, opened 2018-10-08
[fix] Failed to stop integration #3405

KurtStam:
Fixes https://github.com/syndesisio/syndesis/issues/3405

In the PR review I'd like to understand what the difference is between the deploymentVersion and the version. The deploymentVersion is undefined, but the version contains the correct version. Using that instead fixes the issue but I'm unclear why we have two pieces of data that seem to contain the same information. Maybe I'm missing something, or maybe there is larger issue.

@gashcrumb, @seanforyou23 maybe one of you can shed a light on this?


KurtStam:
ping @gashcrumb, @seanforyou23

gashcrumb:
I can't shed light on this other than there were a lot of model changes throughout the scope of adding deployments/history, these changes really expose how badly we need to have a single source of truth for the models :-)

PR #3771, opened 2018-10-09
fix(performance) Reduce CPU usage of syndesis server

### LINKED ISSUE #3717: Syndesis Server running at 75% CPU
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

## Description
The syndesis-server pod takes a lot of CPU! If left untamed it has a tendency to swipe 3 CPUs. If limited by Openshift to 1 CPU then it runs smoothly with 75% (while its couple of subordinate threads reside on 50%+). See htop screenshot.

By profiling the server, it is possible to get an idea of what is taking up the CPU time. The IntegrationUpdateHandler validates all integrations (including deleted integrations!) by calling validators on each integrations fetched from the DataManager. The trouble is the NoDuplicateValidator then fetches all the integrations from the DataManager in order to test for duplicates:

IntegrationUpdateHandler:
```
List<Integration> integrations = DataManager.fetchAll(Integration.class ...)
for (Integration integration : integrations) {
  ...
  messages.addAll(computeValidatorMessages(supplier, integration));
  ...
}
```

NoDuplicateIntegrationValidator:
```
...
final Set<String> names = dataManager.fetchAll(Integration.class).getItems().stream()
...
```

Fetching integrations from the DataManager is expensive and slow in this case since there is no caching of the integrations (or at least caching is not used since a DataAccessObject is available for the Integration [class](https://github.com/phantomjinx/syndesis/blob/master/app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java#L216)).

phantomjinx:
* install/...
* tools/...
 * Limit the cpu requirements of the syndesis server to 3/4 of a single
   CPU, stopping it taking advantage of 3-4 CPUs otherwise.

* app/...
 * Reduce the number of calls to the DataManager by fetching all the items
   from it once at the beginning of each compute then pass these collections
   into the requisite method calls using the TargetWithVector class.
 * TargetWithVector pairs up the target with its search vector to allow
   for Validator classes to not have to use the DataManager.

fixes #3717

### DIFF
--- a/app/server/dao/src/main/java/io/syndesis/server/dao/validation/extension/NoDuplicateExtensionValidator.java
+++ b/app/server/dao/src/main/java/io/syndesis/server/dao/validation/extension/NoDuplicateExtensionValidator.java
@@ -15,14 +15,14 @@
  */
 package io.syndesis.server.dao.validation.extension;
 
-import io.syndesis.server.dao.manager.DataManager;
-import io.syndesis.common.model.extension.Extension;
-import io.syndesis.common.model.validation.extension.NoDuplicateExtension;
-import org.springframework.beans.factory.annotation.Autowired;
-
+import java.util.Set;
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
-import java.util.Set;
+import org.springframework.beans.factory.annotation.Autowired;
+import io.syndesis.common.model.extension.Extension;
+import io.syndesis.common.model.validation.extension.ExtensionWithDomain;
+import io.syndesis.common.model.validation.extension.NoDuplicateExtension;
+import io.syndesis.server.dao.manager.DataManager;
 
 public class NoDuplicateExtensionValidator implements ConstraintValidator<NoDuplicateExtension, Extension> {
 
@@ -36,6 +36,10 @@ public class NoDuplicateExtensionValidator implements ConstraintValidator<NoDupl
 
     @Override
     public boolean isValid(final Extension value, final ConstraintValidatorContext context) {
+        if (value instanceof ExtensionWithDomain) {
+            return isValid((ExtensionWithDomain) value);
+        }
+
         if (value.getExtensionId() == null) {
             return true;
         }
@@ -52,8 +56,26 @@ public class NoDuplicateExtensionValidator implements ConstraintValidator<NoDupl
                 return false;
             }
         }
-
         return true;
     }
 
+    private boolean isValid(final ExtensionWithDomain value) {
+        Extension target = value.getTarget();
+        if (target.getExtensionId() == null) {
+            return true;
+        }
+
+        for (Extension other : value.getDomain()) {
+            if (other.idEquals(target.getExtensionId())) {
+                continue;
+            }
+
+            boolean installed = other.getStatus().isPresent() && other.getStatus().get() == Extension.Status.Installed;
+            if (installed) {
+                return false;
+            }
+        }
+
+        return true;
+    }
 }
--- a/app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/IntegrationUpdateHandler.java
+++ b/app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/IntegrationUpdateHandler.java
@@ -26,7 +26,6 @@ import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import javax.validation.Validator;
-
 import io.syndesis.common.model.ChangeEvent;
 import io.syndesis.common.model.Kind;
 import io.syndesis.common.model.action.Action;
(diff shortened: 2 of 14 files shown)


phantomjinx:
@rhuss 
@lburgazzoli 
Hi guys, this is a single commit PR but does have 2 elements, ie. server code changes and OS template changes. If you could prioritize review of the changes you are most au fait with I would be most grateful.
Thx!

oscerd:
Aren't we in code freeze?

phantomjinx:
@oscerd 
Sorry, though it was a feature-freeze but we were still bug fixing?

oscerd:
Yes, this a sprint of a bug fixing too. I don't know if this is considered a bug, but it looks more like an enhancement. Not sure.

app/server/dao/src/main/java/io/syndesis/server/dao/validation/extension/NoDuplicateExtensionValidator.java
```diff
@@ -15,38 +15,30 @@
...
+        if (target.getExtensionId() == null) {
             return true;
         }
 
-        Set<String> ids = dataManager.fetchIdsByPropertyValue(Extension.class, "extensionId", value.getExtensionId());
-        if (value.getId().isPresent()) {
-            ids.remove(value.getId().get());
-        }
+        for (Extension other : value.getVector()) {
+            if (other.idEquals(target.getExtensionId()))
```
lburgazzoli: We should always use brakets

phantomjinx:
@oscerd 

Profiling and solving CPU bottlenecks is a bugfix to my mind but happy to shelve it if people prefer.

app/server/dao/src/main/java/io/syndesis/server/dao/validation/TargetWithVector.java
```diff
@@ -0,0 +1,56 @@
...
+        this.vector = vector;
+    }
+
+    public T getTarget() {
+        return target;
+    }
+
+    public Collection<T> getVector() {
+        return vector;
+    }
```
lburgazzoli: Not a huge fan of "Vector", maybe we can use `Stream<T> stream()~

phantomjinx: This is not the java Vector object but the technical term vector as in "I have a seach vector". In that sense stream doesn't make much sense. Also, Stream can only be used once then it is closed, which I think might be a little limiting?

oscerd:
> @oscerd
> 
> Profiling and solving CPU bottlenecks is a bugfix to my mind but happy to shelve it if people prefer.

I agree about bugfixing, I'm just thinking about the fact that this PR changed some stuff. But mine was just a question, no problem from my side.

app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/IntegrationUpdateHandler.java
```diff
@@ -142,7 +142,27 @@ private Connection includeConnector(Connection connection) {
...
         final List<Integration> integrations = dataManager.fetchAll(Integration.class).getItems();
+        final List<IntegrationDeployment> deployments =
+                dataManager.fetchAll(IntegrationDeployment.class).getItems().stream()
+                .filter(d -> !d.getSpec().isDeleted())
+                .collect(Collectors.toList());
+        final Set<Integration> deployedIintegrations = deployments.stream()
+                .map(d -> d.getSpec())
+                .collect(Collectors.toSet());
+        final List<Connection> connections = dataManager.fetchAll(Connection.class).getItems();
+        final Collection<Integration> allIntegrations = new ArrayList<Integration>();
```
lburgazzoli: the size of the list is known so we can construct it with a proper initial capacity

phantomjinx:
@oscerd 
No problem. I'll leave it up to the powers-that-be ;-)

app/server/dao/src/main/java/io/syndesis/server/dao/validation/TargetWithVector.java
lburgazzoli: The usage pattern seems always like "foreach" thus I was thinking about using stream, but that's not a huge issue.

lburgazzoli:
@rhuss can you have a look at the changes done to the templates ?

install/generator/04-syndesis-server.yml.mustache
```diff
@@ -126,6 +126,13 @@
               path: "/health"
               port: 8181
             initialDelaySeconds: 10
+          resources:
```
rhuss: why is this section added here as it is already included some lines before ? If its about adding cpu limit / resources please adapt the existing 'resources' configuration.

lburgazzoli:
@phantomjinx it looks good to me. 

If you have time if would be nice to have a comparison of the resources (cpu, memory) used before and after this patch.

install/generator/04-syndesis-server.yml.mustache
phantomjinx: @zregvart 
No problem. They are part of the commit.

phantomjinx:
Results of changes to CPU usage:

profiling-post-change

- CPU usage has lessened considerably. Occasionally, spikes to 20-50% which is expected with work to perform. However, it is happy to return to a much quieter background level.

profile-visualvm-post-change

- The validate methods no longer appear as bottlenecks. DataManager.fetchAll() does get listed but is a necessary call for retrieving the artefacts for comparison.

lburgazzoli:
so LGTM, now it is up to @rhuss :)

rhuss:
@lburgazzoli looks good to me now, but should we merge it as there is a feature freeze for 7.2 ? // @heiko-braun

heiko-braun:
#3717 is a bug, or at least at the edge to it. I'd say we merge it.

heiko-braun:
@rhuss Feel to free to merge when you have finished the review

Issue #3601, opened 2018-09-17
UX Design Improvements for API Provider 


dongniwang:
Related to https://github.com/syndesisio/syndesis/issues/3140

Question! In this [design](https://redhat.invisionapp.com/share/4JNR7J5DMNY#/screens/316053627) we show a badge with a number. I implemented it as the count of operations coming from the swagger file, but talking with @dongniwang I found that it should have been the number of _integrated_ operations.

But iterating on that, wouldn't both info be useful at the same time? Something like 5/20 for example, to indicate that 5 out of 20 operations have an integration defined.

Somewhat linked to this, I think it would be interesting to show the same thing [here](https://redhat.invisionapp.com/share/CBNWDY58AHG#/screens/318306620) instead of the Santa Hat, it would be a sort of progress indicator for the user while working.

WDYT?

_Originally posted by @riccardo-forina in https://github.com/syndesisio/syndesis/issues/3140#issuecomment-421352496_

cc: @amysueg @sjcox-rh @heiko-braun @paoloantinori @gashcrumb

michael-coker:
@dongniwang how should we handle the badge size? I imagine the badge size should remain constant regardless the count inside it? If so, how many numbers should be able to fit inside, and how many should it be optimized for? (for example, it looks great with 2 numbers in it and spacious if there is only 1 and tight if there are 3)

Or should the circle get bigger/smaller depending on the # inside it?

Or should we consider something like the design @jameslau included when we were talking about it originally? 
![unread-mail-number-iphone](https://user-images.githubusercontent.com/35148959/46636717-c9ce5480-cb1e-11e8-9bc4-366d15494b9d.jpg)

michael-coker:
here's what I have so far.

<img width="319" alt="screen shot 2018-10-08 at 5 27 12 pm" src="https://user-images.githubusercontent.com/35148959/46636877-6ee92d00-cb1f-11e8-97de-31f3cbe92f45.png">

PR #3781, opened 2018-10-09
bug(badge): fix oval shape for api flow count badge, add tooltip

michael-coker:
fixex https://github.com/syndesisio/syndesis/issues/3601

### DIFF
--- a/app/ui/src/app/integration/list/list.component.html
+++ b/app/ui/src/app/integration/list/list.component.html
@@ -34,7 +34,7 @@
           <span class="fa fa-angle-right"></span>
           <span class="icon">
             <img class="syn-icon-small" src="./../../assets/icons/multi-flow.connection.png" aria-hidden="true">
-            <span class="badge">{{ integration.flows.length }}</span>
+            <span class="syn-badge" tooltip="{{ integration.flows.length }} implemented flows" container="body" placement="right">{{ integration.flows.length }}</span>
           </span>
         </ng-template>
       </div>
--- a/app/ui/src/app/integration/list/list.component.scss
+++ b/app/ui/src/app/integration/list/list.component.scss
@@ -23,12 +23,12 @@
         border-radius: 11px;
         top: 0;
         right: 0;
-        height: 22px;
-        min-width: 22px;
+        height: 20px;
+        min-width: 20px;
         line-height: 1;
         transform: translate(75%, -75%);
         color: $color-pf-white;
-        padding: 5px;
+        padding: 4px;
       }
     }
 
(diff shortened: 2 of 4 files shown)


app/ui/src/app/integration/list/list.component.scss
```diff
@@ -16,14 +16,19 @@
 
     .icon {
       position: relative;
-      .badge {
+
+      .syn-badge {
         position: absolute;
         background: #ec7a08;
```
zregvart: Any chance we can make this other color? I know that orange is the new black, but it screams like error to me, we can hope that it's me.

michael-coker: @dongniwang what do you think?

dongniwang: We’re going to gather user feedback on API provider and we can definitely get some feedback on the color of the badge. A more neutral color choice I can think of is perhaps blue? I’ll check with a visual design SME to get some suggestions.

michael-coker: screen shot 2018-10-10 at 8 40 33 am
screen shot 2018-10-10 at 8 40 26 am
screen shot 2018-10-10 at 8 39 46 am

@dongniwang that sounds good to me. here's a screenshot of the dark and light blue, and one with the list row hover state (blue background)

app/ui/src/app/integration/list/list.component.html
```diff
@@ -34,7 +34,7 @@
           <span class="fa fa-angle-right"></span>
           <span class="icon">
             <img class="syn-icon-small" src="./../../assets/icons/multi-flow.connection.png" aria-hidden="true">
-            <span class="badge">{{ integration.flows.length }}</span>
+            <span class="syn-badge" tooltip="{{ integration.flows.length }} implemented flows" container="body" placement="right">{{ integration.flows.length }}</span>
```
riccardo-forina: `{{ integration.flows.length }}` will yield the number of flows, aka operations, coming from the spec. AFAIK there is no clear way to know the number of _implemented_ flows. @nicolaferraro is there any property we can use to figure out if an operation is implemented or not, but checking if the last step returns anything different from a 501?

nicolaferraro: We have talked about that in a call. I remember that 501 should be the "flag" that distinguish a implemented operation from a non-implemented one.

gashcrumb:
Is this ready to get merged?

michael-coker:
@gashcrumb - I believe @dongniwang is still working on how to present the badge.

gashcrumb:
Is it worth maybe merging this one as the groundwork and then just open a PR with the final work later?

PR #3805, opened 2018-10-11
fix(runtime): don't add log step when logging parameters not specified

### LINKED ISSUE #3786: Integration fails to start with with empty log step
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## The problem

If the user adds a logging step to the integration without checking any of the checkboxes (Message Context, Message Body) or providing a message to log. The integration fails to start.

## Expected behavior

We can either validate so that the user specifies something to log, we can provide a default if nothing is provided to log or we can skip the logging if nothing is provided.

## Screenshot

screenshot_2018-10-10 syndesis

1. Either add validation
2. Provide default
3. Or don't add log to route definition in LogStepHandler

[i-log-me-baby-one-more-time-12-qjwsc.log](https://github.com/syndesisio/syndesis/files/2464026/i-log-me-baby-one-more-time-12-qjwsc.log)

zregvart:
Prevents `IllegalArgumentException` on integratino startup by not adding a `log` processor to route when no parameters are given to log.

Fixes #3786

### DIFF
--- a/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/LogStepHandler.java
+++ b/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/LogStepHandler.java
@@ -34,16 +34,21 @@ public class LogStepHandler implements IntegrationStepHandler {
 
     @Override
     public Optional<ProcessorDefinition<?>> handle(Step step, ProcessorDefinition<?> route, IntegrationRouteBuilder builder, String flowIndex, String stepIndex) {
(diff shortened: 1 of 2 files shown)


lburgazzoli:
@zregvart @heiko-braun 

It looks good to me and I'm ok to merge it but as it is marked as a feature I want to be sure we want this to be merged.

heiko-braun:
@lburgazzoli Thanks Luca, I'd say this is an edge case and we are good to merge

PR #3819, opened 2018-10-12
Fix issue with overwritten templates

nicolaferraro:


heiko-braun:
Thanks for the quick turn around and responses to this issue everyone. I appreciate it.

PR #3849, opened 2018-10-16
Update to connection and step icons

sjcox-rh:
Consistent color and visual style for step icons as well as changing color to SQL database and timer connection.


riccardo-forina:
Thanks :)

PR #3867, opened 2018-10-16
fix: only allow the data mapper to be used if the next step that has a datashape is not none/any

### LINKED ISSUE #3859: 'No data type specification was found for subsequent step' error in data mapper
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Having multiple data mapper steps can sometimes lead to the above error when editing one of them. See steps to reproduce and screenshots for one such scenario

## Steps to Reproduce

1. Create a API Provider integration from the Todo API
1. Start implementing the 'Fetch Task' operation
1. Add a Postgres connection, invoking SELECT * FROM todo (yes, this does not make much sense IRL, but still is possible)
1. Add a Data Mapper step between the start connection and the Postgres connection
1. Notice this Data Mapper has the response fields in the Target colum (as the Postgres step does not take any input)
1. Connect 'id' from the Source to the Target, leave other fields unconnected, close the Data Mapper
1. Add a new Data Mapper step, this time between the Postgres step and the finish connection
1. Connect 'completed' and 'task' fields from the Source to the Target, close the Data Mapper
1. Open the first Data Mapper step again, an error is shown instead of the Data Mapper UI

## Screenshots

screenshot_2018-10-16_13-46-12
screenshot_2018-10-16_13-46-45
screenshot_2018-10-16_13-47-06
screenshot_2018-10-16_13-47-23

gashcrumb:
fixes #3859

@igarashitm FYI

### DIFF
--- a/app/ui/src/app/store/step/step.store.ts
+++ b/app/ui/src/app/store/step/step.store.ts
@@ -43,7 +43,7 @@ export class StepStore {
       stepKind: DATA_MAPPER,
       properties: {},
       configuredProperties: undefined
-    }),
+    }, true, false),
     StepStore.requiresOutputDataShape({
       id: undefined,
       connection: undefined,
@@ -278,7 +278,8 @@ $\{in.body.title\} // Evaluates true when body contains title.
         s =>
           s.action &&
           s.action.descriptor.outputDataShape &&
-          s.action.descriptor.outputDataShape.kind !== DataShapeKinds.NONE
+          s.action.descriptor.outputDataShape.kind !== DataShapeKinds.NONE &&
+          s.action.descriptor.outputDataShape.kind !== DataShapeKinds.ANY
       ).length > 0
     );
   }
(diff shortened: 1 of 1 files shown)


app/ui/src/app/store/step/step.store.ts
```diff
@@ -278,7 +278,8 @@ $\{in.body.title\} // Evaluates true when body contains title.
         s =>
           s.action &&
           s.action.descriptor.outputDataShape &&
-          s.action.descriptor.outputDataShape.kind !== DataShapeKinds.NONE
+          s.action.descriptor.outputDataShape.kind !== DataShapeKinds.NONE &&
+          s.action.descriptor.outputDataShape.kind !== DataShapeKinds.ANY
```
igarashitm: Just one thing want to make sure - is this OK for "Describe Data" scenario? `DataShapeKinds.ANY` is kinda intermediate state, once user define a custom DataShape specification via "Describe Data" page, then that DataShape turns into "shaped" state, which is available for adding datamapper step.

igarashitm: If this method is not used to determine showing a warning for Describe Data, then that would be fine.

gashcrumb: Yeah, if the user selects some other kind than "type spec not required" then this field will be like json schema or instance or etc so this check will pass still.

PR #3869, opened 2018-10-16
fix(Template Step) Wording problems and abnormalities

### LINKED ISSUE #3757: Description of Template step incorrectly says it is a custom step
## This is a...

<pre><code>
[ ] Feature request
[X ] Bug report  
</code></pre>

## Description
In the UI, after you click "Add a step", in the list of steps, the description of the Template step is:

&nbsp;&nbsp;&nbsp;Upload a Mustache template to define a custom step

This description should not refer to a Template step as a custom step. We already use "custom step" in the doc to describe an extension that defines a step that operates on data between connections. And in the page for Choosing a Step, you can filter for Custom Steps and you won't get the Template step.  I think this description needs to change to this:

&nbsp;&nbsp;&nbsp;Upload or create a Mustache template to define consistent output data.

Then, when the Upload Template page appears, the text at the top says:

&nbsp;&nbsp;&nbsp;An integration can have one custom step that you define by uploading a template.

I suggest replacing that line with:

&nbsp;&nbsp;&nbsp;A template step takes data from a source and inserts it into the format that is defined in a template that you provide. 

This would remove the only indication that only one Template step in an integration is allowed. I think that's the current behavior -- there is nothing that prevents the user from adding more than one Template step to an integration, as far as I can tell. Also, I can imagine a scenario where I start an integration data with a lot of data, for example from a Salesforce connection, and I want to use one Template step to send some high level data to an an executive and then I want to use a second template to send expanded data to the person who needs to work with the customer. 

However, if there is supposed to be a restriction that an integration can have only one Template step, that should probably be a separate issue. 

@gaughan   @dongniwang @phantomjinx

phantomjinx:
* Corrects wording in the UI that does not reflect the true functionality

fixes #3757


app/ui/src/app/integration/edit-page/step-configure/step-configure.component.html
```diff
@@ -74,7 +74,7 @@
                       <div class="title">
                         <h1>Upload Template</h1>
                       </div>
-                      <p>An integration can have one custom step that you define by uploading a template.</p>
+                      <p>A template step takes data from a source and inserts it into the format that is defined in a template that you provide.</p>
```
michael-coker: should this be in the dictionary file?

app/ui/src/app/store/step/step.store.ts
```diff
@@ -62,7 +62,7 @@ export class StepStore {
       action: undefined,
       name: 'Template',
       stepKind: TEMPLATE,
-      description: 'Upload a Mustache template to define a custom step',
+      description: 'Upload or create a Mustache template to define consistent output data.',
```
michael-coker: this, too? Not sure if there's a rule for what stuff should be in the dictionary.

gashcrumb: Yeah, definitely.  I think this will be a bit more tricky because it'd require injecting the i18n service into the store, so again I'm fine leaving it as-is for now, probably worth creating a separate issue specific to this class since there's more than just this one step defined in the step store, all of the steps defined in this class should be using the dictionary.

PR #3873, opened 2018-10-17
Fixes 3408, added external Prometheus configuration to documentation

dhirajsb:


dhirajsb:
@tplevko I haven't added instructions for using Prometheus operator to install an external instance, since I believe there is a separate document for that from https://issues.jboss.org/browse/ENTESB-9412
@fbolton if there is a link we can provide to that please let me know and I'll add it to this doc.

PR #3878, opened 2018-10-17
chore(server): update Spring and Boot

zregvart:
Updates Spring Framework and Spring Boot to the latest versions.


zregvart:
> I believe we should be in sync with what the camel version we are using is declaring

Yeah I just wanted to quickly update the version, a lot of test failures this caused. This needs more work.

PR #3883, opened 2018-10-17
fix: manually trigger data mapper line machine component after init

### LINKED ISSUE #3833: DataMapper : inconsistent field mapping visiualization
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

## The problem

I define the mapping between two large objects:
- a SalesForce Case object
- a ServiceNow Incident object 

When defining the map, everything work as expected but when I'm getting back to my integration to change the mapping, the lines connecting the fields to map are inconsistent.

## Screenshot

syndesis

1. Create an integration between SalesForce and ServiceNow
2. Map Case to Incident
3. Publish the integration
4. Edit the integration

gashcrumb:
fixes #3833


gashcrumb:
@riccardo-forina yeah I looked into that, but it's synchronous with the change detection cycle apparently.  

The other option would have been to use `runOutsideOfAngular` and inject the `ChangeDetectorRef` and manually trigger the detector, but `setTimeout` is overloaded to do this anyways, hence why it's such a popular sledgehammer I guess :-)

Issue #3760, opened 2018-10-08
Template step - can't upload a file nor create template in the editor


mastepan:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

I can't upload a template file (nothing happens after selecting a file)
&
Missing template editor:
![screenshot from 2018-10-05 09-13-03](https://user-images.githubusercontent.com/8707251/46594790-7e438880-cad5-11e8-85a1-9fe006362461.png)
![screenshot from 2018-10-05 09-14-21](https://user-images.githubusercontent.com/8707251/46594793-84396980-cad5-11e8-972f-536f168fccb5.png)

phantomjinx:
@mastepan 
It would appear that something in the codemirror module dependency does not like the minification routine used in the productionalization/release operations.

@gashcrumb 
Are you able to shed some light on the minification routine, please?

gashcrumb:
@kahboom can you work with @phantomjinx on this one while I'm out?

mastepan:
I've just updated Syndesis to the latest version and the editor and file upload started to work.
I'm closing this as it seems to be ok.

mastepan:
This issue is present again...I can reproduce it in Chrome and Firefox.

- Editor is unresponsive
- Uploading a template file doesn't fill the editor
- Uploading a template file doesn't enable **Done** button (after 2nd upload try Firefox enables it. Chrome doesn't)



Syndesis:
    1.5-SNAPSHOT

Build ID:
    dd612b39-ed84-477b-9b06-9583a5932028

Commit ID:
    61eea3921d5f03920934677d47747ebdcd0a85f5

heiko-braun:
@phantomjinx This has been reopened by QE, can you take a look please?

phantomjinx:
Latest investigation results:
There seems to be some deficiency in codemirror's [lint addon](https://github.com/codemirror/CodeMirror/blob/master/addon/lint/lint.js#L177) that causes the observed error when production builds uglify the code.

[This](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/264) seems to be a similar issue.

[...]

PR #3886, opened 2018-10-17
fix(ui) Patch codemirror's lint script file

phantomjinx:
* ReferenceError occurs when codemirror's lint.js is uglified. Since it
  will take some time for a PR to be accepted and a new version released,
  uses package-patch to make the small code change and fix the problem.

* package-patch will auto-run on post-install to ensure the patch is
  applied to the file. See https://www.npmjs.com/package/patch-package
  for details.

#3760

### DIFF
--- a/app/ui/package.json
+++ b/app/ui/package.json
@@ -28,7 +28,9 @@
     "sw": "sw-precache --root=dist --config=sw-precache-config.js",
     "docker:build": "yarn ng build --aot --prod && docker build -f docker/Dockerfile -t syndesis/syndesis-ui:latest . && docker tag syndesis/syndesis-ui:latest docker.io/syndesis/syndesis-ui:latest",
     "snyk-protect": "snyk protect",
-    "prepare": "if [ -z $OFFLINE_BUILD ]; then npm run snyk-protect; yarn snyk-protect; fi"
+    "package-patch": "patch-package --use-yarn",
+    "prepare": "if [ -z $OFFLINE_BUILD ]; then npm run snyk-protect; yarn snyk-protect; fi; yarn package-patch",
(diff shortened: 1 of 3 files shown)


gashcrumb:
@phantomjinx use `yarn` instead of `npm` if possible, `npm` isn't even installed on the CI machine.  It should be a direct switch I think.

gashcrumb:
Looks like you want to use the `--use-yarn` flag so it knows to use `yarn` vs `npm`

phantomjinx:
@gashcrumb 

So in order to workaround yarn's lack of postinstall support (explained [here](https://github.com/ds300/patch-package#why-use-postinstall-postinstall-with-yarn)), [patch-package](https://github.com/ds300/patch-package) recommends to install either [postinstall-prepare](https://github.com/ds300/patch-package) or [postinstall-postinstall](https://github.com/ds300/postinstall-postinstall) (both appear to be the same package but simply have a rename). However, these packages  do [depend](https://github.com/ds300/postinstall-postinstall/blob/master/run.js#L14) on npm hence the error in circleci builds.

The workaround is to uninstall it and just use the prepare hook. This means that the patch is not applied on a _yarn remove_ but given this specific patch is only required on a release install (full uglify) then that shouldn't be an issue at the moment. I have opened an [issue](https://github.com/ds300/postinstall-postinstall/issues/2) with postinstall-postinstall so assuming that gets fixes we can reinstall in the future.

gashcrumb:
> The workaround is to uninstall it and just use the prepare hook. This means that the patch is not applied on a yarn remove but given this specific patch is only required on a release install (full uglify) then that shouldn't be an issue at the moment.

Yep, I think that's fine, awesome!

PR #3895, opened 2018-10-18
Rework the response data shapes to support collections

### LINKED ISSUE #3788: Cannot implement API Provider operation returning collection
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem

When an operation in API Provider returns a collection, there is no way to set that up in the actual flow. It is possible to map whatever the items are in the returned array, but only the first element is ever returned when accessing the resulting API.

e.g. using the TODO API specification, which contains the following for GET / : 

```
            "schema": {
              "type": "array",
              "items": {
                "$ref": "#/definitions/Task"
              }
            }
```

I get the following:
screenshot_2018-10-10_12-30-23
screenshot_2018-10-10_12-30-55
screenshot_2018-10-10_12-36-40

zregvart:
This reworks the response data shapes of custom API client connector and API provider so they look like our quasi standard request shapes, i.e. have parameters/body properties.

Should be easier to review if you look from d92c335 onward as 33f5797 just contains JSON formatting to make the diff easier to read.

The gnarly bit is the change to the integration runtime to support Customizers on legacy Connectors (such as REST Swagger used by API client). This was needed to preserve backward compatibility with existing API client connectors that do not have the new response data shape.

Fixes #3788


lburgazzoli:
lgtm

we definitively need to get rid of the old connectors in a future release

app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/ResponseCustomizer.java
```diff
@@ -0,0 +1,100 @@
...
+        this.dataShape = dataShape;
+
+        addResponseConverter = shouldAddResponseConverter(dataShape);
+    }
+
+    static boolean isNullNode(final JsonNode node) {
+        return node == null || node.isNull();
+    }
+
+    static boolean shouldAddResponseConverter(final DataShape dataShape) {
```
zregvart: isUnifiedDataShape

app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/ResponseCustomizer.java
```diff
@@ -0,0 +1,100 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+public final class ResponseCustomizer implements ConnectorComponentCustomizer, OutputDataShapeAware {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ResponseCustomizer.class);
+
+    private boolean addResponseConverter;
```
zregvart: isUnifiedDataShape

KurtStam:
I opened: https://github.com/syndesisio/syndesis/issues/3896 so we can track updating connectors to return arrays as well.

PR #3912, opened 2018-10-19
[fix] API Provider Data Type Mismatch #3807

### LINKED ISSUE #3807: API Provider Data Type Mismatch
See also **https://issues.jboss.org/browse/ENTESB-11568**

## This is a...

<pre><code>
[ ] Feature request
[x ] Bug report  
</code></pre>

## Description
When you provider Todo App swagger to API Provider a data type mismatch occurs on unimplemented flows:

image

image

KurtStam:
Fixes #3807, so we don't need a datamapper


app/server/api-generator/src/main/java/io/syndesis/server/api/generator/swagger/UnifiedJsonDataShapeGenerator.java
```diff
@@ -162,8 +164,11 @@ private static ObjectNode createJsonSchemaWithProperties(final Map<String, Strin
         final List<AbstractSerializableParameter<?>> parameters = defaultProperties.entrySet().stream()//
```
zregvart: And I'd nuke this method too

zregvart:
```
[INFO] There is 1 error reported by Checkstyle 8.10.1 with /syndesis/checkstyle.xml ruleset.
[ERROR] src/main/java/io/syndesis/server/api/generator/swagger/UnifiedJsonDataShapeGenerator.java:[65] (regexp) RegexpSingleline: Line has trailing spaces.
```

@KurtStam pre-commit hooks would take care of that for you, just sayin...

zregvart:
Oh, and we need to follow this up with a change that includes the syndesis JSON schema URI in the data shape so that `ResponseCustomizer` can detect the need to unwrap the response here:

KurtStam:
@zregvart Adding the JSON Schema thingy seems like a different issue to me. Maybe we should open a new GH-issue for it, or am I missing something?

PR #3919, opened 2018-10-19
fhir-connector first spike - tech preview

johnpoth:
Thanks !

cc @rkorytkowski @paoloantinori @gaughan


oscerd:
Lets see if you'll pass PMD on the first iteration :-D

rkorytkowski:
Thanks @johnpoth! I've reviewed the code and tested locally. Looks good to me! Well done!

paoloantinori:
Update for anyone not familiar with this feature:
- we are considering adding it even if we are in code freeze since this is an isolated component that should not affect the stability of the rest of the project
- we are not expecting to have QE involved, it's too late
- the feature can eventually be hidden behind some env var, if we want to be even more conservative

Having said so, we still like to get a code review, eventually pointing out possible risks if you see any.

cc @heiko-braun @aileenc @janstey

app/connector/fhir/src/main/java/io/syndesis/connector/fhir/FhirReadCustomizer.java
```diff
@@ -0,0 +1,96 @@
...
+    private String id;
+    private String resourceType;
+    private String version;
+    private Boolean summary;
+    private FhirContext fhirContext;
+
+    @Override
+    public void customize(ComponentProxyComponent component, Map<String, Object> options) {
+        for (Map.Entry<String, Object> stringObjectEntry : options.entrySet()) {
+            LOG.info("setApiMethod key:" + stringObjectEntry.getKey() + "setApiMethod value" + stringObjectEntry.getValue());
```
lburgazzoli: May be better not to log options as info as they may contain sensitive info, maybe not in this case but as general
camel rule this should be avoided.

PR #3923, opened 2018-10-22
chore(upgrade): Update release script to set the proper upgrade pod tag.

### LINKED ISSUE #3463: [Upgrade] Syndesis operator uses hardcoded "latest" for upgrade pod tag
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

## The problem

Currently, by default, the syndesis specifies upgrade container tag "latest": https://github.com/syndesisio/syndesis/blob/master/install/syndesis.yml#L1635 

But for the community version in the docker hub, there is no such tag: https://hub.docker.com/r/syndesis/syndesis-upgrade/tags/ 

Therefore it causes upgrade to fail because of missing image: 

`Failed to pull image "docker.io/syndesis/syndesis-upgrade:latest": rpc error: code = Unknown desc = Tag latest not found in repository docker.io/syndesis/syndesis-upgrade`

I think the version should be changed from latest to <operator_version> somewhere in the process of building the operator image.

cc @rhuss 

1.
2.
3.
4.

rhuss:
Also removed the SYNDESIS_VERSION parameter for the upgrade template
used by the operator.
The version for this pod will be hardcoded and created when the template is created.
While we could assume operator version == upgrade-pod version in upstream (there everything is in sync),
this is not the case for productisation where build number might differ.

Therefore it has to be ensured already during template creating that the
proper tag (including build number) is set.

The tag for the ugrade pod needs to be provided via --upgrade-tag when
creating the templates.
This has been added for the community release and it shoud be verified
that this option is also used during productisation.

Fixes #3463.

### DIFF
--- a/install/operator/pkg/syndesis/action/upgrade.go
+++ b/install/operator/pkg/syndesis/action/upgrade.go
@@ -172,7 +172,6 @@ func (a *Upgrade) getUpgradeResources(syndesis *v1alpha1.Syndesis) ([]runtime.Ob
 		InstallParams: syndesistemplate.InstallParams{
 			OAuthClientSecret: "-",
 		},
-		SyndesisVersion: a.operatorVersion,
 		UpgradeRegistry: configuration.Registry,
 	})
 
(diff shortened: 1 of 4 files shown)


rhuss:
Duplicate of #3924, not sure how that happened.

Issue #3933, opened 2018-10-22
Update API integration & connection icon, increase size of that icon when used


michael-coker:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
API connection and integration icon should be updated with new icon. By default, that icon is presented too small. Need to increase the size of that icon (only) when it's displayed.

michael-coker:
The CSS that increases the icon visually is just temporary until we revisit our icon designs. 

cc @dongniwang

PR #3934, opened 2018-10-22
fix(api-icon): update api integration & connection icon, increase size visually

michael-coker:
fixes https://github.com/syndesisio/syndesis/issues/3933


michael-coker:
thanks @jameslau! Cool I'll keep that in mind. Generally speaking, I try not to rely on something like a screenshot during the verification, and pull the branch down and verify it visually on a local build instead. You never know if someone took a screenshot of their local build and forgot to commit a file or something, so their PR doesn't look like the screenshot. As part of the visual review, you would typically also want to verify that the design actually matches the implementation. In this issue, it would have been fair to ask for a design or specifications as to why I changed to the sizes that I did, and what the new icon is supposed to look like. And then on top of that, be sure to review the code itself. That way, you should be able to see the settings themselves pretty quickly, the before and after sizes, and know where in the app the changes occur based on the filenames and codeblocks modified in the templates. And then be sure to verify the validity of the code, like where I added something like `[ngClass]="{ 'api-provider' : (connection.id == 'api-provider') }"`, is that going to break anything? What if `connection` or `id` is undefined? Is my CSS scoped well? Is there a more efficient or best practice way to do anything I did? And then, if there are some parts of the PR you aren't comfortable verifying and merging into the master branch (like angular or other changes you don't understand the impact of), feel free to leave a comment in the PR (but don't "approve" it) stating that what you checked looks good if there are other reviewers assigned who can check the remaining parts, check with the person who submitted the PR and let them know which parts you are/aren't comfortable reviewing, or reach out on #syndesis-ui on IRC for help.

PR #3936, opened 2018-10-23
fix(server): make CORS great again

riccardo-forina:
While working on some experiment, I noticed that the current implementation of CORS was broken due to CORS checks being handled through a filter instead of using the HttpSecurity model.

Long story short: using a filter it's not possible to skip some of the checks performed as by the HttpSecurity configuration; preflight requests (OPTIONS) where being refused because the CSRF check was performed on them (which is wrong).

Now cors checks are chained to other checks, and preflight requests are permitted as by the `CorsConfigurationSource` rules.

This doesn't mean that it's now possible to do CORS requests against the default entrypoint of the app (eg. https://syndesis.192.168.64.16.nip.io), since the oauth-proxy is doing something weird I haven't grasped yet.
If one needs to talk directly with the server, it's required to expose it with a public route from the Openshift console, writing the authentication e CSRF headers as the oauth-proxy would do (eg. use the `X-Forwarded-Access-Token: token` header in place of the `Authentication: Bearer token` one).
And potentially, also a new oauth client needs to be created to allow implicit flows login against Openshift.

I'm not suggesting to do this by default, but it's very useful for a FE developer to be able to interact with the server outside the cluster...

_Screen of the "Routes - external traffic" configuration mentioned above_
server-external-traffic

### DIFF
--- a/app/server/runtime/src/main/java/io/syndesis/server/runtime/SecurityConfiguration.java
+++ b/app/server/runtime/src/main/java/io/syndesis/server/runtime/SecurityConfiguration.java
@@ -36,6 +36,11 @@ import org.springframework.security.web.authentication.preauth.PreAuthenticatedA
 import org.springframework.security.web.authentication.preauth.PreAuthenticatedGrantedAuthoritiesUserDetailsService;
 import org.springframework.security.web.authentication.preauth.PreAuthenticatedGrantedAuthoritiesWebAuthenticationDetails;
 import org.springframework.security.web.authentication.preauth.RequestHeaderAuthenticationFilter;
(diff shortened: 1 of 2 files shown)


riccardo-forina:
Flagged as WIP since we are in code freeze and I have stepped outside my comfort zone here :)

gashcrumb:
I think this is okay to merge at this point, @zregvart WDYT can I consider that approval a +1?  :-)

zregvart:
> I think this is okay to merge at this point, @zregvart WDYT can I consider that approval a +1? :-)

Better merge it this week :smile:

PR #3938, opened 2018-10-23
 fix(db): increase `autovacuum_vacuum_cost_limit` 

### LINKED ISSUE #3922: Autovacuum not reclaiming disk space
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description

We had a Syndesis instance (1.4.10, 76bc7c6f389f65ba0518bdd05823d247f4870a90) running for ~3 days (2 days 17 hours) with 6 integrations generating 730K messages.

The `syndesis-db` (1GiB) volume was on the verge of running out of disk space (12MiB left). And `VACUUM FULL jsondb` could not be run:

```
ERROR:  could not extend file "base/16385/16845": No space left on device
HINT:  Check free disk space.
STATEMENT:  VACUUM FULL ANALYSE jsondb;
```

Fetching activities resulted in 504 (Gateway timeout).

With ~30K rows in `jsondb` table the overwhelming majority of disk space was taken up with deleted rows. That were not vacuumed by autovacuum.

The only way to get the stream of activities was to copy the table content to a new table, drop the `jsondb` table, rename the new table to `jsondb` and recreate indexes.

i.e.
```
> SELECT * INTO jsondb2 FROM jsondb;
> DROP TABLE jsondb;
> ALTER TABLE jsondb2 RENAME TO jsondb;
> CREATE UNIQUE INDEX jsondb_pkey ON jsondb USING btree (path);
> CREATE INDEX jsondb_idx ON jsondb USING btree (idx, value) WHERE (idx IS NOT NULL)
```

I think we should try to tune the `autovacuum` so that the disk space is reclaimed (autovacuum will never perform `VACUUM FULL`) or bring back the DB maintenance job to perform `VACUUM FULL`.

The level of severity depends on how we envision the users using Syndesis, if this level of activity is acceptable (looks to me like it should be), then this feels like `p0`.

zregvart:
This doubles th `autovacuum_vacuum_cost_limit` from `1000` to `2000`
which empirically looks like a good value so that the autovacuum is not
throttled.

Before (with 1000):
```
LOG:  automatic vacuum of table "syndesis.public.jsondb": index scans: 0
	pages: 0 removed, 72236 remain, 1996 skipped due to pins
	tuples: 0 removed, 855896 remain, 797779 are dead but not yet removable
	buffer usage: 89805 hits, 70138 misses, 359 dirtied
	avg read rate: 4.438 MB/s, avg write rate: 0.023 MB/s
	system usage: CPU 0.55s/0.25u sec elapsed 123.46 sec
```

After (with 2000):
```
LOG:  automatic vacuum of table "syndesis.public.jsondb": index scans: 1
        pages: 0 removed, 1859 remain, 0 skipped due to pins
        tuples: 2915 removed, 22780 remain, 0 are dead but not yet removable
        buffer usage: 2611 hits, 431 misses, 6 dirtied
        avg read rate: 81.327 MB/s, avg write rate: 1.132 MB/s
        system usage: CPU 0.00s/0.01u sec elapsed 0.04 sec
```

Fixes #3922 

Also:

chore(db): improve PostgreSQL logging

This turns off the logging collector so the log messages are logged to
`stderr` making it more 12-factory. Also configures the log prefix to
include the timesatmp, application, command, SQLSTATE and session id.

For example:
```
2018-10-23 08:54:55 UTC psql CREATE ROLE 42710 5bcee1df.3c ERROR: role "sampledb" already exists
```

### DIFF
--- a/install/generator/04-syndesis-db.yml.mustache
+++ b/install/generator/04-syndesis-db.yml.mustache
@@ -83,7 +83,7 @@
       autovacuum_analyze_threshold = 10
       autovacuum_analyze_scale_factor = 0.05
       autovacuum_vacuum_cost_delay = 10ms
-      autovacuum_vacuum_cost_limit = 1000
+      autovacuum_vacuum_cost_limit = 2000
 
 - apiVersion: v1
   kind: Service
--- a/install/generator/04-syndesis-db.yml.mustache
+++ b/install/generator/04-syndesis-db.yml.mustache
@@ -76,6 +76,8 @@
   data:
     syndesis-postgresql.conf: |
       log_autovacuum_min_duration = 0
+      log_line_prefix = '%t %a %i %e %c '
+      logging_collector = off
       autovacuum_max_workers = 6
       autovacuum_naptime = 15s
       autovacuum_vacuum_threshold = 25
--- a/install/syndesis-dev.yml
+++ b/install/syndesis-dev.yml
@@ -537,6 +537,8 @@ objects:
   data:
     syndesis-postgresql.conf: |
       log_autovacuum_min_duration = 0
+      log_line_prefix = '%t %a %i %e %c '
+      logging_collector = off
       autovacuum_max_workers = 6
       autovacuum_naptime = 15s
       autovacuum_vacuum_threshold = 25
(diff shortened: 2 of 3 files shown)


zregvart:
The test has been running for 21 hours now, and the disk space hasn't increased dramatically (from 160MB used to 161MB), logs show that autovacuum is being trigered and `pg_stat_all_tables` has `autovacuum_count` at `85` (ATM), and `n_dead_tup` at around ~200. The highest `n_dead_tup` I've seen has been < `2000`.
I'm merging this to include it in a release we can confirm via more tests.

cc @mmelko

PR #3941, opened 2018-10-23
Several cleanups for the operator

zregvart:
1. ran `go fmt`
2. add `go fmt` `pre-commit` hook
3. fixed tests
4. tests are now run with `syndesis` script

### DIFF
--- a/tools/bin/commands/util/operator_funcs
+++ b/tools/bin/commands/util/operator_funcs
@@ -9,20 +9,6 @@ OPERATOR_IMAGE="syndesis/syndesis-operator"
 # globs to test
 TEST_TARGETS="./cmd/... ./pkg/..."
 
-# $GOPATH volume to share between docker runs
-DOCKER_GOPATH_VOLUME=$(mktemp -d ${TMPDIR:-/tmp}/gopath.XXXXX)
-
-# $XDG_CACHE_HOME ($GOCACHE) volume to share between docker runs
-DOCKER_CACHE_VOLUME=$(mktemp -d ${TMPDIR:-/tmp}/gocache.XXXXX)
-
-CLEANUP_DIRS=()
-
-cleanup() {
-  rm -rf ${CLEANUP_DIRS[@]}
-}
-
-trap cleanup EXIT
-
 dep_ensure() {
     local operator_dir=${1}
     local do_local=${2:-}
@@ -37,7 +23,15 @@ dep_ensure() {
         dep ensure -vendor-only -v
     else
         echo "Running 'dep ensure' with $GO_BUILDER_IMAGE"
-        required_dirs
+        local docker_gopath_volume=$(mktemp -d ${TMPDIR:-/tmp}/gopath.XXXXX)
+        local docker_cache_volume=$(mktemp -d ${TMPDIR:-/tmp}/gocache.XXXXX)
+        trap "rm -rf $docker_cache_volume $docker_gopath_volume" EXIT
+
+        mkdir_and_check_owner "$operator_dir/dep-cache"
+        mkdir_and_check_owner "$docker_cache_volume"
+        mkdir_and_check_owner "$docker_gopath_volume/pkg/dep"
+        mkdir_and_check_owner "$docker_gopath_volume/src/github.com/syndesisio/syndesis/install/operator"
+
         local extra_opts=""
         if [ -n "$as_user" ]; then
             extra_opts="$extra_opts -u $as_user"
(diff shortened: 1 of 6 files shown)


tools/bin/commands/util/operator_funcs
```diff
@@ -20,25 +37,52 @@ dep_ensure() {
...
     fi
 }
 
+required_dirs() {
+    require_dir "$operator_dir/dep-cache"
+    require_dir "$DOCKER_CACHE_VOLUME"
+    require_dir "$DOCKER_GOPATH_VOLUME/pkg/dep"
+    require_dir "$DOCKER_GOPATH_VOLUME/src/github.com/syndesisio/syndesis/install/operator"
+
+    CLEANUP_DIRS+=($DOCKER_GOPATH_VOLUME)
```
rhuss: You shouldn't remove cache dirs as they considerably will increase the build time. Maybe an option `--clean` would be better which would remove that dirs _before_ calling.

tools/bin/commands/util/operator_funcs
```diff
@@ -105,7 +163,7 @@ create_operator_image() {
 
     local build_dir=$(prepare_docker_context "$operator_dir")
     pushd $build_dir > /dev/null
-    trap "rm -rf $build_dir" EXIT
+    CLEANUP_DIRS+=($build_dir)
```
rhuss: this works if the function is not evaled via `$()`. Otherwise, CLEANUP_DIRS can't be easily updated.

tools/bin/commands/util/operator_funcs
```diff
@@ -20,25 +37,52 @@ dep_ensure() {
...
                    -v $operator_dir/dep-cache:/gopath/pkg/dep:z \
                    -v $operator_dir:/gopath/src/github.com/syndesisio/syndesis/install/operator:z \
                    $GO_BUILDER_IMAGE \
                    dep ensure -vendor-only -v
     fi
 }
 
+required_dirs() {
+    require_dir "$operator_dir/dep-cache"
+    require_dir "$DOCKER_CACHE_VOLUME"
```
rhuss: in the scripts we should avoid global variables as much as possible. Typically they are only used for constants, so would it be an issue to pass them via arguments around ?

tools/bin/commands/util/operator_funcs
```diff
@@ -6,6 +6,23 @@ GO_BUILDER_IMAGE="syndesis/godev:1.10"
 # Name of the operator image
 OPERATOR_IMAGE="syndesis/syndesis-operator"
 
+# globs to test
+TEST_TARGETS="./cmd/... ./pkg/..."
+
+# $GOPATH volume to share between docker runs
+DOCKER_GOPATH_VOLUME=$(mktemp -d ${TMPDIR:-/tmp}/gopath.XXXXX)
```
rhuss: Shouldn't that be set only when no `--local` is given ? Also I would add them already here to $CLEANUP_DIRS so is clear where it belongs together.

tools/bin/commands/util/operator_funcs
rhuss: Hasn't this directory just been created above, so its already there and empty ?

tools/bin/commands/util/operator_funcs
```diff
@@ -20,25 +37,52 @@ dep_ensure() {
...
+
+    CLEANUP_DIRS+=($DOCKER_GOPATH_VOLUME)
+    CLEANUP_DIRS+=($DOCKER_CACHE_VOLUME)
+}
+
+require_dir() {
+    local dir=${1}
+    [ -d "$dir" ] || mkdir -p $dir
+    if [ "$(stat --format '%u:%g' $dir)" != "$(id -u)":"$(id -g)" ]; then
+        echo "ERROR: $dir is not owned by the current user"
```
rhuss: why is the group important ?

tools/bin/commands/util/operator_funcs
rhuss: When you add here $DOCKER_GOPATH_VOLUME here to the CLEANUP_DIRS then it won't get delete when you use --local (but it's still created)

tools/bin/commands/util/operator_funcs
```diff
@@ -20,25 +37,52 @@ dep_ensure() {
...
+        exit 1
+    fi
+}
+
 gopath_dir() {
+    set +o nounset
     if [ -z "$GOPATH" ]; then
-        echo "ERROR: You have to set your GOPATH environment variable"
-        exit 1
+        if [ ! -d "$HOME/go" ]; then
```
rhuss: Never heard about that convention, but not harmful I guess ;-)

tools/bin/commands/util/operator_funcs
zregvart: Call me too cautious

tools/bin/commands/util/operator_funcs
zregvart: This is how we hipsters roll, $GOPATH that so yesterday

rhuss:
@zregvart I just merged your other PR, hence these conflicts. Could you please resolve them and then we can merge this PR, too. Thanks!

zregvart:
@rhuss got your hot rebase over here, approve while it's hot

Issue #3942, opened 2018-10-23
Remove # of flows badge from API integration icon


michael-coker:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
No longer need to display the badge on the API integration icon.

## Screenshot
<img width="1576" alt="screen shot 2018-10-23 at 10 05 45 am" src="https://user-images.githubusercontent.com/35148959/47370396-40388e00-d6ab-11e8-8303-99873e36e20c.png">
<img width="1576" alt="screen shot 2018-10-23 at 10 05 42 am" src="https://user-images.githubusercontent.com/35148959/47370397-40388e00-d6ab-11e8-9746-9d7d2543a033.png">

cc @gashcrumb @riccardo-forina @zregvart @gaughan @amysueg @dongniwang

dongniwang:
I updated it to Feature request instead of Bug report. 

Just wanted to provide the background story here, we have a few issues opened regarding the badge visual representation and the meaning of the badge.

- #3695 - the orange color indicates warning/error
- #3601 - what does the number mean? 

@syndesisio/uxd had a discussion on this and we recommend removing the badge altogether on both integration list page and integration details page. Here are the reasons:

- On the integration details page, we have help text underneath the return path icon indicating how many flows that integration has. 
- On the list page, we feel that the orange color badge draw too much attention from users while it should convey secondary information. 
- The meaning of the badge is not obvious to users at a first glance as the visual representation of it resemble notification (the badge is located at the top right corner of the icon). @michael-coker did some work on adding a tooltip to the badge but it requires additional action from users. 
- We have the API Provider icon and the return path icon to help users understand an integration is API provider integration, and users can always go to the integration details page to find more information about a particular integration. 
- We'll gather feedback from users regarding the badge usage/color as part of the feedback gathering user research. 

@gaughan - Do you have any other suggestions? 

@gashcrumb @zregvart @riccardo-forina - What do you all think about this?

gaughan:
@dongniwang agree on the orange part and the gathering feedback on it from user research. I think ultimately it would be nice to indicate the number of flows created automatically from the definition.

dongniwang:
@gaughan @zregvart @gashcrumb - Do we have the green light to remove it for now then? Please advice. Thank you!

zregvart:
[...]

PR #3943, opened 2018-10-23
fix(integration): remove flow count from api integration icon

michael-coker:
fixes https://github.com/syndesisio/syndesis/issues/3942


michael-coker:
@zregvart @gashcrumb looks like this is OK to merge - https://github.com/syndesisio/syndesis/issues/3942#issuecomment-432655625

PR #397, opened 2017-11-21
feat: Swagger Connector

zregvart:
Adds a Initial implementation of a Connector that uses Swagger
specification to invoke REST services.

Fixes #308

PR #3979, opened 2018-10-26
fix(integration): show return path icon as end connection in API integration

michael-coker:
fixes https://github.com/syndesisio/syndesis/issues/397


gashcrumb:
We'll definitely need  a PR on the 1.5.x branch too for this one

PR #3984, opened 2018-10-29
The use of OAuthClient

### LINKED ISSUE #3576: SSO Enhancements
See also **https://issues.jboss.org/browse/ENTESB-11525**

## This is a...

<pre><code>
[X] Feature request
[ ] Bug report  
</code></pre>

# Allow the 'authorize access' prompt at first login to be skipped

## Description
Syndesis should be configured with a full 'OAuthClient' in OpenShift. This will avoid the 'authorize access' prompt if grantMethod is set to 'auto'.

zregvart:
This removes the need for OAuth client service account and replaces it with OAuthClient object.

I've verified it manually to work. Having no idea what might break, i.e. if the provisioning to a cluster can break, would be good to take a closer look.

Fixes #3576


zregvart:
@rhuss @chirino @nicolaferraro any thoughts?

pretty-please

rhuss:
Sorry @zregvart still no time today. But could you please remind me tomorrow again when I'm back from Munich ?

zregvart:
> Sorry @zregvart still no time today. But could you please remind me tomorrow again when I'm back from Munich ?

Knock, knock @rhuss :laughing:

install/operator/deploy/syndesis-operator.yml
```diff
@@ -125,6 +146,24 @@ subjects:
...
+metadata:
+  name: syndesis-operator:oauth
+  labels:
+    app: syndesis
+    syndesis.io/app: syndesis
+    syndesis.io/type: operator
+    syndesis.io/component: syndesis-operator
+subjects:
+- kind: ServiceAccount
+  namespace: syndesis
```
zregvart: @rhuss is this ok?

zregvart: Probably should be removed, no `namespace` used for `RoleBinding`

install/operator/deploy/syndesis-operator.yml
```diff
@@ -125,6 +146,23 @@ subjects:
 roleRef:
   name: edit
 ---
+kind: ClusterRoleBinding
```
heiko-braun: @rhuss this is currently not present on any cluster, right? And thinking this further, we lack privileges to add new cluster level roles, so we have to run this by the OSO team.

zregvart: It's a hard requirement for this feature...

rhuss: Now I drilled down, sorry that I couldn't comment on this earlier. I really feel we must not have a cluster role binding required to install the operator.

This might change with the use of the OLM, but now this puts us in even more trouble (like the later fixes for multiple installations with creating multi-cluster roles, one per project, is a hackish workaround. That kind of cluster role really feels like a plain role, which is project dependent.

Personally, if I would have the choice and I would have to select between this feature (if it is not realizable differently) and requiring cluster role permissions, I would forgo the feature (its about not having to grant permissions as user, right ?)

rhuss: And I really, really apologize to not being able to make the review in a timely manner. shame on me. Just meeting/travel/whatnot overload. sorry.

Issue #3989, opened 2018-10-29
API client connector "review actions" step has a card nested in a card


michael-coker:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
"Review actions" step when creating an api client connector has a card nested in a card

## Expected behavior
That step should look like the rest of the steps in the wizard. Just a single card.

## Screenshot
<img width="1593" alt="screen shot 2018-10-29 at 4 14 34 pm" src="https://user-images.githubusercontent.com/35148959/47680828-fd3c5600-db95-11e8-9c9b-e461db00a3a4.png">

## Tasks involved / Steps to Reproduce

1. Create api client connector
2. go to "review actions" step

michael-coker:
cc @dongniwang

PR #3990, opened 2018-10-29
fix(customizations): remove nested card, unused css in api client connector review step.

michael-coker:
fixes https://github.com/syndesisio/syndesis/issues/3989


gashcrumb:
Rebase should get the checks working, does this need to go into 1.5.x?

zregvart:
I've reabased hope you don't mind, I wouldn't call this critical for 1.5.x but I think it poses no risk to have it there.

gashcrumb:
@zregvart Wonder why `meta` and `server` still seem to fail and then seemingly work?

gashcrumb:
@zregvart ah, never mind, transient failure...

PR #4004, opened 2018-10-31
fix: consider all container objects when unwrapping response

### LINKED ISSUE #3788: Cannot implement API Provider operation returning collection

zregvart:
When unwrapping, i.e. converting from Syndesis wrapped format with body and parameters, we should consider arrays as well as objects for JSON serialization of the body.

Fixes #3788

### DIFF
--- a/app/connector/support/processor/src/main/java/io/syndesis/connector/support/processor/HttpRequestUnwrapperProcessor.java
+++ b/app/connector/support/processor/src/main/java/io/syndesis/connector/support/processor/HttpRequestUnwrapperProcessor.java
@@ -69,7 +69,7 @@ public final class HttpRequestUnwrapperProcessor implements Processor {
                     return;
                 }
 
-                if (bodyData.isObject()) {
+                if (bodyData.isContainerNode()) {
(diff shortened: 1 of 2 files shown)


paoloantinori:
Thanks Zoran, just a reminder, if needed,  that during this Sprint we can merge only to address blockers.

heiko-braun:
@paoloantinori We have already merged `1.5.x` and I think changes like can go into master, just not `1.5.x`

heiko-braun:
And btw, #3788 is blocking :)

PR #4012, opened 2018-11-01
fix(template-step): Removes need for body prefix in template

### LINKED ISSUE #3894: Remove the need for Template Step variables to be prefixed with 'body.'
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description
Due to the way Mustache handles scoping of variables and their values, to be populated correctly, they must be located in a map nested inside the scoping map (keyed using 'body'). As such, for the template to work correctly the variables need to be all prefixed with 'body.'. This is an implementation detail and should really be unnecessary.

phantomjinx:
* TemplateStepHandler
 * Adds preprocessing to the template content to add in the 'body.' prefix
   to each symbol if not already present

* MustacheTemplatePreProcessor
 * Parses each token in the template string to check validity and prepend
   body prefix to symbols

* Update tests including adding extra tests that check the syntax of the
  template

* mustache-mode.ts
 * Modifies the regex for the symbol format in the UI's template step
   editor. This removes the body prefix from being required in the symbols.
 * Moves the syntax checking portion to the bottom of the loop since its
   current position means the last word / symbol is never checked.

fixes #3894

### DIFF
--- a/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/TemplateStepHandler.java
+++ b/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/TemplateStepHandler.java
@@ -36,21 +36,18 @@ import io.syndesis.common.model.DataShapeKinds;
 import io.syndesis.common.model.action.Action;
 import io.syndesis.common.model.integration.Step;
 import io.syndesis.common.model.integration.StepKind;
-import io.syndesis.common.util.StringConstants;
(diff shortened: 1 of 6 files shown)


phantomjinx:
@lburgazzoli @gashcrumb 
Heys guys, given there is both a java and javascript element to this, if you would like to review just your preferred elements as before, I would be grateful. Thx!

gashcrumb:
LGTM!  I'll let @lburgazzoli click the approve button :-)

Issue #3917, opened 2018-10-19
Upgrade Spring Boot due to vulnerabilities found


KurtStam:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[*] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
> 8 org.springframework:spring-core vulnerabilities found in app/pom.xml 3 days ago
> Remediation
> 
> Upgrade org.springframework:spring-core to version 4.3.18 or later.

We also need to coordinate versions with Camel, so this is a little more complex then changing the version number in the pom. On a quick try @zregvart reported some tests started failing. We prob want to do this sooner rather then later.

![yummy](https://user-images.githubusercontent.com/35576/47214139-4d006d80-d39d-11e8-827f-23b552787c21.jpg)

https://github.com/syndesisio/syndesis/network/alert/app/pom.xml/org.springframework:spring-core/open

zregvart:
This was done in #3878. Here's [the failed](https://circleci.com/gh/syndesisio/syndesis/46899) CI build.

Just for posterity the root cause seems to be:
```
java.lang.IllegalStateException: Failed to load ApplicationContext
Caused by: org.springframework.context.ApplicationContextException: Unable to start embedded container; nested exception is java.lang.NoSuchMethodError: io.undertow.servlet.api.DeploymentInfo.addServletContainerInitializer(Lio/undertow/servlet/api/ServletContainerInitializerInfo;)Lio/undertow/servlet/api/DeploymentInfo;
Caused by: java.lang.NoSuchMethodError: io.undertow.servlet.api.DeploymentInfo.addServletContainerInitializer(Lio/undertow/servlet/api/ServletContainerInitializerInfo;)Lio/undertow/servlet/api/DeploymentInfo;
```

KurtStam:
@zregvart says: This was done in #3878. Here's the [failed CI build](https://circleci.com/gh/syndesisio/syndesis/46899).

Just for posterity the root cause seems to be:

java.lang.IllegalStateException: Failed to load ApplicationContext
Caused by: org.springframework.context.ApplicationContextException: Unable to start embedded container; nested exception is java.lang.NoSuchMethodError: io.undertow.servlet.api.DeploymentInfo.addServletContainerInitializer(Lio/undertow/servlet/api/ServletContainerInitializerInfo;)Lio/undertow/servlet/api/DeploymentInfo;
[...]

PR #4023, opened 2018-11-05
[fix] Upgrade Spring Boot due to vulnerabilities found #3917

KurtStam:


oscerd:
The problem is always the same. We need to be in sync with Fuse Camel.

https://github.com/jboss-fuse/camel/blob/2.21.x.redhat-7-x/parent/pom.xml#L636

KurtStam:
@oscerd https://issues.jboss.org/browse/ENTESB-9322 references 1.5.16. Making sure that works for us.

oscerd:
We are in the same situation on standalone and openshift. As management want by the way. Approved.

PR #4035, opened 2018-11-06
Fix #4034: Resource id value cannot be passed for fhir read action

### LINKED ISSUE #4034: Resource id value cannot be passed for fhir read action
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description

When adding a mapping to id/value or id/valueString for the FHIR read action the integration fails with resourceType is missing. Resource type must be set from the connection details.

In addition we should simplify setting the id/value to be a top level property called "id" and have other options hidden under "complexId" property. Adding "version" property and getting rid of "summary" at the top level.

rkorytkowski:
#4034

### DIFF
--- a/app/connector/fhir/pom.xml
+++ b/app/connector/fhir/pom.xml
@@ -86,40 +86,6 @@
       <groupId>org.apache.camel</groupId>
       <artifactId>camel-fhir</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.apache.camel</groupId>
-      <artifactId>camel-fhir-api</artifactId>
-    </dependency>
-    <dependency>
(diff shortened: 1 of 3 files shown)


zregvart:
Seems we can now remove `camel-fhir-api` dependency from the `connector-fhir`:

```
[INFO] --- maven-dependency-plugin:3.0.2:analyze-only (default) @ connector-fhir ---
[WARNING] Unused declared dependencies found:
[WARNING]    org.apache.camel:camel-fhir-api:jar:2.21.0.fuse-720029:compile
```

rkorytkowski:
@zregvart removed transitive dependencies...

rkorytkowski:
PMD issues... will look into them.

PR #4037, opened 2018-11-06
chore: introduce wrapped JSON schema identifier

### LINKED ISSUE #3914: Add $id property to the JSON schema URI in the DataShape
## This is a...

<pre><code>
[ ] Feature request
[*] Bug report  
</code></pre>

## Description
We need to follow this up with a change that includes the syndesis JSON schema URI in the data shape so that `ResponseCustomizer` can detect the need to unwrap the response here:

zregvart:
This introduces the wrapped JSON schema identifier of `io:syndesis:wrapped`, we use it to detect if the response needs to be _wrapped_ i.e. convert `{"hello": "world"}` to `{"body": {"hello: "world"}}`.

Also tries to stay backward compatible. Because of this there is a possibility to detect some payloads as wrapped, and that's why it's better to rely on the schema identifier.

Fixes #3914


zregvart:
review happened in #4039

PR #4039, opened 2018-11-06
chore: introduce wrapped JSON schema identifier (1.5.x backport)

### LINKED ISSUE #3914: Add $id property to the JSON schema URI in the DataShape

zregvart:
This introduces the wrapped JSON schema identifier of `io:syndesis:wrapped`, we use it to detect if the response needs to be _wrapped_ i.e. convert `{"hello": "world"}` to `{"body": {"hello: "world"}}`.

Also tries to stay backward compatible. Because of this there is a possibility to detect some payloads as wrapped, and that's why it's better to rely on the schema identifier.

Fixes #3914

(cherry picked from commit 2bc2488680b716bce80068154979308fb08d1d98)


app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/ResponseCustomizer.java
```diff
@@ -57,6 +57,37 @@ public void setOutputDataShape(final DataShape dataShape) {
         isUnifiedDataShape = isUnifiedDataShape(dataShape);
     }
```
KurtStam: At what point can we remove checking the legacy way? Should we add a deprecate warning?

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/swagger/UnifiedJsonDataShapeGenerator.java
```diff
@@ -288,6 +288,8 @@ private static DataShape unifiedJsonSchema(final String name, final String descr
         }
 
         final ObjectNode unifiedSchema = JsonSchemaHelper.newJsonObjectSchema();
+        // we're using Draft 4, in Draft 6+ this should be `$id`
```
KurtStam: Can we move to Draft 6+ now?

zregvart: up to Atlasmap I think, and for another PR

app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/ResponseCustomizer.java
zregvart: The data shapes could be baked into Integration (like the model), I'd prefer to drop it, but I'm unsure if the integrations stop working at that point. It depends on how many integrations are living with the data shapes in JSON schema without the `id`.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/swagger/UnifiedJsonDataShapeGenerator.java
KurtStam: How about opening and issue in AltasMap for it and reference that here?

app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/ResponseCustomizer.java
KurtStam: You're saying to support legacy forever?

zregvart: @KurtStam what's your suggestion?

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/swagger/UnifiedJsonDataShapeGenerator.java
zregvart: Actually it us:

https://github.com/atlasmap/atlasmap/blob/70020a7f78feec388efb289352dc0593ef4eeee4/runtime/modules/json/core/src/main/java/io/atlasmap/json/inspect/SchemaInspector.java#L90

https://github.com/atlasmap/atlasmap/blob/70020a7f78feec388efb289352dc0593ef4eeee4/runtime/modules/json/core/src/test/resources/inspect/schema/flatprimitive-base-unrooted.json#L2

zregvart: Right, Mr. Smartypants

app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/ResponseCustomizer.java
zregvart: Actually, I just realized we added this in 1.5, so removing it won't impact anyone really. I'm going to drop it.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/swagger/UnifiedJsonDataShapeGenerator.java
KurtStam: Capt Obvious to you

PR #4042, opened 2018-11-06
Update to Telegram json file to change text in labels and descriptions.

TovaCohen:
I made some updates to the json file for the Telegram connector. One update corrects a typo. The other updates hopefully clarify the text and apply consistent style formats. 
Ready to merge.


app/connector/telegram/src/main/resources/META-INF/syndesis/connector/telegram.json
```diff
@@ -72,15 +72,15 @@
       "type": "MAVEN"
     }
   ],
-  "description": "Send and receive messages to/from a Telegram chat",
+  "description": "Receive and send messageses.",
```
oscerd: I think this is wrong, otherwise this looks good.

TovaCohen: Glad you caught that!

oscerd:
Yes, sure. I think it should be fine. Not sure if the release for 1.5.x has been already built. @heiko-braun @paoloantinori ?

paoloantinori:
we are not planning to add other commits to 1.5.x release, but since it often happens that exceptional situations pop up at the last second, what we can do is to keep track of commits we can consider do include, if the opportunity happens.

that's the best we can do for this specific request.

heiko-braun:
I think there will be another CR, but I need @janstey to confirm this. However it doesn't justify the inclusion since `1.5.x` is in blockers only mode. 

I see this fix is simple, unlikely with side-effects, but I would still prefer to stick to the "rules". Before the release is after the release. The next window comes quickly and we clan include it there.

PR #4059, opened 2018-11-08
fix: use HeaderFilterStrategy with all HTTP

### LINKED ISSUE #3865: Remove all non Syndesis message headers at the API provider start processor
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
We need to remove all non-Syndesis Camel message headers on reception of a HTTP request in the API provider-based integrations. If we don't they will be used by subsequent connectors/Camel components and they will cause issues.

Typical example is the `Host` header which will make the API client fail the request due to virtual host mismatch at the API service.

zregvart:
Previous attempts of removing Camel message headers did not pan out successful, this is mainly because it's a race between the Activity tracing interceptor that adds this to around every interceptor and the header-removal logic.

With this, a new approach of using/configuring HeaderFilterStrategy to the servlet (webhook & API provider), HTTP and API client connector eliminates the need for removing Camel message headers and deals with the issue on the HTTP protocol binding level.

As the HTTP4 component RestProducerFactory implementation overwrites any HeaderFilterStrategy configured on the component/endpoint a subclass was necessary to make this function.

So with this fix all HTTP interactions: be it HTTP requests sent out of integrations (HTTP connector, API client connector) or any HTTP responses received from integrations (webhook, API provider) should not
contain `Syndesis.*` HTTP headers regardless of their presence in the Camel message headers.

Fixes #3865

### DIFF
--- a/app/connector/rest-swagger/src/main/resources/camel-connector.json
+++ b/app/connector/rest-swagger/src/main/resources/camel-connector.json
@@ -16,7 +16,7 @@
   "outputDataType" : "json",
   "componentOptions" : [ "host", "basePath" ],
   "endpointValues" : {
-    "componentName" : "http4"
+    "componentName" : "connector-rest-swagger-http4"
   },
   "endpointOptions" : [ "operationId" ],
   "connectorProperties" : {
--- a/app/connector/support/processor/src/main/java/io/syndesis/connector/support/processor/SyndesisHttpConfiguration.java
+++ b/app/connector/support/processor/src/main/java/io/syndesis/connector/support/processor/SyndesisHttpConfiguration.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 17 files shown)


zregvart:
cc @paoloantinori @gashcrumb (I need to back port this to 1.5.x after review)

app/connector/support/processor/src/main/java/io/syndesis/connector/support/processor/SyndesisHttpConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Conditional;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@Conditional(SyndesisHttpConfiguration.NeedsHeaderFilterStrategy.class)
+@AutoConfigureBefore(name = "org.apache.camel.component.servlet.springboot.ServletComponentAutoConfiguration")
+public class SyndesisHttpConfiguration {
+
+    static class NeedsHeaderFilterStrategy extends AnyNestedCondition {
```
KurtStam: Add some comments as to for what these things are triggered

app/connector/rest-swagger/src/main/resources/camel-connector.json
```diff
@@ -16,7 +16,8 @@
   "outputDataType" : "json",
   "componentOptions" : [ "host", "basePath" ],
   "endpointValues" : {
-    "componentName" : "http4"
+    "componentName" : "connector-rest-swagger-http4",
+    "headerFilterStrategy" : "syndesisHeaderStrategy"
```
zregvart: remove

zregvart:
@KurtStam should be good, fancy another look?

PR #4067, opened 2018-11-09
[fix] Inform user when integration limit is reached when attempting t…

### LINKED ISSUE #246: Inform user when integration limit is reached when attempting to publish or activate
|</img> @rhuss | [2017-10-25](https://github.com/syndesisio/syndesis-ui/issues/1133) | UX Design  |
|-|-|-|

Currently, there is only a server side check for a maximum number of integrations. However, if the limit is reached, then some operations fails but the indication is only in the logs.

Instead, the UI controls should be disabled when the limit of actively running integrations is reached, i.e. the buttons "Publish Integration". When disabled, a tooltip should explain why it is disabled and that another integration needs to be stoped to publish / activate the integration.

The question is from where to get the limit. In the simplest case this would be from the ui configuration but for maintainability reason, it should come from the backend. Wonder whether we need some new endpoint or whether some kind of global 'confguration' endpoint already exists.

KurtStam:
…o publish or activate #246 

partial fix #246 
- Adds a /users/~/quota

```
{
maxIntegrationsPerUser: 2,
maxDeploymentsPerUser: 2,
usedIntegrationsPerUser: 2,
usedDeploymentsPerUser: 2
}
```

- Sets populates the statusMessage in the Integration deployment entity that is returned on the PUT to https://syndesis.192.168.42.91.nip.io/api/v1/integrations/{id}/deployments

screen shot 2018-11-09 at 2 54 45 pm

@gashcrumb Maybe you can use the statusMessage and display it in the UI?

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/Kind.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/Kind.java
@@ -56,6 +56,7 @@ public enum Kind {
     Permission(io.syndesis.common.model.user.Permission.class),
     Role(io.syndesis.common.model.user.Role.class),
     User(io.syndesis.common.model.user.User.class),
+    //Quota(io.syndesis.common.model.user.Quota.class),
 
     ConnectionBulletinBoard(io.syndesis.common.model.bulletin.ConnectionBulletinBoard.class),
     IntegrationBulletinBoard(io.syndesis.common.model.bulletin.IntegrationBulletinBoard.class),
--- a/app/common/model/src/main/java/io/syndesis/common/model/user/Quota.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/user/Quota.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.common.model.user;
+
+import java.io.Serializable;
+import java.util.Optional;
+
+import org.immutables.value.Value;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+import io.syndesis.common.model.ToJson;
+
+@Value.Immutable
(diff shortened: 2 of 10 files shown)


app/common/model/src/main/java/io/syndesis/common/model/Kind.java
```diff
@@ -56,6 +56,7 @@
     Permission(io.syndesis.common.model.user.Permission.class),
     Role(io.syndesis.common.model.user.Role.class),
     User(io.syndesis.common.model.user.User.class),
+    Quota(io.syndesis.common.model.user.Quota.class),
```
zregvart: do we need this, I kinda thought it was only needed for DB access

app/common/model/src/main/java/io/syndesis/common/model/user/Quota.java
```diff
@@ -0,0 +1,50 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+import io.syndesis.common.model.Kind;
+import io.syndesis.common.model.ToJson;
+import io.syndesis.common.model.WithKind;
+
+@Value.Immutable
+@JsonDeserialize(builder = Quota.Builder.class)
+@SuppressWarnings("immutables")
+public interface Quota extends WithKind, ToJson, Serializable {
```
zregvart: do we need `WithKind`?

app/common/model/src/main/java/io/syndesis/common/model/Kind.java
KurtStam: It's needed to generate the Immutable.

KurtStam:
> You musn't change the OpenShift templates directly, only via changes to install/generator/*.mustache and then run install/generator/run.sh otherwise the changes will be overwritten.

I knew that.. good catch

gashcrumb:
@KurtStam cool!  There's a missing space in your status message *however* I wouldn't worry about it too much as we'd want the string to be pulled from the i18n dictionary anyways, so you could use a much simpler string for the statusMessage.

But are you sure the PUT shouldn't just return an error status code instead?  Otherwise then I have to check for an error/warning in a non-error code path on my side, it's less ideal as it'll complicate that code a bit.

Also would you want me to push the related UI changes to this branch, or shall I do it in a subsequent PR?  Either way works for me.

KurtStam:
@gashcrumb The space was fixed already :), the screenshot wasn't. 

In the current design, it's not really an error condition: You're saving an integration, then lateron a controller will start building and deploying (and it finds out you used up your quota). Maybe we should just always display that status message somewhere in the UI so the user knows what's going. We can also add status updates from the controller onto the stream socket.

gashcrumb:
Gotcha, fair enough.  I'll work with Dongi to figure out that bit, so I think we should just merge this now and I'll create a second PR later on to finish up and close out the issue.

KurtStam:
OK will merge. As a sidenote. If you could increase the quota (or shutdown an other integration), the new integration would deploy by itself. So it's really just in a holding pattern. We just never tell the user..

Issue #4055, opened 2018-11-08
Missing icon custom connector after import


KurtStam:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[*] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
The icon should be exported along with the integration, so that on import it reappears

<img width="599" alt="screen shot 2018-11-08 at 11 09 07 am" src="https://user-images.githubusercontent.com/35576/48191951-c230e400-e346-11e8-9600-22615298ea17.png">

Todo API icon is missing on fresh install and re-import.

https://github.com/syndesisio/syndesis-quickstarts/raw/master/db-2-api-connector/DB-2-APIConnector-export.zip

zregvart:
@KurtStam can you take a look at the `src` of that image, should be something like:

```html
<img class="syn-icon-small" src="./../../assets/icons/multi-flow.connection.png" alt="1 Flow">
```

KurtStam:
@zregvart here you go

> <img _ngcontent-c8="" class="syn-icon-medium" src="/api/v1/connectors/i-LQcy-E1cre0V-4vObkwz/icon?db:i-LQcyc_Pcre0V-4vObkxz">

zregvart:
The thing is that the UI replaces icon (see #3974) and I guess then posts it as the integration is being saved. I think this should have been done on the backend when the integration is created so we can reference a predefined icon on the backend. 

Seems to me that what we have has two drawbacks:
 - we get a new icon but identical with each flow stored in the database
 - we need to implement support for exporting action icons in integration export

Since #3974 was a p0, should we consider this a p0 also and rework this?

cc @paoloantinori @gashcrumb

KurtStam:
Ready for QE

mastepan:
Fix verified on 1.6.1-20181203

PR #4088, opened 2018-11-14
[fix] Missing icon custom connector after import #4055

KurtStam:
Fixes issue 4055. With this fix it exports the icon model as well as the image as file. On integration import the model is restored into the JSONDB and the image is restored into the (Sql)FileStore if it didn't exist already.

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/support/IntegrationSupportHandler.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/support/IntegrationSupportHandler.java
@@ -188,7 +188,7 @@ public class IntegrationSupportHandler {
         }
         LOG.debug("Extensions: {}", extensions);
         LOG.debug("Icons: {}", icons);
-        
+
         return out -> {
             try (ZipOutputStream tos = new ZipOutputStream(out) ) {
                 ModelExport exportObject = ModelExport.of(Schema.VERSION);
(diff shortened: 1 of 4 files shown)


app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/support/IntegrationSupportHandler.java
```diff
@@ -156,27 +162,33 @@ public StreamingOutput export(@NotNull @QueryParam("id") @ApiParam(required = tr
...
                     .flatMap(flow -> flow.getSteps().stream())
-                    .collect(Collectors.toList()), true)
-                .stream()
+                    .collect(Collectors.toList()), true);
+            dependencies.stream()
                 .filter(Dependency::isExtension)
                 .map(Dependency::getId)
                 .forEach(extensions::add);
+            dependencies.stream()
+            .filter(Dependency::isIcon)
```
zregvart: nit: could be done in one pass via something like `filter(d -> d.isExtension() || d.isIcon())`

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/support/IntegrationSupportHandler.java
```diff
@@ -277,6 +300,19 @@ private void addResourcesToExport(final JsonDB export, final Integration integra
                     }
                 }
 
+                // import custom icons
+                if( entry.getName().startsWith("icons/") ) {
```
zregvart: nit: funny formatting

zregvart:
This is from the build:
```
[INFO] PMD Failure: io.syndesis.server.endpoint.v1.handler.integration.support.IntegrationSupportHandler:266 Rule:NPathComplexity Priority:3 The method 'importIntegration(SecurityContext, InputStream)' has an NPath complexity of 245.
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/support/IntegrationSupportHandler.java
KurtStam: Agreed but I didn't want to reformat the entire file.

PR #4101, opened 2018-11-16
fix: make sure that we don't overwrite spring.factories

zregvart:
Removes code generation and adds current generated files to `src/main/resources`. This way the META-INF/spring.factories` will not be overwritten by the code generation.


zregvart:
Actually let me filter the versions in JSON descriptors

zregvart:
@oscerd I think this should be OK now, can you take another look?

oscerd:
LGTM

PR #4110, opened 2018-11-19
Fix #4109: Add support for maven-mirror to syndesis build

### LINKED ISSUE #4109: Add support for maven-mirror to syndesis build
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description

Enhance the build script to accept the maven-mirror parameter and the MAVEN_MIRROR_URL environment variable allowing to use a maven mirror for builds. If you setup a maven proxy within OpenShift, you can speed up builds by caching build dependencies as described at https://docs.openshift.com/container-platform/3.5/dev_guide/app_tutorials/maven_tutorial.html

rkorytkowski:


tools/bin/commands/build
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-
+set -eu
```
rhuss: normally such kind of global settings should be dealt with in the top level "syndesis" script (where there is already a set -eu https://github.com/syndesisio/syndesis/blob/0420a964654c8ac448476f394858234ae0ca61d9/tools/bin/syndesis#L20-L21

Modules ideally should only hold functions.

Issue #3649, opened 2018-09-21
Extending the capabilities of the Template Step


phantomjinx:
Building on the features produced in [#3047](https://github.com/syndesisio/syndesis/issues/3047), enhance the Template Step to:
- Include the use of alternative template libraries, ie. FreeMarker and Velocity;
- Follow the UI modifications as laid out in [UX Design v3](https://redhat.invisionapp.com/share/MKNX2N6SVE6) (as opposed to v3.1 which only provided for mustache);
- Dynamic runtime dependencies whereby the choice of library dictates which camel library the runtime integrations are build against, ie. camel-mustache, camel-velocity, camel-freemarker (currently camel-mustache is always included).

mmuzikar:
Currently verifying this feature, I believe this might be part of #3598 but I'd like to explicitly state that comments are included in the output of the template step.

tplevko:
Verified https://issues.jboss.org/browse/FUSEQE-2943

PR #4111, opened 2018-11-19
Enhancing the template step

phantomjinx:
Functionality extended as per [#3649](https://github.com/syndesisio/syndesis/issues/3649).

@gashcrumb @lburgazzoli
Has both UI js elements and backend java elements.

This is for 7.3 so no rush to review and merge (just wanted to get on with it while familiar with code). If any reservations then plenty of time to get it right so feel free to review-with-fine-tooth-comb.


app/common/model/src/main/java/io/syndesis/common/model/integration/step/template/TemplateStepLanguage.java
```diff
@@ -0,0 +1,224 @@
...
+ * and {@link TemplateStepPreProcessor} implementation defined
+ * for use by the template step handler and project generator.
+ */
+public class TemplateStepLanguage {
+
+    public static final String LANGUAGE_PROPERTY = "language";
+
+    public static final TemplateStepLanguage MUSTACHE = new TemplateStepLanguage(
+             "MUSTACHE",
+             "org.apache.camel:camel-mustache:2.21.0.fuse-720007",
```
lburgazzoli: We should not have hard-coded version

phantomjinx: @lburgazzoli 
That's the GAV that goes into the maven dependencies (dynamically). How should that be referenced instead as it seems that the 3 elements (group, id, version) are all required?

app/common/model/src/main/java/io/syndesis/common/model/integration/step/template/TemplateStepLanguage.java
```diff
@@ -0,0 +1,224 @@
...
+            return step;
+        }
+
+        Map<String, String> properties = step.getConfiguredProperties();
+        TemplateStepLanguage language = stepLanguage(
+                                                                    properties.get(LANGUAGE_PROPERTY));
+
+        Dependency dependency = Dependency.maven(language.mavenDependency());
+        return step.builder()
+                     .dependencies(Collections.singleton(dependency))
```
lburgazzoli: maybe here you can use `addDependency` as it will add dependencies to those already set on the step, unless the goal is to always replace them.

phantomjinx: They do need replacing as they are mutually exclusive. For example, if I change the language syntax of an existing step then I want to the language dependency replaced completely.

app/common/model/src/main/java/io/syndesis/common/model/integration/step/template/TemplateStepLanguage.java
phantomjinx: I'll try it without the version then and see if everything is happy then.

PR #4116, opened 2018-11-19
fix(tooltip glyphicon): tooltip misaligned with glyphicon.

### LINKED ISSUE #4094: Tooltips are off-center
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem
Our tooltip arrow is off center relative to the thing it points to.

## Expected behavior
The arrow should be centered vertically or horizontally depending on the tooltip position.

## Screenshot
screen shot 2018-11-15 at 8 35 52 am
screen shot 2018-11-15 at 8 38 20 am

1. You can see the first screenshot when creating an integration, or anywhere you see the info icon beside a form field label.
2. The second is on the Settings page, but you might be able to find other places with top/bottom tooltip placement.

jameslau:
fixes #4094

- glyphicon css needed to be overwritten to help align itself with the tooltip on hover.

screen shot 2018-11-19 at 3 26 03 pm

### DIFF
--- a/app/ui/src/scss/utils/_overrides.scss
+++ b/app/ui/src/scss/utils/_overrides.scss
@@ -154,3 +154,8 @@ body.cards-pf {
     }
   }
 }
+
+// Tooltip styling override
+.glyphicon-info-sign {
+  display: block;
+}


mcoker:
@jameslau looks like it's still an issue

screen shot 2018-11-19 at 3 04 14 pm

It looks like there is a larger issue with the position of the tooltip than altering one of the types of elements that triggers it. It also looks like the arrow itself in the tooltip is off-centered inside of the tooltip.

mcoker:
In the tooltip in your screenshot, there is more space above the arrow in the tooltip than below it, and in my screenshot there is more space to the left of the arrow than to the right.

jameslau:
Thanks for pointing that out @mcoker! I'll take another look at it.

mcoker:
@jameslau you bet :) and it probably wasn't clear in my original issue that it looks like this might be app-wide, or at leasts exists outside of the specific instances I've included screenshots of. Those screenshots were just a couple of places I've seen it. I expect it's off-center in other parts of the app, too.

mastepan:
I can't see this fully fixed in 1.6.1-20181212. 

This still works the same way:
> screen shot 2018-11-19 at 3 04 14 pm

jameslau:
Hi @mastepan, I'll take another look at it with @mcoker. Thanks!

mcoker:
@mastepan I looked into this more, and it looks like an issue with the ngx-bootstrap library and how it positions that arrow. I [submitted a PR to ngx-bootstrap](https://github.com/valor-software/ngx-bootstrap/pull/4850) to fix it and it was merged, but is still being verified. Once that fix is included in a release of ngx-bootstrap, we can consume it in syndesis and revisit this alignment. Also, if they haven't included this fix in a new release of ngx-bootstrap by the end of this sprint, we'll just put in a manual fix until it's fixed upstream.

mcoker:
This is similar to my comment on https://github.com/syndesisio/syndesis/pull/4121.

PR #4121, opened 2018-11-20
fix(tooltip): adding CSS override for tooltip on hover above and below parent div.

### LINKED ISSUE #4094: Tooltips are off-center

jameslau:
fixes #4094.

- Culprit with the misalignment was due to margin settings. It needed an override.

**before css override**
screen shot 2018-11-20 at 3 24 07 pm

**after css override**
screen shot 2018-11-20 at 3 21 47 pm

### DIFF
--- a/app/ui/src/scss/utils/_overrides.scss
+++ b/app/ui/src/scss/utils/_overrides.scss
@@ -159,3 +159,15 @@ body.cards-pf {
 .glyphicon-info-sign {
   display: block;
 }
+
+// Centering tooltip styling above and below parent div
+.bs-tooltip-top .arrow,
+.bs-tooltip-bottom .arrow {
+  margin: 0 !important;
+  transform: translateX(-50%);
+}
+.bs-tooltip-left .arrow,
+.bs-tooltip-right .arrow {
+  margin: 0 !important;
+  transform: translateY(-50%);
+}


app/ui/src/scss/utils/_overrides.scss
```diff
@@ -158,4 +158,16 @@ body.cards-pf {
 // Tooltip styling override
 .glyphicon-info-sign {
```
mcoker: I don't think this is necessary anymore right?

mcoker:
Looks like these two popovers aren't centered relative to the element? I'm not sure, maybe they are :) But the arrow doesn't align with the double arrow in the IVP expand/collapse button or the + icon in the empty connection circle icon.

Can you also just poke around the app and see if you can find more instances?

screen shot 2018-11-20 at 4 50 47 pm
screen shot 2018-11-20 at 4 49 59 pm

mastepan:
@jameslau any progress with this one? Just asking if we can test it soon :)

mcoker:
@mastepan Thanks for following up. I'm going to go ahead and close this one for now. This impacts both the tooltips (https://github.com/syndesisio/syndesis/issues/4094) and popovers (https://github.com/syndesisio/syndesis/issues/4122) in fuse online. We discovered this is an upstream issue with ngx-bootstrap, so I have an issue/PR open with them to address it. If they aren't able to address it before the end of the next sprint, we'll submit another PR to fix both the tooltips and popovers.

mcoker:
@mastepan this was fixed in https://github.com/syndesisio/syndesis/pull/4250

PR #4144, opened 2018-11-23
chore: Fix SYNDESIS_REGISTRY env var removal for productisation

rhuss:


heiko-braun:
@rhuss please backport to 1.5.x as well, just in case

rhuss:
This has been already applied as part of #4155 , so closing it here.

PR #4155, opened 2018-11-23
fix: Several issues affecting productisation

rhuss:


avano:
@rhuss I think the upgrade scripts tweaks should be done in `master` as well, as currently the upgrade won't work with oc version > 3.9

rhuss:
@avano good point, created #4220 for a backport/forward port to master.

PR #4157, opened 2018-11-23
Upgrades Jackson to 2.9.7

zregvart:
Bit of detail in the commit messages:

fix: include all nodes in JSONDB test

Seems that the test was written with the idea that all nodes, even
`null` ones will be included in serialization/deserialization, so this
configures the right inclusion. As surfaced by Jackson upgrade.

fix: preloads the Swagger 2.0 JSON schema

Seems that Jackson upgrade triggered Swagger 2.0 schema download from
`http://swagger.io/v2/schema.json#`, this is unwanted we provide the
schema file embeded in the artefact.

fix: remove accessTokenExpiresAt.description

It is empty and in `io.syndesis.common.util.Json` we serialize only non
empty nodes. Change introduced with Jackson 2.9.7

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -54,7 +54,7 @@
 
     <hibernate.validator.version>5.3.5.Final</hibernate.validator.version>
 
-    <jackson.version>2.8.11</jackson.version>
+    <jackson.version>2.9.7</jackson.version>
     <json-patch.version>1.9</json-patch.version>
     <kubernetes.client.version>3.1.4.fuse-710001</kubernetes.client.version>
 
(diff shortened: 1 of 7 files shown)


zregvart:
cc @syndesisio/ui-api: we changed a version of our JSON library, if you get any funny issues it could relate to this.

One thing I noticed is that the definition of non-empty changed so keys with `""` values will not be present. It's easy to change, but I think we wanted it from the beginning like this, so I just fixed tests on the backend.

https://github.com/syndesisio/syndesis/blob/2c9bf265a5cd3fd2fb3ed48f6a278e85700373da/app/common/util/src/main/java/io/syndesis/common/util/Json.java#L43

PR #4163, opened 2018-11-26
[chore] upgrading to latest activemq

KurtStam:


zregvart:
Finding your way to conventional commits @KurtStam? What's up with the squares?

PR #4174, opened 2018-11-28
[closes #4173] Fix parsing minor version

### LINKED ISSUE #4173: [Upgrade] Cannot extract minor version from "1.1"
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

## Description
Just a cosmetical issue, the script can't parse minor version from string like "1.0"

avano:
#4173

### DIFF
--- a/tools/upgrade/migration/preflight.sh
+++ b/tools/upgrade/migration/preflight.sh
@@ -56,6 +56,11 @@ extract_minor_version() {
         echo "latest"
         return
     fi
+    if [[ $version == [0-9]\.[0-9] ]]; then
+        # Already minor version, no need to parse
+        echo $version
+        return
+    fi
     local minor_version=$(echo $version | sed 's/^\([0-9]*\.[0-9]*\)\.[0-9]*\(-.*\)*$/\1/')
     if [ "$minor_version" = "$version" ]; then
         echo "ERROR: Cannot extract minor version from '$version'"


tools/upgrade/migration/preflight.sh
```diff
@@ -56,6 +56,11 @@ extract_minor_version() {
         echo "latest"
         return
     fi
+    if [[ $version == [0-9]\.[0-9] ]]; then
```
rhuss: This doesn't work for multi digit versions like 1.10. (Also it uses a bashism which I tend to avoid, but thats only me :). Instead I would make the sed expression more robust. E..g instead of `minor_version = version` check, just check whether minor_version matches [0-9]+.[0-9]+.

avano: @rhuss bump :-) what do you thinj about that one ^?

rhuss: "expr match" is unfortunately not Posix compliant. From my man page

> According to the POSIX standard, the use of string arguments length, substr, index, or match produces undefined results. In this version of expr, these arguments are treated just as their respective string values.

and for me on macOs yields:

```
$ expr match "1.2" "^[0-9]\+\.[0-9]\+"
expr: syntax error
```

What's about 

```
local minor_version=$(echo "${version}XX" | sed 's/^\([0-9]*\.[0-9]*\).*$/\1/')
if [ $minor_version = "${version}XX" ]; then 
    echo "ERROR: Cannot extract minor version from '$version'"
fi

```

?

avano: how simple can that be :man_facepalming: I'll fix that, thank you Mr. Shell Ninja! :-)

PR #4189, opened 2018-11-29
Update version in dictionary

TovaCohen:
Apologies that this is so late. 
I updated the version number in the dictionary file. Without this update, when Fuse Online 7.2 is released, Fuse Online UI help links to user documentation will target the 7.1 user doc and not the 7.2 user doc. 

And, oh gosh, no, I do not want to merge 170 commits, just my one commit. I don't know what I did wrong.


heiko-braun:
@TovaCohen Which is the one file you wanted to change?

TovaCohen:
syndesis/app/ui/src/assets/dictionary/en-GB.json

Just search for "7.1" and change it to "7.2". There is just one instance than needs to change.

PR #4190, opened 2018-11-29
Update 3620: Google Sheets Connector

christophd:
Update on Google sheets connector
- Adding basic chart support for creating charts in spreadsheets
- Add option to split spreadsheet value ranges so each row is handled in a separate exchange


christophd:
@oscerd how about pushing those changes to Camel upstream? Use another issue than CAMEL-12950 as it has been resolved with v2.23.0 release

oscerd:
Good idea, please raise an upstream JIRA issue for this and open a PR :-)

PR #4197, opened 2018-12-01
Several upgrades to address issues reported by Snyk

zregvart:
Please have a look at the versions, some of them have small some of them have large jumps. I've tested locally and it seems to be OK.


gashcrumb:
I haven't been able to test it locally still but I trust that if it doesn't blow up for you then it should be fine, we can always revert if it's not.

PR #4209, opened 2018-12-07
feat: Have the syndesis build -m operator use build the operator on docker without needing multi-stage builds.

chirino:


tools/bin/commands/util/operator_funcs
```diff
@@ -113,58 +113,25 @@ build_operator_local() {
...
-               -v "$docker_gopath_volume":/gopath:z \
-               -v "$docker_cache_volume":/.cache:z \
-               -v $operator_dir/dep-cache:/gopath/pkg/dep:z \
-               -v $operator_dir:/gopath/src/github.com/syndesisio/syndesis/install/operator:z \
-               -e CGO_ENABLED=0 \
-               $GO_BUILDER_IMAGE \
-               sh -c "$cmd"
+    cd $operator_dir
+    docker build -t syndesis-operator-builder . -f Dockerfile-builder
+    docker run syndesis-operator-builder cat /syndesis-operator > syndesis-operator
```
rhuss: Nice trick, but I would prefer 'docker cp' on the finished builder container to avoid potential encoding / pipe issues on different Linux. I think it makes it cleaner and a bit more robust to use native Docker means:

```bash
local id=$(docker create syndesis-operator-builder)
docker cp $id:/syndesis-operator syndesis-operator
```

chirino: Why would encoding be taking place?  I've not seen any issues with encoding BTW.  Also we would need to delete the container right?  Doing a quick run seems easier.

rhuss: `docker run` is `docker create` + `docker start` and you don't clean up the container either afterwards (you could add a --rm though).

As mentioned, it's just a gut feeling to stick to a `docker cp` instead of a `cat` as feels more binary-ish, but if you insist of doing it that way, I'm fine, too. Its an easy change we could add if it breaks anyway, no need to waste energy here.

zregvart:
Not sure why @pure-bot[bot] didn't merge this one...

Issue #4206, opened 2018-12-05
Connector data shape inspections not generated in clean build


christophd:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

When doing a clean build the connector inspections for Java typed input/output data shapes are not generated. 

This is leads to data mapper steps not being able to determine source and target data shapes at runtime. You will get `No source data type was found. Data Mapper requires at least one data type aware step prior to itself` errors when adding data mapper steps to your integrations.

christophd:
I think this is caused by the `generate-connector-inspections` mojo in `connector-support-maven-plugin` Maven plugin. The plugin execution has been linked to the `generate-resources` phase where the required connector `*.json` has not been copied to the classes folder yet.

Before the plugin execution has been bound to the `prepare-package` phase.

christophd:
@zregvart any idea why we have changed the default phase for this plugin execution?

christophd:
I think we should bind the plugin execution to `copy-resources` phase to make sure that the connector meta `.json` has been copied to the classes output directory before the plugin executes. Or go back to `prepare-package` phase.

zregvart:
> I think we should bind the plugin execution to `process-resources` phase to make sure that the connector meta `.json` has been copied to the classes output directory before the plugin executes. Or go back to `prepare-package` phase.

Sounds like an easy fix.

christophd:
Fixed with #4210

PR #4210, opened 2018-12-10
#4206 Fix generation of connector data shape inspections

christophd:
Make sure that connector meta `*.json` has been copied to target classes when plugin performs inspection generation. Bind plugin execution to phase `compile` because generation Mojo is using source class files for its inspection, too.

### DIFF
--- a/app/connector/activemq/pom.xml
+++ b/app/connector/activemq/pom.xml
@@ -203,7 +203,7 @@
         <executions>
           <execution>
             <id>inspections</id>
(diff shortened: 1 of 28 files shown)


zregvart:
@christophd would `process-classes` be more suitable?

christophd:
@zregvart I agree `process-classes` is the better option as the phase is located right after compilation of classes. Let me change that and add this to the PR.

christophd:
@zregvart done

PR #4221, opened 2018-12-12
Doc 1.5 changes

heiko-braun:
This patch carries over the docs changes from master into 1.5.x


heiko-braun:
@TovaCohen Can you double check that this covers the intended contents?

heiko-braun:
Thanks @TovaCohen Good catch, let me check how these sneaked in.

PR #4233, opened 2018-12-20
Add kudu connector

lgarciaaco:


heiko-braun:
@lgarciaaco Isn't that already merged? See #4231

lgarciaaco:
@heiko-braun it was intended that way, but I think @zregvart pushed to master and changes didn't go through, sorry for the confusion!!

PR #4235, opened 2018-12-21
Enable integration testing on Google sheets connector

christophd:
Integration tests have been disabled because of required connectivity to productive Google sheets API. I enabled those tests to run with each build by implementing a simulated Google sheets server that integration tests connect with instead. 

The test server automatically starts on a random port at test runtime and provides full OAuth2 capabilities.

In order to make this possible I had to use configurable root URL and trust certificates on Google sheets client library.

In addition to that I have prepared Google sheets server simulation in Openshift deployment. The connector needs to provide some server properties adjustable via Syndesis UI for the root URL and server certificate validation.


oscerd:
Really nice. This would be super cool to add to the other google connectors!

app/connector/google-sheets/src/test/resources/test-options.properties
```diff
@@ -18,8 +18,8 @@
...
 ## Application client id and secret
-clientId=
-clientSecret=
+clientId=syndesis-client
+clientSecret=syndesis
 applicationName=camel-google-sheets/1.0
-accessToken=
-refreshToken=
+accessToken=cd887efc-7c7d-4e8e-9580-f7502123badf
+refreshToken=bdbbe5ec-6081-4c6c-8974-9c4abfc0fdcc
```
zregvart: These are not valid, right?

app/connector/google-sheets/pom.xml
```diff
@@ -37,6 +37,7 @@
 
     <google-api-client-version>1.22.0</google-api-client-version>
     <google-api-services-sheets-version>v4-rev551-1.22.0</google-api-services-sheets-version>
+    <citrus.version>2.7.8</citrus.version>
```
zregvart: would pull this into connector POM

app/connector/google-sheets/pom.xml
```diff
@@ -168,6 +169,94 @@
...
+      <artifactId>spring-security-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.consol.citrus</groupId>
```
zregvart: it would probably be a good idea to pull citrus dependencies into dependencyManagement of the connector parent POM

app/connector/google-sheets/src/main/resources/META-INF/syndesis/connector/sheets.json
```diff
@@ -820,18 +806,97 @@
...
+      "kind": "parameter",
       "raw": true,
       "required": true,
       "secret": false,
+      "tags": [
+        "oauth-additional-query-parameters"
+      ],
+      "type": "hidden"
+    },
+    "validateCertificates": {
```
zregvart: perhaps we would like to prevent users from changing this (make it `hidden`)?

christophd: Yeah, this feels a bit ugly. But how would a tester then be able to set this property when using a simulator deployed to Openshift in the test environment?

app/connector/google-sheets/src/test/resources/test-options.properties
christophd: Yes, right

PR #4250, opened 2019-01-08
chore: version upgrades import cleanups

zregvart:
Upgrades a number of versions (look in package.json) and while doing that makes the assumption of the caret[1] version explicit by using tilde versions[2] instead.

Cleans up a number of deep imports.

Refactors RxJS deprecation warnings away caused by @ngrx/store upgrade (probably).

Could not update ngx-bootstrap to 3.1.x, encountered big issues with that, most likely due to the build change (they adopted ng-packagr).

[1] https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004
[2] https://docs.npmjs.com/misc/semver#tilde-ranges-123-12-1


zregvart:
cc @syndesisio/ui-api

gashcrumb:
For `ngx-bootstrap` if I recall correctly we kinda have to stick with the exact version patternfly ships with, otherwise we see odd runtime-only breakages, or breakages only with the production console build etc.  So just as well that upgrade couldn't be done :-)

gashcrumb:
Just an update, it took me awhile to get a running environment to test this.  I have it running and so far so good!  I've a couple more things I would like to test this morning still.

@rareddy I think this could have an impact on your work integrating the beetle studio bits, so just a heads up.

gashcrumb:
Oh right, I also meant to try the 3.1.x upgrade for `ngx-bootstrap` for @mcoker, will try that too today.

zregvart:
> Oh right, I also meant to try the 3.1.x upgrade for `ngx-bootstrap` for @mcoker, will try that too today.

Good luck!

rareddy:
@gashcrumb for the beetle studio since we are slated to be going through a lot of UI changes IMO we will adapt to new versions you have.

gashcrumb:
@mcoker @zregvart pushed a commit that gets ngx-bootstrap 3.1.4 installed.  The tooltip that shows up when you hover doesn't look right, I suspect we could have a local workaround in place perhaps that's effecting this:

image

zregvart:
Thanks @gashcrumb @mcoker :raised_hands:

Issue #4266, opened 2019-01-10
Unstable GoogleSheets integration test breaks build


christophd:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

GoogleSheets integration test is unstable failing from time to time with following error:

```
[INFO] Running org.apache.camel.component.google.sheets.stream.SheetsStreamConsumerIntegrationTest
[ERROR] Tests run: 3, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 8.897 s <<< FAILURE! - in org.apache.camel.component.google.sheets.stream.SheetsStreamConsumerIntegrationTest
[ERROR] testConsumeValueRange(org.apache.camel.component.google.sheets.stream.SheetsStreamConsumerIntegrationTest)  Time elapsed: 7.586 s  <<< ERROR!
org.apache.camel.CamelExecutionException: Exception occurred during execution on the exchange: Exchange[ID-8a10ca970798-1547073284681-14-1]
	at org.apache.camel.component.google.sheets.stream.SheetsStreamConsumerIntegrationTest.testConsumeValueRange(SheetsStreamConsumerIntegrationTest.java:59)
Caused by: org.apache.camel.RuntimeCamelException: java.lang.IllegalArgumentException: no JSON input found
Caused by: java.lang.IllegalArgumentException: no JSON input found

[ERROR] testConsumeValueRange(org.apache.camel.component.google.sheets.stream.SheetsStreamConsumerIntegrationTest)  Time elapsed: 7.586 s  <<< ERROR!
com.consol.citrus.exceptions.TestCaseFailedException: java.lang.NullPointerException
Caused by: com.consol.citrus.exceptions.CitrusRuntimeException: java.lang.NullPointerException
Caused by: java.lang.NullPointerException
```

christophd:
Maybe leave this issue open until we have seen some builds to complete without that error. Maybe one week?

christophd:
Issue seems to be solved. There were no further failing builds with that particular reason

christophd:
We have to further investigate on this as @alexkieling came up with some failing tests locally:

```
[INFO] Running org.apache.camel.component.google.sheets.stream.SheetsStreamConsumerIntegrationTest
[...]

PR #4267, opened 2019-01-10
#4266 Fix unstable build

christophd:
Try to fix unstable build by making sure that GoogleSheetsApi mock server is started before tests run. I could not reproduce that failing behavior locally but it is likely the cause for the failing build on CircleCi.


oscerd:
ping @zregvart

Issue #4255, opened 2019-01-08
Support deleting operations


zregvart:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

When editing OpenAPI specification of a API provider based integration, user can opt to delete an operation. Deleting the operation should result in the removal of a flow from the integration.

We need a new endpoint on the backend to receive the updated OpenAPI specification of the integration and update the integration.

The frontend needs to invoke that endpoint when the user is done editing the OpenAPI specification in Apicurio and reload the integration.

asmigala:
Verified with CR3

PR #4270, opened 2019-01-10
Ability to edit API provider's OpenAPI specification

### LINKED ISSUE #4253: Support data shape change
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description

When the user modifies the OpenAPI specification of a API provide based integration in a way that will modify the data shapes, a mapping step might need to be updated. 

Currently we get an error when opening the mapping step (`Could not find field from document for mapped field ...`).

We need the ability to let the user know that the data shape was changed in a incompatible way. If the data shape was changed but in a compatible way I don't think we need to notify the user for that.

I think this is functionality needs some support in Atlasmap.

Overall it it could work like this:
 - given an integration,
    - go through all the flows of an integration
      - go through all mapping steps of a flow
         - check if the existing mapping is still valid given the updated data shapes

I think we need to notify the user both in the operation (flow) list (perhaps a exclamation icon with a tooltip saying that mapping steps need updating) and within the integration editor (perhaps a exclamation icon next to the mapping step icon in the integration view pane).

This functionality is also beneficial to other use cases, for instance when user defines the data shape (e.g. when using message queue connection) and later, after mapping, changes that data shape.

cc @igarashitm, @syndesisio/uxd

zregvart:
This is an effort to implement the functionality of editing OpenAPI specification from which the API provider-based integration is created and have the changes reflected in the integration flows. Creating a PR just to get anyone interested in an early review to take a look.

Still to do

- [x] Implement API endpoint receive the updated OpenAPI specification and update the flows
- [x] Store the updated OpenAPI specification resource
- [x] Implement API to fetch the stored OpenAPI specification resource
- [x] Implement editing of OpenAPI specification in the UI
  - [x] Add the button to edit OpenAPI specification in the integration operation list view
  - [x] Fetch the OpenAPI specification and feed it to Apicurio
  - [x] On save in Apicurio post to the endpoint to update the OpenAPI specification and the Integration flows
  - [x] On save reload the Integration

Fixes #4253, #4254, #4255, #4256 #4323


lburgazzoli:
> And sorry for all the reformatting in `common-model` was hurting my OCD.

There's nothing wrong with reformatting code, but I think we should avoid that as much as possible. Of course I have to blame myself because I've done that too :) but massive reformatting hurts reviews as it's very hard to distinguish between cosmetic and functional changes. 

IMHO, it would be better to do it in a separate commit/follow up PR.

zregvart:
@gashcrumb if you have the time you might want to take a look at 2e5ffcad825bd2a7e43222a2c1d2be8a138de1e5 and d81648b98a919d6bd1dff6aeadb2817707dd5f95, I've done a first pass at implementing this. Quite pragmatic and somewhat functional.

Issue #3918, opened 2018-10-19
Application dashboards


gaughan:
As an expert user of Fuse Online I want to be able to browse to my Grafana instance and view metrics pre configured fro my Fuse environment.
I want my Grafana instance to be display metric per integration deployed.
Metrics we would like to see (per intergration)
	CPU usage
	Memory usage
	Average time to execute
	Min time to execute
	Max time to execute
	Average payload size	
	Min payload size
	Max payload size

Sample dashboard:

![image](https://user-images.githubusercontent.com/674767/47219372-a3a97f80-d37c-11e8-98a3-5014fc727fd0.png)

DoD
	Grafana should be installed and preconfigured for prometheus
	Grafana should be automatically connected to prometheus
	A single Fuse datasource is preconfigured in Grafana
	A single Fuse Dashboard is available exposing all prometheus fuse metrics and loads on navigating to the instance
	Documentation of how to create a custom dashboard is created

The feature owner has confirmed one of the following:
--There is an initial draft of the user doc for this feature.
--The writer acknowledges receiving enough information from engineering to write the initial draft of the user doc for this feature.

alexkieling:
Merged into https://github.com/syndesisio/syndesis/issues/4146

alexkieling:
I'm assuming we want to show JVM metrics (memory, threads, etc.) and Camel metrics (messages processed, processing time, etc.).

gaughan:
Are JVM metrics handled in the infrastructure?

alexkieling:
There is a infrastructure dashboard that shows the JVM metrics exposed by the Syndesis Server and Syndesis Meta applications.
The application dashboard will show the JVM and Camel metrics of each integration. We plan to add a drop-down to the top of the dashboard to select what integration we want to see the metrics.

gaughan:
Sounds good, thanks alex

zregvart:
@mmelko has soak tests running on staging (in `syndesis-test` project), would be good to have these tested there also.

alexkieling:
@heiko-braun If this feature is targeted at expert Fuse Online user, shouldn't it be available in the Integration Metrics page?

mkralik3:
[...]

PR #4277, opened 2019-01-14
feat: Add Grafana to syndesis install #3918.

jamesnetherton:
Initial stab at adding Grafana to the install process for #3918. The finer detail for some of the config options will likely need optimizing at some point. But, this PR serves as a baseline for getting something up and running, which others can use when working on related monitoring / metrics tasks.


install/operator/pkg/syndesis/configuration/configuration_test.go
```diff
@@ -88,6 +88,15 @@ func TestSpecificConfig(t *testing.T) {
 						VolumeCapacity: "7Gi",
 					},
 				},
+				Grafana: v1alpha1.GrafanaConfiguration{
+					Resources: v1alpha1.Resources{
+						ResourceRequirements: v12.ResourceRequirements{
+							Limits: v12.ResourceList{
+								"memory": resource.MustParse("2Gi"),
```
lgarciaaco: Isn't this around 2Gb?? I would suggest something closer to the 256Mb.

jamesnetherton: Yeah, it's just a test case so I picked a random value. I can modify it to 256Mb though.

I actually created a new template param `GRAFANA_MEMORY_LIMIT` which defaults to 512Mb which I can also modify to 256Mb if you think it's better.

lgarciaaco: Oh I see, 512Mb is more reasonable dough ... great stuff

lgarciaaco: 512Mb makes more sense, great stuff!

Issue #4294, opened 2019-01-15
Refactor integration editor's current flow service


gashcrumb:
As part of supporting the API provider enhancements we need to refactor the current flow service.  Currently any changes done to the integration are *mostly* handled by this guy in one event handler function.

Planning on pulling out each block of code in the event handler into separate static functions so it's easier to reuse that logic elsewhere.

Also as part of this I'll verify that there aren't other controllers in the editor that make changes to the integration object directly outside of the current flow service, cases like that should really be refactored out, and that logic should also follow the same pattern as above.

@riccardo-forina FYI, and feel free to add any other concerns that I should be aware of.

gashcrumb:
@riccardo-forina wonder, think it'd be helpful to also extract out functions like `getSubsequentConnections` or `getSubsequentSteps` etc?

riccardo-forina:
yeah, I think so! if it's easy to do I mean

gashcrumb:
Almost done, I want to get the current flow functions I mentioned previously and I also spotted some code in some of the controllers that I think would be good to extract out as well.

PR #4296, opened 2019-01-16
chore: refactor out the meat of the event handlers in current flow service

gashcrumb:
relates to #4294

@riccardo-forina FYI, feedback welcome!

planning on finishing up the last 4 event handler blocks that mutate the step array directly, then there's all the positional functions in the current flow service that controllers use to examine neighboring steps etc. like `filterConnectionsByPosition` and `getStartStep`, `getEndStep`, `getSubsequentConnections` and so on...


gashcrumb:
@riccardo-forina Pushed an update, if you have a look at the current flow service it uses the new functions to update the integration, passing in the integration and flow ID etc.  Lemme know what you think, this was mostly what I had in mind at least for some initial refactoring but if there's anything else give me a heads up, otherwise I'd be keen on merging this soon-ish.

Issue #4147, opened 2018-11-23
Infrastructure Dashboard


heiko-braun:
As an administrator of Fuse on Openshift and/or Fuse Online, I want to be able to use some OOTB dashboards. These OOTB dashboards allow me to monitor my Prometheus infrastructure metrics for my Fuse infrastructure. This allows me to quickly check on the health of my Fuse install.

alexkieling:
Merged into https://github.com/syndesisio/syndesis/issues/4145

PR #4327, opened 2019-01-22
feat(monitoring): Use mustache templating for Grafana dashboard ConfigMap #4147

jamesnetherton:
Would be nice to have #4325 merged before this, but there's no strict dependency between the PRs.


jamesnetherton:
@lgarciaaco please can you take a peek at this.

PR #4328, opened 2019-01-22
Fix #4278 Support basic and bearer token authentication in FHIR

### LINKED ISSUE #4278: Support basic and bearer token authentication in FHIR
## This is a...

<pre><code>
[x ] Feature request
[ ] Bug report  
</code></pre>

## Description

FHIR connector needs to support basic and bearer token (oauth) authentication.

rkorytkowski:
I have an issue with building this locally with checks enabled. It fails due to:
`[WARNING] /home/rafal/.m2/repository/io/syndesis/common/common-model/1.6-SNAPSHOT/common-model-1.6-SNAPSHOT.jar(/io/syndesis/common/model/WithConfigurationProperties.class):[-1,-1] Cannot find annotation method 'value()' in type 'edu.umd.cs.findbugs.annotations.SuppressFBWarnings': class file for edu.umd.cs.findbugs.annotations.SuppressFBWarnings not found`

It started to happen after adding tests.

I have no idea why it happens. Does anyone have a clue?

### DIFF
--- a/app/connector/fhir/pom.xml
+++ b/app/connector/fhir/pom.xml
@@ -63,6 +63,7 @@
     <dependency>
       <groupId>io.syndesis.integration</groupId>
       <artifactId>integration-component-proxy</artifactId>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>io.syndesis.common</groupId>
@@ -72,10 +73,12 @@
     <dependency>
       <groupId>org.apache.camel</groupId>
       <artifactId>camel-core</artifactId>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
+      <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.camel</groupId>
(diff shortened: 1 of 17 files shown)


rkorytkowski:
@zregvart worked! Have you seen it before? How did you figure it out?

rkorytkowski:
PR for #4278

zregvart:
> @zregvart worked! Have you seen it before? How did you figure it out?

Saw it before, easy way to get an idea what's missing is to do a `fc:edu.umd.cs.findbugs.annotations.SuppressFBWarnings` search on https://search.maven.org and see if any dependencies show up that are missing in `mvn dependency:tree`. Since this is a SpotBugs annotation it is [not required at runtime](https://github.com/spotbugs/spotbugs/blob/e6ea7a71cd4ab8a349c5c573adfca26c0acc2bdb/spotbugs-annotations/src/main/java/edu/umd/cs/findbugs/annotations/SuppressFBWarnings.java#L33), we need to set the `provided` scope.

rkorytkowski:
@zregvart thanks for the explanation. I was mistaken by the fact that IntelliJ was showing me that the class was already on classpath, but I missed that the package was different and only name matched!

rkorytkowski:
It seems like a failure due to connectivity issue. I've force-pushed to rebuild.

Issue #4232, opened 2018-12-20
SPIKE: Investigate the removal of Springboot for integrations (runtime)


heiko-braun:
In order to simplify the usage of Camel-K, the removal of SpringBoot from integrations runtimes could help a lot. This would allow us to prevent version mismatches and the potential need to support different SpringBoot versions. 

This task is about investigating the feasibility and effort to make it happen in order to take a decision for 7.3.

lburgazzoli:
We may think to split the integration runtime bits related to spring in a "starter" artifact so we can include the bits that are needed according to the environment.

lburgazzoli:
@KurtStam as follow up of our conversation I would expect that we end up with the following modules: (naming is just an indication)

- integration/runtime
- integration/runtime-adapter-spring-boot
- integration/runtime-adapter-camel-k

The syndesis runtime bits should be part of integration/runtime and it should have no dependencies on spring-boot or camel-k whereas the runtime-adapter-* have the dependencies on the target runtime.

This would allow us to:
- deploy "classic" integration by including runtime-adapter-spring-boot
- deploy camel-k integration including  runtime-adapter-camel-k

KurtStam:
@lburgazzoli makes sense to me. As a first step, and before splitting things out, I'm looking into removing spring-boot from the current integration/runtime first or at least from the RouteBuilder.

KurtStam:
![springboot-runtime](https://user-images.githubusercontent.com/35576/51752287-267d5d00-2085-11e9-8d62-4142ce9a2e3b.png)

tplevko:
As this is was a refactoring task, QE work is probably not necessary on this. Closing.

PR #4336, opened 2019-01-23
[Refactoring] isolating springboot in the runtime

KurtStam:
Removing the dependency of the RouteBuilder on SpringBoot so it can be reused in CamelK #4232. First step is a simple refactoring.


KurtStam:
@lburgazzoli Cool - will do

zregvart:
bot's not merging for some reason, so I merged the thing

Issue #4326, opened 2019-01-22
UXD - Create UX design to allow users to tag an integration for release to a target environment


dhirajsb:
## This is a...

<pre><code>
[X] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

Need UX design to allow users to tag an integration by clicking on a button which generates a new release tag internally which is used by backend CI/CD APIs. 
Internally tags are kept in a map with environment name keys, mapped to generated tag, timestamp, etc.
User should be able to select a target environment name, or add one on the fly. A new tag is generated when user clicks on a button. User may also optionally also delete an environment name, which should pop up a confirmation dialog. 
The backend service will respond with a `ContinuousDeliveryEnvironment` json with the following fields:
```
{
"name": "test",
"integrationId": "<integration-id>",
"releaseTag": "<some-tag-string>",
"taggedAt": "<timestamp>",
"lastExportedAt": "<timestamp>"
}
```
The `lastExportedAt` timestamp is used internally to compare whether an integration has been tagged for release since last being exported to an environment.

heiko-braun:
@dongniwang @amysueg Can we get your thoughts on this? It's required for 7.3 and we only have a few weeks left.

dongniwang:
@heiko-braun - We will need some more information to understand the request.

dhirajsb:
@gaughan Gary, do you have feedback on the UI based export trigger proposed here?

gaughan:
Sure, are tags part of the first version? We could add a "Mark for CI/CD" option in the kebab menu of version history of the integration

dongniwang:
@dhirajsb @gaughan - Design is ready for review here. Please provide your comments in InVision. https://redhat.invisionapp.com/share/95QAAR3EZ8F

I'm still not certain what would be a good way to confirm deletion of an environment name.

dhirajsb:
@dongniwang looks good, added a comment on the delete button

gaughan:
nice, is delete the right term, remove is more accurate maybe?

dongniwang:
I understand where you're coming from @gaughan, but I think "delete" and "remove" are both the same in this context...

cc @TovaCohen here and maybe she can suggest a word?

dongniwang:
[...]

PR #4344, opened 2019-01-25
Fixes #4274, fixes #4275 Initial version of public import/export API

### LINKED ISSUE #4274: Backend service for exporting multiple/all integrations for CI/CD
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description

Create a backend public endpoint for exporting multiple/all integrations for CI/CD, including any dependent extensions. 
This service should support filtering integrations that have been marked for release using custom tags  that users can change using a UI button. 
Three tags can used to accomplish this:
* a timestamp to show when it was last exported
* a timestamp to show when it was last marked for release
* a timestamp that contains a unique hash, which can be used to easily determine newly imported integrations in the import CI/CD service

An input parameter could control whether this service should use the tags above to filter integrations or export ALL integrations. This may be useful to perhaps create a backup of the entire cluster or to replicate a cluster at the beginning of a CI/CD environment setup.

### LINKED ISSUE #4275: Backend service for importing multiple integrations for CI/CD
## This is a...

<pre><code>
[X] Feature request
[ ] Bug report  
</code></pre>

## Description

Create a backend service that imports multiple integrations at a time. This service can use the tags created by the export service to identify which integrations have changed in the target cluster. 
This service can also create it's own tag with a timestamp showing when the integration/extension/connection was imported.  

An API parameter could control whether the service may attempt to start integrations that have been successfully imported. Alternatively, this service may not perform any lifecycle operations, as that may be something the CI/CD pipeline may choose to do itself using a separate public facing lifecycle endpoint. 

The first option to auto-start integrations may be easier to implement in the first version.

dhirajsb:


dhirajsb:
@chirino @zregvart can you add your review? I'm sure there are things to improve/clarify in this initial version.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiParam;
+import io.syndesis.common.model.connection.ConnectionOverview;
+import io.syndesis.common.model.integration.ContinuousDeliveryEnvironment;
+import io.syndesis.common.model.integration.ContinuousDeliveryImportResults;
+import io.syndesis.common.model.integration.IntegrationDeployment;
+import io.syndesis.common.model.monitoring.IntegrationDeploymentStateDetails;
+
+@Api("continuous-delivery")
```
zregvart: Continuous delivery is a practice, this feature might enable that practice but it might be also used for different use cases. Naming this continuous delivery is a mistake IMHO.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+@Path("/public/continuousdelivery")
```
zregvart:
```suggestion
@Path("/public/integrations")
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+    @GET
+    @Path("environments")
+    @Produces(MediaType.APPLICATION_JSON)
+    List<String> getReleaseEnvironments();
+
+    /**
+     * List all tags associated with this integration.
+     */
+    @GET
+    @Path("releasetags/{id}")
```
zregvart:
```suggestion
    @Path("{id}/tags")
```

or

```suggestion
    @Path("{id}/releases")
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    Map<String, ContinuousDeliveryEnvironment> getReleaseTags(@NotNull @PathParam("id") @ApiParam(required = true) String integrationId);
+
+    /**
+     * Delete an environment tag associated with this integration.
+     */
+    @DELETE
+    @Path("releasetags/{id}/{env}")
```
zregvart:
```suggestion
    @Path("{id}/environments/{env}")
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+    void deleteReleaseTag(@NotNull @PathParam("id") @ApiParam(required = true) String integrationId, @NotNull @PathParam("env") @ApiParam(required = true) String environment);
+
+    /**
+     * Tag an integration for release to target environments.
+     */
+    @POST
+    @Path("releasetags/{id}")
```
zregvart:
```suggestion
    @Path("{id}/releases")
```

Not sure if we need both `getReleaseTags` and `tagForRelease`.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+    Map<String, ContinuousDeliveryEnvironment> tagForRelease(@NotNull @PathParam("id") @ApiParam(required = true) String integrationId,
+                                                @NotNull @ApiParam(required = true) List<String> environments);
+
+    /**
+     * Export integrations to a target environment.
+     */
+    @GET
+    @Path("export/{env}/export.zip")
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
+    StreamingOutput exportResources(@NotNull @PathParam("environment") @ApiParam(required = true) String environment,
```
zregvart: Is this needed?

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+                                          @NotNull @PathParam("version") @ApiParam(required = true) final int version);
+
+    /**
+     * DTO for {@link org.jboss.resteasy.plugins.providers.multipart.MultipartFormDataInput} for importResources.
+     */
+    class ImportFormDataInput {
+        @FormParam("importfile")
+        private InputStream importFile;
+
+        @FormParam("connectionparamsfile")
```
zregvart:
```suggestion
        @FormParam("parameters")
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+    @Path("integrations/{id}/deployments/{version}")
+    @Produces(MediaType.APPLICATION_JSON)
+    void stopIntegration(@Context final SecurityContext sec, @NotNull @PathParam("id") @ApiParam(required = true) final String integrationId,
+                                          @NotNull @PathParam("version") @ApiParam(required = true) final int version);
+
+    /**
+     * DTO for {@link org.jboss.resteasy.plugins.providers.multipart.MultipartFormDataInput} for importResources.
+     */
+    class ImportFormDataInput {
+        @FormParam("importfile")
```
zregvart:
```suggestion
        @FormParam("data")
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+        private InputStream importFile;
+
+        @FormParam("connectionparamsfile")
+        private InputStream paramsFile;
+
+        @FormParam("environment")
+        private String environment;
```
zregvart: Needed?

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+    @Path("export/{env}/export.zip")
+    @Produces(MediaType.APPLICATION_OCTET_STREAM)
+    StreamingOutput exportResources(@NotNull @PathParam("environment") @ApiParam(required = true) String environment,
+                           @QueryParam("all") @ApiParam() boolean exportAll) throws IOException;
+
+    /**
+     * Import integrations into a target environment.
+     */
+    @POST
+    @Path("import")
```
zregvart:
```suggestion
    @Path("integrations")
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+    ContinuousDeliveryImportResults importResources(@Context SecurityContext sec, @NotNull @ApiParam ImportFormDataInput formInput) throws IOException;
+
+    /**
+     * Configure a connection.
+     */
+    @POST
+    @Path("connections/{id}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    ConnectionOverview configureConnection(@Context SecurityContext sec, @NotNull @PathParam("id") @ApiParam String connectionId, @NotNull @ApiParam Map<String, String> properties) throws IOException;
```
zregvart: Is this needed?

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+    IntegrationDeployment publishIntegration(@Context final SecurityContext sec, @NotNull @PathParam("id") @ApiParam(required = true) final String integrationId);
+
+    /**
+     * Stop Integration.
+     */
+    @POST
+    @Path("integrations/{id}/deployments/{version}")
+    @Produces(MediaType.APPLICATION_JSON)
+    void stopIntegration(@Context final SecurityContext sec, @NotNull @PathParam("id") @ApiParam(required = true) final String integrationId,
+                                          @NotNull @PathParam("version") @ApiParam(required = true) final int version);
```
zregvart: Is `version` needed?

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/NoopContinuousDeliveryHandlerImpl.java
```diff
@@ -0,0 +1,88 @@
...
+
+import io.syndesis.common.model.connection.ConnectionOverview;
+import io.syndesis.common.model.integration.ContinuousDeliveryEnvironment;
+import io.syndesis.common.model.integration.ContinuousDeliveryImportResults;
+import io.syndesis.common.model.integration.IntegrationDeployment;
+import io.syndesis.common.model.monitoring.IntegrationDeploymentStateDetails;
+
+@Component
+@ConditionalOnProperty(value = "continuousdelivery.enabled", havingValue = "false", matchIfMissing=true)
+public class NoopContinuousDeliveryHandlerImpl implements ContinuousDeliveryHandler {
```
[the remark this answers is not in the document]
dhirajsb: Yeah, that was a copied usage from providers, I have removed it now.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: Yeah, it can figure it out on it's own, so removed it

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
```diff
@@ -0,0 +1,184 @@
...
+import io.swagger.annotations.ApiParam;
+import io.syndesis.common.model.connection.ConnectionOverview;
+import io.syndesis.common.model.integration.ContinuousDeliveryEnvironment;
+import io.syndesis.common.model.integration.ContinuousDeliveryImportResults;
+import io.syndesis.common.model.integration.IntegrationDeployment;
+import io.syndesis.common.model.monitoring.IntegrationDeploymentStateDetails;
+
+@Api("continuous-delivery")
+@Path("/public/continuousdelivery")
+public interface ContinuousDeliveryHandler {
```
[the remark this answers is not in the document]
dhirajsb: As in, why not implement the methods directly in the handler?

install/generator/09-syndesis-public-oauth-proxy.yml.mustache
```diff
@@ -0,0 +1,184 @@
...
+  kind: ServiceAccount
+  metadata:
+    name: syndesis-public-oauthproxy
+    labels:
+      app: syndesis
+      syndesis.io/app: syndesis
+      syndesis.io/type: infrastructure
+      syndesis.io/component: syndesis-public-oauthproxy
+- apiVersion: authorization.openshift.io/v1
+  kind: ClusterRoleBinding
```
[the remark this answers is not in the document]
dhirajsb: @rhuss @heiko-braun this brings up the issue that this proxy and the public API should really only be installed by someone who has cluster level privileges. Looking at it another way, since the API exposes a special endpoint that allows long lived service account token access, it should be something that a cluster admin is aware of. 
Any ideas on how we can make this an optional thing in syndesis install, so only a user with the right privileges would install this?

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: Yeah, I was looking from it inside out. I am inclined to just call it something like `SyndesisPublicAPI` for now. We can always refactor that one interface when the number of endpoints grows in the future.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: Changed it to use the format `integrations/{id}/tags`

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: Changed to use `environments/{id}/tags`

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: The separate endpoints are for the UI workflow from https://github.com/syndesisio/syndesis/issues/4326

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: Renamed to properties, since they are connection properties in a properties file format.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: @chirino suggested that the CD pipeline might want to do import/export separate from configuring connections, e.g. calling this endpoint for each connection to be configured.

install/generator/09-syndesis-public-oauth-proxy.yml.mustache
```diff
@@ -0,0 +1,184 @@
...
+        - name: syndesis-public-oauthproxy
+{{^WithDockerImages}}
+          image: ' '{{/WithDockerImages}}{{#WithDockerImages}}
+          image: ${SYNDESIS_REGISTRY}/{{Images.OAuthProxyImagePrefix}}/{{ Images.Support.OAuthProxy }}:{{ Tags.OAuthProxy }}{{/WithDockerImages}}
+          args:
+            - --provider=openshift
+{{^WithOAuthClient}}
+            - --client-id=system:serviceaccount:${OPENSHIFT_PROJECT}:syndesis-oauth-client{{/WithOAuthClient}}{{#WithOAuthClient}}
+            - --client-id=syndesis-${OPENSHIFT_PROJECT}{{/WithOAuthClient}}
+            - --client-secret=${OPENSHIFT_OAUTH_CLIENT_SECRET}
```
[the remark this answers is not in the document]
rhuss: Makes sense, but this adds another mandatory parameter to the template, and required for the SA used here to be created upfront (like with the other OAuth proxy). So we need to update the documentation, OpenShift Online installation script, OCP documentation, operator usage, etc. 

If this is ok, we should do this.

install/README.md
```diff
@@ -42,12 +42,14 @@ All template parameters are required. Most of them have sane defaults, but some
 | Parameter | Description |
 | --------- | ----------- |
 | **ROUTE_HOSTNAME** | The external hostname to access Syndesis |
+| **PUBLIC_API_ROUTE_HOSTNAME** | The external hostname to access Syndesis public API |
```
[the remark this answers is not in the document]
rhuss: @heiko-braun also a heads up here, that ever new mandatory parameter requires an update of the deployment script on the openshift-online side.

install/syndesis-privileged.yml
```diff
@@ -1821,3 +1828,181 @@ objects:
...
+  metadata:
+    labels:
+      app: syndesis
+      syndesis.io/app: syndesis
+      syndesis.io/type: infrastructure
+    annotations:
+      console.alpha.openshift.io/overview-app-route: "true"
+    name: syndesis-public-api
+  spec:
+    host: ${PUBLIC_API_ROUTE_HOSTNAME}
```
rhuss: Can't we just reuse the original ROUTE_NAME ?

install/syndesis-privileged.yml
```diff
@@ -469,7 +471,10 @@ objects:
           "logging": false
         },
         "branding": {
-          "logoDarkBg": "assets/images/syndesis_logo_icon_darkbkg.png",
+          "logoWhiteBg": "assets/images/syndesis-logo-svg-white.svg",
+          "logoDarkBg": "assets/images/syndesis-logo-svg-white.svg",
+          "iconWhiteBg": "assets/images/glasses_logo_square.png",
+          "iconDarkBg": "assets/images/glasses_logo_square.png",
```
rhuss: Not how this comes in here.

install/generator/09-syndesis-public-oauth-proxy.yml.mustache
```diff
@@ -0,0 +1,177 @@
...
+          kind: ImageStreamTag
+          name: {{ Images.Support.OAuthProxy }}:{{ Tags.OAuthProxy }}
+          namespace: ${IMAGE_STREAM_NAMESPACE}
+      type: ImageChange
+{{/WithDockerImages}}
+    - type: ConfigChange{{#WithOAuthClient}}
+
+- apiVersion: oauth.openshift.io/v1
+  grantMethod: auto
+  kind: OAuthClient
```
rhuss: @dhirajsb @zregvart is that related to your change or does the import/export also needs an OAuthClient ? then we are running in the same issue with the other PR.

rhuss: Actually it looks like that the API access can also work with the OAuthClient (otherwise it wouldn't be omittted in the non-privilegded template). So I gues this OAuthClient can just be removed here.

heiko-braun:
@dhirajsb Regarding the additionl cluster level permissions needed if the things are baked into the template:

> carefully check which permission are the least required and if this is on cluster-admin level, remove it from the templates and document/script the extra steps required which needs to be performed as clusteradmin.

I agree with @rhuss, we should keep this out of the template to prevent the permissons become a strict requirement at installation time. We cannot expect this permission to be granted in all environments. 

Can we make this a post installation step, that becomes documented and will only be needed if people want to leverage the import/export API's?

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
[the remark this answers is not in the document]
zregvart: Yeah, we have no need for different implementations of the handler in foreseeable future...

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
zregvart: I'd split this into three feature sets, one for import/export (addressed here), one for managing connection configuration and one for managing integration state. Code-wise that would be three handlers I think...

gashcrumb:
This branch seems to need a rebase

Anyways I pushed a start to the dialog, hoping @elvisisking could maybe help move it along starting next week.  The current impl should at least show environments but none of the editing or adding is implemented yet, and there's some tricky stuff in that design which definitely makes it a lot more complicated as you dig into the details.

gashcrumb:
@kahboom, wonder if you'd be able to portion off some time to team up with @elvisisking on this one?

install/generator/09-syndesis-public-oauth-proxy.yml.mustache
```diff
@@ -0,0 +1,184 @@
...
+            - --pass-user-bearer-token
+            - --skip-provider-button
+            - --approval-prompt=auto
+            - --skip-auth-regex=/logout
+            - --skip-auth-regex=/[^/]+\.(png|jpg|eot|svg|ttf|woff|woff2)
+            - --skip-auth-regex=/api/v1/swagger.*
+            - --skip-auth-regex=/api/v1/index.html
+            - --skip-auth-regex=/api/v1/credentials/callback
+            - --skip-auth-regex=/api/v1/version
+            - --skip-auth-preflight
```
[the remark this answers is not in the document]
dhirajsb: Agreed, removed redundant skip auth parameters in another commit.

install/generator/09-syndesis-public-oauth-proxy.yml.mustache
```diff
@@ -0,0 +1,184 @@
...
+            - --openshift-ca=/etc/pki/tls/certs/ca-bundle.crt
+            - --openshift-ca=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
+            - --openshift-sar={"namespace":"syndesis","resource":"pods","verb":"get"}
+            - --openshift-delegate-urls={"/api/v1/public":{"namespace":"syndesis","resource":"pods","verb":"get"}}
+          env:
+          - name: OAUTH_COOKIE_SECRET
+            valueFrom:
+              secretKeyRef:
+                name: syndesis-oauth-proxy-cookie-secret
+                key: oauthCookieSecret
```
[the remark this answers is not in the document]
dhirajsb: Good question. That's what I thought and had removed these parameters and cookie secret. But, for some reason, the delegate mode does not work without these. At startup it prints a message:
```
Delegation of authentication and authorization to OpenShift is enabled for bearer tokens and client certificates.
```
There does not seem to be a way to enable it only for bearer tokens. Weird!!!

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: We could do it that way. But I was concerned that we would end up with many handlers with one or two shallow methods. Do we want to have to give out lots of simple Open API docs, or a single public API doc that captures everything?

install/syndesis-privileged.yml
dhirajsb: I put that back after merge with master. Evidently it wasn't generated like you said using the install.../run script. 
I didn't want to rock any boats, so I put it back. :)

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
zregvart: We have single (well, two) OpenAPI docs in the runtime right now for a lot of handlers, so that should not be an issue, checkout:

https://github.com/syndesisio/syndesis/blob/1f349f4b87f389b4cb50428002b03d3255082be2/app/server/runtime/pom.xml#L256-L297

install/generator/09-syndesis-public-oauth-proxy.yml.mustache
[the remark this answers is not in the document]
dhirajsb: I agree with the concerns raised about this. I'm refactoring the public proxy out of the installer into its own template to be run separately to create a public proxy with required permissions.

app/server/runtime/src/main/resources/application.yml
```diff
@@ -146,3 +146,6 @@ metrics:
 
 monitoring:
   kind: default
+
+continuousdelivery:
```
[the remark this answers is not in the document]
dhirajsb: Done, moved it to features.public-api.enabled

install/generator/04-syndesis-server.yml.mustache
```diff
@@ -215,6 +215,8 @@
           enabled: true
       monitoring:
         kind: default
+      continuousdelivery:
+        enabled: true
```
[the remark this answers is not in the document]
dhirajsb: Moved to features section as recommended.

install/syndesis-privileged.yml
dhirajsb: Regenerated syndesis.yml using `/install/generator/run.sh`

install/generator/09-syndesis-public-oauth-proxy.yml.mustache
dhirajsb: Removed

install/generator/09-syndesis-public-oauth-proxy.yml.mustache
```diff
@@ -0,0 +1,184 @@
...
+{{^WithDockerImages}}
+          image: ' '{{/WithDockerImages}}{{#WithDockerImages}}
+          image: ${SYNDESIS_REGISTRY}/{{Images.OAuthProxyImagePrefix}}/{{ Images.Support.OAuthProxy }}:{{ Tags.OAuthProxy }}{{/WithDockerImages}}
+          args:
+            - --provider=openshift
+{{^WithOAuthClient}}
+            - --client-id=system:serviceaccount:${OPENSHIFT_PROJECT}:syndesis-oauth-client{{/WithOAuthClient}}{{#WithOAuthClient}}
+            - --client-id=syndesis-${OPENSHIFT_PROJECT}{{/WithOAuthClient}}
+            - --client-secret=${OPENSHIFT_OAUTH_CLIENT_SECRET}
+            - --upstream=http://syndesis-server/api/v1/public/
```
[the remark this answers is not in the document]
dhirajsb: I had looked into this earlier, but there doesn't seem to be a way to block any subpaths in the proxy command line options, or maybe I missed it somehow. 
Anyway, the UI bit added for creating tags, does use some of the endpoints that are part of the public API.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: I was using this to tag the imported integrations with an `importedAt` timestamp, which might come in handy in the future. 
Otherwise, we could drop this here and the associated timestamp in the db.

install/generator/04-syndesis-server.yml.mustache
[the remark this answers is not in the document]
dhirajsb: Moved under `features`.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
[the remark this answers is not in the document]
dhirajsb: Will remove the interface.

dhirajsb:
@gashcrumb this PR is ready to merge for my part, do you have more UI changes to make in this PR or create a separate Pr for that?

install/generator/09-syndesis-public-oauth-proxy.yml.mustache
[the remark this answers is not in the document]
dhirajsb: Moved public proxy out into it's own template `syndesis-public-oauthproxy`, also added instructions in README.md on how to install using the template.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: Convenience filter for initial CD implementations. Will be elaborated upon in the future when other requirements come forward.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: Added package `io.syndesis.server.endpoint.v1.handler.external` to swagger docs. 
The public API handler can be refactored into multiple handlers in the future if the number of APIs increases significantly.

install/README.md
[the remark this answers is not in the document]
dhirajsb: Moved to a separate template, so doesn't affect current install setups.

install/generator/01-header.yml.mustache
```diff
@@ -19,6 +19,8 @@ metadata:
 parameters:
 - name: ROUTE_HOSTNAME
   description: The external hostname to access Syndesis
+- name: PUBLIC_API_ROUTE_HOSTNAME
+  description: The external hostname to access Syndesis public API
```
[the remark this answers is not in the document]
dhirajsb: Moved to a separate template.

install/syndesis-privileged.yml
dhirajsb: Had some issues making this work with OpenShift router. Will investigate in the future on how to setup a route to two oauth proxies and filter by uri path.

install/generator/09-syndesis-public-oauth-proxy.yml.mustache
```diff
@@ -0,0 +1,184 @@
...
+- apiVersion: v1
+  kind: Service
+  metadata:
+    labels:
+      app: syndesis
+      syndesis.io/app: syndesis
+      syndesis.io/type: infrastructure
+      syndesis.io/component: syndesis-public-oauthproxy
+    annotations:
+      service.alpha.openshift.io/serving-cert-secret-name: syndesis-public-oauthproxy-tls
```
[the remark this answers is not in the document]
dhirajsb: Will investigate in the future if the need arises.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: Renamed to `PublicApiHandler`.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: Renamed to prefix `public`.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java
dhirajsb: Used as a convenience filter for simple CD scripts in the first version.

PR #4347, opened 2019-01-25
feature(syndesis): Add Camel-K integration into "syndesis install" and "syndesis minishift"

rhuss:


tools/bin/commands/util/operator_funcs
```diff
@@ -101,7 +101,7 @@ build_operator_local() {
         echo "Running tests"
         go test $TEST_TARGETS
     fi
-    if [ "$do_linux" ] && [ $(isOSX) ]; then
+    if [ "$do_linux" ] && isMacOs; then
```
zregvart:
```suggestion
    if [ "$do_linux" ] && $(isMacOs); then
```

rhuss: @zregvart why ?

isMacOs returns either "true" or "false" which are valid Unix commands and get executed.

rhuss: $(isMacOs) would fire also for "false" (as this is a non-0 length string)

rhuss: I know, we used it differently, but I'd like to change to this kind of usage for boolean checks as it more clear than requiring to return "true" (or anything larger than 1 char) for true, and "" for false (has it is done in hasOption method). But this would be a larger change.

In fact I would even go for changing to getopts instead of our own grown option checks.

rhuss: And finally, I would rewrite in go (if people would feel comfortable with go)

zregvart: > @zregvart why ?

I could be misunderstanding something, based on:

```shell
wat() {
  echo "false"
}

if [ "true" ] && wat; then
  echo 1
fi

if [ "true" ] && $(wat); then
  echo 2
fi
```

Outputs:
```
false
1
```

I would expect it only to output:
```
false
```

rhuss: Yes, you shoudl use true and false without quote.

```
if  false ....
```

that's its how it is interpolaed by the shell when you call 

```
if isMacOs; then ..
```

(note the missing test operator [ ... ] !)

rhuss: sorry, just re-read you sample. trying now ...

rhuss: you are right. I have to eval it, but without test operator.

lburgazzoli:
looks like there are some unrelated commits, isn't it ?

zregvart:
Probably rebase issue

zregvart:
Or GitHub suggestion did something unexpected...

rhuss:
Yeah history does not look good. Also, looks like that purebot doesn't like to merge again. 

I will rebase to a single commit and find out what's up with pure bot again.

lburgazzoli:
lgtm

Issue #4096, opened 2018-11-16
Some Tooltips on Setting page are missing


nding-anges:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
On Setting page, the tooltips of 'Client ID' and 'Client Secret, etc...'on Salesforce and SAP Concur  are missing.  

<img width="1278" alt="2018-11-16 3 04 20" src="https://user-images.githubusercontent.com/34368385/48603375-2724b380-e9b1-11e8-9bcf-dc8e0a22e791.png">

<img width="1297" alt="2018-11-16 3 09 07" src="https://user-images.githubusercontent.com/34368385/48603489-8b477780-e9b1-11e8-8f6d-923fd4cb9303.png">

nding-anges:
@TovaCohen Please help us to review the tooltips. Thanks~

dongniwang:
@zregvart @gashcrumb - Do you have any idea as to why some of the form fields have the "i" icon right next to the field labels while some of them have help text below the form fields?

gashcrumb:
Yeah, some of these forms are configured to use `labelHint` (which makes the 'i' icon show up) vs `description` in the properties object, which drives the text *under* the input field.  We could probably stand to align all of these to one, though the `labelHint` means that you get a hover tooltip which works terribly when there's lots of text, like some of those fields under the Salesforce oauth connector etc.

TovaCohen:
Apologies for not responding. I've been out sick the last week, and will be out this week. But I will look at this as soon as I can after Thanksgiving.  Can someone point me to a file that contains the tooltips that need editing?

TovaCohen:
And I forgot to specify this issue number in these pull requests: 
Twitter: https://github.com/syndesisio/syndesis/pull/4358 
Salesforce: https://github.com/syndesisio/syndesis/pull/4357

PR #4358, opened 2019-01-26
Update Twitter field names and hints on Settings page (#4096)

TovaCohen:
This updates the display names as well as the hints for Twitter fields on the Settings page. 
In the staging site, in the Settings page, Twitter entry, the field order is:

Consumer Secret
Concumer Key

This should be reversed, but I don't know how to do that.

### DIFF
--- a/app/connector/twitter/src/main/resources/META-INF/syndesis/connector/twitter.json
+++ b/app/connector/twitter/src/main/resources/META-INF/syndesis/connector/twitter.json
@@ -139,12 +139,14 @@
     "consumerKey": {
       "componentProperty": true,
       "deprecated": false,
-      "displayName": "Consumer Key",
+      "displayName": "Consumer API Key",
       "group": "security",
       "javaType": "java.lang.String",
       "kind": "property",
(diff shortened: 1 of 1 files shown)


app/connector/twitter/src/main/resources/META-INF/syndesis/connector/twitter.json
```diff
@@ -139,12 +139,14 @@
...
-      "displayName": "Consumer Key",
+      "displayName": "Consumer API Key",
       "group": "security",
       "javaType": "java.lang.String",
       "kind": "property",
       "label": "security",
-      "labelHint": "The consumer key",
+      "labelHint": "The Consumer API key that Twitter provides when you register a client application.
+
+",
```
zregvart: These newlines cause the build to fail, can you please put the quote on the same line as the `labelHit` property.

TovaCohen:
I'm running into conflicts trying to push and I can't figure out how to resolve them. 
So I'm closing this PR and will create a new one with all the updates and the new lines removed.

PR #4373, opened 2019-01-28
Match ascii-doctor-version with the version in app/pom.xml

cunningt:
Getting an error building doc 

[INFO] --- dependency-management-maven-plugin:0.7:analyze (default) @ syndesis-docs ---
[WARNING] Version mismatch for plugin org.asciidoctor:asciidoctor-maven-plugin, managed version 1.5.7.1 does not match project version 1.5.3

Should align asciidoctor-maven-plugin versions.

### DIFF
--- a/doc/pom.xml
+++ b/doc/pom.xml
@@ -41,7 +41,7 @@
             <plugin>
                 <groupId>org.asciidoctor</groupId>
                 <artifactId>asciidoctor-maven-plugin</artifactId>
-                <version>1.5.3</version>
+                <version>1.5.7.1</version>
                 <executions>
                     <execution>
                         <phase>process-resources</phase>


christophd:
Can we just use the managed version coming from Syndesis parent Maven POM?

I have had same error and removed the explicit version in docs POM see open PR https://github.com/syndesisio/syndesis/pull/4374

https://github.com/syndesisio/syndesis/pull/4374/commits/8794795c5a31e24629ea601da0c7d0558a0fd7f2

oscerd:
> Can we just use the managed version coming from Syndesis parent Maven POM?
> 
> I have had same error and removed the explicit version in docs POM see open PR #4374
> 
> [8794795](https://github.com/syndesisio/syndesis/commit/8794795c5a31e24629ea601da0c7d0558a0fd7f2)

I think we can use the managed version

Issue #4302, opened 2019-01-17
Google Sheets: Initial draft of user doc


christophd:
Write initial draft of the user doc for the connector

PR #4378, opened 2019-01-29
refactor(build): use consistent asciidoctor-maven-plugin version

zregvart:
This should fix issue raised in https://github.com/syndesisio/syndesis/pull/4373#issuecomment-458461513


zregvart:
Already included in #4302

PR #4382, opened 2019-01-29
feat: Combine step/connection selection into one page

### LINKED ISSUE #4322: Implement combined add step/connection page
The API provider editor doesn't have separate pages for adding a step or connection, instead it's just one page:

Full page -> https://redhat.invisionapp.com/share/BHPQXBH6W8Y#/screens/338299142

Not clear if this wireframe should incorporate elements of [this most recent design](https://redhat.invisionapp.com/share/X6K0SKUWZFG#/screens).  It probably should at least have some kind of filter/search.

This also means the popups in the IVP can be removed.

gashcrumb:
fixes #4322

This simplifies how the editor loads it's integration and also combines the separate step and connection selection pages into one.  And some general cleanup here and there too.

image


gashcrumb:
Note there's a bunch of changes in some templates due to IDE reformatting, so it's not *quite* as big as it looks, honest!

PR #4423, opened 2019-02-04
feat(monitoring): Create Syndesis REST API dashboard

alexkieling:
Convert Syndesis Server REST API into a generic REST API dashboard with drop-downs for 'Namespace' and 'Application' (server or meta)


alexkieling:
Fix #4366

Issue #4249, opened 2019-01-07
Apply PF4 markup to Masthead and Vertical Navigation


amysueg:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
The consoles that are part of the Agile Integration product should share a common look and feel. We'd like to start by applying a common visual theme to the masthead and vertical navigation of the consoles, using underlying PF4 markup and PF3 CSS.

## Expected behavior
Use PF4 markup for the masthead and vertical navigation of Syndesis; apply PF3 CSS styling until all consoles in the Integration product are ready to unveil PF4 styling.

FYI @gaughan @mcoker @jameslau @dongniwang

amysueg:
1. This year we have a goal to move our Integration consoles (Fuse, AMQ, 3scale) to Patternfly4. This will be a gradual adoption and one that we propose to coordinate between the console teams and developers. 
2. To begin the move, we propose an initial milestone to move to a PF4 look and feel for the masthead and vertical navigation of the console UI. And probably the login page as well.
3. Because it is possible for PF4 and PF3 to live together harmoniously in the same application, the recommended approach is to use a PF3 compatibility mode (CSS) but to change the underlying mark-up (HTML) to use PF4.  
4.  When all consoles are ready to adopt the PF4 look and feel, then we can drop the PF3 CSS, and PF4 styling will be applied.

In the case of 3scale, they had *no* Patternfly at all. In that case,  the approach was that Michael assisted Thomas to adopt PF3 HTML and CSS styles to produce a PF3 look and feel that more or less matches the existing PF3 UI in Fuse Online and in OpenShift. They had initially considered that 3scale would write the PF4 HTML and then apply a PF3 CSS to it; I believe that was abandoned when Thomas decided it wouldn't be that onerous to re-write the code for PF4.

jameslau:
Thank you for the clarification @amysueg!

gashcrumb:
> And probably the login page as well.

Note the login page comes from openshift and is out of our control.

jameslau:
Based on recent meetings, we've come to a consensus that the masthead will go forward for 7.3 release with PF4 styling while sitting on top of a PF3 markup (HTML) infrastructure. The vertical navigation will be released at a much later time in 2019. Please ping @amysueg and @dongniwang for further questions and/or comments on the progress for this issue.

[...]

PR #4433, opened 2019-02-04
New Syndesis Masthead Update

jameslau:
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report
</code></pre>

## Implementation

The Syndesis application is moving forward with a new masthead design that matches with the new PF4 beta release design specs. 

## Screenshot
screen shot 2019-02-04 at 3 44 54 pm

screen shot 2019-02-04 at 3 51 20 pm

## Notable Issue:
Our internal or main body sections have a slight padding issue with will be corrected. There is a PF ticket that address the namespace reset styling conflict that effects multiple CSS styling systems (link to issue: https://github.com/patternfly/patternfly-next/issues/1368)

fixes: #4249


dongniwang:
When do we expect https://github.com/patternfly/patternfly-next/issues/1368 to be resolved? @mcoker @jameslau 

Without the background color makes it really hard to see what's on the screen...

gashcrumb:
Looks like this PR needs a rebase when you get a chance

app/ui/src/scss/components/_masthead.scss
```diff
@@ -0,0 +1,88 @@
+@import 'syndesis-sass';
+
+// .pf-c-page {
```
gashcrumb: What's this commented out stuff for?  Ideally we should avoid committing commented out code, unless this is for enabling at some future point.  If so it'd be awesome if you could add a comment about that to the commented out block, thereby increasing the amount of commenting :-)

jameslau: Hey @gashcrumb, good catch. That's remnant code from a branch that I branched off from a previous masthead and vert nav branch. I'll remove it right now.

mcoker:
@dongniwang just submitted a PR in pf4 core - https://github.com/patternfly/patternfly-next/pull/1383

app/ui/src/scss/utils/_overrides.scss
```diff
@@ -15,33 +15,31 @@ body.cards-pf {
   margin-bottom: 10px;
 }
 
-.layout-pf.layout-pf-fixed .container-pf-nav-pf-vertical.collapsed-nav {
-  margin-left: 0;
+.pf-c-page {
+  display: block;
+  grid-template-columns: none;
```
mcoker: Do you need these? If it isn't `display: grid;`, I would expect the `grid-template-*` properties to be ignored.

app/ui/src/scss/utils/_overrides.scss
```diff
@@ -15,33 +15,31 @@ body.cards-pf {
...
+  grid-template-rows: none;
+  grid-template-areas: none;
 }
 
-.navbar {
-  a {
-    color: #ffffff;
-  }
+.pf-c-page__header {
+  grid-area: none;
```
mcoker: Same here, if the parent isn't `display: grid;`, I would expect this to be ignored

app/ui/src/scss/utils/_overrides.scss
jameslau: By not adding `none` to `grid-template-columns` into the css you get these properties creep up:

screen shot 2019-02-07 at 4 00 13 am

mcoker: Ah yeah, but those only apply to a `grid` parent. If you set `display: block;` on `.pf-c-page` (overriding `display: grid;`), those properties shouldn't do anything.

app/ui/src/app/app.component.html
```diff
@@ -1,246 +1,243 @@
...
....4L669,189.32h.41V59.2h13.32v148h-16.8L586,79.28h-.41V207.15H572.24Z"/><path class="cls-3" d="M760,59.2c48.56,0,76.23,25,76.23,74s-27.67,74-76.23,74H720V59.2Zm3.27,135.66c41.6,0,58.61-24.39,58.61-61.69,0-43.64-25.61-62.91-67.62-61.68h-20.9V194.86Z"/><path class="cls-3" d="M870.19,207.15V59.2H958.1V71.49H883.51v53.08h71.73v12.29H883.51v58h77.67v12.29Z"/><path class="cls-3" d="M1030.44,210.23c-23.56,0-43.44-17.22-44.05-41.6h14.34c.41,22.74,17.21,29.3,37.09,29.3,16.6,0,35.66-9.63,35.66-28.48,0-14.76-10.66-21.73-23.16-25.82-24-8-60.45-15.58-60.45-47.55,0-25.82,23.77-40,47.13-40,19.47,0,44.88,6.77,48.16,35.25h-14.35c-.41-18.24-17-23-32.17-23-15.37,0-34.43,7.79-34.43,25.82,0,42,83.61,23,83.61,74C1087.82,200.39,1060.16,210.23,1030.44,210.23Z"/><path class="cls-3" d="M1121.63,207.15V58.58H1135V207.15Z"/><path class="cls-3" d="M1212.62,210.23c-23.56,0-43.44-17.22-44.06-41.6h14.35c.41,22.74,17.21,29.3,37.09,29.3,16.6,0,35.66-9.63,35.66-28.48,0-14.76-10.66-21.73-23.16-25.82-24-8-60.45-15.58-60.45-47.55,0-25.82,23.77-40,47.13-40,19.47,0,44.88,6.77,48.16,35.25H1253c-.41-18.24-17-23-32.17-23-15.37,0-34.43,7.79-34.43,25.82,0,42,83.61,23,83.61,74C1270,200.39,1242.34,210.23,1212.62,210.23Z"/></svg>
```
zregvart: Are we all aware that this SVG needs to change based on the configuration in `config.json`?

jameslau: Hi @zregvart, do you mean the `png` file that currently exists in the `config.json` file? I see that they are different. Could you explain to me what that file does exactly?

seanforyou23: I think it is referring to what this logic was doing

https://github.com/syndesisio/syndesis/blob/daa24cf7791aba89340e6fb50a948c691e5e0413/app/ui/src/app/app.component.html#L9-L12

https://github.com/syndesisio/syndesis/blob/daa24cf7791aba89340e6fb50a948c691e5e0413/app/ui/src/app/app.component.ts#L110

https://github.com/syndesisio/syndesis/blob/daa24cf7791aba89340e6fb50a948c691e5e0413/app/ui/src/app/app.component.ts#L38

Do I understand this correctly? Instead of specifying this image explicitly, instead, consult with the settings to pull the correct one from config.json? Hope I didn't complicate things..

jameslau: Thank you for clearing that up @seanforyou23. I did not know. How should we proceed with the changes if we're going to use `SVG`s instead of `png` or any other conventional `img` file? Can we insert the `SVG` into the `app.component.ts` file?

seanforyou23: I think we can do the same type of check as before and keep using the config in the same way, just using the svg instead of png

zregvart: The only requirement that I see is that the logo and the app name needs to change. The upstream version is called Syndesis, and the productized version is called Fuse Online, the logo also changes with that. I don’t think there’s any issue in getting the productized version in SVG. Just need to make sure that the ability to swap out the branding remains the same so we don’t end up showing the community logo and branding in the product.

jameslau: @zregvart is there an automate function for this option or some form of a if/else to detect when it is Syndesis or Fuse Online? It wouldn't be efficient to load every instance of the images depending on the usage.

zregvart: Not sure how it's done when building the product. I do know that the values in `config.json` change. This is the end result:

https://github.com/syndesisio/fuse-online-install/blob/36261b43e34217ba12208c247b41bd9e80098ca3/resources/fuse-online-template.yml#L346-L355

jameslau: Hi @zregvart, I think the latest masthead commit will have the correct changes for this. @seanforyou23 also did some rework on the markup so that the `config.json` registers correctly without the oddity property values.

mcoker:
Thanks @jameslau. Some comments:

* I'm not seeing the background-image in the header.

* I would expect the left of the hamburger icon to be aligned with the left of the vertical nav text. 
  * Looks like it's mis-aligned in both desktop and mobile.

screen shot 2019-02-12 at 9 41 13 am
screen shot 2019-02-12 at 9 41 08 am

* you don't want `.pf-c-dropdown__toggle-icon` on the `.fa-question-circle` icon in the header unless there is text beside that icon.

* Looks like the font-size and font-family changed. Is that on purpose? 
  * not sure if you tried this, but you might set `$pf-global--enable-reset: false;` in the SCSS before PF4 is loaded, which will disable PF4's default reset SCSS and leave all the global styles alone. It will likely still set `font-size` and `font-family` (it sets these properties - https://github.com/mcoker/patternfly-next/blob/master/src/patternfly/_base.scss#L56-L61) on `.pf-c-page`, which will apply to the whole page since that class wraps the page, but you could simply reset those properties to PF3 styles to retain the PF3 font properties.

* The header icons collapse odd with double kebab icons

screen shot 2019-02-12 at 9 51 05 am
screen shot 2019-02-12 at 9 51 01 am

* should the navigation close on a narrow viewport? it currently does in fuse, and it does in the PF4 page layout by default, too, though it isn't here.

dongniwang:
Thanks for the comments @mcoker! 

- The background image - If this is optional and a product level decision, should consult @amysueg to make sure we are doing this (with or without background image) consistently across all products. 
- Hamburger icon alignment - If this is an easy fix and we can get this in on time, it would be a nice-to-have. 
- As for the `dropdown__toggle-icon` and the `.fa-question-circle` icon, please follow PF4 guideline and I think this should be fixed. 
- Font-size and font-family - Let's keep PF3 font properties for now until we officially move to PF4. That said, keep it the same as it is in the app today. 
- Double kebab icons - This should be fixed. 
- Navigation collapsing on a narrow viewport - This is also a nice-to-have if it's not a huge effort and we can get it in on time. But eventually we should fix this or make sure the behavior is consitent.

jameslau:
Thanks for the feedback @mcoker / @dongniwang!

mcoker:
Also wanted to mention that the dropdown menus in the header should use `.pf-m-align-right` since they're on the right edge of the screen. More about that here - https://pf4.patternfly.org/components/Dropdown/examples/

And that we might want to make a decision about how to handle the menus in the masthead on mobile. In patternfly3, the masthead dropdown menus are hidden and shown in the vertical navigation instead. In patternfly4, the dropdown menus collapse into a kebab menu. Currently we have both - the new masthead menus collapse into kebabs _and_ the user/help menus are shown in the vertical navigation.

amysueg:
Regarding @dongniwang 's comment above, for now, let's leave off the background image on the header and I'll recommend we do that for all the consoles adopting PF4 for the masthead. @jameslau

amysueg:
Update on the background image for the masthead-- @catrobson will advise on which background image to use.

mcoker:
Also I think we want to use [pf-icon-help](https://pf4.patternfly.org/icons) for the help icon in the header.

seanforyou23:
Good news is the masthead is pretty much finished up. Bad news is I goofed up the rebase here pretty badly. Shall we close this PR and open a new one with clean history?

gashcrumb:
Oh yeah, good call...  Probably could rebase that branch and force push to fix it but if it's just as easy to recreate it then go for it...

PR #4436, opened 2019-02-05
[api provider - collections] enhance connector/extension maven plugin to support collections type for data shapes

### LINKED ISSUE #4426: [api provider - collections] enhance connector/extension maven plugin to support collections type for data shapes

lburgazzoli:
Fixes #4426

@zregvart @christophd 

This is an initial prototype of the connector-maven-plugin to support collections for java data shapes which require some more polish but seems to work.

The new data shape definition (backward compatible) may looks like:

```json
{
    "outputDataShape": {
        "description": "...",
        "kind": "java",
        "type": "twitter4j.Status",
        "collectionType": "List",
        "collectionClass": "java.util.ArryList",
        "metadata": {
            "variant": "collection"
        },
        "variants": [
            {
                "kind": "java",
                "type": "twitter4j.Status",
                "metadata": {
                    "variant": "element",
                    "compression": "true"
                }
            }
        ]
    }
}
```

1. there are two new optional fields available: `collectionType` and `collectionClass` you can use to instruct the inspector generator about if and what type of collection the connector emits.
2. there is a new optional field `variants` that can holds nested shapes (exact same model) that can be used to provide alternative shapes, in our case the one related to the element.
3. the `metadata` field can be used to add additional information used to look-up the right variant as well as some info used to influence the inspector (i.e. if you add `"compression": "true"`, the generated inspection is stored as gzip+base64 to squeeze some bits)

An example of the plugin applied to the twitter connector can be found [here](https://gist.github.com/lburgazzoli/2aca8a4b5cddf4a0b03de55e9d63f831)

The `DataShape` interface has been adapted and enriched with two helpers:
```java


    default Optional<DataShape> findVariant(Predicate<DataShape> predicate) {
        if (predicate.test(this)) {
            return Optional.of(this);
        }

        return getVariants().stream().filter(predicate).findFirst();
    }

    default Optional<DataShape> findVariantByMeta(String key, String val) {
        return findVariant(ds -> {
            if (key == null || val == null) {
                return false;
            }

            return Objects.equals(
                ds.getMetadata().get(key),
                val
            );
        });
    }
```
Note that the root data shape is always included so you can use the same method no matter of the step from which you are looking it up i.e.

- `shape.findVariantByMeta("variant", "element")` can be used for the split step
- `shape.findVariantByMeta("variant", "collection")` can be used for the aggregate one


lburgazzoli:
added support for extensions

lburgazzoli:
@christophd @zregvart 

I've added support for extensions too, now ready for general review

zregvart:
Fails on license checks:
```
[INFO] Checking licenses...
[WARNING] Missing header in: /workspace/app/extension/maven-plugin/src/it/settings.xml
[WARNING] Missing header in: /workspace/app/extension/maven-plugin/src/it/with-data-shape/verify.groovy
[WARNING] Missing header in: /workspace/app/extension/maven-plugin/src/it/with-data-shape-variants/verify.groovy
```

lburgazzoli:
@zregvart @christophd I think we need an additional approval to trigger purebot

PR #4440, opened 2019-02-05
Fixes #4429 Suppport multi-selection for form fields

### LINKED ISSUE #4429: Suppport multi-selection for form fields
## This is a...

<pre><code>
[x ] Feature request
[ ] Bug report  
</code></pre>

## Description

In FHIR connector we need the user to be able to specify a list of contained resource types in a form of a multiselect field on a configure action step.

A multiselect field is supported by the underlying ng-dynamic-forms library, but not by our UI and backend. 
The change in UI is straightforward and backwards compatible.

In the backend we need to modify https://github.com/syndesisio/syndesis/blob/master/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectionActionHandler.java#L95 to accept Map<String, Object> properties since multiple choices will be deserialized as String[] and not String.

It will result in us having to refactor a number of public methods e.g. EncryptionComponent.encryptPropertyValues/.decrypt, WithConfigurationProperties.* and possibly a few more, which I missed at first glance. It's a rather straightforward refactoring, but not sure what's our policy on refactoring public methods in a backwards incompatible way. Are these part of our API contract e.g. when implementing extensions?

Alternately we could just modify ConnectionActionHandler.enrichWithMetadata to accept Map<String, Object> and convert it to Map<String, String> by converting any String[] value to a String using some custom separator like '|#|'. It's far from ideal, but does not require any refactoring and I did use this method initially as a workaround.

Which way to go? @paoloantinori, @zregvart, others?

rkorytkowski:
https://github.com/syndesisio/syndesis/issues/4429

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/connection/ConfigurationProperty.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/connection/ConfigurationProperty.java
@@ -32,6 +32,11 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 @SuppressWarnings("immutables")
 public interface ConfigurationProperty extends WithTags, Ordered, Serializable {
 
+    /**
+     * Separator for properties supporting multiple values
+     */
+    String MULTIPLE_SEPARATOR = ",";
+
     class Builder extends ImmutableConfigurationProperty.Builder {
         // make ImmutableConfigurationProperty.Builder accessible
     }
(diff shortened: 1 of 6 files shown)


gashcrumb:
Looks like this could use a rebase, it'd be good to get this in soon so it could get tested.

rkorytkowski:
Rebased...

Issue #4397, opened 2019-01-30
OData output results needs separating to allow for mapping and transmission to producers


phantomjinx:
The consumer odata connector produces results as json. However, in line with the underlying olingo4 component it has been implemented to provide all results in a single message. Unfortunately, this means that the data cannot be mapped to services such as gmail or slack. Therefore, need to split the results (consumer actually contains a consume.splitResults property) to facilitate correct mapping and communication.

zregvart:
Related to #4291

PR #4445, opened 2019-02-05
fix(#4397): split odata results object into separate messages

phantomjinx:
* ODataComponent
 * Adds the splitResult property to consumer which pokes the olingo4
   consumer to try and split the ClientEntitySet into individual ClientEntities.

* Olingo4Consumer
 * Implements the splitResult method. This to be pushed upstream

* ODataStartCustomizer
* *Serializer
 * Fixes the mapping serializers so that the resulting json can be
   correctly mapped by the data-mapper

* ODataMetaDataRetrieval|Extension
 * In order to do data-mapping correctly, json needs to generate a
   dynamic json-schema to map against. Generates the schema from the
   Edm metadata gleaned from the OData service when the configuration
   wizard is completed.


christophd:
Looks good to me! I would consider to return a singleton list of that one item here: https://github.com/syndesisio/syndesis/pull/4445/files#diff-fdb85cc529948bdfc6e24d459d386429R112

This would make sense to me as you also return an empty list when body is not of expected type and the customizer then returns a list type in any case.

PR #4447, opened 2019-02-05
fix(s2i): Removed unnecessary assembly

### LINKED ISSUE #4118: Docker image size of s2i increases over time
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description

While rebuilding the syndesis-s2i Docker image during development the image size increases from build to build. 1st build gives me an image size of 1.29 GB, 2nd build 1.8 GB and so on. 

Saw that behavior both in s2i-s2i build within Minishift and locally with --docker flag.

Increasing image size is a threat to the disk space of the Minishift VM so reset of Minishift becomes mandatory at some time.

rhuss:
Added some comments to the Dockerfile to clarify the process.
Fixes #4118.


rhuss:
Added some comments to the Dockerfile and removed the unneeded assembly.

Please remove your local ~/.m2/repository/io/syndesis/s21 directory to get rid of this zip locally, too.

The whole approach doesn't really scale as (a) the 1.25 GB size will always increase with the number of connectors and (b) this image needs to be loaded by every node where a build is running. So, there is a considerable time required for a first build on a node (which is non-deterministic)

christophd:
Looks good to me! Can we delete the file `src/assembly/repository.xml` then or is it used somewhere else?

rhuss:
True, this can be removed as well. Let me update the PR.

Issue #4309, opened 2019-01-18
[runtime] reduce integration definition size 


lburgazzoli:
The integration definition json is quite huge and contains lot of duplicated information well as bits that are not used by the runtime. 

We should remove the garbage before generating a prject/crd

rhuss:
Isn't it just that we pick only those parts from the integration definition which are relevant for a Camel-K resource ?

KurtStam:
I've removed all actions from the connectors as the action in use is already attached to the Step. Now the connection > connector adds very little to the size of the integration anymore. It'd be interesting to know if there are other scenarios where it still makes the integration large in size. Is there a way we can monitor the size in our QE tests?

tplevko:
@KurtStam I asked @avano to look into it. There are some possibilities we discussed, he'll let you know once we have some results.

avano:
@kurtstam if I got it right, it is about the size of the final integration.json file in the integration pod, is that correct?

we can implement such checks in our testsuite, but do you know where such big payloads can occur? we tried to brainstorm with @tplevko but we couldn't find any other place.

btw it is everything for the issue or will more commits follow?

KurtStam:
@avano - correct, I added a pruning step which runs just before it get's packaged up for the runtime. Big payloads used to occur for in particular for api-connectors such as the concur connector, where there are tons of actions per connector. Now we only send over the action in use.

all commits are in.

avano:
Hi Kurt, everything looks ok to me. Do you think this is worth automating? The only (naive) thing that came to our minds was to set a threshold for each step and then compare the integration size with (countOfSteps * threshold).

Does something better come to your mind?

PR #4450, opened 2019-02-06
fix: reduce integration definition size #4309

KurtStam:
I'm simply pruning all actions from the integration that is about to be generated to address #4309.

### DIFF
--- a/app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGeneratorHelper.java
+++ b/app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGeneratorHelper.java
@@ -24,6 +24,7 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.ListIterator;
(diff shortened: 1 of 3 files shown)


lburgazzoli:
wondering if we really need to keep the action inside the connector as such info is available on the step too

zregvart:
Seems that we need to pin another early access Camel release.

```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] Non-resolvable import POM: Could not find artifact org.apache.camel:camel-spring-boot-dependencies:pom:2.21.0.fuse-730033 in jboss-fuse (https://origin-repository.jboss.org/nexus/content/groups/ea) @ line 970, column 19
 @ 
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project io.syndesis:syndesis-parent:1.6-SNAPSHOT (/workspace/app/pom.xml) has 1 error
[ERROR]     Non-resolvable import POM: Could not find artifact org.apache.camel:camel-spring-boot-dependencies:pom:2.21.0.fuse-730033 in jboss-fuse (https://origin-repository.jboss.org/nexus/content/groups/ea) @ line 970, column 19 -> [Help 2]
```

I think @oscerd can do this for us.

oscerd:
I already pinned the 042

KurtStam:
> wondering if we really need to keep the action inside the connector as such info is available on the step too

@lburgazzoli good catch, we can in fact remove all actions.

KurtStam:
@zregvart, I think now that _all_ actions are removed we reducing the footprint significantly!

PR #4139, opened 2018-11-23
Fix #4138: Add support for all DSTU3 FHIR resources

rkorytkowski:
https://github.com/syndesisio/syndesis/issues/4138

PR #4454, opened 2019-02-06
SSO Enhancements revert

### LINKED ISSUE #3576: SSO Enhancements

zregvart:
Fixes #3576


install/syndesis.yml
```diff
@@ -469,6 +469,9 @@ objects:
           "logging": false
         },
         "branding": {
+          "logoWhiteBg": "",
+          "logoDarkBg": "",
+          "iconWhiteBg": "assets/images/FuseOnlineLogo_Black.svg",
```
rhuss: This change looks a bit unrelated.

install/syndesis.yml
```diff
@@ -996,7 +999,6 @@ objects:
       syndesis.io/type: infrastructure
     annotations:
       console.alpha.openshift.io/overview-app-route: "true"
-      haproxy.router.openshift.io/timeout: 60s
```
rhuss: Any idea why this was added in the first place and now removed ?

install/generator/04-syndesis-oauth-proxy.yml.mustache
```diff
@@ -75,9 +75,7 @@
           image: ${SYNDESIS_REGISTRY}/{{Images.OAuthProxyImagePrefix}}/{{ Images.Support.OAuthProxy }}:{{ Tags.OAuthProxy }}{{/WithDockerImages}}
           args:
             - --provider=openshift
-{{^WithOAuthClient}}
-            - --client-id=system:serviceaccount:${OPENSHIFT_PROJECT}:syndesis-oauth-client{{/WithOAuthClient}}{{#WithOAuthClient}}
-            - --client-id=syndesis-${OPENSHIFT_PROJECT}{{/WithOAuthClient}}
+            - --client-id=system:serviceaccount:${OPENSHIFT_PROJECT}:syndesis-oauth-client
             - --client-secret=${OPENSHIFT_OAUTH_CLIENT_SECRET}
```
rhuss: iiur this env var OPENSHIFT_OAUTH_CLIENT_SECRET is not needed when using an OAuthClient and could have been removed from the templates parameters ? (this is just a question to validate my understanding of how it worked)

rhuss: I just checked, so currently, this variable is currently generated automatically and reused when creating the OAuthClient. Howeve, I think it should be mandatory and the template should fail if this secret is not set (so removing the auto generation). Because the required service account will for sure have a different token. So the advantage would be that the template cannot be instantiated if the token is not provided, and this more or less requires the user to create the SA first and extract the token. Otherwise, the installation will just fail with an error that the SA is missing (as it has to be created from the outside). We could add some minor instructions to this template parameter so its clear that an SA has to be created in advance.

But maybe one could add a token alread to an SA, too ? Then we could the SA directly here like you did with the OAuthClient.

Just thinking a loud. Said that, I'm happy as it is as the templates are going away eventually anyway.

install/syndesis.yml
zregvart: This is a derived resource. Previous commit(s) didn't regenerate it.

install/syndesis.yml
zregvart: Seems to be added as a side effect of #4139 without adding it to the templates in `install/generator/`. @rkorytkowski if we need this it should be part of the templates in `install/generator/`

rkorytkowski: It's no longer needed. I was able to get the processing time significantly lower than the 30s default.

Issue #4331, opened 2019-01-22
Implicit split step not saved to DB


christophd:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
Some connectors (sql, aws-s3, servicenow) use an implicit split step for automatically iterating over a collection of results. The implicit split is added to the connector descriptor as an empty Json node (e.g. [sql-connector](https://github.com/syndesisio/syndesis/blob/master/app/connector/sql/src/main/resources/META-INF/syndesis/connector/sql.json#L90)). 

This empty split node never makes it to the jsondb. So when the syndesis server is reading an integration definition from the DB the implicit split step is gone.

## How to reproduce

- Create an integration with one of the start connections (sql, aws-s3, servicenow) 
- Stop the integration
- Rebuild syndesis-server or kill the server pod (to clear the cached integration definitions on the server)
- Start the integration

Result: Implicit split is not performed and the integration runs on the initial collection

christophd:
A quick fix for newly created integrations would be to add default properties to the empty split node in the connector descriptors for sql, aws-s3 and servicenow

```json
    "split": {
          "language": "simple",
          "expression": "${body}"
    }
```

The split definition will then be saved to the jsondb.

Though this will not fix existing integrations that already have been saved to the db (without the empty split node).

lburgazzoli:
We do need to add an update step to fix this.

@rhuss do we have some docs about this ?

christophd:
Funny thing is that the ui is always provided with the correct empty split node as these integration definitions seem to be built directly from the connector descriptor json files in META-INF read from classpath.

lburgazzoli:
I guess we do not need to expose the split step but instead we need to work on the new foreach https://github.com/syndesisio/syndesis/issues/4312. 

This bug impacts the current version but it should not require any UI work

gashcrumb:
Sure!  You may though want the UI to at least be able to properly render *something*.  If there's no data in a step it probably just shows some empty node or something.

christophd:
The split node is only a marker for the backend to automatically add a step at integration runtime. The implicit split is not visible to the user. The implicit split marker is also not part of the steps provided to the ui so we will not have to render that.

christophd:
We can close that issue as the implicit split has been removed completely. No need to save to DB anymore.

PR #4455, opened 2019-02-06
Implicit split integration migration

christophd:
With new split/aggregate steps we need to migrate old integrations that have been using the implicit split in connector descriptor.

This PR introduces a DB migration that automatically adds split and aggregate steps for these integrations.

- Updates the schema version to 28
- Add up-28.js migration script
- Removed implicit split and its handlers

The migration script operates on integrations that start with sql, servicenow, and aws-s3 connectors as these were making use of implicit split. The implicit split has never been saved to the jsondb (see https://github.com/syndesisio/syndesis/issues/4331) therefore the script identifies those integrations via connector id rather than looking for the split json node.


app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/SplitAggregateStepHandler.java
```diff
@@ -82,10 +82,10 @@ public boolean canHandle(Step step) {
         return Optional.of(route);
     }
 
-    public static class EndHandler implements IntegrationStepHandler {
+    public static class AggregateStepHandler implements IntegrationStepHandler {
```
lburgazzoli: I know that is quite a dummy class and it make sense now but I would move the aggregate step (and related utilities) to a separate class so we can have a generic and reusable step to be used for next iteration

christophd: Agreed! Pushed it to this PR

PR #4463, opened 2019-02-06
Feat(ui): integration editor UI tweaks

riccardo-forina:
## Changes

* new breadcrumbs
* updated toolbar, now built using PF3 toolbar
* updated api provider creation pages to use cards
* openapi editor reworked to be compatible with an in-card usage (might be broken elsewhere)
* reworked the main editor html/css structure to ease building full height scrollable panels

### DIFF
--- a/app/ui/src/app/integration/edit-page/current-flow.service.ts
+++ b/app/ui/src/app/integration/edit-page/current-flow.service.ts
@@ -90,6 +90,14 @@ export class CurrentFlowService {
     this._integration = undefined;
   }
 
+  isSaved(): boolean {
+    /**
+     * TODO: check if this is the proper way to check if the integration has been
+     * saved or not
+     */
+    return !!this.integration.id;
+  }
+
   isValid(): boolean {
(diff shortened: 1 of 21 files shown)


app/ui/src/app/integration/edit-page/current-flow.service.ts
```diff
@@ -90,6 +90,14 @@ export class CurrentFlowService {
     this._integration = undefined;
   }
 
+  isSaved(): boolean {
+    /**
+     * TODO: check if this is the proper way to check if the integration has been
```
gashcrumb: So far yes, this is all we use.

We *could* also use a metadata item to flag this, as well add a dirty flag actually which could be handy in the future, then it'd be in the `integration.metadata` map.  I can't recall if it allows values other than strings so that could be only one gotcha.

gashcrumb: BTW I'm not suggesting you do it for this PR :-)

PR #4467, opened 2019-02-07
feat(ui): API Provider editor pane-based layout

riccardo-forina:
## Changes

* redone the layout of the pages to support scrollable panes
* wrapped editor steps in cards

screenshot 2019-02-07 at 16 12 19
screenshot 2019-02-07 at 16 12 11
screenshot 2019-02-07 at 16 12 05
screenshot 2019-02-07 at 16 11 47
screenshot 2019-02-07 at 16 11 43
screenshot 2019-02-07 at 16 11 31

## Issues

width-issue

In this screencap the browser started at ~1200px and got up to ~1300px, but the main content (the rightmost pane) doesn't work great with that little space that's left. 
@dongniwang any suggestion?

Also, the connections grid doesn't look great on a white BG.

### DIFF
--- a/app/ui/src/app/common/openapi/editor/editor.component.scss
+++ b/app/ui/src/app/common/openapi/editor/editor.component.scss
@@ -7,6 +7,9 @@
     display: flex;
     flex-flow: column;
 
+    border-top: 2px solid transparent;
+    box-shadow: 0 1px 1px rgba(3, 3, 3, 0.175);
+
     &__header {
       display: flex;
       justify-content: space-between;
--- a/app/ui/src/app/common/ui-patternfly/list-toolbar/list-toolbar.component.scss
+++ b/app/ui/src/app/common/ui-patternfly/list-toolbar/list-toolbar.component.scss
@@ -3,3 +3,8 @@
     border-right: 0 !important;
   }
 }
+
+:host-context(.card-pf-heading) /deep/ .toolbar-pf {
+  border-bottom: 0 !important;
+  box-shadow: none !important;
+}
--- a/app/ui/src/app/integration/api-provider/editor-page/index.ts
+++ b/app/ui/src/app/integration/api-provider/editor-page/index.ts
@@ -1,0 +1,2 @@
+export * from './integration-api-provider-editor-page.component';
+export * from './integration-api-provider-editor-toolbar.component';
--- a/app/ui/src/app/integration/api-provider/editor-page/integration-api-provider-editor-page.component.html
+++ b/app/ui/src/app/integration/api-provider/editor-page/integration-api-provider-editor-page.component.html
@@ -3,52 +3,34 @@
   [content]="content"
 >
   <ng-template #content>
-    <div class="row">
-      <div class="col-md-12">
-        <ol class="breadcrumb">
+    <div class="integrations edit row">
+      <div class="header">
+        <ol class="syn-breadcrumb">
           <li>
-            &lt;&lt;<a [routerLink]="['/integrations']">All Integrations</a>
-          </li>
-          <li>
-            <a
-              [routerLink]="[
-                '/integrations',
-                currentFlowService.integration.id
-              ]"
-              >{{ currentFlowService.integration.name }}</a
-            >
+            <i class="fa fa-angle-double-left"></i>
+            <a [routerLink]="['/integrations']">All Integrations</a>
           </li>
         </ol>
-      </div>
-    </div>
-    <div class="row">
-      <div class="col-md-12">
         <syndesis-integration-flow-toolbar
           [isApiProvider]="true"
           [hideOperationsButton]="false"
         ></syndesis-integration-flow-toolbar>
       </div>
-    </div>
-    <div class="row">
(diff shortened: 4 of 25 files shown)


gashcrumb:
Some suggestings off the top of my head I'd say the vertical padding to the left and right of the operation list would need to go for starters, as well as reduce the gap between the IVP and the operation list.  The divider is too wide between the IVP and main content area.  Maybe there's excess much padding in the operation list items as well :-/

We could think about changing the step select page to use a list view I guess, I don't think that'd look right though.

riccardo-forina:
I played around using 1080px as the target width to support. I removed some paddings and reduced the columns to 300px each (from the 400px it was). The flow column is expanded to take into account the new design that takes more horizontal space than the current collapsed IVP.

1080-bs-grid

The grid is still cramped.

I then dumped the mediaquery based grid and tried a flex based one, and it's slightly better .

1080-flex-items
 
The cards' heights aren't consistent between rows, though. A grid built using the `display: grid` could address this, but I'm not sure about cross-browser support.

Anyway, I'm not sure we want to go "full custom" here. It would be nice to leverage PF components/design as much as possible I think...

gashcrumb:
I mean, when you're editing a particular flow do you really need the list of operations visible on the page in a list like that at all times?

I could envision a select element of some kind in the toolbar area to the right of the integration name to pick the current operation which would let us get rid of that operation list panel.

gashcrumb:
The main thing is if it's squished for 1080px, it'll be worse at 1024px or 800px when you're talking about viewing the console on a projector.

gashcrumb:
Also one thing to take into account, on other OSs you can have scrollbars which will take up more horizontal space.

riccardo-forina:
Updated the description.

gashcrumb:
K, with the last couple commits we're here for the API provider editor:

image

And the operation list page:

image

I wonder about that header...

dongniwang:
@gashcrumb Thanks for the update! Which header are you referring to?

mcoker:
@gashcrumb @riccardo-forina is there a particular part of this that you'd like me to look at?

riccardo-forina:
I guess the header @gashcrumb was referring to is the one with the operation path and name.
I tried a card-based approach, WDYT?

screenshot 2019-02-08 at 12 34 39
screenshot 2019-02-08 at 12 34 36
screenshot 2019-02-08 at 12 34 31
screenshot 2019-02-08 at 12 34 28

If we like it I guess all that's left is to update the IVP to show the box instead of the circle.

riccardo-forina:
Also maybe we could move the quick switch operation dropdown from the toolbar to the card title.

dongniwang:
@riccardo-forina - I do feel like it makes more sense to keep the dropdown in the toolbar area. 

I like the card approach. But maybe we can try a lighter grey for the card background? Right now it's blending with the main background. 

image

riccardo-forina:
Sure thing, like this maybe?

screenshot 2019-02-08 at 14 21 53

dongniwang:
@riccardo-forina, looks good to me. And per our discussion about the background colors, let's make sure to revisit this for PF4 React implementation. Thank you!

dongniwang:
Hey @riccardo-forina - Thinking about it some more and talking to @amysueg, we feel that it's fine to keep the original grey background color for the "Add to integration" area. Although it may look like a cut-out because the two backgrounds are blending together, it's probably ok for now. 

Here's a screenshot to make it more clear: 
- Remove the expand/collpase bar to text based control 
- Remove the card in card layout for "Choose a Step" page to make sure we have enough contrast for the step cards. 

Let me know if you have any other questions. 

screen shot 2019-02-08 at 11 55 18 am

riccardo-forina:
Thanks for this!

Just one thing, the "Choose a Step" page is shared with the editor for
normal integration, and all the subsequent steps are now wrapped in cards
as well.

I assume we want to stay consistent between every step and so we should
remove the card everywhere, right?

Il giorno ven 8 feb 2019 alle 18:00 Dongni (Iris) Wang <
notifications@github.com> ha scritto:

dongniwang:
@riccardo-forina - Do you mean pages like the following screenshot? Or are you talking about other pages?

screen shot 2019-02-08 at 12 08 29 pm

riccardo-forina:
Yeah exactly, here is the full list:

- new/edit integration: select connection
- new/edit integration: Select action
- new/edit integration : Configure step
- new/edit api provider operation: select connection
- new/edit api provider operation: Select action
- new/edit api provider operation: Configure step

Il giorno ven 8 feb 2019 alle 18:10 Dongni (Iris) Wang <
notifications@github.com> ha scritto:

dongniwang:
Yeah, I think visually, we are just removing the white background of the toolbar. 

screen shot 2019-02-08 at 12 17 34 pm

And the configure step page, should stay the same visually. 

I understand this is proably adding some work for you, let me know if you want to talk through anything. 

Technically, we can still use card component, but instead of white background, give it a grey? @mcoker @seanforyou23 @gashcrumb may know better.

Issue #4405, opened 2019-01-31
[api provider - collections] Introduce data shapes for split and aggregate steps


christophd:
New split and aggregate steps should provide a data shape with meta data retrieval.

The split step should provide a single element data shape representation of the step output sitting right before the split.

The aggregate step should provide a collection of the last step output data shape sitting in the split/aggregate.

The data shapes for these steps will be retrieved by meta data dynamic lookup.

mastepan:
Reopening for the QE to verify this if needed. It was in Done state previously so we have a task for this.

Issue #3896, opened 2018-10-18
Connectors now should support returning an Array


KurtStam:
## This is a...

<pre><code>
[*] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
With https://github.com/syndesisio/syndesis/pull/3895 addressed, connectors like SQL, Twitter, etc should now support returning an array of entities, on top of the split functionality we have in place currently. This should be a setting when selecting the action.

For example SQL using "SELECT * FROM todo", currently sends a message for each row matched (and in a flow you only get the first match), should have the option to return an array with allow rows. So if this is set the datashapeOutput body will be of type array.

lburgazzoli:
We get the first match only because of the sql post processor explicit discard all the rows but the first
The Split EIP should also work for intermediate steps if needed (of course in addition to allow returning an array)

KurtStam:
@lburgazzoli what's the Split EIP? We don't ship a split Step last time I checked. BTW my concern is about the definition of the DataShape.

lburgazzoli:
The Split EIP is what it is used under the hoods as example from the sql actions that have "From" as pattern so the resulting collection is split into multiple exchanges, see https://github.com/syndesisio/syndesis/blob/master/app/connector/sql/src/main/resources/META-INF/syndesis/connector/sql.json#L148.

The data shape do not need to change to use the Split EIP as DataMapper will receive a new exchange for each matching row with a body that matches the DataShape. 

Btw, I've nothing against returning an array was just to let you know that there is also the possibility to automatically split a collection if needed.

KurtStam:
@lburgazzoli yes well not completely 'automatically' right, because the outputShape is different when using array. So I think we need to add a setting on the Connection configuration screen so that the user can set what output shape is expected: 

1. single entities & split, or 
2. array and no split.

lburgazzoli:
I'm not sure if the data shape need to be different as imho it should describe the type of elements of the collection, not the collection itself. 

We could add:
- a tag to the action telling that a collection can be returned
- an attribute/tag to the step to configure if split has to be applied or not

The DataMapper UI can then use the step attribute to determine if the datashape  is part or not of a collection.

Of course that's only my humble opinion :)

zregvart:
[...]

PR #4510, opened 2019-02-11
[#4405] Metadata data shapes for split/aggregate steps

christophd:
As follow-up of https://github.com/syndesisio/syndesis/issues/4426  introduced new server endpoint `/steps/{kind}/descriptor` that should be called by UI when split/aggregate steps are added to an integration.

Requests to that endpoint should provide proper input/output data shapes according to the step right before the split or aggregate. The server endpoint evaluates a matching data shape variant (element or collection) for the split/aggregate step and returns that variant as output data shape for the split/aggregate step in the response.

When no explicit variant (element or collection) is given the server tries to automatically convert the given data shape to collection or single element shape. This auto conversion of types has been implemented for data shapes 'json-schema' and 'json-instance'. 

Data shapes of kind 'java' and 'xml' do require an explicit variant marked as element or collection variant.


christophd:
@zregvart @lburgazzoli I decided to not go from server-endpoint to server-meta with an explicit API call as we have all information for the step metadata retrieval (data shapes) available in server-endpoint. 

I think separation between server-endpoint and server-meta is based on the fact that only server-meta has access to connectors. In this case we do not need connector specifics so I stayed within server-endpoint boundaries in order to save me the API call to server-meta.

What do you think? Should we move that step metadata retrieval to server-meta for some reason?

app/common/model/src/test/java/io/syndesis/common/model/DataShapeBuilderTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.zip.GZIPInputStream;
+
+import org.junit.Assert;
```
zregvart: Would use AssertJ's `assertThat` instead of `assertEquals`, but thats just a style preference

app/common/model/src/test/java/io/syndesis/common/model/DataShapeBuilderTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+    private void assertCompressedSpecification(String expected, String compressed) throws IOException {
+        byte[] data = Base64.getDecoder().decode(compressed);
+
+        try (InputStream is = new GZIPInputStream(new ByteArrayInputStream(data))) {
+            Assert.assertEquals(expected, readToString(is));
+        }
+    }
+
+    private String readToString(InputStream is) throws IOException {
```
zregvart: `io.syndesis.common.util.IOStreams::readText`

app/extension/maven-plugin/src/main/java/io/syndesis/extension/maven/plugin/GenerateMetadataMojo.java
```diff
@@ -581,26 +577,24 @@ private void generateAtlasMapInspections() throws MojoExecutionException {
...
+                DataShape.Builder builder = new DataShape.Builder()
+                                                    .createFrom(dataShape.get())
+                                                    .specification(inspection);
+
+                builder.variants(dataShape.get().getVariants()
+                                                 .stream()
+                                                 .map(variant -> {
+                                                     try {
+                                                         return generateInspections(actionId, Optional.of(variant));
+                                                     } catch (Exception e) {
```
zregvart: Bit ugly with catching any Exception, might cause bugs to be unnoticed...

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -0,0 +1,98 @@
...
+import io.syndesis.common.model.connection.DynamicActionMetadata;
+import io.syndesis.common.model.integration.StepKind;
+import io.syndesis.common.util.Json;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author Christoph Deppisch
+ */
+public class AggregateMetadataHandler implements StepMetadataHandler {
```
zregvart: perhaps package private, would it work?

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/StepMetadataHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+package io.syndesis.server.endpoint.v1.handler.meta;
+
+import io.syndesis.common.model.connection.DynamicActionMetadata;
+import io.syndesis.common.model.integration.StepKind;
+
+/**
+ * @author Christoph Deppisch
+ */
+public interface StepMetadataHandler {
```
zregvart: perhaps package private and the implementations also

lburgazzoli:
looks good @christophd is it safe to merge it in master or you expect side effects ?

app/common/model/src/main/java/io/syndesis/common/model/DataShape.java
```diff
@@ -90,5 +99,94 @@
     }
 
     class Builder extends ImmutableDataShape.Builder {
+
+        public static final String COMPRESSION_METADATA_KEY = "compression";
+        private static final String COMPRESSED_METADATA_KEY = "compressed";
+
+        public Builder decompress() {
+            DataShape currentShape = build();
+            if (Objects.equals(currentShape.getMetadata().get(COMPRESSION_METADATA_KEY), "true") && Objects.equals(currentShape.getMetadata().getOrDefault(COMPRESSED_METADATA_KEY, "true"), "true")) {
```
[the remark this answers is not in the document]
christophd: It is checked for COMPRESSION_METADATA_KEY and COMPRESSED_METADATA_KEY where the first is for compression enabled at all and the 2nd to mark that compression has already been performed in order to avoid multiple compressions.

However, I will simplify the predicate a bit

app/common/model/src/main/java/io/syndesis/common/model/DataShape.java
```diff
@@ -90,5 +99,94 @@
...
+                            .variants(currentShape.getVariants()
+                                      .stream()
+                                      .map(variant -> new DataShape.Builder()
+                                               .createFrom(variant)
+                                               .decompress()
+                                               .build())
+                                      .collect(Collectors.toList()))
+                            .specification(readToString(is));
+                } catch (IOException e) {
+                    throw new IllegalArgumentException("Failed to decompress data shape", e);
```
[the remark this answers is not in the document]
christophd: 👍 will do

app/common/model/src/main/java/io/syndesis/common/model/DataShape.java
```diff
@@ -90,5 +99,94 @@
...
+                                        .build())
+                                .collect(Collectors.toList()));
+            }
+
+            return this;
+        }
+
+        public Builder compress() {
+            DataShape currentShape = build();
+            if (Objects.equals(currentShape.getMetadata().get(COMPRESSION_METADATA_KEY), "true") && Objects.equals(currentShape.getMetadata().getOrDefault(COMPRESSED_METADATA_KEY, "false"), "false")) {
```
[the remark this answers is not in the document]
christophd: Will also simplify that expression a bit to be less confusing.

Both COMPRESSION_METADATA_KEY and COMPRESSED_METADATA_KEY are checked here where the first is for compression enabled and the 2nd to make sure that compression/decompression is only performed once.

app/common/model/src/test/java/io/syndesis/common/model/DataShapeBuilderTest.java
christophd: True, I keep switching between those based on what the test has been using before but `assertThat` should be the first choice for new tests

app/common/model/src/test/java/io/syndesis/common/model/DataShapeBuilderTest.java
christophd: Ah nice. Haven't seen this one. 

I have had `org.apache.commons.io.IOUtils` before but this introduces a new compile time dependency to `syndesis-model` and to all modules that depend on `syndesis-model`.

The common util helper in Syndesis matches perfectly, thanks!

app/extension/maven-plugin/src/main/java/io/syndesis/extension/maven/plugin/GenerateMetadataMojo.java
christophd: true, at least log the error as a warning

christophd:
@lburgazzoli I think it is save to merge as it is just a new server endpoint that is not called by ui at the moment.

We might need to sync UI then with https://github.com/syndesisio/syndesis/issues/3896 where we adjust the connector data shapes to use collection types instead of single element types

lburgazzoli:
so LGTM, I let @zregvart approve when he think it is ready

zregvart:
@christophd are you're doing some clean-up in this or in subsequent PR?

christophd:
@zregvart already pushed some improvements to this PR yesterday. The refactoring to go via syndesis-meta preferably will be done in a subsequent PR.

PR #4513, opened 2019-02-11
Bump version of jackson-databind to 2.8.11.3

jonathanchristison:
See https://github.com/FasterXML/jackson-databind/blob/jackson-databind-2.8.11.3/release-notes/VERSION#L6 for security related fixes


zregvart:
@jonathanchristison I assume you want this on `master` branch also?

zregvart:
> @jonathanchristison I assume you want this on `master` branch also?

Oh, I see we use `2.9.8` on `master`, so we should be good...

PR #4460, opened 2019-02-06
feature(operator): First step to upgrade to Operator SDK 0.5.0

rhuss:
At least it compiles ;-) Still to check: Template handling (i.e. how to get the Config from the client) has probably been broken

PR #4517, opened 2019-02-11
feat(install): Support for addon resources in syndesis-operator

### LINKED ISSUE #4381: Support for "addon" resources in syndesis-operator
For integrating with external systems (e.g. Prometheus, Grafana) it would be very valuable when the Syndesis operator would install arbitrary resources stored in a directory within the operator image.

A sample workflow would look like this:

* Within the Operator image, there is a directory `/addons` which can contain any possible Kubernetes resource in YAML foramt. This can also be also custom resources like a `ServiceMonitor` for the Prometheus Operator. The syndesis-operator build process should be modified to pick up those file from _somewhere_ and put it into this directory.
* When the syndesis-operator is running and someone deploys a `Syndesis` resource, after the operator has installed syndesis, it checks this directory and for every resource found in this directory:
  - Adds an `ownerReference` to the `Syndesis` resource
  - Adds a label/annotation indicating that it has been installed on behalf of the operator (e.g. `syndesis.io/addon-resource`)
  - Applies this resource to the current namespace

This ensures, that when the Syndesis resources are deleted, also the addon resources are deleted.

The Installation of an Operator would do the following steps;

* The new operator gets started. During startup, and when it detects an upgrade situation (e.g. a Syndesis resource is already installed) it checks for all resources with a label `syndesis.io/addon-resource' and then for each delete the resource.
* When an existing installation needs to be updated, just install the addon resources from the new operator image after the update process has finished.

@jamesnetherton @alexkieling @lgarciaaco I hope this summarizes our discussion accurately. 

@heiko-braun this feature is requested by Integreatly so that Syndesis could be monitored by an external, integreatly provided Prometheus instance. Is there a way that we can schedule this work into one of our next releases ?

astefanutti:
Fixes #4381.

We may want to either:
* Add an option to deactivate add-on by default
* Or do not fail-fast on add-on resource creation error

### DIFF
--- a/install/operator/pkg/controller/syndesis/syndesis_controller.go
+++ b/install/operator/pkg/controller/syndesis/syndesis_controller.go
@@ -2,13 +2,13 @@ package syndesis
 
 import (
 	"context"
-	"github.com/syndesisio/syndesis/install/operator/pkg/syndesis/action"
-	"k8s.io/apimachinery/pkg/types"
 	"reflect"
 
-	syndesisv1alpha1 "github.com/syndesisio/syndesis/install/operator/pkg/apis/syndesis/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -16,6 +16,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	syndesisv1alpha1 "github.com/syndesisio/syndesis/install/operator/pkg/apis/syndesis/v1alpha1"
+	"github.com/syndesisio/syndesis/install/operator/pkg/syndesis/action"
 )
 
 var log = logf.Log.WithName("controller")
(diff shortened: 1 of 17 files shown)


heiko-braun:
@rhuss Can you take a look at this?

rhuss:
@astefanutti is this based on top of #4460 or on the state on master ?

astefanutti:
@rhuss it's based on `master` branch. Once #4460 is merged I can rebase this PR.

rhuss:
@astefanutti ok. I expect to be in a mergeable state tomorrow evening with #4460 I will review then this PR after your rebase.

astefanutti:
@rhuss thanks for the update. Sounds good to me!

astefanutti:
I've rebased the PR on top of #4460 and changed the original behaviour so that the `--addons` option must be explicitly set for the addons to be applied.

install/operator/pkg/controller/syndesis/syndesis_controller.go
```diff
@@ -39,12 +42,24 @@ func init() {
...
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileSyndesis{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
+	clientset, err := kubernetes.NewForConfig(mgr.GetConfig())
+	if err != nil {
+		return nil, err
+	}
+	return &ReconcileSyndesis{
+		apis:   clientset,
```
rhuss: Why are these apis needed ?

rhuss:
Looks good but conflicts a bit with the refactorings of #4574 where the client and scheme are put into a baseAction struct.

tbh didn't get it what this `apis` field is for. Is a scheme not good enough ?

astefanutti:
The `apis` field stores a references to the Kubernetes interface, under the `apis` endpoint, that gives access to the discovery API. It's used in the upgrade action to get the supported types to then delete the addons resources. While it is possible to create that interface client for each upgrade, it can be useful to initialise it a start time so that it's available in addition to the operator SDK client (with cache, informers, limited endpoints...).

rhuss:
sounds good. let's merge it.

PR #4522, opened 2019-02-12
fix(monitoring): Add missing roles for monitoring resources

astefanutti:


heiko-braun:
@astefanutti What happens if these permission are not given or present?

heiko-braun:
@astefanutti Fail fast seem to be to rigid. We cannot assume the other operator and needed permission to present in all environments. Can you change it to warning statements instead?

astefanutti:
@heiko-braun would you mind commenting on #4381, which is where the code responsible for addons creation is. I did not make a decision about error handling in order to gather opinion / feedback first, though I agree we should either:
* Add an option to deactivate add-on by default
* Or do not fail-fast on add-on resource creation error, at least for missing permission or missing CRDs.

Anyway, I think the permissions added by this PR can be safely added to the default deployment.

PR #4541, opened 2019-02-13
fix(license:format) Fix call to license:format by adding license pluign to extension & integration BOM

rhuss:


zregvart:
Not sure about this, the license plugin should be used only from parent, i.e.

    ./mvnw -N license:format

rhuss:
@zregvart yeah, but this doesn't work as the bom's affected have no parent. That's the reason why this is not checker nor can be formatted. But feel free to close this if that's not correct.

zregvart:
This configuration:

https://github.com/syndesisio/syndesis/blob/48c95d51abcb0e19a16789521a660dc910fb9510/app/pom.xml#L587-L590

should include any file in a `src` directory and any `pom.xml` files anywhere underneath the `app` directory.

zregvart:
And that's why we only execute it from the parent and do not inherit it in other modules that do inherit from parent:

https://github.com/syndesisio/syndesis/blob/48c95d51abcb0e19a16789521a660dc910fb9510/app/pom.xml#L575

PR #4546, opened 2019-02-13
OData connector returns collection data shape

christophd:
OData connector now using json-schema array collection type where applicable


lburgazzoli:
@phantomjinx can you have a look ?

phantomjinx:
Looks fine. As long as it will still data map ok to the likes of gmail then I am happy. ;-)

christophd:
@phantomjinx can you provide a sample integration with odata and gmail that should be tested as a regression test?

phantomjinx:
@christophd 

Syndesis Export : [TriPinIntegration-export.zip](https://drive.google.com/open?id=1AqdwSz-JlR1Y5w9bjbWdRkcUkrWQB0YA)

Issue #4406, opened 2019-01-31
UI provides data shape for split and aggregate steps


christophd:
As part of https://github.com/syndesisio/syndesis/issues/4405 we need the UI to call syndesis-meta when split and aggregate steps are added to the integration in order to receive dynamic output data shapes for these steps.

For split step the UI should provide the output data shape of the step sitting right before the split. Meta server will evaluate and return a proper output data shape for the split step representing a single element in the collection.

For aggregate step the UI should provide the output data shape of the last step in the split/aggregate to the server. Meta server will evaluate and return a proper output data shape for the aggregate step.

lburgazzoli:
@paoloantinori @gashcrumb FYI

gashcrumb:
Do we have an issue also for creating the split/aggregate step entries in UI's step store by any chance, or is this issue covering that too?  If the latter, what should the split and aggregate step definitions look like?  Or should I just [rename these](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/store/step/step.store.ts#L135-L154)?

christophd:
The rename to split/aggregate is pending with PR https://github.com/syndesisio/syndesis/pull/4455 (commit https://github.com/syndesisio/syndesis/pull/4455/commits/3aa8b013d633683cc6da90c92158bb319f1a3e1e)

The step definitions will be very simple as in a first iteration the user will not have any options available on those.

christophd:
@gashcrumb renamed steps are now available on master

gashcrumb:
Do we have endpoints for this in place?

christophd:
There is a new server endpoint `/steps/{kind}/descriptor` where `{kind}` path parameter is the step kind (one of: endpoint, connector, expressionFilter, ruleFilter, extension, mapper, split, aggregate, log, headers, template)

The endpoint receives a POST request with a json payload representing the current step meta data. The step meta data should provide input and output data shapes:

For instance:
```
{
  "inputShape" : {
    "kind" : "json-schema",
    "type" : "type",
    "name" : "name",
    "description" : "description",
    "specification" : "{ ... }"
  },
  "outputShape" : {
    "kind" : "none",
    "type" : "type"
  }
}
```

For `split` step the input and output shapes of the preceding connection should be provided as is.
For `aggregate` step the input and output shapes of the preceding step should be provided as is.

The endpoint responds with an adjusted json object representing the input and output data shapes that should be set to the respective step (split/aggregate) that the request has been called for.

[...]

PR #4547, opened 2019-02-13
SQL connector returns collection data shape

christophd:
SQL connector now using json-schema array collection type as output data shape for sql query actions.

Store procedure actions stay as is as the underlying Camel component is not able to provide a list of results but only a single entry as result set.

### DIFF
--- a/app/connector/sql/src/main/java/io/syndesis/connector/sql/SqlMetadataRetrieval.java
+++ b/app/connector/sql/src/main/java/io/syndesis/connector/sql/SqlMetadataRetrieval.java
@@ -46,6 +46,9 @@ import org.apache.camel.component.extension.MetaDataExtension.MetaData;
 
 @SuppressWarnings("PMD.GodClass")
 public final class SqlMetadataRetrieval extends ComponentMetadataRetrieval {
+
+    private static final String JSON_SCHEMA_ORG_SCHEMA = "http://json-schema.org/schema#";
+
     static final String PROCEDURE_NAME = "procedureName";
     static final String PROCEDURE_TEMPLATE = "template";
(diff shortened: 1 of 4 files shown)


lburgazzoli:
@KurtStam can you have a look ?

KurtStam:
@christophd, just to double check, it's now ALWAYS returning a collection?

christophd:
@KurtStam yes that is correct

KurtStam:
@christophd Looks good. What's left to do (WIP?)?

christophd:
@KurtStam thanks, just want to make sure that merging to master produces a useful state. maybe we need some UI work done first before changing the data shapes. Need to check if this can be merged before that

lburgazzoli:
yes we need UI to be ready before merging otherwise we'll have troubles using the connector

gashcrumb:
> yes we need UI to be ready before merging otherwise we'll have troubles using the connector

@lburgazzoli @christophd It'd be good to update one of these though so I actually know what needs to be addressed.  Also this would probably help me out in implementing #4406 unless there's a better connection to test with.

lburgazzoli:
I think this one is a good candidate however it may break lot of things if merged, any chance to run syndesis back-end from this pr ?

gashcrumb:
> I think this one is a good candidate however it may break lot of things if merged, any chance to run syndesis back-end from this pr ?

Yeah, I'll have to test it out later today I guess.  Is there an issue BTW tracking what the potential UI impact is?

gashcrumb:
Tested a backend with this running and I don't see huge breakages.  Obviously adding a data mapper without the split step will show things but as a quick test you can still make integrations and save them.

PR #4558, opened 2019-02-13
 feat: fetch descriptors for split/aggregate steps

### LINKED ISSUE #4406: UI provides data shape for split and aggregate steps

gashcrumb:
fixes #4406

Using a connector that returns an array (using the code in #4547 running in the backend):

image

### DIFF
--- a/app/ui/src/app/integration/edit-page/current-flow.service.ts
+++ b/app/ui/src/app/integration/edit-page/current-flow.service.ts
@@ -12,9 +12,16 @@ import {
   Flow,
   Flows,
   StepOrConnection,
+  Action,
 } from '@syndesis/ui/platform';
 import { log, getCategory } from '@syndesis/ui/logging';
-import { IntegrationStore, DATA_MAPPER, StepStore } from '@syndesis/ui/store';
+import {
+  IntegrationStore,
+  DATA_MAPPER,
+  StepStore,
+  SPLIT,
+  AGGREGATE,
+} from '@syndesis/ui/store';
 import {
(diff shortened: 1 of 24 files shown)


app/ui/src/app/integration/edit-page/current-flow.service.ts
```diff
@@ -578,6 +587,40 @@ export class CurrentFlowService {
...
+    step: Step,
+    then: (step: Step) => void
+  ): any {
+    switch (step.stepKind) {
+      case SPLIT:
+      case AGGREGATE:
+        // A split step needs the data shape of the previous thing with a data shape
+        const prev = this.getPreviousStepWithDataShape(position);
+        this.integrationSupportService
+          .getStepDescriptor(SPLIT, {
```
christophd: @gashcrumb I think `step.stepKind` should be passed to `getStepDescriptor` as a first argument

christophd: see PR for this: https://github.com/syndesisio/syndesis/pull/4561

PR #4561, opened 2019-02-14
#4406 Fixed step kind argument when fetching step descriptor

christophd:


gashcrumb:
Whoops missed that, thanks!

Issue #4151, opened 2018-11-23
Add Prometheus alerts and SOPs


heiko-braun:
As an administrator of Fuse on Openshift and/or Fuse Online, I want to be able to use some OOTB alerts. These OOTB alerts sett eh thresholds on my Prometheus metric or metrics to inform me that action may need to take place.
Each alert should have a corresponding SOP (standard operating procedure) The focus of the SOP is to quickly navigate the steps required to get the system to back to an acceptable state.

Example pod X has restarted multiple times in 24hrs. Alert is triggered: email sent to a@b.c
SOP: 1. login to cluster, 2 check list of running pods includes x->y 3 clear log files in location z

PR #4565, opened 2019-02-14
feat(monitoring): Reference SOP documentation from alerts

jamesnetherton:
Ref #4151

### DIFF
--- a/doc/managing_environments/topics/alerting_sop.adoc
+++ b/doc/managing_environments/topics/alerting_sop.adoc
@@ -1,0 +1,119 @@
+[id='alerting-standard-operating-procedures']
+= Standard Operating Procedures
+
+== Syndesis Infrastructure - Database Alerts
+
+=== FuseOnlineDatabaseInstanceDown
+
+*Alert name*
+
+FuseOnlineDatabaseHighRollbackRate
+
+*Description*
+
+The syndesis-db pod is not running or is not serving requests.
+
+**Process steps**
+
+1) Log in to the OpenShift cluster.
+
+2) Switch to project namespace identified in the alert message.
+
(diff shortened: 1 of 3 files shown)


lgarciaaco:
We need to set clear expectations for these alerts. I am going to take  *FuseOnlineDatabaseHighRollbackRate* and *FuseOnlineDatabaseHighConnectionCount* as example. These alerts **SHOULD not** be exposed to NOC, mainly because there is nothing they can do to solve the problem (probably not even engineers from our team can do something more than further investigation). I am not saying that we don't need the metrics, but more as a support for further debugging (probably a dashboard is enough).

*FuseOnlineDatabaseHighRollbackRate* and *FuseOnlineDatabaseHighConnectionCount* will cause the application using the database to start throwing errors, and that is what we want to monitor/alert upon.

Imho, only golden signals should be monitored atm: https://www.infoq.com/articles/monitoring-SRE-golden-signals

Maybe this is not the right place but I just wanted to start the discussion :)

jamesnetherton:
Sure - makes sense. I felt compelled to add them as I didn't have much in the way of alerting for the DB. But you're right. We should focus on symptoms, not causes. The DB dashboard will cover those metrics anyway.

If you think we should definitely remove them, I'll raise a PR shortly to do that.

lgarciaaco:
I think *FuseOnlineRestApiHighEndpointErrorRate*, *FuseOnlineRestApiHighEndpointLatency* and *FuseOnlineDatabaseInstanceDown* are the ones we want to receive alerts from, the rest should be dashboarded or the alerts should go to a channel for history purposes.

jamesnetherton:
Ok - agreed. 

Let's aim to keep things simple to being with. I'll remove the other redundant alerts for now. They are already covered comprehensively by the dashboards.

jamesnetherton:
PR updated to reflect the reduced set of alerts as we discussed.

PR #4574, opened 2019-02-14
refactor(operator): Some operator optimizations

rhuss:


rhuss:
@astefanutti If you could this PR a quick look ? It's about refactoring the action structure, making the signature a bit smalled, and using the same context.Context everywher (although I really dont exactly know hat it is used for ;-)

install/operator/pkg/syndesis/action/upgrade.go
```diff
@@ -92,10 +97,28 @@ func (a *upgrade) Execute(scheme *runtime.Scheme, cl Client, syndesis *v1alpha1.
...
 
-				err = createOrReplaceForce(cl, res, true)
+				err = createOrReplaceForce(ctx, a.client, res, true)
+				if err != nil {
+					return err
+				}
+			}
+
+			// Install addons
+			if addonsDir := *configuration.AddonsDirLocation; len(addonsDir) > 0 {
```
astefanutti: Looks like the install addons code got duplicated :)

rhuss: Good catch !

PR #4576, opened 2019-02-14
#4405 Aggregate step should provide input shape of subsequent step

christophd:
Aggregate step should not only provide output shape of previous step but also the input shape of the subsequent step.

If subsequent step defines a collection shape the server step metadata lookup provides the single element variant of that shape.


christophd:
@gashcrumb sorry for the hacky ui adjustments!

PR #4579, opened 2019-02-15
feat: finish up ci/cd dialog functions (almost)

### LINKED ISSUE #4494: Implement UXD design for add to CI/CD dialog
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description

Implement this design -> https://redhat.invisionapp.com/share/E9QCV528WVN#/screens/345130937

Using this info -> https://github.com/syndesisio/syndesis/issues/4326

backend API -> https://github.com/syndesisio/syndesis/blob/f178db73b5124913ffeaf84456b92b0e4cb2b78b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/continuousdelivery/ContinuousDeliveryHandler.java

Object definition -> https://github.com/syndesisio/syndesis/blob/f178db73b5124913ffeaf84456b92b0e4cb2b78b/app/common/model/src/main/java/io/syndesis/common/model/integration/ContinuousDeliveryEnvironment.java

gashcrumb:
fixes #4494 

Well, almost.  The rename function doesn't work, I'm not sure if I just don't have the request data set up or what.  For example I wound up making myself an environment literally named `\{ "environment": "foo", "newEnvironment": "bar" \}`, oops.  But from what I can see the API expects just a string, but I send it that and it's rejected.

The "remove" functionality works I think as described.  The "add" button next to the dialog kinda didn't make sense to me but I think the way it behaves is as described.

Anyhoo, @riccardo-forina @dhirajsb if you get a chance to play with it that'd be awesome, it's like 90% there I think...

image

### DIFF
--- a/app/ui/src/app/core/providers/integration-support-provider.service.ts
+++ b/app/ui/src/app/core/providers/integration-support-provider.service.ts
@@ -192,19 +192,24 @@ export class IntegrationSupportProviderService extends IntegrationSupportService
   ): Observable<Map<String, ContinuousDeliveryEnvironment>> {
     return this.apiHttpService
       .setEndpointUrl(integrationEndpoints.tags, { integrationId })
-      .post<any>({ environments });
+      .post<any>(environments);
   }
 
   untagIntegration(integrationId: string, env: string): Observable<void> {
     return this.apiHttpService
-      .setEndpointUrl(integrationEndpoints.deleteTag, { integrationId, env })
+      .setEndpointUrl(integrationEndpoints.deleteTag, {
+        integrationId,
+        env: encodeURIComponent(env),
+      })
       .delete();
   }
 
   renameEnvironment(oldEnv: string, newEnv: string): Observable<void> {
     return this.apiHttpService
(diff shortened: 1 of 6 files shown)


dhirajsb:
I tried testing locally but all links, `Add new`, `Edit` and `Remove` point to `https://syndesis.192.168.64.42.nip.io/null` and reload the ui when clicked.

dhirajsb:
Sorry, it was an issue with my environment. Actions `Add new`, and `Remove` are indeed working. `Rename` doesn't seem to be working and adds braces around the new value for some reason. For example, the value `testing` becomes `{"testing"}`.

dhirajsb:
Ok, I tested a little more, and it looks like it's complaining that the request `Content-Type` should be set to 'application/json' instead of 'plain/text'.

riccardo-forina:
Thanks @dhirajsb, adding the content-type header fixed the issue.
@gashcrumb I made the form take the full width, but other than that it LGTM.

gashcrumb:
Awesome, thanks guys!

Issue #4563, opened 2019-02-14
Update split/aggregate step icons


christophd:
Update icons to official step icons for split and aggregate

zregvart:
The icons should be placed here:

https://github.com/syndesisio/syndesis/tree/master/ux/connections_icons/svg

Since these are step icons (not connector icons), the script to optimize/manage the SVGs probably needs to be updated also.

christophd:
@zregvart do you have time to look at this?

zregvart:
Can do, sometime next week I think.

tplevko:
Verified, closing.

PR #4585, opened 2019-02-15
#4563 Update split and aggregate step icons

christophd:
Using official split and aggregate step icons


christophd:
@zregvart do we need icon optimization here?

zregvart:
> @zregvart do we need icon optimization here?

Not strictly needed as those are served from the web server not from the database but it is beneficial.

christophd:
@zregvart can we approve this PR and optimize later on?

Issue #4596, opened 2019-02-15
pf4 masthead dropdown and alignment issues


mcoker:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
When the user menu collapses in a narrow viewport, the "logout" link should move into the kebab with a separator between the help menu and the login link.

Also the left edge of the nav text should align with the left edge of the hamburger icon.

## Tasks involved / Steps to Reproduce

1. Shrink the viewport down and notice how the "logout" link moves to the kebab menu, but there is no separator. There should be a separator there.
2. Look at the left edge of the vertical nav text and see that it doesn't align with the left edge of the hamburger icon on desktop and mobile. Those should align.

tplevko:
Verified

PR #4597, opened 2019-02-15
fix(masthead): add separator in masthead dropdown, align nav text with hamburger

mcoker:
fixes https://github.com/syndesisio/syndesis/issues/4596


gashcrumb:
I guess @seanforyou23 has to approve it too for the bot to be happy :-)

Issue #4257, opened 2019-01-09
 API Provider: Collection Support 


heiko-braun:

christophd:
There are lots of different issues and discussions around this epic so I want to collect/summarize what we are going to implement:

* https://github.com/syndesisio/syndesis/issues/4346 Migrate integrations
  * Auto translate implicit split that is used by connectors (sql, asw-s3, servicenow) to new split/aggregate steps
  * Either do this via DB migration script or auto translate when providing old integrations from DB to UI
  * User might be provided with a warning/info that the integration has been migrated/auto translated
* https://github.com/syndesisio/syndesis/issues/3896 Add collection support to connectors
  * Connectors such as SQL, Twitter will return collections only
  * User can choose to add new split/aggregate steps in order to operate on individual items in these collections or work with the collection itself
  * Old integrations will be provided with auto split/aggregate steps
  * Connectors continue to describe the type of elements of the collection, not the collection itself
* https://github.com/syndesisio/syndesis/issues/4337 Data shape definition for collections
  * Add flag to data shape definitions that mark data shape to be a collection
  * Data mapper is able to provide topmost collection support for these kind of data shapes
* https://github.com/syndesisio/syndesis/issues/4349 Implement UX
  * Introduce new split/aggregate steps in UI
  * New steps do not hold any properties (for now - might change in future with split expression, aggregation strategy)
  * New split and aggregate steps are added separately to an integration
  * When split is added integration shows an info that aggregation might be needed, too
  * Aggregate step is only available in case a split step is present
  * Aggregate can only be added after a split step
  * Backend automatically provides new split aggregate steps for old integrations that have been using implicit split functionality
  * UI will display a info that the integration has been automatically migrated/translated from implicit split to new split/aggregate
  * Users can add aggregate step as finishing step of an integration
  * Users can add aggregate step after an existing finishing step
* https://github.com/syndesisio/syndesis/issues/4355 Out message capturing adjustments
  * Data mappers inside a split/aggregate see all captured out messages as possible source
  * Data mappers placed after an aggregate step do not see the inside messages within split/aggregate but only the overall aggregate out message and all out messages before the split

asmigala:
Verified with CR3

PR #4602, opened 2019-02-15
Fix Json type source and target conversion for topmost collection support

christophd:
The problem was that Syndesis operates on List<String> bodies when dealing with topmost collections for Json schema based data shapes.

Atlasmap data mapper expects a Json array String representation of that topmost collection. So we need to convert the source documents marked as Json type documents that are passed to mapper in `OutMessageCaptureProcessor.CAPTURED_OUT_MESSAGES_MAP` or in the exchange itself.

Also we need to convert mapper output target document form Json array String representation back to List<String> where each element is a Json Object String.

I have added this conversion to the datamapper step handler. The handler is provided with the atlasmap mapping definition and reads all Json typed source and target document ids. Based on those and on the List<String> nature of the documents we do a conversion before and after mapper processing.

This is related to #4257 and fixes https://github.com/atlasmap/atlasmap/issues/759

### DIFF
--- a/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/DataMapperStepHandler.java
+++ b/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/DataMapperStepHandler.java
@@ -15,17 +15,40 @@
  */
 package io.syndesis.integration.runtime.handlers;
 
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.databind.JsonNode;
+import io.syndesis.common.model.integration.Step;
+import io.syndesis.common.model.integration.StepKind;
+import io.syndesis.common.util.Json;
 import io.syndesis.integration.runtime.IntegrationRouteBuilder;
 import io.syndesis.integration.runtime.IntegrationStepHandler;
 import io.syndesis.integration.runtime.capture.OutMessageCaptureProcessor;
-import io.syndesis.common.model.integration.Step;
-import io.syndesis.common.model.integration.StepKind;
+import org.apache.camel.Exchange;
+import org.apache.camel.Message;
+import org.apache.camel.Processor;
 import org.apache.camel.model.ProcessorDefinition;
 import org.apache.camel.util.ObjectHelper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+@SuppressWarnings("PMD.AvoidReassigningParameters")
 public class DataMapperStepHandler implements IntegrationStepHandler {
+
+    private static final Logger LOG = LoggerFactory.getLogger(DataMapperStepHandler.class);
+
+    private static final String ATLASMAP_MODEL_VERSION = "v2";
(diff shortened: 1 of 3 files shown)


app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/DataMapperStepHandler.java
```diff
@@ -15,17 +15,40 @@
...
-import io.syndesis.common.model.integration.StepKind;
+import org.apache.camel.Exchange;
+import org.apache.camel.Message;
+import org.apache.camel.Processor;
 import org.apache.camel.model.ProcessorDefinition;
 import org.apache.camel.util.ObjectHelper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+@SuppressWarnings("PMD.AvoidReassigningParameters")
```
zregvart: :slightly_frowning_face:

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/DataMapperStepHandler.java
```diff
@@ -35,8 +58,126 @@ public boolean canHandle(Step step) {
...
         ObjectHelper.notNull(route, "route");
 
-        return Optional.of(
-            route.toF("atlas:mapping-flow-%s-step-%s.json?encoding=UTF-8&sourceMapName=" + OutMessageCaptureProcessor.CAPTURED_OUT_MESSAGES_MAP, flowIndex, stepIndex)
-        );
+        List<Map<String, Object>> dataSources = getAtlasmapDataSources(step.getConfiguredProperties());
+
+        route = handleJsonTypeSourceProcessor(route, dataSources).orElse(route);
+        route = route.toF("atlas:mapping-flow-%s-step-%s.json?encoding=UTF-8&sourceMapName=" + OutMessageCaptureProcessor.CAPTURED_OUT_MESSAGES_MAP, flowIndex, stepIndex);
+        route = handleJsonTypeTargetProcessor(route, dataSources).orElse(route);
```
zregvart: I wouldn't use `Optional` to return from private methods makes the logic hard to read IMHO.

christophd: Totally agree, thanks! Pushed improvements to this PR

christophd:
Fixes #https://github.com/syndesisio/syndesis/issues/4607

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/DataMapperStepHandler.java
```diff
@@ -74,17 +73,15 @@ public boolean canHandle(Step step) {
      * @param dataSources
      * @return
      */
-    private Optional<ProcessorDefinition<?>> handleJsonTypeSourceProcessor(ProcessorDefinition<?> route, List<Map<String, Object>> dataSources) {
+    private void addJsonTypeSourceProcessor(ProcessorDefinition<?> route, List<Map<String, Object>> dataSources) {
         List<String> jsonTypeSourceIds = dataSources.stream()
                                                     .filter(s -> ATLASMAP_JSON_DATA_SOURCE.equals(s.get("jsonType")) && "SOURCE".equals(s.get("dataSourceType")))
                                                     .map(s -> Optional.ofNullable(s.get("id")).map(Object::toString).orElse(""))
```
lburgazzoli: wondering if this could be a filter, or an empty id has a meaning ?

christophd: good point! added the filter

christophd:
@zregvart @lburgazzoli I fixed the code style issues and incorporated requested changes. Thanks for pointing me to these!

Issue #4590, opened 2019-02-15
Unable to build master with clean maven repo and flash profile


avano:
## This is a...

<pre><code>
[ ] Feature request
[X] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
`./tools/bin/syndesis build --backend --flash`

```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Connector .......................................... SUCCESS [  5.160 s]
[INFO] Connector :: Support :: Util ....................... SUCCESS [  1.908 s]
[INFO] Connector :: Support :: Verifier ................... SUCCESS [  1.437 s]
[INFO] Connector :: Support :: Maven Plugin ............... SUCCESS [ 11.115 s]
[INFO] Connector :: Twitter ............................... SUCCESS [  1.271 s]
[INFO] Connector :: FTP ................................... SUCCESS [  0.362 s]
[INFO] Connector :: SFTP .................................. SUCCESS [  0.319 s]
[INFO] Connector :: Dropbox ............................... SUCCESS [  4.144 s]
[INFO] Connector :: Support :: Test ....................... SUCCESS [  0.481 s]
[INFO] Connector :: AWS S3 ................................ SUCCESS [01:12 min]
[INFO] Connector :: ActiveMQ .............................. SUCCESS [ 13.917 s]
[INFO] Connector :: AMQP .................................. SUCCESS [ 16.412 s]
[INFO] Connector :: SQL ................................... SUCCESS [  6.795 s]
[INFO] Connector :: MQTT .................................. SUCCESS [  2.786 s]
[INFO] Connector :: Support :: Processor .................. SUCCESS [  0.827 s]
[INFO] Connector :: HTTP .................................. SUCCESS [  3.435 s]
[INFO] Connector :: Salesforce ............................ SUCCESS [ 11.603 s]
[INFO] Connector :: Slack ................................. SUCCESS [  1.683 s]
[INFO] Connector :: Gmail ................................. SUCCESS [  8.550 s]
[INFO] Connector :: Timer ................................. SUCCESS [  3.892 s]
[INFO] Connector :: ServiceNow ............................ SUCCESS [  9.216 s]
[INFO] Connector :: Concur ................................ SUCCESS [  0.361 s]
[INFO] Connector :: Kafka ................................. SUCCESS [  3.315 s]
[...]

PR #4621, opened 2019-02-18
fix(#4590): Fix dependency on integration-runtime-tests in odata

phantomjinx:
* Make all test dependencies in odata active if basepom.check.skip-all
  is not defined or false, ie. if flash profile active.

### DIFF
--- a/app/connector/odata/pom.xml
+++ b/app/connector/odata/pom.xml
@@ -44,12 +44,6 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
-        <groupId>io.syndesis.integration</groupId>
-        <artifactId>integration-runtime</artifactId>
-        <version>${project.version}</version>
-        <classifier>tests</classifier>
-      </dependency>
-      <dependency>
         <groupId>org.apache.httpcomponents</groupId>
         <artifactId>httpclient</artifactId>
         <version>4.5.3</version>
@@ -230,12 +224,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>io.syndesis.integration</groupId>
-      <artifactId>integration-runtime</artifactId>
-      <classifier>tests</classifier>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>com.github.spotbugs</groupId>
       <artifactId>spotbugs-annotations</artifactId>
       <scope>test</scope>
@@ -377,7 +365,7 @@
           </execution>
         </executions>
       </plugin>
-       <plugin>
+      <plugin>
         <artifactId>maven-dependency-plugin</artifactId>
         <executions>
           <execution>
--- a/app/connector/odata/src/test/java/io/syndesis/connector/odata/AbstractODataTest.java
+++ b/app/connector/odata/src/test/java/io/syndesis/connector/odata/AbstractODataTest.java
@@ -17,8 +17,10 @@ package io.syndesis.connector.odata;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.util.List;
 import java.util.Properties;
 import org.apache.camel.CamelContext;
@@ -44,11 +46,14 @@ import io.syndesis.common.model.integration.Flow;
 import io.syndesis.common.model.integration.Integration;
 import io.syndesis.common.model.integration.Step;
 import io.syndesis.common.model.integration.StepKind;
+import io.syndesis.common.util.Resources;
 import io.syndesis.connector.odata.server.ODataTestServer;
 import io.syndesis.connector.odata.server.ODataTestServer.Options;
-import io.syndesis.integration.runtime.IntegrationTestSupport;
+import io.syndesis.integration.runtime.IntegrationRouteBuilder;
+import io.syndesis.integration.runtime.IntegrationStepHandler;
+import io.syndesis.integration.runtime.logging.ActivityTracker;
 
-public abstract class AbstractODataTest extends IntegrationTestSupport implements ODataConstants {
+public abstract class AbstractODataTest implements ODataConstants {
 
     protected static final int MOCK_TIMEOUT_MILLISECONDS = 60000;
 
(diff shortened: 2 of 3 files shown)


phantomjinx:
@zregvart 
When
- basepom.check.skip-all is true then the flash profile has been invoked and no tests are run at all. Therefore, none of the test dependencies are required.
- basepon.check.skip-all is false or not defined then the flash profile has NOT been invoked and all tests are run. Therefore all test dependencies are required, including an expectation that integration-runtime.tests has been built and is available.

zregvart:
Other connectors also depend on `integration-runtime` in `test` scope. I think we need to address what the root cause of the issue is rather than fixing the symptoms.

phantomjinx:
@zregvart 

> Other connectors also depend on `integration-runtime` in `test` scope. I think we need to address what the root cause of the issue is rather than fixing the symptoms.

You're right other connectors depend on integration-runtime in the test scope but the odata connector is the only one that depends on integration-runtime-tests in the test scope, since the unit tests inherit from [IntegrationTestSupport](https://github.com/syndesisio/syndesis/blob/master/app/integration/runtime/src/test/java/io/syndesis/integration/runtime/IntegrationTestSupport.java).

zregvart:
> You're right other connectors depend on integration-runtime in the test scope but the odata connector is the only one that depends on integration-runtime-tests in the test scope, since the unit tests inherit from [IntegrationTestSupport](https://github.com/syndesisio/syndesis/blob/master/app/integration/runtime/src/test/java/io/syndesis/integration/runtime/IntegrationTestSupport.java).

This is the root issue, that class is in `src/test/java` so it is not meant to be reused outside of the `integration-runtime` test suite. It either needs to be refactored and moved to `src/main/java` or the tests in the OData connector need to be refactored not to use it.

phantomjinx:
@zregvart 
How about https://maven.apache.org/guides/mini/guide-attached-tests.html ??

phantomjinx:
Of course, the other thing to consider is that a flash build no longer downloads any of the dependencies required for the tests and as such this is far more streamlined and efficient. Would the use of this profiling based on that property not be useful across all projects?

zregvart:
> @zregvart
> How about [maven.apache.org/guides/mini/guide-attached-tests.html](https://maven.apache.org/guides/mini/guide-attached-tests.html) ??

This sounds like an easy fix to to try.

zregvart:
> Of course, the other thing to consider is that a flash build no longer downloads any of the dependencies required for the tests and as such this is far more streamlined and efficient. Would the use of this profiling based on that property not be useful across all projects?

To me profiles scream workarounds, and they should not be. In general profiles should be used as little as possible (IMHO).

Can you expand on "flash build no longer downloads any of the dependencies required for the tests"?

phantomjinx:
@zregvart 
Sorry missed your reply and will directly answer now. The class IntegrationTestSupport is purely for tests and provides some really useful methods that other modules can take advantage of. Therefore, to consider:
- IntegrationTestSupport is not a runtime class but only for tests;
- Maven provides means to share test code for an occasion such as this;
- It would seem pointless to have to duplicate the class' code just to avoid a test dependency.

phantomjinx:
> To me profiles scream workarounds, and they should not be. In general profiles should be used as little as possible (IMHO).

Agree with you on the whole and careful use of profiles is preferable.

> 
> Can you expand on "flash build no longer downloads any of the dependencies required for the tests"?

If flash is switched on then all odata-connector dependencies are ignored / not downloaded. Since the tests are not required and never run then likewise these test dependencies are never required so never fetched into the m2 repository (of course some may be there from other modules but you see my point).

zregvart:
> If flash is switched on then all odata-connector dependencies are ignored / not downloaded. Since the tests are not required and never run then likewise these test dependencies are never required so never fetched into the m2 repository (of course some may be there from other modules but you see my point).

I completely and utterly do not understand this. Are you saying that dependencies are not resolved by Maven when `flash` profile is activated?

phantomjinx:
I completely and utterly do not understand this. Are you saying that dependencies are not resolved by Maven when flash profile is activated?

Yes. Those placed inside the profile in this pom are only resolved if the property basepom.check.skip-all is false or not defined. Therefore, the result is all test dependencies are ignored in a flash profiled build.

phantomjinx:
> > @zregvart
> > How about [maven.apache.org/guides/mini/guide-attached-tests.html](https://maven.apache.org/guides/mini/guide-attached-tests.html) ??
> 
> This sounds like an easy fix to to try.

So looking into this a bit further ...

A test-jar is only created if the test archive is not empty. When a flash profile build is enacted, all tests are skipped and all test compilation is skipped so no test archive. Therefore, test-jar creation is skipped as well. This means that unless the specific class (and any fellows) are moved to main, a profile is necessary is odata-connector to avoid trying to resolve the dependency, ie.

1. A flash build is enacted;
2. All test compilation and running is skipped so no test classes or test jars created at all;
3. OData connector pom (atm) has a dependency to resolve on the integration-runtime-tests jar, regardless of which profile is enacted;
4. OData connector errors because it tries to resolve integration-runtime-tests dependency but that has not been compiled so not available.

Therefore, 3 choices I can see at the moment:
- Use a profile as I have done in this PR to avoid resolving the problematic dependency during a flash build;
- Add the useful test classes to the main jar of integration-runtime where they will always be compiled and available;
- Move the useful test classes to a common test module that always gets compiled but does not pollute the integration-runtime jar;

Thoughts??

@zregvart @oscerd @lburgazzoli @anyonelse

zregvart:
> Thoughts??

Pretty much what was already said in https://github.com/syndesisio/syndesis/pull/4621#issuecomment-464772612

Except for profiles, they should not be used.

phantomjinx:
> > Thoughts??
> 
> Pretty much what was already said in [#4621 (comment)](https://github.com/syndesisio/syndesis/pull/4621#issuecomment-464772612)
> 
> Except for profiles, they should not be used.

Thanks @zregvart 

With respect I would appreciate some others to chip in on this as I have tried to explain and further discuss the problem and merits of the proposed solution but seem to hit the metaphorical brick wall of 'profiles should not be used.' when it addresses the rather unique problem, ie. a profile is causing a dependency error and a further profile fixes it. Therefore, if someone else would like to provide guidance on how to fix this or provide a superseding PR then that would be appreciated.

gashcrumb:
Seeing as it's master that's broken I think it's acceptable to go for a quick fix to unbreak master and have a longer term plan, is that feasible for this PR?

zregvart:
I wonder who will then provide a proper fix and when. I'm fine with someone else taking on the responsibility to take this on.

phantomjinx:
Just pushed an updated PR that severs the IntegrationTestSupport connection by duplicating code into the odata-connector module. Hopefully, that should satisfy everybody.

Can take this one or [4625](https://github.com/syndesisio/syndesis/pull/4625).

phantomjinx:
Job done.

PR #4627, opened 2019-02-19
chore: upgrade immutables

### LINKED ISSUE #4624: Can no longer compile syndesis in Eclipse workspace
Something appears to have changed recently with respect to common-util. Up until today, I was able to get this module to compile in Eclipse by adding the following paths as source folders:
- src/main/java
- src/test/java
- target/generated-test-sources/test-annotations

The final folder provides the source classes generated from the Immutable annotations. Thus, it is only available after an mvn compile of the module. Running mvn through eclipse and refreshing the project achieves this. However, on refreshing the project:
- the class ImmutableTestValue cannot compile since it implements SkipNullsTest.TestValue;
- the class SkipNullsTest cannot compile since it has a method relying on ImmutableTestValue.builder();

Thus, a circular dependency problem. It would appear that running mvn on the command line avoids this problem and everything works as expected but Eclipse cannot handle this.

- How can Eclipse be configured to properly handle importing/building syndesis?
- If Eclipse is no longer able to handle building syndesis, what is a preferred IDE to be used instead?

zregvart:
Upgrades immutables to a version that resolves the Eclipse issue.

Ref https://github.com/immutables/immutables/issues/891

Fixes #4624

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -118,7 +118,7 @@
     <resteasy.version>3.6.1.Final</resteasy.version>
     <resteasy-spring-boot-starter.version>1.0.1.Final</resteasy-spring-boot-starter.version>
 
-    <immutables.version>2.7.4</immutables.version>
+    <immutables.version>2.7.5</immutables.version>
 
     <jdbi.version>2.78</jdbi.version>
     <postgresql.version>9.1-901-1.jdbc4</postgresql.version>


phantomjinx:
@zregvart 

Confirmed this is required to fix Immutables in Eclipse but also need [m2e-apt](https://immutables.github.io/apt.html) installed as well. Only with both has my workspace breathed a sigh of relief.

PR #4643, opened 2019-02-20
fix(server): robustly parse OpenAPI documents

### LINKED ISSUE #4542: Unable to use OpenAPI document from URL to create API provider/custom API connector
## This is a...

<pre><code>
[ ] Feature request
[ x ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
</code></pre>

## Description

I tried to create a new CustomAPI connector and API provider using the petstore swagger.json from this URL: http://petstore.swagger.io/v2/swagger.json

There is this message displayed:
```Unable to read OpenAPI document from: http://petstore.swagger.io/v2/swagger.json ```

The server pod error message:
```
2019-02-13 07:55:46.548 ERROR [-,76d7f5dca706a24c,76d7f5dca706a24c,false] 1 --- [  XNIO-3 task-6] i.s.s.a.g.swagger.SwaggerAPIGenerator    : An error occurred while trying to validate a API
--
```

The local copy of petstore swagger.json works as expected.  
Server log is attached: 
[serverlog.txt](https://github.com/syndesisio/syndesis/files/2859367/serverlog.txt)

## Screenshot
unable_to_read_openapi_document

### PR DESCRIPTION
**Pull request by zregvart**
Adds more robustness to the OpenAPI parsing.

Fixes #4542

### DIFF
--- a/app/ui/src/app/customizations/api-connector/api-connector.reducer.ts
+++ b/app/ui/src/app/customizations/api-connector/api-connector.reducer.ts
@@ -4,7 +4,8 @@ import { PlatformState } from '@syndesis/ui/platform';
 import {
   ApiConnectorState,
   ApiConnectorWizardStep,
-  CustomApiConnectorRequest
+  CustomApiConnectorRequest,
+  ApiConnectorData
 } from '@syndesis/ui/customizations/api-connector/api-connector.models';
 import {
   ApiConnectorActions,
@@ -214,10 +215,8 @@ export function apiConnectorReducer(
           properties: action.payload.properties,
           warnings: action.payload.warnings,
           errors: action.payload.errors,
-          name: action.payload.name ? action.payload.name
-                                    : action.configuredProperties.specification.info.title,
-          description: action.payload.description ? action.payload.description
-                                                  : action.configuredProperties.specification.info.description
+          name: action.payload.name,
+          description: action.payload.description
         },
         loading: false,
         hasErrors: false,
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by zregvart** [comment]
I'm kinda OK with the backend stuff, can someone take a look at the changes in the UI code? (thanks :speak_no_evil:)

File: app/ui/src/app/customizations/api-connector/api-connector.reducer.ts [hunk 1]
```diff
@@ -215,9 +216,17 @@ export function apiConnectorReducer(
           warnings: action.payload.warnings,
           errors: action.payload.errors,
           name: action.payload.name ? action.payload.name
-                                    : action.configuredProperties.specification.info.title,
+                                    : action.configuredProperties
+                                      && action.configuredProperties.specification
+                                      && action.configuredProperties.specification.info
+                                      && action.configuredProperties.specification.info.title
```
riccardo-forina: (nothing to suggest here, just maybe that we _really_ need a better model definition?)

zregvart: Actually that bit can be refactored, not sure why we dig into the specification there, there are properties set on the backend, let me see if that works...

zregvart: Yeah, backend guarantees to return those properties even if they're not specified, in that case the value is `"unspecified"`.

zregvart:
And I need to note, there's a bit of functionality change here, if the supplied OpenAPI document fails validation with errors (not warnings) - for example (wrong) format issues, schema issues (schema verifies required properties, for instance), failing to fetch the schema from URL - the user will placed on the first (upload/provide URL) screen in the flow. I don't think it makes sense to go to a screen with edit/review if the supplied OpenAPI document is not valid.

cc @dongniwang @gaughan

gashcrumb:
Yup, will review today

dongniwang:
@zregvart 

> And I need to note, there's a bit of functionality change here, if the supplied OpenAPI document fails validation with errors (not warnings) - for example (wrong) format issues, schema issues (schema verifies required properties, for instance), failing to fetch the schema from URL - the user will placed on the first (upload/provide URL) screen in the flow. I don't think it makes sense to go to a screen with edit/review if the supplied OpenAPI document is not valid.

Sounds good to me!

PR #4644, opened 2019-02-20
Add kudu install via openshift template

lgarciaaco:
QA is in need to install kudu within syndesis project in order to run automated tests. After discussion with @rhuss, we agree to deliver a static openshift template with which QA can work in the meantime (maybe in the future we implement a way to optional add components from the syndesis installation)


lgarciaaco:
@zregvart sure!

zregvart:
I think we can leave this as community bit, and no need to run it through productization, i.e. we don't need to support it in the product.

Does that work for you @rhuss @chirino @heiko-braun?

PR #4648, opened 2019-02-21
doc: fix list indent

tadayosi:


zregvart:
@TovaCohen do you mind having a look at this change?

gashcrumb:
@TovaCohen @tadayosi marked this WIP for now so you guys can resolve any issues...

doc/developing_extensions/topics/developing_extensions.adoc
```diff
@@ -908,9 +908,9 @@ public class AWSS3DeleteObjectCustomizer implements ComponentProxyCustomizer {
 +
 Defines the class implementing https://github.com/syndesisio/syndesis/blob/master/app/integration/component-proxy/src/main/java/io/syndesis/integration/component/proxy/ComponentProxyFactory.java[ComponentProxyFactory] used to create and/or configure the underling component/endpoint.  It can be set on _action_ or global level with the _action_ one having precedence.
 +
-To customize the behavior of the https://github.com/syndesisio/syndesis/blob/master/app/integration/component-proxy/src/main/java/io/syndesis/integration/component/proxy/ComponentProxyComponent.java[ComponentProxyComponent] you can override on of the following methods:
+To customize the behavior of the https://github.com/syndesisio/syndesis/blob/master/app/integration/component-proxy/src/main/java/io/syndesis/integration/component/proxy/ComponentProxyComponent.java[ComponentProxyComponent] you can override the following methods:
```
TovaCohen: Is this paragraph supposed to be a sub paragraph of the paragraph that describes connectorFactory? I'm wondering whether this paragraph should not be indented at all. It should have its own heading, something like, "Customizing behavior of ComponentProxyComponent" and then you do not need to change the levels of the bullets. But I may be wrong.  If I'm right, then you just need to remove the plus sign that is on line 910. And then the single-asterisk indents will be correct.

tadayosi:
@TovaCohen I still believe my fix is correct for two reasons:

1. `ComponentProxyFactory` has only one method which returns `ComponentProxyComponent`, so this paragraph should be considered to belong to the bullet item `connectorFactory`. You can see the example in the bullet item `configureDelegateEndpoint` still talks about `ComponentProxyFactory`

2. After the last bullet item `configureDelegateEndpoint` of the sub paragraph, you read this:
    
    > **The items above** can be set on action level or at global level and behave as follow:
    
    If you put the sub paragraph out of the the bullet item `connectorFactory`, the meaning of **the items above** changes and become inconsistent with what follows.

TovaCohen:
Thanks Tadayoshi! Let's get this merged.
Looking at this made me realize that we duplicate this content and I need to address that. 
I'll pull your update into my fork and make sure that the doc in the integration guide is correct.

Issue #4629, opened 2019-02-19
Unable to use basic filter with collection (was: Inconsistent behavior with SQL connector and message split)


avano:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

I have following integration:

```
1. SQL - Periodic invocation - Select * from contact
2. AMQ - Publish messages to some queue
```

In this scenario, the list from SQL connector is automatically split into multiple messages for some reason even when I didn't specify the `split` step. So when I have 2 contacts in the database, I receive 2 messages in the queue.

Now I want to add a basic filter to the integration - since the split magic is being done, I might not need the explicit "split" step, so I create:

```
1. SQL - Periodic invocation - Select * from contact
2. Basic filter - last_name contains something (now I know something is wrong because I don't have the autocomplete for the field)
3. AMQ - Publish messages to some queue
```  

Obviously, the integration fails with:

```
org.apache.camel.spring.boot.CamelSpringBootInitializationException: java.lang.IllegalArgumentException: java.lang.IllegalStateException: No step properties defined for rule filter step
	at org.apache.camel.spring.boot.RoutesCollector.onApplicationEvent(RoutesCollector.java:250) ~[camel-spring-boot-2.21.0.fuse-730042.jar!/:2.21.0.fuse-730042]
	at org.apache.camel.spring.boot.RoutesCollector.onApplicationEvent(RoutesCollector.java:57) ~[camel-spring-boot-2.21.0.fuse-730042.jar!/:2.21.0.fuse-730042]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener(SimpleApplicationEventMulticaster.java:172) ~[spring-context-4.3.19.RELEASE.jar!/:4.3.19.RELEASE]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:165) ~[spring-context-4.3.19.RELEASE.jar!/:4.3.19.RELEASE]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:139) ~[spring-context-4.3.19.RELEASE.jar!/:4.3.19.RELEASE]
[...]

PR #4649, opened 2019-02-21
#4629 Fix collection support on filter step

christophd:
Basic filter step was not able to handle collection shapes for autocompletion of available fields. Also filter was not able to handle collection typed bodies at runtime.

I made sure that filter is able to work with collection typed bodies and that basic filter autocompletion is working for collection typed shapes. The autocompletion fields are described with `[]` prefix to indicate that the field should use an index identifying the item in the collection. Also I introduced `size()` field as a first collection field so user is able to filter messages based on collection size.

bildschirmfoto 2019-02-21 um 10 52 15


app/integration/runtime/src/main/java/io/syndesis/integration/runtime/util/JsonSimplePredicate.java
```diff
@@ -79,7 +80,20 @@ public JsonSimplePredicate(final String expression, final CamelContext context)
...
     public boolean matches(final Exchange exchange) {
-        final Message payload = exchange.getIn();
+        // Clone the exchange and set the JSON message converted to a Map /
+        // List as in message.
+        // The intention is that only this predicate acts on the converted
+        // value, but the original in-message still continues to carry the
+        // same format.
+        // The predicated is supposed to be read only with respect to the
+        // incoming message.
+        final Exchange exchangeForProcessing = ExchangeHelper.createCopy(exchange, true);
```
zregvart: I'm not certain of this: if we're running with stream caching (I think we are), this will work, otherwise it might exhaust the any streams present in the message(s) and fail on next exchange processing. Perhaps a test could prove this?

:duck: seems that we were reading the stream below, so I guess it's OK...

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/util/JsonSimplePredicate.java
```diff
@@ -165,4 +189,10 @@ static String toOgnl(final Matcher matcher) {
         return ognl.toString();
     }
 
+    private static boolean isCollectionPath(String expression) {
+        return COLLECTION_PATHS.stream()
```
zregvart: Perhaps leave `COLLECTION_PATHS` as `String` until we need more paths? It's a constant with only one element currently...

gashcrumb:
So is this ready to merge @christophd or are you going to implement @zregvart's suggestion?

christophd:
I am about to push some more changes according to the mentioned review

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/util/JsonSimplePredicate.java
christophd: @zregvart good point! I added a cache of the input stream and added some tests. Can you please have another look? I pushed the changes to this PR

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/util/JsonSimplePredicate.java
christophd: I kinda like the idea to have everything in place to only add more collection paths in future. But this does not have to be a constant though. Pushed changes to this PR

PR #4665, opened 2019-02-22
refactor(connector): update timer's `defaultValue`

### LINKED ISSUE #4491: Stranged default when using a Cron Timer
grafik

The text field is initialized with '+' as separators. Shouldn't this be " " ?

BTW, from an UX point of view it would be far better to allow 5 input fields with labels like "minute" "hour" .... for every column, I never can memorize the order of these columns.

zregvart:
This should give us a more usable value for the cron expression in the timer.

Fixes #4491

### DIFF
--- a/app/common/util/src/main/java/io/syndesis/common/util/Json.java
+++ b/app/common/util/src/main/java/io/syndesis/common/util/Json.java
@@ -15,8 +15,10 @@
  */
 package io.syndesis.common.util;
 
-import java.util.Map;
+import java.io.IOException;
+import java.io.InputStream;
 import java.util.HashMap;
+import java.util.Map;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonParser;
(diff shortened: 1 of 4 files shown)


zregvart:
@oscerd hmm, perhaps a Customizer then that URL encodes the property, let's see...

oscerd:
Yeah, I was talking about this
https://github.com/apache/camel/blob/master/components/camel-quartz2/src/main/docs/quartz2-component.adoc

oscerd:
Maybe we can manipulate the value as you said

zregvart:
Let me do some tests/digging into this, perhaps we can make it nicer and have it working also. Strange idea if you ask me, "nice and working".

oscerd:
image

zregvart:
@oscerd turns out no customizer is needed (I've added a test for both variants).

The `DefaultComponent::createEndpoint` for non-raw URIs will use `URISupport::parseParameters` which does urldecoding so both cases are supported out of the box.

Issue #4612, opened 2019-02-18
Invalid character


mkralik3:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
Everywhere when is a button for editing is an invalid character.

E.g.

Edit name of integration (  )
![image](https://user-images.githubusercontent.com/16251792/52941860-0f6f1980-336a-11e9-8f48-30c34dcaaeb3.png)

Edit name and description of integration.
![image](https://user-images.githubusercontent.com/16251792/52941938-2e6dab80-336a-11e9-971c-cd804be7e115.png)

Edit name and description of Connection
![image](https://user-images.githubusercontent.com/16251792/52941988-49d8b680-336a-11e9-84cc-3ce77c346ed0.png)

![image](https://user-images.githubusercontent.com/16251792/52942090-93c19c80-336a-11e9-98a6-25ba4ab22463.png)

![image](https://user-images.githubusercontent.com/16251792/52942102-9ae8aa80-336a-11e9-8f63-c67b727941bb.png)

![image](https://user-images.githubusercontent.com/16251792/52942242-ea2edb00-336a-11e9-970c-3103c15388fb.png)

heiko-braun:
@mkralik3  `p0` means blocking. Is this what you had in mind?

mkralik3:
@heiko-braun  Sorry I didn't noticed , missclick, I meant  P1

heiko-braun:
@mkralik3 thanks for the clarification

dlabaj:
I'll take a look at this @seanforyou23

seanforyou23:
@dlabaj just a heads up, the latest push to https://github.com/patternfly/patternfly-next/pull/1467 allows us to resolve this by simply not including the styles that cause this problem. It's all coming full circle right about now :) I will lump this fix in with https://github.com/syndesisio/syndesis/pull/4669

Issue #4608, opened 2019-02-18
consolidate font sizes in syndesis 


mmelko:
Introducing new font brought some inconsistency in font sizes and visibility of connection names:
![screenshot_20190218_094516](https://user-images.githubusercontent.com/6814482/52938782-d5e6e000-3362-11e9-94d6-e3188d0e2985.png)
![screenshot_20190218_095532](https://user-images.githubusercontent.com/6814482/52939010-5dccea00-3363-11e9-8e44-38f962d3c286.png)

gashcrumb:
So I've seen quite a lot impacted by introducing patternfly 4 with the masthead PR:

1)  font-awesome icons are broken
2)  Fonts in general are all over the place
3)  Extra padding here and there where there wasn't padding
4)  Misalignment of elements here and there where this wasn't happening

I'm starting to wonder if we shouldn't consider backing out that change, or are we going to wind up with a lot of extra custom CSS to suppress the above?

seanforyou23:
Don't give up just yet! 😀

Had some delays getting going on these follow up issues, but I think we can get them all resolved and in a way that's properly supported on the PF side and without it taking forever. I know they went through this exact exercise recently over in OSIO so I plan to team up with the wiz that made that happen tomorrow. I'll see if we can expose some overrides or something like that as an include from the PF dist. That way the overrides are out of our hair. We may have to denote certain html containers with some extra classes, but that's a much better solution I think. 

Either way we'll shoot for as little touch points as possible. Keep you posted!

seanforyou23:
Just an update - we're making good progress on this and looks like the general approach has the thumbs up from the PF team. I'm ironing out a final detail on that side and thinking it will be merged in the next day. Then, we'll just need to update the corresponding PR in syndesis and give it a final test run. Thanks for being patient!

seanforyou23:
@mmelko this should be fixed now, let me know if you see any issues!

mmelko:
It looks good now. Thanks!

PR #4669, opened 2019-02-22
fix(fonts): consolidate font styles/sizes

seanforyou23:
This PR makes an adjustment to what all we're importing from patternfly4-core. Previously, patternfly-base.css came with extra baggage that we didn't want (namely fonts). With these changes, the entire application again uses only PF3 fonts, and many of the various spacing issues should also be resolved.

To keep things clean on our end, a change in patternfly core was required - https://github.com/patternfly/patternfly-next/pull/1467 - we need this merged in pf-core before we can merge this PR. To see it all in action, pair these two branches together.

Fixes https://github.com/syndesisio/syndesis/issues/4608
Fixes https://github.com/syndesisio/syndesis/issues/4612

screen shot 2019-02-22 at 8 28 22 am

screen shot 2019-02-22 at 8 24 08 am

screen shot 2019-02-22 at 8 28 34 am

import only the pf4 css modules we need
clean up font styles

### DIFF
--- a/app/ui/angular.json
+++ b/app/ui/angular.json
@@ -47,7 +47,7 @@
               "node_modules/patternfly/dist/css/patternfly.css",
               "node_modules/patternfly/dist/css/patternfly-additions.css",
               "node_modules/patternfly-ng/dist/css/patternfly-ng.min.css",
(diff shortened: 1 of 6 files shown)


dlabaj:
LGTM

Issue #4618, opened 2019-02-18
[collection support]: Data mapper should select aggregate input shape as target


christophd:
When a data mapper is placed in front of an aggregate step the mapper does not select the shape of the aggregate as target document. Instead the next connector step input is selected as target shape.

![Bildschirmfoto 2019-02-18 um 14.07.02.png](https://images.zenhubusercontent.com/5be31e5e6d41381168d0a8e7/74190bdd-3057-4f8d-9190-2da58320a155)

Steps to reproduce:
- add finish connection
- add aggregate step before that connection
- add data mapper before aggregate step

Data mapper should pick aggregate input shape as target.

gashcrumb:
So I created this integration and this seems to behave as I'd expect:

![image](https://user-images.githubusercontent.com/351660/53037667-c8a42100-3448-11e9-88c8-24d1ba634638.png)

I verified in the data mapper host component via the debugger that in this case the aggregate step is picked up and supplied to the data mapper.  

Why don't I add some behavior though to the aggregate step to indicate where it got it's data shape from?

gashcrumb:
Found a way to reproduce it here, AMQ connections are handy since you can specify json instance and make up your own data type.  And, as it turns out the issue only shows up when the data mapper is first added.  When you go and revisit the step afterwards it looks correct.

christophd:
Yes it is a very special procedure to reproduce the problem. For me it was adding the mapper 1st then the aggregate. We have many ways to write the integrations right!?

christophd:
Another thing that comes to my mind. What about the last connection having that warning sign that a data mapping is required. Can we get rid of this? The aggregate right before the connection provides that required shape. And I tested that it works without an additional mapping between aggregate and the finish connection. Is a lower prio issue I guess though.

mmuzikar:
While verifying this issue I've run into a problem when aggregate step doesn't select the subsequent data shape as the output, if you change it (by removing step or adding another one). So you end up in a state like this. It is fixed by removing the aggregate step and then adding it again.
![image](https://user-images.githubusercontent.com/46345469/54270424-af8e1c00-457f-11e9-8465-534b2e4624b8.png)

heiko-braun:
@mmuzikar I think it might help to create a separate issue for the problem you see in https://github.com/syndesisio/syndesis/issues/4618#issuecomment-472357130

mmuzikar:
Opened #4866 regarding the issue with updating datashape in aggregate. So this issue is verified.

PR #4673, opened 2019-02-25
#4618 Data mapper select target doc when step.id is undefined

christophd:
You can reproduce this when adding a data mapper at a position right before an aggregate step. The data mapper did not select the input of the aggregate but the input of the step subsequent to the aggregate.

This was because both data mapper step and aggregate step did have an `step.id === undefined` at that time. Following from that the logic in data mapper host component by mistake thought that the 1st subsequent step is the data mapper itself and used the step subsequent to that for populating the target doc.

So I have added another check on the `step.stepKind` to check wether the 1st step in line is the data mapper itself.

I do not know though why the aggregate step has a `step.id === undefined` maybe that is the root cause that should be fix in first place.


gashcrumb:
I think I've verified that steps don't appear to get an ID set properly when they're first added, just having trouble with the Syndesis install at the moment to ensure that I'm right.

I think we should fix the IDs instead of merging this PR, so closing this one and hope to get a related PR opened soon-ish.

Issue #4655, opened 2019-02-21
Help menu stays visible after selecting an item


mcada:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

Help menu stays open after selection of an item and we have to click on the icon again to close it. I have never seen such behavior for similar features - do we want it like this or should we make the menu disappear when we select an item?

## Expected behavior

When I open help menu and select an item, the menu should disappear.

## Screenshot

![out](https://user-images.githubusercontent.com/14313995/53172401-85cb8000-35e5-11e9-8785-f42c49968d64.gif)

amysueg:
Can we close this item since #4675 has been merged? @mcada @heiko-braun @dlabaj

heiko-braun:
Moving to `Done`

amysueg:
Closing.

amysueg:
Closing.

PR #4675, opened 2019-02-25
fixed(masthead dropdown): Added code to handle collapsing dropdown on…

dlabaj:
Fixes Issue #4655 .  Code was added to handle when a user clicks outside of the drop down in the masthead, the drop down will now close.


dlabaj:
@seanforyou23 Will add it there as well.  Thanks.

PR #4698, opened 2019-02-28
refactor(server): perform usage accounting of connections and extensions on integration update

### LINKED ISSUE #3931: Missing Connection/Extension Property Change Event When Integrations Are Created Or Deleted
## This is a...

<pre><code>
[ ] Feature request
[x ] Bug report  
</code></pre>

## Description

Connections and extensions have a `uses` integer property that identifies how many integrations are using it. When an integration is created that uses a connection or extension, a change event for that connection and/or extension should be fired. I don't believe that event is fired as the UI object cache/store does not get refreshed.

See #391 and #3930.

zregvart:
Adds a new `ResourceUpdateHandler` to update the connection usage when integration is updated and persists the change in the database. This way an event that the connection is updated is fired and broadcast to the UI.

Impementation contains a debounce of 3 seconds so that the backend database is not overwhelmed by updates. Some care has been taken not to interact with the database too much, though all integrations and all connections need to be fetched with the current interface.

Fixes #3931


zregvart:
@syndesisio/backend can anyone have a quick look at this, does change things a bit:

 - no longer doing a n+1 query on load of connection(s) and extension(s)
 - `uses` is computed asynchronously on integration update (with 3sec debounce)
 - extension use is defined as being part of an integration: used to be tied to integration deployment, not sure why, I think it should be considered used if it's part of an unpublished integration (could lead to issues if deleted when the integration _using_ it loaded)

zregvart:
Actually @gaughan, is there a requirement that only extensions that are used in published integrations are counted as used?

With this that changes and connections and extensions are counted when added to the integration, regardless of the integration being deployed or not.

zregvart:
@lburgazzoli @nicolaferraro any thoughts?

nicolaferraro:
Not sure about the workflow when using deployments. I think there's a deployment also when the integration is stopped after being deployed the first time, so moving it to integration will affect the behavior with draft integrations only, right? I don't remember there was a specific requirement back then.. and I feel Integration is the right place..

zregvart:
Thanks for the review @nicolaferraro I'll merge this in, it can always be rolled back/fixed if needed.

Issue #4718, opened 2019-03-01
Validate connector json files while building


phantomjinx:
The connector json files have a large number of repeating elements and a sensitivity to stray commas. As such it would be useful if these files were validated for being well-formed and against a json-schema.

PR #4719, opened 2019-03-01
fix(#4718): Validate connector json files

phantomjinx:
* support/shared
 * Creates a module that allows distribution of the single connector schema

* connector/*/pom.xml
 * Uses maven-remote-resources plugin to bring in the support-shared
   module that provides the connector schema
 * Uses the yaml-json-validator plugin check the connector plugin is
   well-formed then validates against the connector schema.

### DIFF
--- a/app/connector/support/maven-plugin/pom.xml
+++ b/app/connector/support/maven-plugin/pom.xml
@@ -76,6 +76,18 @@
     </dependency>
 
     <dependency>
+        <groupId>com.github.fge</groupId>
+        <artifactId>json-schema-core</artifactId>
+        <version>1.2.5</version>
+    </dependency>
+
+    <dependency>
+        <groupId>com.github.fge</groupId>
+        <artifactId>json-schema-validator</artifactId>
+        <version>2.2.6</version>
+    </dependency>
+
+    <dependency>
       <groupId>org.immutables</groupId>
       <artifactId>value</artifactId>
       <scope>provided</scope>
--- a/app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
+++ b/app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
@@ -16,7 +16,9 @@
 package io.syndesis.connector.support.maven;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.security.AccessController;
@@ -25,18 +27,6 @@ import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
-
-import com.fasterxml.jackson.core.PrettyPrinter;
-import io.atlasmap.core.DefaultAtlasConversionService;
-import io.atlasmap.java.inspect.ClassInspectionService;
-import io.atlasmap.java.v2.JavaClass;
-import io.atlasmap.v2.CollectionType;
-import io.syndesis.common.model.DataShape;
-import io.syndesis.common.model.DataShapeKinds;
-import io.syndesis.common.model.action.ConnectorAction;
-import io.syndesis.common.model.action.ConnectorDescriptor;
-import io.syndesis.common.model.connection.Connector;
-import io.syndesis.common.util.Json;
 import org.apache.maven.artifact.DependencyResolutionRequiredException;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
@@ -47,6 +37,25 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.shared.utils.StringUtils;
+import com.fasterxml.jackson.core.PrettyPrinter;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.github.fge.jsonschema.core.exceptions.ProcessingException;
+import com.github.fge.jsonschema.core.load.Dereferencing;
+import com.github.fge.jsonschema.core.load.configuration.LoadingConfiguration;
+import com.github.fge.jsonschema.core.report.ProcessingMessage;
+import com.github.fge.jsonschema.core.report.ProcessingReport;
+import com.github.fge.jsonschema.main.JsonSchema;
+import com.github.fge.jsonschema.main.JsonSchemaFactory;
+import io.atlasmap.core.DefaultAtlasConversionService;
+import io.atlasmap.java.inspect.ClassInspectionService;
+import io.atlasmap.java.v2.JavaClass;
+import io.atlasmap.v2.CollectionType;
+import io.syndesis.common.model.DataShape;
+import io.syndesis.common.model.DataShapeKinds;
+import io.syndesis.common.model.action.ConnectorAction;
+import io.syndesis.common.model.action.ConnectorDescriptor;
+import io.syndesis.common.model.connection.Connector;
+import io.syndesis.common.util.Json;
 
 @Mojo(
     name = "generate-connector-inspections",
@@ -56,6 +65,8 @@ import org.apache.maven.shared.utils.StringUtils;
 public class GenerateConnectorInspectionsMojo extends AbstractMojo {
     private static final String CONNECTORS_META_PATH = "META-INF/syndesis/connector/";
 
+    private static final String CONNECTOR_SCHEMA_FILE = "connector-schema.json";
+
     @Parameter(readonly = true, defaultValue = "${project}")
     private MavenProject project;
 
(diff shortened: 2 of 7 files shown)


app/connector/activemq/pom.xml
```diff
@@ -197,6 +197,24 @@
           </delimiters>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-remote-resources-plugin</artifactId>
```
zregvart: Perhaps instead of using this plugin we can simply point to the file in the repository for `jsonSchema` below?

phantomjinx: Refactored commit to extend maven-support-plugin.

phantomjinx:
@zregvart 
D'you mind having another look at this (reworked as per our conversation).
Thx.

app/connector/support/maven-plugin/pom.xml
```diff
@@ -29,6 +29,21 @@
   <name>Connector :: Support :: Maven Plugin</name>
   <packaging>maven-plugin</packaging>
 
+  <dependencyManagement>
```
zregvart: If no other POM is using these dependencies there's no need for declaring them in `<dependencyManagement>`, just put them as `<dependencies>` below. If other POMs will be using these declare them in the parent POM.

app/connector/support/maven-plugin/pom.xml
```diff
@@ -75,6 +90,16 @@
       <artifactId>jackson-databind</artifactId>
     </dependency>
 
+    <dependency>
+        <groupId>com.github.fge</groupId>
+        <artifactId>json-schema-core</artifactId>
+    </dependency>
```
zregvart: See above, just add version here. Might want to format on two spaces.
```suggestion
      <version>1.2.5</version>
    </dependency>
```

app/connector/support/maven-plugin/pom.xml
```diff
@@ -75,6 +90,16 @@
...
+
+    <dependency>
+        <groupId>com.github.fge</groupId>
+        <artifactId>json-schema-validator</artifactId>
+    </dependency>
```
zregvart: See above, just add version here. Might want to format on two spaces.
```suggestion
      <version>2.2.6</version>
    </dependency>
```

app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
```diff
@@ -102,6 +117,78 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+    /**
+     * Validate the given file against the connection schema
+     *
+     * @param jsonStream {@link InputStream} of the file to validate.
+     *                                          Method will close the stream after use.
+     * @return the {@link JsonNode} of the file content or throw exception if an error
+     * @throws MojoExecutionException
+     */
+    public JsonNode validateWithSchema(String fileName, InputStream jsonStream) throws MojoExecutionException {
+        initJsonSchema();
```
zregvart: Perhaps do the schema initialisation in the constructor?

app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
```diff
@@ -102,6 +117,78 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         }
     }
 
+    private void initJsonSchema() throws MojoExecutionException {
+        if (jsonSchema != null) {
+            return; // already initialised
+        }
+
+        InputStream schemaStream = null;
+        try {
```
zregvart: Use try-with-resources.

```suggestion
        try (InputStream schemaStream = GenerateConnectorInspectionsMojo.class.getResourceAsStream("/" + CONNECTOR_SCHEMA_FILE) {
```

app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
```diff
@@ -102,6 +117,78 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+            schemaStream = getClass().getResourceAsStream("/" + CONNECTOR_SCHEMA_FILE);
+            if (schemaStream == null) {
+                throw new IOException("Schema Initialisation Error: file not available");
```
zregvart: Will this ever happen? Won't the tests fail if the `connector-schema.json` is not present.

app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
```diff
@@ -80,7 +93,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
                 }
 
                 for (File file: files) {
-                    Connector connector = Json.reader().forType(Connector.class).readValue(file);
+                    JsonNode jsonNode = validateWithSchema(file.getName(), new FileInputStream(file));
```
zregvart: I would pass in `file` instead of `FileInputStream`, to make the responsibility of `validateWithSchema` to handle IO issues

app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
```diff
@@ -80,7 +93,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+
+                    Connector connector = Json.reader().forType(Connector.class).readValue(jsonNode);
```
zregvart: I think you can use `convertValue`:
```suggestion
                    Connector connector = Json.convertValue(jsonNode, Connector.class);
```

app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
```diff
@@ -102,6 +117,78 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+     *
+     * @param jsonStream {@link InputStream} of the file to validate.
+     *                                          Method will close the stream after use.
+     * @return the {@link JsonNode} of the file content or throw exception if an error
+     * @throws MojoExecutionException
+     */
+    public JsonNode validateWithSchema(String fileName, InputStream jsonStream) throws MojoExecutionException {
+        initJsonSchema();
+
+        try {
```
zregvart: If `validateWithSchema` receives a `File` instead of `InputStream` we can do try-with-resources here.

app/connector/support/maven-plugin/src/test/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojoTest.java
```diff
@@ -125,6 +130,49 @@ public void generateJavaInspectionWithVariantTest() throws Exception {
...
+    @Test
+    public void unformedFileAgainstSchema() throws Exception {
+        GenerateConnectorInspectionsMojo mojo = new GenerateConnectorInspectionsMojo();
+
+        InputStream fileStream = getClass().getResourceAsStream("/my-unformed-connector.json");
+        assertNotNull(fileStream);
+        try {
+            JsonNode jsonNode = mojo.validateWithSchema("testFile", fileStream);
+            assertNotNull(jsonNode);
+            fail("Should throw an exception at line 9");
```
zregvart: I find these kinds of tests are better by using AssertJ and exception assertions, see http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#exception-assertion

app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
phantomjinx: I did try that in both cases of accessing InputStreams. However, the code validations post-compile kept flagging them as possibilities for resource-leakage and failed the build. I looked into trying to ignore them / fix them but couldn't find anything obvious so reverted to a format that was guaranteed to work.

app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
phantomjinx: It did up until it the schema file was put in the correct location. Just being thorough out of habit, I guess.

app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
phantomjinx: See above comment about code validations post-compile.

zregvart: Let me checkout the branch and have a look, we use try-with-resources all over the place, should work...

zregvart: Yeah I see the same thing. The root cause seems to be https://github.com/spotbugs/spotbugs/issues/756, using Java 8 doesn't help either. Let's merge this and keep an eye on that issue and revisit when/if fixed...

Issue #4674, opened 2019-02-25
Move postgres exporter to syndesis-db pod


lgarciaaco:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

Having the postgres prometheus exporter as a sidecar is a good practice. As part of the move, change the `readinessProve` and `livenessProbe` to tcp. This will solve https://github.com/syndesisio/syndesis/issues/4652 and https://github.com/syndesisio/syndesis/issues/4410.

heiko-braun:
@lgarciaaco  I assume you plan to tackle it this sprint?

lgarciaaco:
I am working on it ...

PR #4725, opened 2019-03-03
Install db-metrics as sidecar

lgarciaaco:
Addressing https://github.com/syndesisio/syndesis/issues/4674


install/generator/04-syndesis-db-metrics.yml.mustache
```diff
@@ -1,122 +0,0 @@
...
-                description: "Disk space used by the database"
-- apiVersion: v1
-  kind: Service
-  metadata:
-    labels:
-      app: syndesis
-      syndesis.io/app: syndesis
-      syndesis.io/type: infrastructure
-      syndesis.io/component: syndesis-db-metrics
-    name: syndesis-db-metrics
```
jamesnetherton: We need to carry this service over into `syndesis-db.yml.mustache`. Else the Prometheus ServiceMonitor will have no endpoint to scrape.

Or we expose 9187 on the existing syndesis-db service. But, `install/addons/syndesis-db-servicemonitor.yml` would need modifying if we do that.

install/generator/04-syndesis-db.yml.mustache
```diff
@@ -227,4 +286,12 @@
...
-    - type: ConfigChange
+    - imageChangeParams:
+        automatic: true
+        containerNames:
+        - syndesis-db-metrics
+        from:
+          kind: ImageStreamTag
+          name: {{ Images.Support.PostgresExporter }}:{{ Tags.PostgresExporter }}
+          namespace: ${IMAGE_STREAM_NAMESPACE}
+      type: ImageChange
```
jamesnetherton: File needs a new line at the end. When I ran the generator it produced an invalid YAML template with `type: ImageChange- apiVersion: v1` at line 773.

install/syndesis.yml
```diff
@@ -802,7 +706,41 @@ objects:
             name: syndesis-sampledb-config
           - mountPath: /opt/app-root/src/postgresql-cfg/
             name: syndesis-db-conf
+        - capabilities: {}
+          env:
+          - name: DATA_SOURCE_NAME
+            value: postgresql://${POSTGRESQL_USER}:${POSTGRESQL_PASSWORD}@syndesis-db:5432/syndesis?sslmode=disable
```
jamesnetherton: Now that the exporter is running in the same pod as the DB, we can connect via `localhost`, right?

PR #4729, opened 2019-03-04
fix: Missing Json instance inspection for basic filter options

### LINKED ISSUE #4162: The basic filter doesn't show any hints for the webhook property
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
When I have (e.g.) the slack connection, the Basic filter gives me the hint, which values I can add to the property name.
e.g.
image

But when I have a webhook which is defined with e.g. JSON instance {"first_name":"John","company":"Red Hat"} like:
image
the Basic filter doesn't give me any hint.
image

When I add _first_name_ to the input element, the integration works.

For users, it can be confusing when for some connection it works and for some, it doesn't work. They could think that when the Basic filter doesn't show properties, something is wrong in the steps before.
I would expect that Basic filter shows me **first_name** and **company**.

christophd:
Fixes #4162

Also adds nested Json array support for options in both Json instance and schema inspector.

### DIFF
--- a/app/server/inspector/src/main/java/io/syndesis/server/inspector/JsonInstanceInspector.java
+++ b/app/server/inspector/src/main/java/io/syndesis/server/inspector/JsonInstanceInspector.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


app/server/inspector/src/main/java/io/syndesis/server/inspector/JsonInstanceInspector.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(JsonInstanceInspector.class);
+
+    private static final String ARRAY_CONTEXT = "[]";
+    static final List<String> COLLECTION_PATHS = Collections.singletonList("size()");
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public List<String> getPaths(String kind, String type, String specification, Optional<byte[]> exemplar) {
+        if (specification == null) {
```
zregvart: Should we test on `"json-instance".equals(kind)`?

zregvart: Perhaps also in the `JsonSchemaInspector`

christophd: This check on the kind is already done in the `supports` method which is called when evaluating matching inspectors

Issue #4762, opened 2019-03-05
FuseOnlineDatabaseInstanceDown alert is not activated when Syndesis-db is down


mkralik3:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

I have those alerts (https://github.com/syndesisio/syndesis/issues/4151 , https://github.com/syndesisio/syndesis/issues/4499):
![image](https://user-images.githubusercontent.com/16251792/53820191-8fe96900-3f6b-11e9-9060-89928be2f6cc.png)

I scale syndesis-db to 0 due to simulate database down state for alerting. After some time, when the server cannot connect to db, the latency of RestAPI is higher(10s)
![image](https://user-images.githubusercontent.com/16251792/53820296-c8894280-3f6b-11e9-9d9c-69f6c7f7f941.png)
Because of that, The FuseOnlineRestApiHighEndpointLatency and FuseOnlineRestApiHighEndpointErrorRate alerts are in the pending state
![image](https://user-images.githubusercontent.com/16251792/53820390-f66e8700-3f6b-11e9-8233-f5531c052dd5.png)
and after some times, they are fired
![image](https://user-images.githubusercontent.com/16251792/53820464-1aca6380-3f6c-11e9-9355-99ae92776bd1.png)

However, the [FuseOnlineDatabaseInstanceDown]( https://github.com/jamesnetherton/syndesis/blob/ec10dc4ec4ceb5bf90dd84e07fc7fd6c3ab9ddfe/doc/managing_environments/topics/alerting_sop.adoc#fuseonlinedatabaseinstancedown) alert is still inactive. In my POV, when syndesis-db pod is scaled down to 0, this alert should fired too.

jamesnetherton:
For 7.3 I think we should maybe remove the `FuseOnlineDatabaseInstanceDown` rule and revisit in 7.4.

I realised we have a general issue configuring rules that monitor for a job's availability using `up`, and in this case `pg_up`, for environments where Fuse Online could be installed in multiple namespaces (E.g integr8ly).

If a pod is scaled to 0, then its endpoint is removed from the service that the Prometheus ServiceMonitor is trying to scrape. The typical query to use for alerting in this case would be something like:

`absent(up{job="syndesis-db"})`. 

[...]

PR #4780, opened 2019-03-06
fix(monitoring): Fix FuseOnlineDatabaseInstanceDown alert rule

jamesnetherton:
After #4674 was implemented, the FuseOnlineDatabaseInstanceDown Prometheus alert rule no longer works. So this PR fixes that issue, but with some caveats that I documented in #4762.

### DIFF
--- a/doc/managing_environments/topics/alerting_sop.adoc
+++ b/doc/managing_environments/topics/alerting_sop.adoc
@@ -7,11 +7,11 @@
 
 *Alert name*
 
-FuseOnlineDatabaseHighRollbackRate
+FuseOnlineDatabaseInstanceDown
 
 *Description*
 
-The syndesis-db pod is not running or is not serving requests.
+The postgresql container within the syndesis-db pod is not running or is not serving requests.
 
 **Process steps**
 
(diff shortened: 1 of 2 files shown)


heiko-braun:
@lgarciaaco Mind a review?

doc/managing_environments/topics/alerting_sop.adoc
```diff
@@ -25,13 +25,38 @@ The syndesis-db pod is not running or is not serving requests.
...
+2) Switch to project namespace where Fuse Online is installed.
+
+3) If the syndesis-db pod is still running, capture logs for analysis.
+
+[source,bash,options="nowrap"]
+----
+oc logs syndesis-db-1-7nwh9 -c syndesis-db-metrics > syndesis-db-metrics.log
+----
+
+4) Start or restart the syndesis-db pod.
```
lgarciaaco: I think we need to be carefully here, but for now it should be ok, we can always refine this instructions as we gain some experience.

lgarciaaco:
@jamesnetherton since you aren't around, I am removing the WIP tag later today ...

PR #4804, opened 2019-03-08
refactor: make rest-swagger connector modern

### LINKED ISSUE #2436: Update API client connector to use the latest integration runtime

zregvart:
Fixes #2436


app/integration/component-proxy/src/main/java/io/syndesis/integration/component/proxy/ComponentCustomizer.java
```diff
@@ -44,23 +44,37 @@ default void consumeOption(Map<String, Object> options, String name, Consumer<Ob
...
-            }
+            try {
+                if (val instanceof String) {
+                    val = camelContext.resolvePropertyPlaceholders((String)val);
+                }
 
-            consumer.accept(
-                converter.mandatoryConvertTo(type, val)
-            );
+                if (type.isInstance(val)) {
```
lburgazzoli: this does not hurt but I think the type converter does that by default

zregvart: No need to invoke one if not needed, that was the idea...

lburgazzoli: yes but it is the same check the type converter does, anyway not an issue

app/integration/component-proxy/src/main/java/io/syndesis/integration/component/proxy/ComponentCustomizer.java
```diff
@@ -44,23 +44,37 @@ default void consumeOption(Map<String, Object> options, String name, Consumer<Ob
         }
     }
 
+    default void consumeStringOption(Map<String, Object> options, String name, Consumer<String> consumer) {
+        Object val = options.remove(name);
+        if (val != null) {
+            consumer.accept(String.valueOf(val));
```
lburgazzoli: So this would bypass any custom type converter registered for translating to a string, so I guess it has only to be used when you know that the type is a string right ? if so it it is worth documenting it

zregvart: It is called `consumeStringOption` and it accepts `Consumer<String>`, kinda apparent to me the the caller is asking for a `String`, no?

lburgazzoli: I mean that it does not involve the type converter so the result may be unexpected

zregvart: Aah, that bit, got it, yeah, delegating to `consumeOption` might be better...

zregvart:
OK, so the build is green, @valdar @lburgazzoli @KurtStam I know it's a big PR, if you have the time can you take a look?

PR #4823, opened 2019-03-11
fix: force bootstrap-select version to 1.13.6

zregvart:
Ref:
 - https://github.com/patternfly/patternfly/issues/1167
 - https://github.com/snapappointments/bootstrap-select/issues/2199

### DIFF
--- a/app/ui/package.json
+++ b/app/ui/package.json
@@ -137,7 +137,8 @@
   },
   "resolutions": {
     "ngx-bootstrap": "3.1.4",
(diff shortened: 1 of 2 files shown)


zregvart:
@paoloantinori @heiko-braun @tplevko there is a security issue with `bootstrap-select` in the version on `1.6.0`, this fixed for `1.7.x`. Should I do a backport, I have no idea if this will manifest in our use cases.

PR #4831, opened 2019-03-11
fix: disable `Next` when the given OpenAPI document has errors

### LINKED ISSUE #4437: StackOverflowError when trying to create custom API connector based on OpenHAB API
## This is a...

<pre><code>
[ ] Feature request
[ x ] Bug report  
</code></pre>

## Description

I was trying to create a custom API connector using the OpenHAB API specification, it however seems to be not possible. 

The swagger.json example can be found here:
http://demo.openhab.org:8080/rest/swagger.json

Attached the server log.
[server_log.txt](https://github.com/syndesisio/syndesis/files/2831265/server_log.txt)

zregvart:
Kinda; there's a weird issue with `<button disabled="disabled">` that isn't really disabled. Not sure what's going on here. Can someone from @syndesisio/ui-api take a look at this?

Peek 2019-03-11 18-04

Fixes #4437

### DIFF
--- a/app/ui/src/app/common/button.component.ts
+++ b/app/ui/src/app/common/button.component.ts
@@ -3,30 +3,53 @@ import { Component, Input } from '@angular/core';
 @Component({
   selector: 'syndesis-button',
   template: `
-    <button [attr.type]="type" class="submit syndesis-button syn-form__submit btn"
+    <button
+      [attr.type]="type"
+      class="submit syndesis-button syn-form__submit btn"
       [class.btn-primary]="theme == 'primary'"
       [class.btn-default]="theme == 'default'"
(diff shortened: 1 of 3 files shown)


gashcrumb:
Yeah, I think because technically `syndesis-button` isn't an actual form control the usual `disabled` handling doesn't actually kick on for that click handler.  And technically the click event isn't actually being handled by the button, it's being received by the `syndesis-button` node.

Quick fix would be to change the click handling to check for errors before firing the event in the click handler.

The longer more involved fix would be ensuring that the click function is properly delegated to the contained button that `syndesis-button` wraps.

gashcrumb:
@zregvart think the added commit should to the trick.

zregvart:
thanks @gashcrumb

PR #4842, opened 2019-03-12
Fixes #4789, fixed public integration state APIs to handle all integration target states

### LINKED ISSUE #4789: Stop integration via Public API doesn't work
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
I have a running integration. When I call
```
curl -k -L -H -v -X PUT "Content-Type: application/json" -H "SYNDESIS-XSRF-TOKEN: awesome" -H 'Authorization: Bearer <TOKEN>' https://public-syndesis.my-minishift.syndesis.io/api/v1/public/integrations/timer-to-log/deployments/stop 
```
It returns 
```
{"errorCode":403,"userMsg":"Given request is not acceptable","developerMsg":"Integration timer-to-log is not published"}
```

On the other hand, when the integration is Unpublished, it returns `204 No Content`. In my POV, it should return the error to notify a user that the integration is not published. 

output

dhirajsb:


heiko-braun:
We need this on 1.6.x

Issue #4833, opened 2019-03-11
Should OData remove its split option?


phantomjinx:
Since the introduction of the split step, the usefulness of the split option internal to the odata connector has come into question. Essentially, the split step is doing the same job in that the results of the odata query are split into separate messages and presented individually to the rest of the route.
Up until now, the use of either option has presented no issues. However, a flaw has been identified in the odata split option, namely that even though the result is split the json is still being described by an array schema, ie. the dynamic creation of the json schema does not take into account the fact that the results have been split. This in turn causes a problem for the datamapper:
- Data Mapper shows on the source side an extra parent array node with the entity schema as its child;
- Mapping then publishing causes failure;
This can be rectified but now begs the question should it in favour of removal of the split option, leaving the function to be performed by the split step.

phantomjinx:
@christophd @oscerd @heiko-braun @lburgazzoli 
Guys, I would value your opinions on this, thx.

christophd:
> a flaw has been identified in the odata split option, namely that even though the result is split the json is still being described by an array schema, ie. the dynamic creation of the json schema does not take into account the fact that the results have been split

We can sort this out in the dynamic meta data lookup so this will be in sync with the split option set in the connector.

This leaves the desicion weather to keep the connector split option or not just to the point of usefulness. 

In case we decide to not keep it we should definitely backport that change to the 7.3 release in order to avoid another db migration of existing integrations using the split option. 

Also we have to adapt the final decision to the Google Sheets connector as we have the split option there, too.

christophd:
Regarding usefulness I personally think that the split option on the connector is very straight forward and easy to use for a citizen integrator.

Using the split step requires more insight and understanding of message processing in my opinion. You have to understand where to place the split step and what you want to do. 

Also at the moment split option and split step are identical in functionality. But split step may be much more powerful in the future with custom split expressions and aggregate strategies provided by the user.

Split option on the connector will remain the easy to use and comfortable way to process each element for the whole integration

jsafarik:
@phantomjinx just tried to verify if odata read with split enabled -> Data Mapper -> DB still causes failure, and it does. Should that have a new issue created?
I looked at the output of the entity when split is enable, and the entity is still inside array that is inside array: `[ [ {entity} ] ]`. is this the right behavior?

phantomjinx:
@jsafarik 
[...]

PR #4847, opened 2019-03-12
Issue 4833 odata metadata ret

phantomjinx:


phantomjinx:
@christophd 
This is the change required to fix the problem found in #4833. Are you happy for this to go forward?
(Will probably need to be backported to 1.6 branch as well)

phantomjinx:
@christophd 
There is only 1 commit in this PR now. The other changes should not have been there (local commits I keep around when developing) so I have now force-pushed and cleaned up the PR. You should now only see a single commit addressing issue #4833.

PR #4849, opened 2019-03-12
AtlasMap is unable to find mapping 

### LINKED ISSUE #4846: AtlasMap is unable to find mapping
This is because we do sanitize the resource name like: https://github.com/syndesisio/syndesis/blob/1.6.x/app/server/controller/src/main/java/io/syndesis/server/controller/integration/camelk/CamelKPublishHandler.java#L344

lburgazzoli:
Fixes #4846


heiko-braun:
we need tnis on 1.6.x

lburgazzoli:
it is on 1.6.x

Issue #4848, opened 2019-03-12
New exception with OData's read operation


phantomjinx:
I got new exception with OData's read operation:
`o.a.c.component.olingo4.Olingo4Consumer  : Consumer Consumer[olingo4-olingo4-0-0://read] failed polling endpoint: olingo4-olingo4-0-0://read. Will try again at next poll. Caused by: [org.apache.camel.RuntimeCamelException - org.apache.olingo.client.api.communication.ODataClientErrorException: Path segment not supported: City) [HTTP/1.1 501 Not Implemented]]
org.apache.camel.RuntimeCamelException: org.apache.olingo.client.api.communication.ODataClientErrorException: Path segment not supported: City) [HTTP/1.1 501 Not Implemented]`

Not sure if this is caused by the exception reported in this GH issue or not, either way the read operation is not working properly for me (I am able to get the whole collection and single entity, but not individual property of entity).

### Steps to reproduce:
 1. Create integration with OData connector as start step
 2. Choose read action
 3. Choose "Airports" as Resource collection
 4. Enter `('KLAX')/Location` in the Entity Key Predicate

It would be useful if the connector logged to pod what Request URL it's trying to send.

_Originally posted by @jsafarik in https://github.com/syndesisio/syndesis/issues/4791#issuecomment-471488857_

phantomjinx:
@jsafarik 
The problem is caused not just by the splitting (that is an extra problem I've found while doing a fix) but also in the lack of handling of the key predicate. Up until this point, the ODataComponent assumed only integer predicates of the form _(1)_ so did not take into account the addition of quotes or indeed sub-predicates, ie. (1)/Location. I have a PR in the works that addresses both of these omissions.

jsafarik:
@phantomjinx ok, thanks for clearing it up :)

jsafarik:
Testing this today, I found out that reading a simple property is ok (ie. https://services.odata.org/V4/TripPinServiceRW/People('russellwhyte')/LastName) but reading a complex one (ie. https://services.odata.org/V4/TripPinServiceRW/People('russellwhyte')/AddressInfo) results in `Consumer Consumer[olingo4-olingo4-0-0://read] failed polling endpoint: olingo4-olingo4-0-0://read. Will try again at next poll. Caused by: [org.apache.camel.RuntimeCamelException - org.apache.olingo.commons.api.ex.ODataException: Unsupported property: AddressInfo]` again.

phantomjinx:
[...]

PR #4850, opened 2019-03-12
fix(#4848): Odata Connector to handle more complex predicates

phantomjinx:
Backport for #4848.

* The connector should be able to handle more complex key predicates that
  have subections but also return values rather than entities as their
  query results.

* ODataComponent
 * Checks the format and validity of the key predicate, including
   * sub predicate, eg. (1)/Location
   * ensures quote marks, if required, and brackets

* AbstractODataCustomizer
 * Remove assumption of an entity being returned in the result since it can
   be a value

* Olingo4Consumer
 * Split should handle both entities and values

* Tests for more complex predicates and return of values.


phantomjinx:
Sorry. Confusing myself.

Issue #4840, opened 2019-03-12
AtlasMap mapping definition not bound to the mapping step


lburgazzoli:
While testing camel-k and atlasmap I've noticed that clicking to the done button after having defined the mapping seems to do noting but after a while the mapping editor is closed. 

When the integration is the deployed, the mapping step lacks mapping definition.

![image](https://user-images.githubusercontent.com/1868933/54177656-4cbf5680-4493-11e9-8442-53e8afdfd59d.png)

riccardo-forina:
I tried to reproduce this locally. I updated the backend - on branch `1.6.x` btw - and run the UI in dev mode. I can create an integration as usual, with the mapping and all. The payload looks correct.
So I can't reproduce this really.

There is only one thing that looks off, and it's this error when entering the configuration page of any connector action:
![Screenshot 2019-03-12 at 13 03 07](https://user-images.githubusercontent.com/966316/54198605-3df29700-44c7-11e9-800f-3ca0187c2449.png)

gashcrumb:
Yeah, that bad request is new.  The UI hits that endpoint to fetch data shapes and the config properties object.

Actually nowadays it also hits that endpoint whenever steps are added and removed.  Error responses like this just get ignored though when the request is made in this fashion.

lburgazzoli:
@gashcrumb so I did another test deleting everything (minishift, browser caches) and now I do see the same errors as @riccardo-forina 

![image](https://user-images.githubusercontent.com/1868933/54207774-0f7eb700-44db-11e9-8145-6332918bbe09.png)

but I got the mapping step properly configured but now it is @nicolaferraro that has the same issue :)

zregvart:
That `400 (Bad Request)` must leave some trace of an error somewhere, either in the meta pod or the server pod or even in the response itself. When I tried to reproduce this I focused on trying to make the call to meta pod fail, and in that case I do get `400 (Bad Request)` but with `_meta.message` and `_meta.type` which in turn end up being handled in the UI and present a warning message. So no way to proceed with action editing.

gashcrumb:
Had you maybe not rebuilt your UI image in awhile or something perhaps?

I think the bad request is a red herring, it's getting made as the component initializes.  Looks like the UI isn't sending an empty object or anything at this point, so could be causing the bad request.  But then a second request gets made using an empty object, that succeeds alright and then it's fine from there.  I'll see about tracking down that initial request though.

nicolaferraro:
[...]

PR #4853, opened 2019-03-12
fix: suppress empty backend call to avoid HTTP 400

gashcrumb:
relates to stuff seen in #4840

@zregvart @lburgazzoli FYI


zregvart:
The build is hanging on CircleCI, from what looks like an outage of one of the configured Maven repositories (https://repo.spring.io/milestone)

zregvart:
oh, the build resumed...

Issue #4462, opened 2019-02-06
The syndesis-server pod fails to start on OCP4


anik120:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[*] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

I deployed the Syndesis Operator on an OpenShift cluster created by the OpenShift installer. The operator was deployed successfully after the following commands:
```
$ oc new-project syndesis
$ oc apply -f deploy/syndesis-crd.yml
$ oc apply -f deploy/syndesis-operator.yml
$ oc get templates -n syndesis

$ oc process syndesis-operator -p NAMESPACE=syndesis | oc create -f -
```

I then applied the Syndesis CR:
```
$ oc apply -f deploy/syndesis.yml
```

 

I then performed a search for available routes and found that a syndesis and todo route existed:
```
$ oc get routes
NAME HOST/PORT PATH SERVICES PORT TERMINATION WILDCARD
syndesis syndesis-syndesis2.apps.agreene2.devcluster.openshift.com syndesis-oauthproxy 8443 reencrypt/Redirect None
todo todo-syndesis-syndesis2.apps.agreene2.devcluster.openshift.com / todo 8080 edge/Allow None
```

Visiting the syndesis route showed that requests for integrations and reservations were failing:
```

    Failing Integration Requests: syndesis-syndesis.apps.clusteranik120.devcluster.openshift.com/api/v1/metrics/integrations
    Failing Reservation Requests: syndesis-syndesis.apps.clusteranik120.devcluster.openshift.com/api/v1/event/reservations

```

A quick review of the running pods showed that the syndesis-server pod kept restarting and could not reach a running state, the logs from a short lived pod are attached as [syndesis-server.txt](https://github.com/syndesisio/syndesis/files/2838207/syndesis-server.txt)

heiko-braun:
Looks like it boils down to

```
Caused by: java.io.FileNotFoundException: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt (No such file or directory)
```

You said the cluster was created "created by the OpenShift installer.". What installer and OCP version?

anik120:
@heiko-braun 
> What installer and OCP version?

I'm using the [Openshift installer] (https://github.com/openshift/installer) to install a OCP 4.0 cluster

anik120:
[...]

PR #4855, opened 2019-03-12
Fixes #4854 and #4462. Provides better OpenShift 4 compatibility.

### LINKED ISSUE #4854: OCP 4: syndesis-server exception: Caused by: java.net.UnknownHostException: openshift.default.svc: Name or service not known
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

## Description
On openshift 4, I see this error in the sydnesis-server logs:
`Caused by: java.net.UnknownHostException: openshift.default.svc: Name or service not known`

Using a terminal against the server pod:
```
sh-4.2$ ping openshift.default.svc
ping: openshift.default.svc: Name or service not known
sh-4.2$ ping kubernetes.default.svc
PING kubernetes.default.svc.cluster.local (172.30.0.1) 56(84) bytes of data.
From ip-10-131-0-1.ec2.internal (10.131.0.1) icmp_seq=1 Destination Host Unreachable
From ip-10-131-0-1.ec2.internal (10.131.0.1) icmp_seq=2 Destination Host Unreachable
From ip-10-131-0-1.ec2.internal (10.131.0.1) icmp_seq=3 Destination Host Unreachable```
```

chirino:


zregvart:
I don't think we should be using `service-ca.crt` according to this, `setCaCertFile`, is set to `ca.crt`:

https://github.com/fabric8io/kubernetes-client/blob/8b285c8ac24a5823997bd9671d1b98941719a975/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java#L389

Might be that all we need is to point the `serviceCaCertFile` to `ca.crt`, or even that we do not need to configure `setCaCertFile` as by default it will get set to `ca.crt`.

app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftConfigurationProperties.java
```diff
@@ -27,13 +30,11 @@
...
 
-    public static final String SERVICE_CA_CERT_FILE = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt";
+    public String serviceCaCertFile = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt";
 
     private boolean enabled;
 
-    private String openShiftHost = "https://openshift.default.svc";
-
-    private final OpenShiftConfig openShiftClientConfig = new OpenShiftConfigBuilder().withMasterUrl(openShiftHost).withCaCertFile(SERVICE_CA_CERT_FILE).build();
+    private String openShiftHost = "https://kubernetes.default.svc";
```
heiko-braun: Does this work on 3.11? (Thinking about back porting to 1.6.x)

chirino: yes.  it works.

chirino:
> I don't think we should be using `service-ca.crt` according to this, `setCaCertFile`, is set to `ca.crt`:
> 
> https://github.com/fabric8io/kubernetes-client/blob/8b285c8ac24a5823997bd9671d1b98941719a975/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/Config.java#L389
> 
> Might be that all we need is to point the `serviceCaCertFile` to `ca.crt`, or even that we do not need to configure `setCaCertFile` as by default it will get set to `ca.crt`.

Your likely correct.  I left it as it was as I was not testing against 3.x and was unsure if changing that would break things.

chirino:
QE can test this change now by patching the image stream:

`oc patch -n ${TARGET_PROJECT} is syndesis-server --type='json' -p='[{"op": "replace", "path": "/spec/tags/0/from/name", "value":"quay.io/hchirino/syndesis-server:latest"}]'`

PR #4881, opened 2019-03-13
fix: tidy up a bit and ensure the aggregate step properly gets the subsequent data shape

### LINKED ISSUE #4866: Aggregate step does not update its data shape
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
The aggregate step does not update its datashape when subsequent steps change, split steps handles changes correctly. 
image
It can be fixed by removing the aggregate step and adding it again. 
## Steps to reproduce
- Start: db "select * from todo"
- End: db add_lead
- add split and aggregate steps between start and end
- add template after aggregate (template used by example, use whatever comes in handy)
- aggregate step still says it outputs SQL or SQL procedure parameters
- delete and place aggregate again
- aggregate outputs json-schema for template

gashcrumb:
fixes #4866

Also reduces events posted by the template step and reworks the code where we evaluate split/aggregate so the logic is easier to follow.

@christophd lemme know if this change needs any adjusting.

### DIFF
--- a/app/ui/src/app/integration/edit-page/current-flow.service.ts
+++ b/app/ui/src/app/integration/edit-page/current-flow.service.ts
@@ -649,7 +649,6 @@ export class CurrentFlowService {
         then();
       }
     };
-
     this.currentFlow.steps.forEach((step, position) => {
       outstanding = outstanding + 1;
       switch (step.stepKind) {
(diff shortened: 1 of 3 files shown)


app/ui/src/app/integration/edit-page/current-flow.service.ts
```diff
@@ -716,43 +695,78 @@ export class CurrentFlowService {
...
-            err => {
-              // we'll just pass through
-              sub.unsubscribe();
-              then(step);
-            }
-          );
+        this.fetchStepDescriptor(
+          step,
+          subsequent.action.descriptor.inputDataShape,
+          subsequent.action.descriptor.outputDataShape,
```
christophd: Aggregate should be provided with the outputDataShape of the previous step and inputDataShape of the subsequent step

gashcrumb: Perfect, thanks will update today!

gashcrumb:
@christophd Updated that if you could just double-check that'd be awesome.

@heiko-braun this fix should really go to the 1.6.x branch too, otherwise the aggregate step is kinda broken.

heiko-braun:
@gashcrumb According to your description I have changed the prio of #4866 to blocking. Please create a backport for 1.6.x.

PR #4886, opened 2019-03-13
backport fix(#4833): modify odata out-data-shape if using split

phantomjinx:
* If split property is set then the result of the query will be a single
  entity or value and not a collection. Therefore, the json schema should
  remain an Object Schema rather than an Array Schema.


phantomjinx:
@zregvart 
Already been merged into master.

zregvart:
@phantomjinx to get this merged to `1.6.x` someone needs to classify #4833 as a p0 bug. I believe those are the rules of engagement, am I right @heiko-braun @paoloantinori @tplevko?

heiko-braun:
Yes, @zregvart is correct. @phantomjinx Can you explain which issue you are trying to fix with that PR?

zregvart:
This fixes #4912

PR #4856, opened 2019-03-12
Fixes #4854 and #4462. Provides better OpenShift 4 compatibility.

chirino:

PR #4891, opened 2019-03-14
refactor: cleanup OpenShiftConfigurationProperties

### LINKED ISSUE #4854: OCP 4: syndesis-server exception: Caused by: java.net.UnknownHostException: openshift.default.svc: Name or service not known

zregvart:
There's no need to specify the `CaCertFile` on the Kubernetes Client
correct CA certificate file is automatically picked up if present.

Fixes #4854, #4462

### DIFF
--- a/app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftConfigurationProperties.java
+++ b/app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftConfigurationProperties.java
@@ -15,8 +15,6 @@
  */
 package io.syndesis.server.openshift;
 
-import java.nio.file.Files;
-import java.nio.file.Paths;
 import java.util.Map;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
@@ -29,11 +27,11 @@ import io.fabric8.openshift.client.OpenShiftConfigBuilder;
 @Validated
 public class OpenShiftConfigurationProperties {
 
-    public String serviceCaCertFile = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt";
-
     private boolean enabled;
 
-    private String openShiftHost = "https://kubernetes.default.svc";
+    private String masterUrlHost = "https://kubernetes.default.svc";
+
+    private OpenShiftConfig openShiftClientConfig = new OpenShiftConfigBuilder().withMasterUrl(masterUrlHost).build();
 
     private String builderImageStreamTag = "s2i-java:2.0";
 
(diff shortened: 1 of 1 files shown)


zregvart:
Since we have #4856 on `1.6.x` when reviewed this should be backported as well.

heiko-braun:
Thanks for checking @zregvart

app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftConfigurationProperties.java
```diff
@@ -29,11 +27,11 @@
...
 public class OpenShiftConfigurationProperties {
 
-    public String serviceCaCertFile = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt";
-
     private boolean enabled;
 
-    private String openShiftHost = "https://kubernetes.default.svc";
+    private String masterUrlHost = "https://kubernetes.default.svc";
+
+    private final OpenShiftConfig openShiftClientConfig = new OpenShiftConfigBuilder().withMasterUrl(masterUrlHost).build();
```
chirino: If you do this then you can't change the master URL via bean properties.

zregvart: Is there a need to change the master URL? I agree we want to make it flexible though if it isn't `kubernetes.default.svc` we're not running on k8s/OpenShift right?

zregvart: Anyhow I can make the change. Shouldn't make much difference...

zregvart:
@chirino added update of `openShiftClientConfig` when `masterUrlHost` changes.

chirino:
Lgtm

PR #4916, opened 2019-03-15
fix: #4867 Add proper error handling for integration publish process

### LINKED ISSUE #4867: Starting an integration fails occasionally with SIGSEGV due to missing prometheus config
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
When publishing an integration, it sometimes fails to start after successful build with the following error:
```
Starting the Java application using /opt/run-java/run-java.sh ...
exec java -javaagent:/opt/jolokia/jolokia.jar=config=/opt/jolokia/etc/jolokia.properties -javaagent:/opt/prometheus/jmx_prometheus_javaagent.jar=9779:/tmp/src/prometheus-config.yml -XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:+ExitOnOutOfMemoryError -cp . -jar /deployments/project-0.1-SNAPSHOT.jar
OpenJDK 64-Bit Server VM warning: If the number of processors is expected to increase from one, then you should configure the number of parallel GC threads appropriately using -XX:ParallelGCThreads=N
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:386)
	at sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:401)
Caused by: java.io.FileNotFoundException: /tmp/src/prometheus-config.yml (No such file or directory)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at java.io.FileReader.<init>(FileReader.java:72)
	at io.prometheus.jmx.shaded.io.prometheus.jmx.JmxCollector.<init>(JmxCollector.java:74)
	at io.prometheus.jmx.shaded.io.prometheus.jmx.JavaAgent.premain(JavaAgent.java:47)
	... 6 more
FATAL ERROR in native method: processing of -javaagent failed
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007fbb5b0e4a47, pid=1, tid=0x00007fbb5bed5700
#
# JRE version: OpenJDK Runtime Environment (8.0_191-b12) (build 1.8.0_191-b12)
# Java VM: OpenJDK 64-Bit Server VM (25.191-b12 mixed mode linux-amd64 compressed oops)
# Problematic frame:
# C  [libc.so.6+0x37a47]  abort+0x297
#
# Core dump written. Default location: /deployments/core or core.1
#
# An error report file with more information is saved as:
# /deployments/hs_err_pid1.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp 
#
[error occurred during error reporting , id 0xb]
```

christophd:
This PR fixes #4867 and adds proper error handling to the integration publishing process. 

When the integration project runtime .tar is created potential errors are now propagated as the deployment state marking the deployment as failed. 

In such a situation the s2i build is still triggered (could not avoid this as the .tar is immediately streamed to s2i build) but the initial project generating error is neither lost nor gets overwritten by subsequent integration state change checks.

I made sure that the deploy step is not triggered in this error situation and that the user is informed via the UI that something went wrong.

Bildschirmfoto 2019-03-15 um 15 41 46

Also the integration deployment overview now properly reflects the deployment history with previous errors and the actual deployment state (pending indicator).

Bildschirmfoto 2019-03-15 um 14 20 56

Also fixed some small issues in UI related to error icons and error labels

### DIFF
--- a/app/server/controller/src/main/java/io/syndesis/server/controller/integration/BaseIntegrationController.java
+++ b/app/server/controller/src/main/java/io/syndesis/server/controller/integration/BaseIntegrationController.java
@@ -15,6 +15,16 @@
  */
 package io.syndesis.server.controller.integration;
 
+import java.io.IOException;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+
 import io.syndesis.common.model.ChangeEvent;
 import io.syndesis.common.model.Kind;
 import io.syndesis.common.model.integration.IntegrationDeployment;
@@ -30,16 +40,6 @@ import io.syndesis.server.dao.manager.DataManager;
 import io.syndesis.server.openshift.OpenShiftService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import java.io.IOException;
-import java.util.Optional;
-import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ThreadFactory;
-import java.util.concurrent.TimeUnit;
 
 /**
  * This class tracks changes to Integrations and attempts to process them so that
@@ -181,7 +181,7 @@ public abstract class BaseIntegrationController implements BackendController {
         IntegrationDeployment reconciled = reconcileDeployment(integrationDeployment);
         IntegrationDeploymentState desiredState = reconciled.getTargetState();
         IntegrationDeploymentState currentState = reconciled.getCurrentState();
-        if (!currentState.equals(desiredState)) {
+        if (currentState != IntegrationDeploymentState.Error && currentState != desiredState) {
                 reconciled.getId().ifPresent(integrationDeploymentId -> {
                     StateChangeHandler statusChangeHandler = handlers.get(desiredState);
                     if (statusChangeHandler != null) {
@@ -191,10 +191,10 @@ public abstract class BaseIntegrationController implements BackendController {
                     }
                 });
         } else if (reconciled.getCurrentState() !=  integrationDeployment.getCurrentState()) {
-          dataManager.update(reconciled);
-          scheduledChecks.remove(getIntegrationMarkerKey(reconciled));
+            dataManager.update(reconciled);
+            scheduledChecks.remove(getIntegrationMarkerKey(reconciled));
         } else {
-          scheduledChecks.remove(getIntegrationMarkerKey(reconciled));
+            scheduledChecks.remove(getIntegrationMarkerKey(reconciled));
         }
     }
 
@@ -217,7 +217,7 @@ public abstract class BaseIntegrationController implements BackendController {
                 final String integrationId = integrationDeployment.getIntegrationId().get();
                 final int deploymentVersion = integrationDeployment.getVersion();
                 LOG.info("IntegrationDeploymentId {} Integration {} : Start processing integration: {}, version: {} with handler: {}", integrationDeploymentId, integrationId, integrationId, deploymentVersion, handler.getClass().getSimpleName());
-                handler.execute(integrationDeployment, update->{
+                handler.execute(integrationDeployment, update-> {
                     if (LOG.isInfoEnabled()) {
                         LOG.info("{} : Setting status to {}{}",
                             getLabel(integrationDeployment),
@@ -231,6 +231,7 @@ public abstract class BaseIntegrationController implements BackendController {
                     final IntegrationDeployment updated = current.builder()
                         .statusMessage(Optional.ofNullable(update.getStatusMessage()))
                         .currentState(update.getState())
+                        .error(update.getError())
                         .stepsDone(update.getStepsPerformed())
                         .build();
                     LOG.trace("Updated {} , Current {}", updated, current);
@@ -297,7 +298,17 @@ public abstract class BaseIntegrationController implements BackendController {
     private IntegrationDeployment reconcileDeployment(IntegrationDeployment integrationDeployment) {
         IntegrationDeploymentState actualState = determineState(integrationDeployment);
         LOG.debug("Actual state: {}, Persisted state: {}, Desired state: {}", actualState, integrationDeployment.getCurrentState(), integrationDeployment.getTargetState());
-        //We also need to compare with current state to make sure we only call the unpublished() once. Calling it multiple times, might loose the steps perfromed, causing an infinite loop.
+
+        // Make sure to not loose error state, only overwrite when actual state in Openshift is published for some reason
+        if (integrationDeployment.getCurrentState() == IntegrationDeploymentState.Error) {
+            if (actualState == IntegrationDeploymentState.Published) {
+                return integrationDeployment.withCurrentState(actualState);
+            } else {
+                return integrationDeployment;
+            }
+        }
+
+        //We also need to compare with current state to make sure we only call the unpublished() once. Calling it multiple times, might loose the steps performed, causing an infinite loop.
         if (actualState == IntegrationDeploymentState.Unpublished && actualState != integrationDeployment.getCurrentState()) {
             return integrationDeployment.unpublished();
         } else {
--- a/app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
+++ b/app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
@@ -22,14 +22,17 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 
 import io.syndesis.common.model.integration.Integration;
 import io.syndesis.common.model.integration.IntegrationDeployment;
+import io.syndesis.common.model.integration.IntegrationDeploymentError;
 import io.syndesis.common.model.integration.IntegrationDeploymentState;
 import io.syndesis.common.util.Labels;
 import io.syndesis.common.util.SyndesisServerException;
+import io.syndesis.integration.api.IntegrationErrorHandler;
 import io.syndesis.integration.api.IntegrationProjectGenerator;
 import io.syndesis.server.controller.StateChangeHandler;
 import io.syndesis.server.controller.StateUpdate;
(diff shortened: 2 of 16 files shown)


gashcrumb:
UI bits LGTM, thanks for that!

app/server/controller/src/main/java/io/syndesis/server/controller/integration/BaseIntegrationController.java
```diff
@@ -181,7 +181,7 @@ private void checkIntegrationStatus(IntegrationDeployment integrationDeployment)
         IntegrationDeployment reconciled = reconcileDeployment(integrationDeployment);
         IntegrationDeploymentState desiredState = reconciled.getTargetState();
         IntegrationDeploymentState currentState = reconciled.getCurrentState();
-        if (!currentState.equals(desiredState)) {
+        if (!currentState.equals(IntegrationDeploymentState.Error) && !currentState.equals(desiredState)) {
```
zregvart: So here we're saying that if the current state is not error we'll reconcile by invoking the handler. @rhuss does that make sense to you? Should we be retrying to reconcile for errors as well?

app/server/controller/src/main/java/io/syndesis/server/controller/integration/BaseIntegrationController.java
```diff
@@ -297,7 +298,17 @@ private boolean stale(StateChangeHandler handler, IntegrationDeployment integrat
     private IntegrationDeployment reconcileDeployment(IntegrationDeployment integrationDeployment) {
         IntegrationDeploymentState actualState = determineState(integrationDeployment);
         LOG.debug("Actual state: {}, Persisted state: {}, Desired state: {}", actualState, integrationDeployment.getCurrentState(), integrationDeployment.getTargetState());
-        //We also need to compare with current state to make sure we only call the unpublished() once. Calling it multiple times, might loose the steps perfromed, causing an infinite loop.
+
+        // Make sure to not loose error state, only overwrite when actual state in Openshift is published for some reason
+        if (integrationDeployment.getCurrentState().equals(IntegrationDeploymentState.Error)) {
```
zregvart:
```suggestion
        if (integrationDeployment.getCurrentState() == IntegrationDeploymentState.Error) {
```

app/server/controller/src/main/java/io/syndesis/server/controller/integration/BaseIntegrationController.java
zregvart:
```suggestion
        if (currentState != IntegrationDeploymentState.Error && currentState != desiredState) {
```

app/server/controller/src/main/java/io/syndesis/server/controller/integration/BaseIntegrationController.java
```diff
@@ -297,7 +298,17 @@ private boolean stale(StateChangeHandler handler, IntegrationDeployment integrat
...
+            if (actualState.equals(IntegrationDeploymentState.Published)) {
```
zregvart:
```suggestion
            if (actualState == IntegrationDeploymentState.Published) {
```

app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
```diff
@@ -177,6 +185,12 @@ private String deploy(IntegrationDeployment integration, DeploymentData data) th
 
     // =================================================================================
 
+    private boolean hasError(IntegrationDeployment integrationDeployment) {
+        return integrationDeployment.getCurrentState().equals(IntegrationDeploymentState.Error) ||
```
zregvart:
```suggestion
        return integrationDeployment.getCurrentState() == IntegrationDeploymentState.Error ||
```

app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
```diff
@@ -238,25 +252,51 @@ private InputStream createProjectFiles(Integration integrationDeployment) {
...
+                logInfo(integrationDeployment, "Skipped step {} because already performed", step);
             }
+        }
+
+        void perform(String step, IoCheckedErrorHandlingFunction<IntegrationDeployment> callable, DeploymentData data) {
+            if (!stepsPerformed.containsKey(step)) {
+                stepsPerformed.put(step, callable.apply(integrationDeployment, data,
+                        (throwable) -> error = new IntegrationDeploymentError.Builder()
+                                                    .type(throwable.getClass().getName())
+                                                    .message(throwable.getMessage())
```
zregvart: Perhaps it would be good to log the stack trace of that `throwable`, I know we do it for build step, could be easily forgotten and then difficult to debug.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/IntegrationOverviewHelper.java
```diff
@@ -84,48 +86,50 @@ static IntegrationOverview toCurrentIntegrationOverview(final Integration integr
...
-                // publish
-                // will have it's status != Unpublished, the reason why we can't
-                // look at the last deployment is because users can choose to
-                // deploy
-                // previous deployments, so we bet that all the Unpublished
-                // integrations are not the ones that the user don't hold the
-                // current state
-                builder.targetState(deployment.getTargetState());
-                builder.currentState(currentState);
+            if (deployment.getCurrentState().equals(IntegrationDeploymentState.Pending) ||
```
zregvart:
```suggestion
            if (deployment.getCurrentState() == IntegrationDeploymentState.Pending ||
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/IntegrationOverviewHelper.java
```diff
@@ -84,48 +86,50 @@ static IntegrationOverview toCurrentIntegrationOverview(final Integration integr
...
+                deployment.getCurrentState().equals(IntegrationDeploymentState.Published)) {
```
zregvart:
```suggestion
                deployment.getCurrentState() == IntegrationDeploymentState.Published) {
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/IntegrationOverviewHelper.java
```diff
@@ -84,48 +86,50 @@ static IntegrationOverview toCurrentIntegrationOverview(final Integration integr
...
                 activeDeployments.add(deployment);
             }
         }
 
-        if (deployed != null) {
-            builder.isDraft(computeDraft(integration, deployed.getSpec()));
+        Optional<IntegrationDeployment> maybePublished = activeDeployments.stream()
+                  .filter(deployment -> deployment.getCurrentState().equals(IntegrationDeploymentState.Published))
```
zregvart:
```suggestion
                  .filter(deployment -> deployment.getCurrentState() == IntegrationDeploymentState.Published)
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/IntegrationOverviewHelper.java
```diff
@@ -84,48 +86,50 @@ static IntegrationOverview toCurrentIntegrationOverview(final Integration integr
...
+                  .findAny();
```
zregvart:
```suggestion
                  .findFirst();
```

`findAny()` is [non-deterministic with regards to ordering](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#findAny--), I think we wan't the first published deployment (and I think that should be the last deployment for that integration)

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/IntegrationOverviewHelper.java
```diff
@@ -84,48 +86,50 @@ static IntegrationOverview toCurrentIntegrationOverview(final Integration integr
...
+
+        IntegrationDeployment exposedDeployment = null;
+        if (maybePublished.isPresent()) {
+            IntegrationDeployment published = maybePublished.get();
+            builder.isDraft(computeDraft(integration, published.getSpec()));
+            builder.version(published.getVersion());
+            exposedDeployment = published;
+        } else if (activeDeployments.size() == 1) {
+            exposedDeployment = activeDeployments.get(0);
```
zregvart: Not 100% sure about correctness as it also might contain integrations in `Pending` state

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/IntegrationOverviewHelper.java
zregvart: Not 100% sure about including `Pending`, I might be wrong it doesn't seem like we did that in the current version.

app/server/controller/src/main/java/io/syndesis/server/controller/integration/BaseIntegrationController.java
christophd: I saw that if we reconcile errors here too the reconcile just overwrites the error and continues with deployment. I also think in case we hit some error we should not try to auto recover from that but wait for another publish request

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/IntegrationOverviewHelper.java
christophd: The user can also redeploy a previous deployment so it is not always the last deployment in the list which is in published state.

Though there should be only one published deployment in that list at the same time so findFirst is fine

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/IntegrationOverviewHelper.java
christophd: I think we should include pending here as it can be any deployment in the list due to redeploy functionality. The UI then displays the current status of the pending deployment.

Publish state wins 1st then any pending deployment then the latest deployment regardless its current state

heiko-braun:
@valdar @christophd i would suggest we create a separate issue and PR for the camel k cases.

christophd:
@valdar @heiko-braun I had a look at the CamelK handlers and I think we do not have the problem there as no s2i image build is involved and all resources are directly added to the integration CRD. In case the CRD build fails we get direct error propagation to the publish process and that's it basically.

zregvart:
1.6.x backport in #4931

Issue #4905, opened 2019-03-15
Datamapper before/after Aggregate - both sides of the mapping are equal (can't map anything)


mastepan:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
**Example 1:**
DB -> Split -> Datamapper -> Aggregate -> Google sheets
**Datamapper offers mapping from DB to DB**
![Screenshot from 2019-03-15 11-00-08](https://user-images.githubusercontent.com/8707251/54424078-6ae6ba00-4712-11e9-93c5-545e2758b944.png)

**Example 2:**
DB -> Split -> Aggregate -> Datamapper-> Google sheets
**Datamapper offers mapping from Google sheets to Google sheets**
![Screenshot from 2019-03-15 11-00-22](https://user-images.githubusercontent.com/8707251/54424051-5c000780-4712-11e9-80b0-a27db35f0723.png)

christophd:
I can not reproduce this behavior on branch 1.6.x (from yesterday evening)

![Bildschirmfoto 2019-03-15 um 13 57 21](https://user-images.githubusercontent.com/195264/54432863-a68d7e00-472a-11e9-852b-e76b7db5eb20.png)

![Bildschirmfoto 2019-03-15 um 13 57 52](https://user-images.githubusercontent.com/195264/54432867-ab523200-472a-11e9-803f-e88eb8ab6e8b.png)

Steps I did:

1) SQL start connection
2) Google Sheets finish connection
3) Split
4) Aggregate
5) Data Mapper

heiko-braun:
@mastepan Moving to done according to @christophd 's comments. Can you verify on recent 1.6.x?

christophd:
@mastepan this issue came in on master a few day ago and now is also on 1.6.x. So no need to retest against 1.6.x, thanks! PR fixing the issue is on its way https://github.com/syndesisio/syndesis/pull/4929

mastepan:
verified

Issue #4879, opened 2019-03-13
Aggregate datashape 


avano:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
I want to create this integration: 

```
sql select
split
  filter (I want only some elements from the collection)
aggregate (I want to continue with the collection)
datamapper (I want to concatenate all last names from the collection into one message)
activemq queue
```

When I use following flow:

```
1. start SQL
2. end AMQ
3. split
4. aggregate
```

The split has the `none -> sql result` datashape (correct), but the aggregate has `amq-in(json instance in my case) -> sql result`.

In this case, the aggregate should have the same datashapes as the `sql select` as I only filtered out some elements.

This results in two datamappers required by the ui:

![6Gnn](https://user-images.githubusercontent.com/7081216/54288195-5508b600-45a7-11e9-82ee-8685e8604d06.png)

avano:
@gashcrumb @christophd any thoughts on this?

avano:
maybe would be solved with #4881

gashcrumb:
Yeah, I think it should, at least it's worth a test once that PR is merged.

christophd:
I think at this point in your described steps 1-4 everything is fine and as expected. At this point there is no data mapper step yet between aggregate and amq so aggregate will have the same shape as amq. Once you add a mapper things change and aggregate will align to the new situation.

christophd:
To be a bit more precise in explanation the aggregate always adopts the input shape of the subsequent step and the output shape of the previous step. 

In your described state the aggregate adopted the shapes according to the steps previous and subsequent to the aggregate which is `amq-in(json instance in my case) -> sql result`.

In addition to adopting the shapes the aggregate transforms the input shape from type single element to type collection.

In case you add a mapper to the picture the aggregate will align its shapes to the new situation.

@avano Can you please give that a try and report any concerns that you are seeing?

avano:
@christophd 

> In addition to adopting the shapes the aggregate transforms the input shape from type single element to type collection.

This is what is **not** happening in my case and maybe it causes the confusion :-)

In my case, when I add aggregate, it has only `SQL Result` datashape and no collection.

avano:
@christophd 

anyway, where the datashape should be place in the case? I would expect to place it between aggregate and amq step, because that is what I want to map really.

[...]

PR #4929, opened 2019-03-18
fix: #4905 #4879 Data shapes for aggregate step mixed up

christophd:
Input and output shape were mixed up when lookup aggregate step meta data


zregvart:
backport in #4932

PR #4946, opened 2019-03-18
fix: prevent the user from moving on until a mapping file is produced

### LINKED ISSUE #4840: AtlasMap mapping definition not bound to the mapping step

gashcrumb:
fixes #4840


zregvart:
Backport for 1.6.x in #4949

Issue #4765, opened 2019-03-05
Integration details page with two shades of grey 


dongniwang:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

With the latest masthead changes, the integration details page now has two sets of grey in the background areas. 

## Expected behavior

We should use only one type of grey for integration details page. 

## Screenshot

<img width="1438" alt="screen shot 2019-03-05 at 1 52 42 pm" src="https://user-images.githubusercontent.com/24943812/53829762-dd0b1200-3f4e-11e9-92d7-900c6648d15f.png">

## Tasks involved / Steps to Reproduce

1.
2.
3.
4.

cc: @seanforyou23 @mcoker

mcoker:
I think you can just change this line to the value below https://github.com/syndesisio/syndesis/blob/master/app/ui/src/scss/utils/_overrides.scss#L39

`--pf-c-page__main-section--BackgroundColor: #e1e1e1;`

tplevko:
Verified by QE

PR #4952, opened 2019-03-18
fix(styles): ensure main page section container applies correct bg color

seanforyou23:
This PR updates `pf-c-page__main-section` to use the proper color reference so that the page background isn't different between various views of the app.

Should help close https://github.com/syndesisio/syndesis/issues/4765

### DIFF
--- a/app/ui/src/scss/utils/_overrides.scss
+++ b/app/ui/src/scss/utils/_overrides.scss
@@ -36,7 +36,7 @@ body.cards-pf {
 .pf-c-page__main {
   z-index: 100;
   .pf-c-page__main-section {
-    background-color: #ededed;
+    background-color: $color-pf-custom-gray;
     margin-top: 76px;
     overflow: auto;
     padding-top: 0;


app/ui/src/scss/utils/_overrides.scss
```diff
@@ -36,7 +36,7 @@ body.cards-pf {
 .pf-c-page__main {
   z-index: 100;
   .pf-c-page__main-section {
-    background-color: #ededed;
+    background-color: $color-pf-custom-gray;
     margin-top: 76px;
     overflow: auto;
     padding-top: 0;
```
mcoker: what do you think about changing this to `padding: 0 20px;`?

gashcrumb:
@seanforyou23 Does this need to go into 7.3?

gashcrumb:
@dongniwang actually same question for you, does this need to be in 7.3 also?

dongniwang:
@gashcrumb - ideally, yes, if that's possible. But it's not the end of the world if we have two different background colors...

cc: @syndesisio/uxd

gashcrumb:
@dongniwang cool, thanks for the heads up!

@seanforyou23 Would you be able to cherry-pick these changes over to the `1.6.x` branch and open a second PR for that branch?  Lemme know if you need any help with that...

app/ui/src/scss/utils/_overrides.scss
seanforyou23: Not against it, but worth noting that fixing the padding in one place seems to break it in others. I'll give it a shot when I move this PR to the other branch as recommended so I can remember why I left it as is. There might be a simple fix in there somewhere.

zregvart:
Backport for 1.6.x in #4958

Issue #4889, opened 2019-03-14
OData connector cant delete/update entity


jsafarik:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

When I do a read from OData service (as start step) and then delete, the delete suggests that I should put datamapper in between, but then there's no way to map it (read operation sends it's datashape, but there's nothing in the delete step). 

![delete_op](https://user-images.githubusercontent.com/46523434/54344931-40c5c700-4642-11e9-9911-10d2ad8cbaf2.png)
when not using a datamapper, the path is created so that the whole body that read produces is taken by the delete operation as Key predicate, which results into wrong resourcePath.

For update, I can map it, but it takes the Key Predicate without single quatation marks (using Airports(KLAX) instead of Airports('KLAX')), which produces error as well.

jsafarik:
@phantomjinx 
Is there a support of if-match or if-none-match header with the delete action?

I am trying to delete an entity from TripPin sample service which requires one of those headers to successfully delete it. 

Not sure if it is an issue of the sample service since the [Protocol specification](http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part1-protocol/odata-v4.0-errata03-os-part1-protocol-complete.html#_Toc453752230) states that the client MAY include one of the headers (so it shouldn't be a requirement).

jsafarik:
So with further looking into that, it seems that there are services that require ETag for data modification (this ETag can be read from headers when doing GET request for entity) and without that ETag specified it is not possible to modify the data (update or delete).

PR #4955, opened 2019-03-18
fix(#4889): Redo input DataShape of OData Delete

phantomjinx:
* odata.json
 * Configure Delete to have a json-schema model on input in line with
   Create and Patch

* ODataMetadateRetrieval
 * Breaks shaping out into individual methods for better readability
 * Modifies Delete datashape to use json-schema with key-predicate

* ...Customizer
 * Delete to extract key predicate from json model
 * Read to observe split property to ensure results are formatted in the
   expected manner. Without this, Read will not map to Delete using the
   DataMapper

* Tests
 * Fixes tests so that, when split, results are not assumed to be in a list
   as this makes little sense when there is most likely 1 result per message

### DIFF
--- a/app/connector/odata/src/main/java/io/syndesis/connector/odata/component/ODataComponent.java
+++ b/app/connector/odata/src/main/java/io/syndesis/connector/odata/component/ODataComponent.java
@@ -32,7 +32,7 @@ import io.syndesis.connector.odata.PropertyBuilder;
 import io.syndesis.integration.component.proxy.ComponentDefinition;
 import io.syndesis.integration.component.proxy.ComponentProxyComponent;
 
-final class ODataComponent extends ComponentProxyComponent implements ODataConstants {
+public final class ODataComponent extends ComponentProxyComponent implements ODataConstants {
 
     private static final Pattern NUMBER_ONLY_PATTERN = Pattern.compile("-?\\d+");
 
--- a/app/connector/odata/src/main/java/io/syndesis/connector/odata/customizer/AbstractODataCustomizer.java
+++ b/app/connector/odata/src/main/java/io/syndesis/connector/odata/customizer/AbstractODataCustomizer.java
@@ -17,10 +17,12 @@ package io.syndesis.connector.odata.customizer;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import org.apache.camel.CamelContext;
 import org.apache.camel.CamelContextAware;
 import org.apache.camel.Message;
+import org.apache.olingo.client.api.domain.ClientCollectionValue;
 import org.apache.olingo.client.api.domain.ClientEntity;
 import org.apache.olingo.client.api.domain.ClientEntitySet;
 import org.apache.olingo.client.api.domain.ClientValue;
@@ -42,6 +44,8 @@ import io.syndesis.integration.component.proxy.ComponentProxyCustomizer;
 public abstract class AbstractODataCustomizer implements ComponentProxyCustomizer, CamelContextAware, ODataConstants {
 
     protected static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    private boolean split;
 
     static {
         SimpleModule module =
@@ -68,19 +72,27 @@ public abstract class AbstractODataCustomizer implements ComponentProxyCustomize
         return this.camelContext;
     }
 
+    public boolean isSplit() {
+        return this.split;
+    }
+
+    protected void setSplit(boolean split) {
+        this.split = split;
+    }
+
     protected void convertMessageToJson(Message in) throws JsonProcessingException {
         if (in.getBody(Object.class) == null) {
             in.setBody(Collections.emptyList());
             return;
         }
 
-        List<String> resultList = new ArrayList<>();
         Object item = in.getBody(Object.class);
         if (item instanceof ClientEntitySet) {
             //
             // If the results have not been split and returned as a
             // ClientEntitySet then split it up into a recognisable list
             //
+            List<String> resultList = new ArrayList<>();
             ClientEntitySet entitySet = (ClientEntitySet) item;
             List<ClientEntity> entities = entitySet.getEntities();
             for (ClientEntity entity : entities) {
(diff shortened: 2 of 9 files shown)


phantomjinx:
@christophd 
Can I particularly get your review of the Customizer changes I've made. Essentially, I have aligned the resulting structure of the odata 'read' with the split property set on the OData Component. Therefore ...
- split = false: Array json-schema is set as Datashape and a json array of results is always returned
- split = true: Entity json-schema is set as Datashape and a json object of each result is returned in different messages.
This has changed since an array was always returned. However, in the situation of the route odata-read->odata-delete that has split=true, the results end up being mapped by AtlasMapping between a json-array and a json-object. This fails with an AtlasException, hence the change required here.
Feel free to ping me for more clarification.

phantomjinx:
Guys,
Can I maybe get a review on this asap, as it references a p0 issue?
Thanks

@heiko-braun @zregvart @lburgazzoli @christophd @oscerd

zregvart:
I think @christophd is already looking into it...

christophd:
Yes, I have a look at it right now. And what I see looks good to me.

I wonder though why the AtlasMapping between a json-array and a json-object did fail with an AtlasException as you mentioned. The odata meta data shape lookup alsready returned the json-schema as object schema when split option was set. So the mapper should align with that in its mapping and suggest an object-to-object mapping. 

Other connectors like GoogleSheets do also set the json-schema to object schema when split is set and provide a list with a single element to the data mapper. Do you think this will fail then, too. Do we have to adjust the GoogleSheets connector, too?

zregvart:
Backport to 1.6.x in #4987

phantomjinx:
> I wonder though why the AtlasMapping between a json-array and a json-object did fail with an AtlasException as you mentioned. The odata meta data shape lookup alsready returned the json-schema as object schema when split option was set. So the mapper should align with that in its mapping and suggest an object-to-object mapping.
> 
So, although the json-schema was set as object, the actual message going through the exchange contained an array containing a single json string. This is where the crux of the error occurred in that AltasMapper was expecting a single json object but got an array.

> Other connectors like GoogleSheets do also set the json-schema to object schema when split is set and provide a list with a single element to the data mapper. Do you think this will fail then, too. Do we have to adjust the GoogleSheets connector, too?
>
Only if the schema set by GoogleSheet's dynamic Shaping is set to one thing yet it Customizers assume another. As long as these are aligned then it shouldn't be a problem.

christophd:
So the issue would also have been fixed with odata delete customizer being able to handle exchange with a single element list as body?

phantomjinx:
Indeed. Although, I think that may have been less optimum given that odata-delete requires a single key-predicate (was originally a simple String but that didn't map correctly -  as seen in the screenshot above).

christophd:
@phantomjinx I just tested that with GoogleSheets while split option enabled and I hit some error in data mapper, too.

```
io.atlasmap.api.AtlasException: Errors: [Unexpected exception is thrown while reading source field: null: docId='-LaQ3UP2ZcoAB-YQIrI3', path='/A'], [Unexpected exception is thrown while reading source field: null: docId='-LaQ3UP2ZcoAB-YQIrI3', path='/A'], [Unexpected exception is thrown while reading source field: null: docId='-LaQ3UP2ZcoAB-YQIrI3', path='/B'], [Unexpected exception is thrown while reading source field: null: docId='-LaQ3UP2ZcoAB-YQIrI3', path='/A'], [Unexpected exception is thrown while reading source field: null: docId='-LaQ3UP2ZcoAB-YQIrI3', path='/B'], [Unexpected exception is thrown while reading source field: null: docId='-LaQ3UP2ZcoAB-YQIrI3', path='/C'], 
	at org.apache.camel.component.atlasmap.AtlasEndpoint.onExchange(AtlasEndpoint.java:266)
	at org.apache.camel.impl.ProcessorEndpoint$1.process(ProcessorEndpoint.java:71)
	at org.apache.camel.util.AsyncProcessorConverterHelper$ProcessorToAsyncProcessorBridge.process(AsyncProcessorConverterHelper.java:61)
	at org.apache.camel.processor.SendProcessor.process(SendProcessor.java:148)
	at org.apache.camel.processor.DelegateAsyncProcessor.process(DelegateAsyncProcessor.java:110)
	at org.apache.camel.processor.RedeliveryErrorHandler.process(RedeliveryErrorHandler.java:548)
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:138)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:101)
	at org.apache.camel.processor.DelegateAsyncProcessor.process(DelegateAsyncProcessor.java:110)
	at io.syndesis.integration.runtime.logging.ActivityTrackingInterceptStrategy$EventProcessor.process(ActivityTrackingInterceptStrategy.java:79)
	at org.apache.camel.processor.DelegateAsyncProcessor.process(DelegateAsyncProcessor.java:110)
	at org.apache.camel.processor.RedeliveryErrorHandler.process(RedeliveryErrorHandler.java:548)
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:138)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:101)
	at org.apache.camel.processor.CamelInternalProcessor.process(CamelInternalProcessor.java:201)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:138)
	at org.apache.camel.processor.Pipeline.process(Pipeline.java:101)
	at org.apache.camel.component.google.sheets.stream.GoogleSheetsStreamConsumer.processBatch(GoogleSheetsStreamConsumer.java:129)
	at org.apache.camel.component.google.sheets.stream.GoogleSheetsStreamConsumer.poll(GoogleSheetsStreamConsumer.java:111)
	at org.apache.camel.impl.ScheduledPollConsumer.doRun(ScheduledPollConsumer.java:174)
	at org.apache.camel.impl.ScheduledPollConsumer.run(ScheduledPollConsumer.java:101)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

@phantomjinx Does it look familiar to you? I guess I have to do the same fix as you did for GoogleSheets. Will open another issue for that.

phantomjinx:
> ```
> io.atlasmap.api.AtlasException: Errors: [Unexpected exception is thrown while reading source field: null: 
> <snip>
>
> @phantomjinx Does it look familiar to you? I guess I have to do the same fix as you did for GoogleSheets. Will open another issue for that.

Yep! That's the one. The 'reading source field: null' is the defining property.

christophd:
@phantomjinx many thanks! I just opened a p0 for GoogleSheets #4994

PR #4956, opened 2019-03-18
fix: hide the log connection in favor of the log step

### LINKED ISSUE #4935: Two "Log" cards appear when the user has to choose a step
## This is a...

<pre><code>
[ ] Feature request
[ x] Bug report  
</code></pre>

## Description

For 7.3, the user now sees two Log cards but has no way to know which card to click. Something must change for 7.3.  See related issue: 
https://github.com/syndesisio/syndesis/issues/4425

gashcrumb:
fixes #4935 

Please make sure that this PR is indeed hiding the thing we want to hide, for reference:

image

Also this PR adds support to hide connections/steps via setting `hide-from-step-select` to true in the step or connection's metadata map.

@zregvart @christophd I wonder, should I be checking the connector too for this test?  I thought the connection metadata should be enough.

### DIFF
--- a/app/ui/src/app/integration/edit-page/flow-functions.ts
+++ b/app/ui/src/app/integration/edit-page/flow-functions.ts
@@ -14,6 +14,7 @@ import {
   ConfigurationProperty,
   createConnectionStep,
   StepOrConnection,
+  HIDE_FROM_STEP_SELECT,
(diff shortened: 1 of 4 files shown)


zregvart:
Backport for 1.6.x in #4973

Issue #4953, opened 2019-03-18
Remove masthead background image 


dongniwang:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
[x] UI update
</code></pre>

## Description
Need to remove the masthead background image per latest conversation. Please make sure there's no background image for both the upstream version and the productized version. 

<img width="1435" alt="Screen Shot 2019-03-18 at 1 17 05 PM" src="https://user-images.githubusercontent.com/24943812/54549363-4b54c900-4980-11e9-9023-b827730779fb.png">
 
cc: @amysueg @mcoker @seanforyou23 @gashcrumb

gashcrumb:
Does this need to get into 7.3?  @heiko-braun FYI

dongniwang:
I want to say yes. @amysueg @catrobson - can you please verify?

dongniwang:
OK. Just confirmed, yes it needs to get into 7.3. 

cc: @amysueg @catrobson

heiko-braun:
@gashcrumb Can you backport the change when we get the PR?

gashcrumb:
Yup!

tplevko:
@gashcrumb the image was removed, however the top of the page now has two shades of gray
![colour_of_header](https://user-images.githubusercontent.com/4180208/54990448-0ef61e00-4fbb-11e9-8f2d-22f21c002992.png)

This is how it looked with Fuse Onlien 7.2.:
![top_fuse_online](https://user-images.githubusercontent.com/4180208/54990477-203f2a80-4fbb-11e9-81d4-1622054a263a.png)

Just want to ask, whether the lighter gray is there on purpose, or it's just an accident.

gashcrumb:
I'm not sure tbh, @dongniwang @seanforyou23 should it be the two shades like that or a single color at the top?

dongniwang:
Adding @mcoker here to help clarify. 

cc: @syndesisio/uxd

seanforyou23:
Ah, shouldn't be two shades of grey. We fixed this in master yesterday, does it also need to go into `1.6.x` branch?

mcoker:
Looks OK to me on master. This was merged yesterday https://github.com/syndesisio/syndesis/pull/5049

[...]

PR #4957, opened 2019-03-18
fix(masthead): remove masthead bg image

seanforyou23:
This PR removes the background image that's visible from the masthead. This change should affect both the productized version as well as upstream Syndesis.

Should help close: https://github.com/syndesisio/syndesis/issues/4953

Screen Shot 2019-03-18 at 3 43 33 PM


dongniwang:
cc: @syndesisio/uxd

PR #4962, opened 2019-03-19
fix: #4879 Data shape meta lookup (split/aggregate) for user defined specs

christophd:
When user defines data shape specifications for a connection (e.g. AMQ, Template, Webhook) the collection/single element meta data variant information is missing on the data shape.

The meta data shape lookup for split/aggregate was not able to handle these specifications and caused errors on Syndesis server. This caused the lookup to return the original non inspected data shapes for split/aggregate leading to inconsistent data shape display in UI.

This PR makes sure that the server is able to handle user defined specifications (json-schema, json-instance) that do not have proper meta data variant information set.

This fix needs to be backported to 1.6.x

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
@@ -29,6 +29,7 @@ import io.syndesis.common.model.integration.StepKind;
 import io.syndesis.common.util.Json;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.util.StringUtils;
 
 /**
  * @author Christoph Deppisch
@@ -47,34 +48,38 @@ class AggregateMetadataHandler implements StepMetadataHandler {
 
     @Override
     public DynamicActionMetadata handle(DynamicActionMetadata metadata) {
-        try {
-            DataShape outputShape = metadata.outputShape();
-            DataShape inputShape = metadata.inputShape();
+        DataShape outputShape = metadata.outputShape();
+        DataShape inputShape = metadata.inputShape();
 
+        try {
             if (metadata.outputShape() != null) {
                 outputShape = adaptOutputShape(new DataShape.Builder()
-                                                        .createFrom(metadata.outputShape())
-                                                        .decompress()
-                                                        .build());
+                        .createFrom(metadata.outputShape())
+                        .decompress()
+                        .build());
             }
+        } catch (IOException e) {
+            LOG.warn("Unable to read input data shape on dynamic meta data inspection", e);
+            LOG.warn("Using original input shape as fallback");
+        }
 
+        try {
             if (metadata.inputShape() != null) {
                 inputShape = adaptInputShape(new DataShape.Builder()
                                                         .createFrom(metadata.inputShape())
                                                         .decompress()
                                                         .build());
             }
-
-            return new DynamicActionMetadata.Builder()
-                    .createFrom(metadata)
-                    .inputShape(inputShape)
-                    .outputShape(outputShape)
-                    .build();
         } catch (IOException e) {
             LOG.warn("Unable to read output data shape on dynamic meta data inspection", e);
+            LOG.warn("Using original output shape as fallback");
         }
 
-        return metadata;
+        return new DynamicActionMetadata.Builder()
+                .createFrom(metadata)
+                .inputShape(inputShape)
+                .outputShape(outputShape)
+                .build();
     }
 
     DataShape adaptInputShape(DataShape dataShape) throws IOException {
@@ -93,31 +98,42 @@ class AggregateMetadataHandler implements StepMetadataHandler {
 
         DataShape collectionShape = dataShape.findVariantByMeta(VARIANT_METADATA_KEY, VARIANT_COLLECTION).orElse(dataShape);
 
-        if (collectionShape.getKind().equals(DataShapeKinds.JSON_SCHEMA)) {
-            String specification = collectionShape.getSpecification();
-            JsonSchema schema = Json.reader().forType(JsonSchema.class).readValue(specification);
+        String specification = collectionShape.getSpecification();
+        if (StringUtils.hasText(specification)) {
+            if (collectionShape.getKind() == DataShapeKinds.JSON_SCHEMA) {
+                JsonSchema schema = Json.reader().forType(JsonSchema.class).readValue(specification);
 
-            if (schema.isArraySchema()) {
-                ArraySchema.Items items = ((ArraySchema) schema).getItems();
-                JsonSchema itemSchema = items.asSingleItems().getSchema();
-                itemSchema.set$schema(schema.get$schema());
-                if (items.isSingleItems()) {
+                if (schema.isArraySchema()) {
+                    ArraySchema.Items items = ((ArraySchema) schema).getItems();
+                    JsonSchema itemSchema = items.asSingleItems().getSchema();
+                    itemSchema.set$schema(schema.get$schema());
+                    if (items.isSingleItems()) {
+                        return new DataShape.Builder().createFrom(collectionShape)
+                                .putMetadata(VARIANT_METADATA_KEY, VARIANT_ELEMENT)
+                                .specification(Json.writer().writeValueAsString(itemSchema))
+                                .addAllVariants(extractVariants(dataShape, collectionShape, VARIANT_COLLECTION))
+                                .build();
+                    }
+                } else {
                     return new DataShape.Builder().createFrom(collectionShape)
-                                                        .putMetadata(VARIANT_METADATA_KEY, VARIANT_ELEMENT)
-                                                        .specification(Json.writer().writeValueAsString(itemSchema))
-                                                        .addAllVariants(extractVariants(dataShape, collectionShape, VARIANT_COLLECTION))
-                                                        .build();
+                            .putMetadata(VARIANT_METADATA_KEY, VARIANT_ELEMENT)
+                            .build();
                 }
-            }
-        } else if (collectionShape.getKind().equals(DataShapeKinds.JSON_INSTANCE)) {
-            String specification = collectionShape.getSpecification();
-            List<Object> items = Json.reader().forType(List.class).readValue(specification);
-            if (!items.isEmpty()) {
-                return new DataShape.Builder().createFrom(collectionShape)
-                                                        .putMetadata(VARIANT_METADATA_KEY, VARIANT_ELEMENT)
-                                                        .specification(Json.writer().writeValueAsString(items.get(0)))
-                                                        .addAllVariants(extractVariants(dataShape, collectionShape, VARIANT_COLLECTION))
-                                                        .build();
+            } else if (collectionShape.getKind() == DataShapeKinds.JSON_INSTANCE) {
+                if (isJsonInstanceArraySpec(specification)) {
+                    List<Object> items = Json.reader().forType(List.class).readValue(specification);
+                    if (!items.isEmpty()) {
+                        return new DataShape.Builder().createFrom(collectionShape)
+                                .putMetadata(VARIANT_METADATA_KEY, VARIANT_ELEMENT)
+                                .specification(Json.writer().writeValueAsString(items.get(0)))
+                                .addAllVariants(extractVariants(dataShape, collectionShape, VARIANT_COLLECTION))
+                                .build();
+                    }
+                } else {
+                    return new DataShape.Builder().createFrom(collectionShape)
+                            .putMetadata(VARIANT_METADATA_KEY, VARIANT_ELEMENT)
+                            .build();
+                }
             }
         }
 
@@ -138,28 +154,40 @@ class AggregateMetadataHandler implements StepMetadataHandler {
         }
 
         DataShape singleElementShape = dataShape.findVariantByMeta(VARIANT_METADATA_KEY, VARIANT_ELEMENT).orElse(dataShape);
+        String specification = singleElementShape.getSpecification();
+        if (StringUtils.hasText(specification)) {
+            if (singleElementShape.getKind() == DataShapeKinds.JSON_SCHEMA) {
+                JsonSchema schema = Json.reader().forType(JsonSchema.class).readValue(specification);
 
-        if (singleElementShape.getKind().equals(DataShapeKinds.JSON_SCHEMA)) {
-            String specification = singleElementShape.getSpecification();
-            JsonSchema schema = Json.reader().forType(JsonSchema.class).readValue(specification);
+                if (schema.isArraySchema()) {
+                    return new DataShape.Builder().createFrom(singleElementShape)
+                            .putMetadata(VARIANT_METADATA_KEY, VARIANT_COLLECTION)
+                            .build();
+                }
 
-            ArraySchema collectionSchema = new ArraySchema();
-            collectionSchema.set$schema(Optional.ofNullable(schema.get$schema()).orElse(JSON_SCHEMA_ORG_SCHEMA));
-            collectionSchema.setItemsSchema(schema);
-            schema.set$schema(null);
+                ArraySchema collectionSchema = new ArraySchema();
+                collectionSchema.set$schema(Optional.ofNullable(schema.get$schema()).orElse(JSON_SCHEMA_ORG_SCHEMA));
+                collectionSchema.setItemsSchema(schema);
+                schema.set$schema(null);
 
-            return new DataShape.Builder().createFrom(singleElementShape)
-                                            .putMetadata(VARIANT_METADATA_KEY, VARIANT_COLLECTION)
-                                            .specification(Json.writer().writeValueAsString(collectionSchema))
-                                            .addAllVariants(extractVariants(dataShape, singleElementShape, VARIANT_ELEMENT))
-                                            .build();
-        } else if (singleElementShape.getKind().equals(DataShapeKinds.JSON_INSTANCE)) {
-            String specification = singleElementShape.getSpecification();
-            return new DataShape.Builder().createFrom(singleElementShape)
-                                            .putMetadata(VARIANT_METADATA_KEY, VARIANT_COLLECTION)
-                                            .specification("[" + specification + "]")
-                                            .addAllVariants(extractVariants(dataShape, singleElementShape, VARIANT_ELEMENT))
-                                            .build();
+                return new DataShape.Builder().createFrom(singleElementShape)
+                        .putMetadata(VARIANT_METADATA_KEY, VARIANT_COLLECTION)
+                        .specification(Json.writer().writeValueAsString(collectionSchema))
+                        .addAllVariants(extractVariants(dataShape, singleElementShape, VARIANT_ELEMENT))
+                        .build();
+            } else if (singleElementShape.getKind() == DataShapeKinds.JSON_INSTANCE) {
+                if (isJsonInstanceArraySpec(specification)) {
+                    return new DataShape.Builder().createFrom(singleElementShape)
+                            .putMetadata(VARIANT_METADATA_KEY, VARIANT_COLLECTION)
+                            .build();
+                } else {
+                    return new DataShape.Builder().createFrom(singleElementShape)
+                            .putMetadata(VARIANT_METADATA_KEY, VARIANT_COLLECTION)
+                            .specification("[" + specification + "]")
+                            .addAllVariants(extractVariants(dataShape, singleElementShape, VARIANT_ELEMENT))
+                            .build();
+                }
+            }
         }
 
         return dataShape;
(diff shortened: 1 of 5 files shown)


heiko-braun:
@zregvart mind a review?

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -47,34 +47,38 @@ public boolean canHandle(StepKind kind) {
...
             if (metadata.outputShape() != null) {
                 outputShape = adaptOutputShape(new DataShape.Builder()
-                                                        .createFrom(metadata.outputShape())
-                                                        .decompress()
-                                                        .build());
+                        .createFrom(metadata.outputShape())
+                        .decompress()
+                        .build());
             }
+        } catch (IOException e) {
```
zregvart: As far as I understand the only case in which IOException is thrown is if the data shape's `specification` can't be parsed as JSON or JSON Schema. I'm fine with completely blowing up rather than silently (apart from those WARN logs), as it is a violation of the contract we have. Not sure if this not resulting in an error will end up being an issue for the UI and the integration.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -47,34 +47,38 @@ public boolean canHandle(StepKind kind) {
...
                                                         .decompress()
                                                         .build());
             }
-
-            return new DynamicActionMetadata.Builder()
-                    .createFrom(metadata)
-                    .inputShape(inputShape)
-                    .outputShape(outputShape)
-                    .build();
         } catch (IOException e) {
```
zregvart: As above, let's just blow up.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -93,31 +97,42 @@ DataShape adaptInputShape(DataShape dataShape) throws IOException {
...
-            JsonSchema schema = Json.reader().forType(JsonSchema.class).readValue(specification);
-
-            if (schema.isArraySchema()) {
-                ArraySchema.Items items = ((ArraySchema) schema).getItems();
-                JsonSchema itemSchema = items.asSingleItems().getSchema();
-                itemSchema.set$schema(schema.get$schema());
-                if (items.isSingleItems()) {
+        String specification = collectionShape.getSpecification();
+        if (specification != null && specification.length() > 0) {
+            if (collectionShape.getKind().equals(DataShapeKinds.JSON_SCHEMA)) {
```
zregvart:
```suggestion
            if (collectionShape.getKind() == DataShapeKinds.JSON_SCHEMA) {
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -47,34 +47,38 @@ public boolean canHandle(StepKind kind) {
...
-                                                        .createFrom(metadata.outputShape())
-                                                        .decompress()
-                                                        .build());
+                        .createFrom(metadata.outputShape())
+                        .decompress()
+                        .build());
             }
+        } catch (IOException e) {
+            LOG.warn("Unable to read input data shape on dynamic meta data inspection", e);
+            LOG.warn("Using original input shape as fallback");
```
zregvart:
```suggestion
            LOG.warn("Using original input shape as fallback");
            throw new UncheckedIOException(e);
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -47,34 +47,38 @@ public boolean canHandle(StepKind kind) {
...
             }
-
-            return new DynamicActionMetadata.Builder()
-                    .createFrom(metadata)
-                    .inputShape(inputShape)
-                    .outputShape(outputShape)
-                    .build();
         } catch (IOException e) {
             LOG.warn("Unable to read output data shape on dynamic meta data inspection", e);
+            LOG.warn("Using original output shape as fallback");
```
zregvart:
```suggestion
            LOG.warn("Using original output shape as fallback");
            throw new UncheckedIOException(e);
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -138,28 +153,40 @@ DataShape adaptOutputShape(DataShape dataShape) throws IOException {
         }
 
         DataShape singleElementShape = dataShape.findVariantByMeta(VARIANT_METADATA_KEY, VARIANT_ELEMENT).orElse(dataShape);
+        String specification = singleElementShape.getSpecification();
+        if (specification != null && specification.length() > 0) {
+            if (singleElementShape.getKind().equals(DataShapeKinds.JSON_SCHEMA)) {
```
zregvart:
```suggestion
            if (singleElementShape.getKind() == DataShapeKinds.JSON_SCHEMA) {
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -93,31 +97,42 @@ DataShape adaptInputShape(DataShape dataShape) throws IOException {
...
+                } else {
                     return new DataShape.Builder().createFrom(collectionShape)
-                                                        .putMetadata(VARIANT_METADATA_KEY, VARIANT_ELEMENT)
-                                                        .specification(Json.writer().writeValueAsString(itemSchema))
-                                                        .addAllVariants(extractVariants(dataShape, collectionShape, VARIANT_COLLECTION))
-                                                        .build();
+                            .putMetadata(VARIANT_METADATA_KEY, VARIANT_ELEMENT)
+                            .build();
+                }
+            } else if (collectionShape.getKind().equals(DataShapeKinds.JSON_INSTANCE)) {
```
zregvart:
```suggestion
            } else if (collectionShape.getKind() == DataShapeKinds.JSON_INSTANCE) {
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -93,31 +97,42 @@ DataShape adaptInputShape(DataShape dataShape) throws IOException {
...
-            String specification = collectionShape.getSpecification();
-            JsonSchema schema = Json.reader().forType(JsonSchema.class).readValue(specification);
-
-            if (schema.isArraySchema()) {
-                ArraySchema.Items items = ((ArraySchema) schema).getItems();
-                JsonSchema itemSchema = items.asSingleItems().getSchema();
-                itemSchema.set$schema(schema.get$schema());
-                if (items.isSingleItems()) {
+        String specification = collectionShape.getSpecification();
+        if (specification != null && specification.length() > 0) {
```
zregvart:
```suggestion
        if (!StringUtils.isEmpty(specification)) {
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -138,28 +153,40 @@ DataShape adaptOutputShape(DataShape dataShape) throws IOException {
         }
 
         DataShape singleElementShape = dataShape.findVariantByMeta(VARIANT_METADATA_KEY, VARIANT_ELEMENT).orElse(dataShape);
+        String specification = singleElementShape.getSpecification();
+        if (specification != null && specification.length() > 0) {
```
zregvart:
```suggestion
        if (!StringUtils.isEmpty(specification)) {
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -47,34 +47,38 @@ public boolean canHandle(StepKind kind) {
...
 
-        return metadata;
+        return new DynamicActionMetadata.Builder()
+                .createFrom(metadata)
+                .inputShape(inputShape)
+                .outputShape(outputShape)
+                .build();
     }
 
     DataShape adaptInputShape(DataShape dataShape) throws IOException {
```
zregvart: I'd probable refactor this method into separate methods for cases where `specification` is given, `specification` is given and `kind` is `JSON_SCHEMA` and kind is `JSON_INSTANCE`.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -138,28 +153,40 @@ DataShape adaptOutputShape(DataShape dataShape) throws IOException {
...
+                collectionSchema.set$schema(Optional.ofNullable(schema.get$schema()).orElse(JSON_SCHEMA_ORG_SCHEMA));
+                collectionSchema.setItemsSchema(schema);
+                schema.set$schema(null);
+
+                return new DataShape.Builder().createFrom(singleElementShape)
+                        .putMetadata(VARIANT_METADATA_KEY, VARIANT_COLLECTION)
+                        .specification(Json.writer().writeValueAsString(collectionSchema))
+                        .addAllVariants(extractVariants(dataShape, singleElementShape, VARIANT_ELEMENT))
+                        .build();
+            } else if (singleElementShape.getKind().equals(DataShapeKinds.JSON_INSTANCE)) {
```
zregvart:
```suggestion
            } else if (singleElementShape.getKind() == DataShapeKinds.JSON_INSTANCE) {
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/SplitMetadataHandler.java
```diff
@@ -73,37 +73,54 @@ public DynamicActionMetadata handle(DynamicActionMetadata metadata) {
 
                 DataShape collectionShape = dataShape.findVariantByMeta(VARIANT_METADATA_KEY, VARIANT_COLLECTION).orElse(dataShape);
 
-                if (collectionShape.getKind().equals(DataShapeKinds.JSON_SCHEMA)) {
-                    String specification = collectionShape.getSpecification();
-                    JsonSchema schema = Json.reader().forType(JsonSchema.class).readValue(specification);
+                String specification = collectionShape.getSpecification();
+                if (specification != null && specification.length() > 0) {
```
zregvart:
```suggestion
                if (!StringUtils.isEmpty(specification)) {
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/SplitMetadataHandler.java
```diff
@@ -73,37 +73,54 @@ public DynamicActionMetadata handle(DynamicActionMetadata metadata) {
...
+                    if (collectionShape.getKind().equals(DataShapeKinds.JSON_SCHEMA)) {
```
zregvart:
```suggestion
                    if (collectionShape.getKind() == DataShapeKinds.JSON_SCHEMA) {
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/SplitMetadataHandler.java
```diff
@@ -73,37 +73,54 @@ public DynamicActionMetadata handle(DynamicActionMetadata metadata) {
...
                                     .outputShape(new DataShape.Builder().createFrom(collectionShape)
-                                                                        .putMetadata(VARIANT_METADATA_KEY, VARIANT_ELEMENT)
-                                                                        .specification(Json.writer().writeValueAsString(itemSchema))
-                                                                        .addAllVariants(extractVariants(dataShape, collectionShape, VARIANT_COLLECTION))
-                                                                        .build())
+                                            .putMetadata(VARIANT_METADATA_KEY, VARIANT_ELEMENT)
+                                            .build())
+                                    .build();
+                        }
+                    } else if (collectionShape.getKind().equals(DataShapeKinds.JSON_INSTANCE)) {
```
zregvart:
```suggestion
                    } else if (collectionShape.getKind() == DataShapeKinds.JSON_INSTANCE) {
```

app/server/endpoint/src/test/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandlerTest.java
```diff
@@ -90,22 +90,7 @@ public void shouldExtractJavaInputElementVariant() throws IOException {
...
-                                .kind(DataShapeKinds.JAVA)
-                                .specification(getSpecification("person-list-spec.json"))
-                                .collectionType("List")
-                                .type(Person.class.getName())
-                                .collectionClassName(List.class.getName())
-                                .putMetadata(StepMetadataHandler.VARIANT_METADATA_KEY, StepMetadataHandler.VARIANT_COLLECTION)
-                                .build())
-                        .addVariant(dummyShape(DataShapeKinds.JSON_INSTANCE))
-                        .build())
+                .outputShape(StepActionHandler.NO_SHAPE)
```
zregvart: We lost a test case for `JAVA` data shape kinds here, should this be a new test case for `NO_SHAPE` instead?

app/server/endpoint/src/test/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandlerTest.java
```diff
@@ -183,22 +161,7 @@ public void shouldExtractJsonSchemaInputElementVariant() throws IOException {
...
-                                .kind(DataShapeKinds.JSON_SCHEMA)
-                                .specification(getSpecification("person-list-schema.json"))
-                                .collectionType("List")
-                                .type(Person.class.getName())
-                                .collectionClassName(List.class.getName())
-                                .putMetadata(StepMetadataHandler.VARIANT_METADATA_KEY, StepMetadataHandler.VARIANT_COLLECTION)
-                                .build())
-                        .addVariant(dummyShape(DataShapeKinds.JAVA))
-                        .build())
+                .outputShape(StepActionHandler.NO_SHAPE)
```
zregvart: Perhaps add a new test case for `NO_SHAPE` and adapt the test for `JSON_SCHEMA`?

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
christophd: The initial reason for me to not blow up on that error was that the call is just a meta data lookup for data shapes. If that call is failing for the mentioned reason that we can not inspect the specification the original specification is returned and the UI should be able to continue. In most cases the integration is able to operate even if this meta data lookup happens to fail.

Maybe come back to that and think over when more lookups are about to come.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
christophd: Very true just force pushed changes, thanks

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
christophd: Very true just force pushed changes, thanks

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
christophd: Very true just force pushed changes, thanks

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
christophd: Very true just force pushed changes, thanks

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
christophd: Very true just force pushed changes, thanks

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
christophd: Very true just force pushed changes, thanks

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/SplitMetadataHandler.java
christophd: Very true just force pushed changes, thanks

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
christophd: Yeah the method is too complicated, will do on master and push in a separate PR

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/SplitMetadataHandler.java
christophd: Very true just force pushed changes, thanks

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/SplitMetadataHandler.java
christophd: Very true just force pushed changes, thanks

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
christophd: Very true just force pushed changes, thanks

app/server/endpoint/src/test/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandlerTest.java
christophd: The test was an actual duplicate with the next test variant where the exact same output shape was used and verified. I changed that so we have input and output shape tested in separate tests

app/server/endpoint/src/test/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandlerTest.java
christophd: The test was an actual duplicate with the next test variant where the exact same output shape was used and verified. I changed that so we have input and output shape tested in separate tests

christophd:
@zregvart thanks for the review. I appreciate all comments including nitpicking! 😅 

I force pushed some changes and the rest is to be done in a separate refactoring on master only as those changes will not need to be backported to 1.6.x

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
zregvart: The risk I'm thinking about is when we provide the wrong (default) data shape to the mapper and this results in an exception at runtime from the mapper.
But really, it should be a rare occurrence and to fix this properly we should have a test harness for connectors that checks the shapes and validations for user provided shapes, so I'm good either way for now.

PR #4974, opened 2019-03-19
Link pull image secret to service accounts (1.6.x backport)

### LINKED ISSUE #4900: Update Operator to link service accounts if a given secret 'syndesis-pull-secret' exists
## Minimal required

The operator should, after he installed the templates at https://github.com/syndesisio/syndesis/blob/fc7ad4c66d819759e24b3f293de9887e981eb81f/install/operator/pkg/syndesis/action/install.go#L89 link the service accounts if a given secret in the operator namespace exist.

**The secret will be called "syndesis-pull-secret".** 

## Optional enhancements 

One could go one step further to add a boolean field to the CRD like 'enablePullAuthentication' which could be used to make this behaviour more explicit. I.e. if this flag is set but the secret is missing, that an error is thrown. If this flag is not set and the secret exists, the secret is ignored.

I suggest implementing this behaviour for 7.4+.

Also one could make this kind of SA update more dynamic and e.g. monitor the CR all the time and when this flag changes in the CR then SA gets adapted. This should be implemented when the CR update functionality is implemented via #3422 .

An advantage of this post-adaption approach (create templates with SAs, modifying SAs afterwards) would be, that the template would not have to be adapted and can be directly used again on OpenShift Online without further change.

// cc @heiko-braun @avano

### LINKED ISSUE #4902: Allow for authenticated S2I builds of the integration runtimes
If pull authentication is enabled, the S2I build for the runtimes need to be able to pull the s2i builder image with authentication.

This is supposed to work with these alternative solutions:

* Add a SA to the "Build" or "BuildConfig" object, which is linked to the secret used for authentication. However it seems that this doesn't work right now.
* Add pull secret to the BuildConfig created in https://github.com/syndesisio/syndesis/blob/ba937b8043a18a90019b5b354f19ff3b10ae033b/app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java#L314-L325

The later approach seems to work (as @avano tested this), but it would require massive changes in the operator and installation scripts. We should strive to get the first solution running, as this allows to switch on authentication on the outside. The SA configured there should be also empowered by install_ocp.sh as described in #4900 .

zregvart:
Fixes #4900
Fixes #4902


astefanutti:
Thanks @zregvart

PR #4975, opened 2019-03-19
fix: #4167 Integration Details History - old version's 'last publishe…

### LINKED ISSUE #4167: Integration Details History - old version's 'last published' field updated on configuration change
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem
When editing an integration there is unexpected change of 'last published' information of the preceding version in Integration's version history.
See the screencast, where Version 3 is updated when Version 4 is created. That is IMO unexpected as Version 3 shouldn't have been updated.

There seems to be conflict in meaning of the field. Should it provide value in the sense of 'This version published on' or 'This version was published until' .... currently we're mixing the information in a single value:
-> for a new version the value holds 'published on'
-> for a replaced version the value holds 'published until'

## Screencast
published

## Expected behavior

I'd expect the value 'last published' to contain timestamp of publishing the integration (time of edit).

KurtStam:
…d' field updated on configuration change

Fixes #4167
Setting the createdAt and using that instead

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -114,6 +114,7 @@
     <json-schema-core.version>1.2.8</json-schema-core.version>
     <json-schema-validator.version>2.2.8</json-schema-validator.version>
     <junit.version>4.12</junit.version>
+    <powermock.version>2.0.0</powermock.version>
 
     <spring-security.version>4.2.6.RELEASE</spring-security.version>
     <resteasy.version>3.6.1.Final</resteasy.version>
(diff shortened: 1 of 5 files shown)


KurtStam:
Squashed my commits. Is there a need to cherry-pick to another branch?

zregvart:
> Squashed my commits. Is there a need to cherry-pick to another branch?

No need, the issue was marked as `p2` and we're backporting only `p0`s; so no.

PR #4979, opened 2019-03-19
fix: don't trigger data shape reconciliation when the mapper initializes

### LINKED ISSUE #4960: datamapper editor does not appear
## This is a...

<pre><code>
[ ] Feature request
[X] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
</code></pre>

## Description

After updating to latest master UI, I'm experiencing the following issue when doing an integration from Telegram to Telegram:

telegram2telegram

This does not happen when doing other integrations such as DB to DB. 
The same behaviour can be observed on staging.

I do not know if this applies to 1.6 too.

gashcrumb:
fixes #4960


lburgazzoli:
@gashcrumb do we need this on 1.6x too ?

zregvart:
Backport for 1.6.x in #4982

PR #5009, opened 2019-03-20
chore: enable CircleCI branch build on `master` and NNN.NNN.x branches only

zregvart:
We're getting a lot of CI churn on branches created in the origin git repository. This enables CircleCI builds only on `master` and release branches.

@rhuss can you take a look at this, I think this might work to prevent CircleCI to build branches created by the backport bot and branches for changes created via GitHub UI, or when folk don't push branches to their own forks for PR creation.


zregvart:
Merged it, we're having really long queue waits on CircleCI...

PR #5020, opened 2019-03-21
fix: #5016 Support predicates in combination with parameters in SQL select

### LINKED ISSUE #5016: Cannot use predicates other than '=' in Invoke SQL step's WHERE clause
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
When creating a PostgresDB - Invoke SQL step, the following statements work:
```
select * from todo where id = :#id
select * from todo where id between 1 and 100
select * from todo where id is null
```
but these cause an error:
```
select * from todo where id < :#id
select * from todo where id > :#id
select * from todo where id != :#id
select * from todo where id between :#id and 100
```
The error in the UI is `Invalid SQL, the number of columns (0) should match the number of number of input parameters (1). Unable to fetch and process metadata `. 

There's also an error in the developer console, but that might be due to #5015:
```"Http failure response for https://syndesis.my-minishift.syndesis.io/api/v1/connections/5/actions/sql-connector: 400 Bad Request"```

image

christophd:
This PR adds support for predicates like (!=,>,<,>=,<=,between ... and) to be used in SQL select statements in combination with parameter placeholders (:#param)

Fixes #5016

### DIFF
--- a/app/connector/sql/src/main/java/io/syndesis/connector/sql/common/SqlStatementParser.java
+++ b/app/connector/sql/src/main/java/io/syndesis/connector/sql/common/SqlStatementParser.java
@@ -205,7 +205,7 @@ public class SqlStatementParser {
 
     List<String> splitSqlStatement(String sql) {
         List<String> sqlArray = new ArrayList<>();
-        String[] segments = sql.split("=|\\,|\\s|\\(|\\)", -1);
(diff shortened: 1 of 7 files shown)


heiko-braun:
@KurtStam Mind a review?

PR #5023, opened 2019-03-21
Unable to use fetches a record from Salesforce by external id

### LINKED ISSUE #5022: Unable to use fetches a record from Salesforce by external id
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

## Description

I created a simple integration that does:

    db -> split -> datamapper -> salesforce -> log

with the aim of retrieving an object from Salesforce using a contact from the contact table (sampledb) but the integration fails for a number of reasons:

- atlasmap is not able to istantiate the object that the salesforce connector expects (SalesforceIdentifier)
- salesforce connector does not check for the message body for being of the expected type and tries to convert it to the expected type as it were a string

lburgazzoli:
Fixes #5022

### DIFF
--- a/app/connector/salesforce/src/main/java/io/syndesis/connector/salesforce/customizer/DataShapeCustomizer.java
+++ b/app/connector/salesforce/src/main/java/io/syndesis/connector/salesforce/customizer/DataShapeCustomizer.java
@@ -105,18 +105,13 @@ public class DataShapeCustomizer implements ComponentProxyCustomizer, CamelConte
                 return;
             }
 
-
-            final Object body = exchange.getIn().getBody();
-            if (type.isInstance(body)) {
-                return;
-            }
-
-            final String bodyAsString = exchange.getIn().getBody(String.class);
-            if (bodyAsString == null) {
+            final Object body = exchange.getIn().getBody(type);
+            if (body != null) {
                 return;
             }
 
             try {
+                final String bodyAsString = exchange.getIn().getBody(String.class);
                 final Object output = Json.reader().forType(type).readValue(bodyAsString);
                 exchange.getIn().setBody(output);
             } catch (final IOException e) {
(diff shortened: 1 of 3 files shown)


app/connector/salesforce/src/main/java/io/syndesis/connector/salesforce/customizer/DataShapeCustomizer.java
```diff
@@ -106,16 +105,20 @@ public void process(final Exchange exchange) {
...
-            final String bodyAsString = message.getBody(String.class);
 
+            final Object body = exchange.getIn().getBody();
+            if (type.isInstance(body)) {
+                return;
+            }
+
+            final String bodyAsString = exchange.getIn().getBody(String.class);
             if (bodyAsString == null) {
                 return;
```
valdar: Isn't this case an error? we just return?

app/connector/salesforce/src/test/java/io/syndesis/connector/salesforce/customizer/DataShapeCustomizerTest.java
```diff
@@ -142,6 +142,19 @@ public void shouldUnmarshallToSpecifiedInputType() throws Exception {
...
+        final Exchange exchange = new DefaultExchange(context);
+        final Message in = exchange.getIn();
+        in.setBody(new SalesforceIdentifier("test"));
+
+        component.getBeforeProducer().process(exchange);
+
+        Assertions.assertThat(in.getBody()).isInstanceOf(SalesforceIdentifier.class);
+        Assertions.assertThat(in.getBody()).hasFieldOrPropertyWithValue("id", "test");
+    }
+
```
valdar: Shouldn't we have 2 tests for the other 2 if?

lburgazzoli:
@valdar should be fixed now

PR #5025, opened 2019-03-21
feat: add a 'ui' command to 'syndesis'

### LINKED ISSUE #5006: Add ui commands to `syndesis` wrapper scripts
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description
Be good to finally bake in the various UI tasks we currently run directly with `yarn` into the `syndesis` script.  Especially if we can get a standard set of commands that would also be applicable for the react port.

@zregvart @christophd any good examples I could work from you could point me to would be appreciated :-)

gashcrumb:
fixes #5006

Also adds proxy support thanks to @riccardo-forina's work in the react repo so by default `--serve` will just run against staging.

I think the `--local` option should work as-is @zregvart lemme know if that needs adjusting.


zregvart:
I think this doesn’t change how local mode works. Not sure about the nuke option, though if it’s helpful for folks I have no objections really...

Issue #5017, opened 2019-03-21
Unable to process a collection POSTed to an API Provider


asmigala:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
When implementing an API which accepts an array of a particular type, it's impossible to process that data in a meaningful way in the API Provider integration, because a split step splits the collection items themselves into (key: value) pairs. 

This is caused by #4862 (and you can find more details there), reporting as a separate issue because it makes it impossible to implement some APIs in API Provider.

heiko-braun:
>  it's impossible to process that data in a meaningful way in the API Provider integration

@asmigala Can you outline possible scenarios how you wold like to process collections? One or two example would help others to understand what you try to achieve.

asmigala:
@heiko-braun the simplest example using our todo API is being able to post a bunch of tasks and have them inserted into a database. The idea is that you would create an operation that accepts Array of Task (this is possible in apicurito), then in the flow, add a split step to get each of the Tasks. After that, an Invoke SQL with `insert into todo (id, task, complete) values (:#id, :#task, :#completed)`.

The problem is that after the split, what I'm getting is not e.g. `{"id": 1, "task": "task1", "completed": 0}`, but rather `{"id": 1}` and then (a separate invocation of the SQL step) `{"task":"task1"}` etc.

heiko-braun:
@asmigala Thanks for the explanation, I now see what you mean.

asmigala:
The issue is still present in 1.6.4.

I'm now getting 
```
org.apache.camel.RuntimeExchangeException: Cannot find key [id] in message body or headers to use when setting named parameter in query [INSERT INTO todo (id, completed, task) VALUES (:?id, :?completed, :?task)] on the exchange: Exchange[i-Lb3ggNjv4h_QK4HUhkFz]
```

even though I have the mapping set up.

Attaching a sample integration.

[TODO Integration-export.zip](https://github.com/syndesisio/syndesis/files/3018427/TODO.Integration-export.zip)

christophd:
@asmigala can you please double check the mapping by going back to the mapper and open the mapping in the editor? I have had this error the other day and it turned out that my mapping was incomplete (I also was sure to have had done the mapping before but it was incomplete)

asmigala:
@christophd 
[...]

PR #5034, opened 2019-03-22
fix: #5017 Enable split step to handle unified Json body

christophd:
API provider, webhook and custom API connectors use a unified Json schema to represent both request parameters and request body. So the request body is not stored as a root element in that Json schema but as a nested property called "body" (see also [unified json docs](https://github.com/syndesisio/syndesis/blob/master/doc/connecting/topics/r_about-json-schema-for-http-requests.adoc))

When this body is a Json array the split step has to split the elements in that body property rather than the root json element.

This PR introduces a new meta data information marking data shapes to follow the unified json schema. The unified meta data is handled in both split step handler and split meta data lookup so the split step automatically performs the split on the json array located in that body property. 

Also the dynamic split meta data shape lookup handles these unified json shapes when converting collection shapes to single element shapes. So the split data shape is provided with the single element shape of the json array in that body property.


app/common/model/src/main/java/io/syndesis/common/model/DataShape.java
```diff
@@ -100,19 +100,16 @@
...
     class Builder extends ImmutableDataShape.Builder {
 
-        public static final String COMPRESSION_METADATA_KEY = "compression";
-        private static final String COMPRESSED_METADATA_KEY = "compressed";
-
         public Builder decompress() {
             DataShape currentShape = build();
-            if ("true".equals(currentShape.getMetadata().get(COMPRESSION_METADATA_KEY)) &&
-                    currentShape.getMetadata().getOrDefault(COMPRESSED_METADATA_KEY, "true").equals("true")) {
+            if ("true".equals(currentShape.getMetadata().get(DataShapeMetaData.COMPRESSION)) &&
```
zregvart:
```suggestion
            final Map<String, String> currentMetadata = currentShape.getMetadata();
            if (Boolean.parseBoolean(currentMetadata.get(DataShapeMetaData.COMPRESSION) &&
```

app/common/model/src/main/java/io/syndesis/common/model/DataShape.java
```diff
@@ -100,19 +100,16 @@
...
+                    currentShape.getMetadata().getOrDefault(DataShapeMetaData.COMPRESSED, "true").equals("true")) {
```
zregvart:
```suggestion
                    Boolean.parseBoolean(currentMetadata().get(DataShapeMetaData.COMPRESSED)) {
```

zregvart: Why do we have `COMPRESSED` and `COMPRESSION` metadata?

app/common/model/src/main/java/io/syndesis/common/model/DataShape.java
```diff
@@ -124,7 +121,7 @@ public Builder decompress() {
                 } catch (IOException e) {
                     throw new IllegalStateException("Failed to decompress data shape", e);
                 }
-            } else if (currentShape.getVariants().stream().anyMatch(variant -> "true".equals(variant.getMetadata().get(COMPRESSION_METADATA_KEY)))) {
+            } else if (currentShape.getVariants().stream().anyMatch(variant -> "true".equals(variant.getMetadata().get(DataShapeMetaData.COMPRESSION)))) {
```
zregvart:
```suggestion
            } else if (currentShape.getVariants().stream().anyMatch(variant -> Boolean.parseBoolean(variant.getMetadata().get(DataShapeMetaData.COMPRESSION)))) {
```

Though I would probably pull this in a helper method like `isCompressed(DataShape dataShape)`

app/common/model/src/main/java/io/syndesis/common/model/DataShape.java
```diff
@@ -141,16 +138,16 @@ public Builder decompress() {
 
         public Builder compress() {
             DataShape currentShape = build();
-            if ("true".equals(currentShape.getMetadata().get(COMPRESSION_METADATA_KEY)) &&
-                    currentShape.getMetadata().getOrDefault(COMPRESSED_METADATA_KEY, "false").equals("false")) {
+            if ("true".equals(currentShape.getMetadata().get(DataShapeMetaData.COMPRESSION)) &&
```
zregvart:
```suggestion
            if (Boolean.parseBoolean(currentMetadata.get(DataShapeMetaData.COMPRESSION)) &&
```

app/common/model/src/main/java/io/syndesis/common/model/DataShape.java
```diff
@@ -141,16 +138,16 @@ public Builder decompress() {
...
+                    currentShape.getMetadata().getOrDefault(DataShapeMetaData.COMPRESSED, "false").equals("false")) {
```
zregvart:
```suggestion
                    !Boolean.parseBoolean(currentMetadata().get(DataShapeMetaData.COMPRESSED)) {
```

app/common/model/src/main/java/io/syndesis/common/model/DataShape.java
```diff
@@ -162,7 +159,7 @@ public Builder compress() {
                 } catch (IOException e) {
                     throw new IllegalStateException("Failed to compress data shape", e);
                 }
-            } else if (currentShape.getVariants().parallelStream().anyMatch(variant -> "true".equals(variant.getMetadata().get(COMPRESSION_METADATA_KEY)))) {
+            } else if (currentShape.getVariants().parallelStream().anyMatch(variant -> "true".equals(variant.getMetadata().get(DataShapeMetaData.COMPRESSION)))) {
```
zregvart:
```suggestion
            } else if (currentShape.getVariants().stream().anyMatch(variant -> Boolean.parseBoolean(variant.getMetadata().get(DataShapeMetaData.COMPRESSION)))) {
```

`parallelStream` should be slower on a small number of elements.

app/common/model/src/main/java/io/syndesis/common/model/DataShapeMetaData.java
```diff
@@ -0,0 +1,39 @@
...
+public final class DataShapeMetaData {
+
+    public static final String VARIANT = "variant";
+    public static final String VARIANT_COLLECTION = "collection";
+    public static final String VARIANT_ELEMENT = "element";
+
+    public static final String UNIFIED = "unified";
+
+    public static final String COMPRESSION = "compression";
+    public static final String COMPRESSED = "compressed";
```
zregvart: At some point we might want to consider using enums for metadata keys. These are bound to end up mixed up.

app/common/util/src/main/java/io/syndesis/common/util/json/JsonUtils.java
```diff
@@ -0,0 +1,111 @@
...
+        return jsonBeans;
+    }
+
+    /**
+     * Converts list of json object strings to a json array string.
+     * @param jsonBeans
+     * @return
+     */
+    public static String jsonBeansToArray(List<?> jsonBeans) {
+        return "[" + jsonBeans.stream().map(Object::toString).collect(Collectors.joining(",")) + "]";
```
zregvart:
```suggestion
        final StringJoiner joiner = new StringJoiner(",", "[", "]");
        for (Object jsonBean : jsonBeans) {
            joiner.add(String.valueOf(jsonBean));
        }
        return joiner.toString();
```

Not sure if `jsonBeans` can contain `null`, this prevents NPE, should perform better, not sure how often this method will be called or how many entries there will be in the `jsonBeans` List.

app/common/model/src/main/java/io/syndesis/common/model/DataShape.java
christophd: COMPRESSION marks that the shape should be compressed. COMPRESSED marks that the shape already has been compressed in order to avoid duplicate compression.

Let me rename COMPRESSED to IS_COMPRESSED.

app/common/model/src/main/java/io/syndesis/common/model/DataShape.java
christophd: Yeah that `isCompressed()` helper is a good idea

app/common/util/src/main/java/io/syndesis/common/util/json/JsonUtils.java
christophd: 👍

PR #5074, opened 2019-03-27
chore: upgrade Spring Boot to 1.5.17.RELEASE

zregvart:
This is the version used in 2.21 version of the Fuse fork[1].

[1] https://repository.jboss.org/nexus/content/groups/ea/org/apache/camel/camel-parent/2.21.0.fuse-730072/camel-parent-2.21.0.fuse-730072.pom


zregvart:
@KurtStam it makes sense that we're aligned here, just double check my assumption.

cc @heiko-braun @apodhrad

PR #5085, opened 2019-03-29
rolling back spring-boot version to 1.5.16.RELEASE

KurtStam:
fix for https://github.com/syndesisio/syndesis/issues/5086


heiko-braun:
@zregvart what about 1.6.x? Is it impacted too?

zregvart:
> @zregvart what about 1.6.x? Is it impacted too?

@heiko-braun this was only on `master`, I'm digging into the root cause as Fuse Standalone on Spring Boot and Fuse on OpenShift seem to be using Spring Boot 1.5.17 (and I guess we should align).

PR #5087, opened 2019-03-29
fix: #5017 Open api formats to be compliant with json schema

christophd:
If we generate json schemas from Open API specifications in API provider we have to make sure to not use unsupported data formats (e.g. int64, int32, byte, float, etc.) in those schemas. This will lead to parsing errors when the schema is read.

This is because Open API introduces data formats that are not supported in Json Schema.

I have added a sanitize step for schemas generated from Open API specifications to not include the unsupported data format information.


app/server/api-generator/src/main/java/io/syndesis/server/api/generator/swagger/util/JsonSchemaHelper.java
```diff
@@ -127,6 +132,34 @@ public static String serializeJson(final JsonNode schemaNode) {
...
+
+        if (node.isArray()) {
+            ArrayNode arrayNode = (ArrayNode) node;
+            Iterator<JsonNode> elements = arrayNode.elements();
+            elements.forEachRemaining(JsonSchemaHelper::sanitize);
+        } else if (node.isObject()) {
+            ObjectNode objectNode = (ObjectNode) node;
+
+            JsonNode formatNode = node.get("format");
+            if (formatNode != null &&
```
zregvart: Perhaps using that `isKnownFormat` helper method.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/swagger/util/JsonSchemaHelper.java
```diff
@@ -127,6 +132,34 @@ public static String serializeJson(final JsonNode schemaNode) {
         }
     }
 
+    public static JsonNode sanitize(final JsonNode node) {
+        if (node == null) {
+            return node;
```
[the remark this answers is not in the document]
christophd: Agreed!

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/swagger/util/JsonSchemaHelper.java
christophd: Thanks, good idea!

christophd:
@zregvart as usual good ideas while nitpicking, many thanks! pushed changes to this PR

PR #5088, opened 2019-03-29
Bump oauth:spring-security-oauth2 test dependency

jonathanchristison:
Make sure 2.0.16.RELEASE is used


christophd:
closing as dependency management is required, see PR #5089 instead

PR #5089, opened 2019-03-29
chore: Ensure spring security oauth2 2.0.16

christophd:
Make sure 2.0.16.RELEASE is used as managed version 2.0.15 is having security issues


christophd:
@jonathanchristison do you need this to be part of 7.3?

jonathanchristison:
@christophd As it's a test dependency I'll be re-scoring this as low, i'd also guess the risk is minimal for the change to go into 7.3

PR #5101, opened 2019-04-02
chore: release speed up

zregvart:
This adds `--skip-tests` and enables by default strict checksum policy (can be disabled with `--no-strict-checksums`.
This should speed up the release if run against the same local repository and with `--skip-tests`.


zregvart:
I'm going to merge this, to test it on CI.

Issue #4466, opened 2019-02-07
SQL Connector: Return generated keys for INSERT statement


jcordes73:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
The SQL connector should return generated keys for INSERT statements, so that these can also be mapped during integration.

Example:

RESTFul PUT statement where an ID is expected as a return parameter and data is actually stored in a PostgreSQL instance.

zregvart:
Currently the only way of doing that is to wrap the `INSERT` statement in a stored procedure.
A while back we had the idea of replacing the SQL parser logic we have in the SQL connector with the use of [`PreparedStatement::getMetadata`](https://docs.oracle.com/javase/7/docs/api/java/sql/PreparedStatement.html#getMetaData()) for generating output data shape and [`PreparedStatement::getParameterMetaData`](https://docs.oracle.com/javase/7/docs/api/java/sql/PreparedStatement.html#getParameterMetaData()) for generating input data shape.

This would allow statements such as:
`INSERT INTO TODO (TASK) VALUES (?) RETURNING ID`

To generate data shapes with the data type of `TASK` as input and `ID` as output.

KurtStam:
Code complete but I can't test end2end b/c stuff is broken at the moment

https://github.com/KurtStam/syndesis/commit/6e0d7895cc5028f2ccba6b8ab27a07803587f715

zregvart:
@KurtStam thanks, though I really think we should replace the SQL parser we have with `getMetadata` and `getParameterMetaData`, i.e. do no parsing on our own. With parsing we'll always be one step behind or we won't support some use cases. With `getMetadata` and `getParameterMetaData` we put all that logic to the one place it is maintained - to the driver and the database's own SQL parser.

The code in https://github.com/KurtStam/syndesis/commit/6e0d7895cc5028f2ccba6b8ab27a07803587f715 relies on a specific/subset database implementation, some databases have a `SERIAL` type, `IDENTITY` or `AUTO_INCREMENT` attributes, some schemas may a `SEQUENCE` for generating keys or triggers. (see http://troels.arvin.dk/db/rdbms/#mix-identity).

KurtStam:
@zregvart 

1. The query isn't actually parsable by the database if the query contains camel paramaters. We're already trying to use the db metadata as much as possible. 

[...]

PR #5133, opened 2019-04-03
fix: one more thing! SQL Connector: Return generated keys for INSERT

KurtStam:
SQL Connector: Return generated keys for INSERT statement #4466


KurtStam:
@christophd thx for noticing the defect. Should be all set now.

KurtStam:
let me redo this.

PR #5138, opened 2019-04-03
Added public API OAuth proxy variables for productization

dhirajsb:


dhirajsb:
@zregvart @heiko-braun I added these variables for productization.

dhirajsb:
@cunningt @djcoleman productization pipeline needs to set these to the same values as at https://github.com/syndesisio/syndesis/blob/1.6.x/install/generator/syndesis-template.go#L120-L149. 
At present the image seems to be set to 'registry.fuse-ignite.openshift.com/openshift/oauth-proxy:v1.1.0.

heiko-braun:
@dhirajsb  i am unclear what you ask for. The oauth proxy version is mangwd externally here: https://github.com/syndesisio/fuse-online-install/blob/master/import_images.pl#L66

dhirajsb:
@heiko-braun Syndesis mustache templates are modified by `syndesis-template.go` script to use productized images. 
But the public OAuth proxy template is standalone and needs to be modified for prod build, where these variables should be replaced to appropriate values.

zregvart:
I guess we need another Syndesis release to pick this up, am I correct?

heiko-braun:
@zregvart no, i think the build from this afternoon is fine. I have talked to @dhirajsb and what he said in https://github.com/syndesisio/syndesis/pull/5138#issuecomment-479900357 doesn't not need to be done. Is this correct @dhirajsb ?

Issue #5036, opened 2019-03-22
Update migration scripts


rhuss:
## This is a...

<pre><code>
[X] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

Currently, the update mechanism allows executing scripts which are stored in a directory named as the target version. However, this approach doesn't work on both, upstream and product because they are using different version schemes.

Since this mechanism is fragile, it should be removed, and update migrations scripts are supposed to be idempotent and the sense they can detect whether they have been already run, and if so, do nothing if running a second time. The directory `migration/resources/any` is reserved for this.

See also #4413 for the original discussion.

Issue #3464, opened 2018-08-28
[Upgrade] Upgrade s2i image stream tag automatically?


avano:
## This is a...

<pre><code>
[X] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

Currently the upgrade scripts do not upgrade configmaps by default: https://github.com/syndesisio/syndesis/blob/master/tools/upgrade/steps/upgrade_40_update_resources#L45-L46

Now when I deploy `1.5.4-20180827` for example, the s2i image stream is defined in `syndesis-server-config` configmap:
https://github.com/syndesisio/syndesis/blob/1.5.4-20180827/install/syndesis.yml#L257

When I upgrade to 1.5.4-20180828, this value in the config map is not changed (because the upgrade script does not change config maps by default). In current situation we should provide the migration script for the config map (we do for prod builds now) for each daily release to upgrade this one single value.

It is just my thought to upgrade it automatically using the upgrade script, or get rid of the value in config map and let the default value be the same tag as syndesis server? No idea if that is possible to do :-)

cc @rhuss @zregvart

## Tasks involved / Steps to Reproduce

1.
2.
3.
4.

rhuss:
I completely agree that we either should update this version by default on every update or think about a solution to not reference that image from the configuration (e.g. by using an environment variable in the deployment config of syndesis-server). 

I think the later is the better solution as one could consider the s2i image to use as part of the installation at not something a user should be able to change in a configmap.

@zregvart @avano ok to move the information to an environment variable in the template ?

avano:
thank you @rhuss ! :-)

rhuss:
Sorry no progress yet, and I can't pick it up in the 7.3 time frame.  Happy for any help, I think the comments on https://github.com/syndesisio/syndesis/issues/3464#issuecomment-418264599 are still valid.

avano:
done in https://github.com/syndesisio/syndesis/pull/5161

PR #5161, opened 2019-04-08
[Upgrade] Use only idempotent scripts + update s2i

avano:
* Removed old scripts that are not related to current (7.4) version
* As all the scripts should be idempotent now, so "all" directory is not
necessary anymore
* Add script that automatically updated s2i is tag based on is tags of
other syndesis components

Solves #5036 #3464


heiko-braun:
Thanks for the contribution @avano ! Is this something you had in mind for `1.6.x`?

avano:
@heiko-braun currently, for each release we needed a migration script to bump the s2i image stream tag with the hardcoded tag - this was done for 1.6.x in https://github.com/syndesisio/syndesis/pull/5083 , so this PR is not needed for 1.6.x, as it is just a enhancement for the future. After this PR we won't need to create the script manually with the hardcoded version, as it is automatically parsed from the template in the upgrade image

avano:
@zregvart thanks, I tweaked the doc for the command a little bit.

> not sure if we'll remember to properly manage, i.e delete old, create new upgrade scripts

basically it can be done in any part of the release, it can be part of bumping the syndesis version, it can be done before creating a tag for the productization and in worst case they can stay, as they shouldn't do any harm - I can take of this

> create either a test or a pre-commit hook that can make sure that we don't forget to create a migration script when changing the templates

if the tests in syndesis-qe are enough, we have them already for a long time :stuck_out_tongue:

Issue #5194, opened 2019-04-12
Provisioning error on openshift online


heiko-braun:
When we try to use the **template based install** (read no operators) on openshift online, it fails.

The installation command:
```
Running system command: oc new-app --template=fuse-ignite/fuse-ignite-1.6 -p ROUTE_HOSTNAME=<URL> -p OPENSHIFT_MASTER=https://api.online-stg.openshift.com -p OPENSHIFT_CONSOLE_URL=https://console.online-stg.openshift.com/console -p OPENSHIFT_PROJECT=fuseproj11926 -p OPENSHIFT_OAUTH_CLIENT_SECRET=$(oc sa get-token syndesis-oauth-client -n fuseproj11926) -p MAX_INTEGRATIONS_PER_USER=5 -p SAR_PROJECT=fuseproj11926 -n fuseproj11926
```

The error:
```
[...]

PR #5199, opened 2019-04-12
Related to #5194: add marker for camel k resources

nicolaferraro:
@heiko-braun if we're going to rebuild syndesis, we may add some markers so we can automatically remove the camel k stuff from the openshift online templates.


astefanutti:
Thanks to yaml comment!

heiko-braun:
@nicolaferraro good idea

PR #5202, opened 2019-04-12
[camel-k] move SyndesisHeaderStrategy to integration/runtime

### LINKED ISSUE #5189: [camel-k] SyndesisHeaderStrategy not loaded into the registry
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

## Description

Some connectors requires an HeaderStrategy named syndesisHeaderStrategy to be present in the registry and it was done using spring boot before but that does not work any more in camel-k so we need to implement a similar strategy in camel-k otherwhise integration using such connector won't even start

We missed to replicate what https://github.com/syndesisio/syndesis/blob/master/app/connector/support/processor/src/main/java/io/syndesis/connector/support/processor/SyndesisHttpConfiguration.java does so we could hit `No bean could be found in the registry for: syndesisHeaderStrategy of type: org.apache.camel.spi.HeaderFilterStrategy` in case we use i.e. the http component.

lburgazzoli:
Fixes  #5189


tplevko:
Should this be backported to 1.6.x?

lburgazzoli:
> Should this be backported to 1.6.x?

yes, but need to do it manually as it requires other commits.

PR #5208, opened 2019-04-12
[camel-k] SyndesisHeaderStrategy not loaded into the registry (1.6.x)

lburgazzoli:


app/connector/http/pom.xml
```diff
@@ -125,12 +125,7 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>io.syndesis.integration</groupId>
-      <artifactId>integration-runtime</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.hibernate</groupId>
+      <groupId>org.hibernate.validator</groupId>
```
zregvart: This is part of the Hibernate validator upgrade, this should not be on `1.6.x` (#5143)

app/server/controller/src/main/java/io/syndesis/server/controller/ControllersConfigurationProperties.java
```diff
@@ -29,6 +35,9 @@
     private int integrationStateCheckInterval = 60;
     private boolean exposeVia3scale;
 
+    @NestedConfigurationProperty
+    private CamelK camelk = new CamelK();
```
zregvart: Are this bits intended in this PR? The title is about `SyndesisHeaderStrategy` and they add a bit more than that.

heiko-braun:
Thanks for your feedback @zregvart. I have added the `wip` label, awaiting @lburgazzoli 's response

heiko-braun:
@lburgazzoli Feel free to remove `wip`, once @zregvart's questions are clarified. Without that, I am reluctant to bring it into 1.6.x

app/connector/http/pom.xml
lburgazzoli: good catch

app/server/controller/src/main/java/io/syndesis/server/controller/ControllersConfigurationProperties.java
lburgazzoli: yes it is, as customizers are loaded using a property we may need this to tweak what customizers are loaded at runtime, now the header strategy customizer is included by default but it was not in the first iteration

lburgazzoli:
@zregvart @heiko-braun PR updated

PR #5211, opened 2019-04-12
chore: Enhance google-sheets meta data with custom column names

christophd:
This PR adds a meta data lookup for Google Sheets connector in order to provide custom column names. By default column names are simply named with their column letter (A,B,C,D). With the meta data lookup you can fetch a header row that contains custom column names. You have to specify the row number of the header row.

Bildschirmfoto 2019-04-12 um 19 56 41

Next property definition step fetches the column names.

Bildschirmfoto 2019-04-12 um 19 56 59

So you can use those names in data mapper which is more comfortable than just using A,B,C,D. See following source document using the custom column names vs. the target using the default column names.

Bildschirmfoto 2019-04-12 um 20 02 43

Users can also skip fetching the header row and customize the names manually or just use the default names.

### DIFF
--- a/app/connector/google-sheets/src/main/java/io/syndesis/connector/sheets/GoogleSheetsConnector.java
+++ b/app/connector/google-sheets/src/main/java/io/syndesis/connector/sheets/GoogleSheetsConnector.java
@@ -16,29 +16,20 @@
 
 package io.syndesis.connector.sheets;
 
-import java.io.ByteArrayInputStream;
-import java.util.Base64;
 import java.util.Map;
 import java.util.Optional;
 
-import com.google.api.client.http.javanet.NetHttpTransport;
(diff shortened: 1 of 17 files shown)


christophd:
@zregvart haha, that has been some almost-ready stash that did not make it to the last release

christophd:
@TovaCohen do you prefer an extra issue for adjusting the changes in the connector docs (for v7.4)? Should be a small change to add the new fields "headerRow" and "columnNames"

TovaCohen:
Hi @christophd, I always need a FUSEDOC Jira to do the work, so I created this one: 

https://issues.jboss.org/browse/FUSEDOC-3129

which points to this issue. I'll be in touch when I do the work. Thanks for letting me know about this!

PR #5229, opened 2019-04-15
connector: move SyndesisHttpConfiguration from connector-support to integration-runtime-sb

lburgazzoli:


lburgazzoli:
@zregvart do you mind having a look ?

this is a follow up of https://github.com/syndesisio/syndesis/pull/5202 so SyndesisHttpConfiguration is moved to integration runtime sb as it make much sense to have it there

PR #5230, opened 2019-04-15
Add a proxy profile for doing yarn build behind a firewall, add repository.xml

cunningt:
We need the ability to be able to run the yarn build through a proxy.    I've added two profiles, proxy and noproxy, to app/ui/pom.xml.      

Have also added a profile to app/s2i for the inclusion of a necessary repository.xml and the license plugin.

### DIFF
--- a/app/s2i/pom.xml
+++ b/app/s2i/pom.xml
@@ -155,6 +155,43 @@
 
   <profiles>
     <profile>
+      <id>addrepository</id>
+      <activation>
+        <property>
+          <name>addrepository</name>
+        </property>
+      </activation>
+      <build>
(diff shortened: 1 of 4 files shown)


zregvart:
I've created a PR for your PR, was easier for me, see: https://github.com/cunningt/syndesis/pull/1

zregvart:
Screenshot_2019-04-16 Yo Dawg Heard You Meme Generator - Imgflip

cunningt:
@zregvart merged your commits in, I'll test it out in this form and if we end up needing the license-maven-plugin definition I'll submit a separate PR for that.

PR #5242, opened 2019-04-16
ENTESB-9863 Upgrade jackson-databind version to 2.8.11.3 in integration-bom

cunningt:
spring-boot-dependencies brings in an old version of jackson-databind, we should make sure it brings in 2.8.11.3


cunningt:
Might be good to cherry pick this to 1.6.x in case we do a 7.3.1

Issue #5181, opened 2019-04-10
Bulk operation(s) for updating descriptors and metadata for steps


gashcrumb:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
Currently when changes are made to the flow in the editor it fetches metadata and descriptors for steps *per step* that needs it individually.  I think a bulk operation that the editor could submit the entire step array would certainly perform better, with the single requests the editor can get more laggy the larger the integration is.

For reference these are the spots where these calls are made currently when changes to the flow are made:

The invocations
https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/edit-page/current-flow.service.ts#L740

https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/edit-page/current-flow.service.ts#L763

And these are triggered from [here](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/edit-page/current-flow.service.ts#L646) and [here](https://github.com/syndesisio/syndesis/blob/master/app/ui/src/app/integration/edit-page/current-flow.service.ts#L696).  It's also complicates the code a bit to have all these requests made while scanning through the step array.

@christophd FYI this'd be *really* nice for 7.4 I think.

christophd:
@gashcrumb yes I think this is a good idea. added my name to it

gashcrumb:
Think this is all set, isn't it?  We can close this.

mastepan:
@gashcrumb please let the issues open for the QE to be able to track them and verify. Thanks

PR #5252, opened 2019-04-18
#5181 Add bulk operation for step descriptor lookup

christophd:
This PR provides a bulk operation for UI to lookup step descriptors with updated data shapes (e.g. adapted shapes for split/aggregate). The backend operation receives the complete list of steps in an integration and performs descriptor lookup for each step. The result is a list of steps with maybe updated data shapes in action descriptors.

Each step descriptor lookup is provided with a dynamic meta data that is created by the respective step meta data handler depending on the step kind.

This way we optimize the number of requests sent form UI frontend to the server backend and improve the robustness of the UI editor.

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
@@ -26,6 +26,7 @@ import io.syndesis.common.model.DataShape;
 import io.syndesis.common.model.DataShapeKinds;
 import io.syndesis.common.model.DataShapeMetaData;
 import io.syndesis.common.model.connection.DynamicActionMetadata;
+import io.syndesis.common.model.integration.Step;
 import io.syndesis.common.model.integration.StepKind;
 import io.syndesis.common.util.Json;
 import io.syndesis.common.util.json.JsonUtils;
@@ -46,6 +47,26 @@ class AggregateMetadataHandler implements StepMetadataHandler {
     @Override
     public boolean canHandle(StepKind kind) {
         return StepKind.aggregate.equals(kind);
+    }
+
+    @Override
+    public DynamicActionMetadata createMetadata(Step step, List<Step> previousSteps, List<Step> subsequentSteps) {
+        DataShape inputShape = subsequentSteps.stream()
+                                            .filter(StepMetadataHelper::hasInputDataShape)
+                                            .findFirst()
+                                            .map(StepMetadataHelper::getInputDataShape)
+                                            .orElse(StepMetadataHelper.NO_SHAPE);
+
+        DataShape outputShape = previousSteps.stream()
+                                            .filter(StepMetadataHelper::hasOutputDataShape)
+                                            .reduce((first, second) -> second)
+                                            .map(StepMetadataHelper::getOutputDataShape)
+                                            .orElse(StepMetadataHelper.NO_SHAPE);
+
+        return new DynamicActionMetadata.Builder()
+                        .inputShape(inputShape)
+                        .outputShape(outputShape)
+                        .build();
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -48,6 +49,26 @@ public boolean canHandle(StepKind kind) {
         return StepKind.aggregate.equals(kind);
     }
 
+    @Override
+    public DynamicActionMetadata createMetadata(Step step, List<Step> previousSteps, List<Step> subsequentSteps) {
+        DataShape inputShape = subsequentSteps.stream()
+                                            .filter(StepMetadataHelper::hasInputDataShape)
+                                            .findFirst()
+                                            .map(StepMetadataHelper::getInputDataShape)
+                                            .orElse(StepMetadataHelper.NO_SHAPE);
```
zregvart: Do you think it would make sense to add these methods to `DataShape`, something like `static Optional<DataShape> findFirstInputIn(List<WithDataShape> items)` (introducting `WithDataShape`). I kinda like when APIs are readable, makes the intent clearer.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/StepMetadataHelper.java
```diff
@@ -0,0 +1,135 @@
...
+final class StepMetadataHelper {
+
+    static final DataShape ANY_SHAPE = new DataShape.Builder().kind(DataShapeKinds.ANY).build();
+    static final DataShape NO_SHAPE = new DataShape.Builder().kind(DataShapeKinds.NONE).build();
+
+    /**
+     * Prevent instantiation of utility class.
+     */
+    private StepMetadataHelper() {
+        super();
```
zregvart: I never understood why IDEs insist on calling `super()`

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/StepMetadataHelper.java
```diff
@@ -0,0 +1,135 @@
...
+import io.syndesis.common.model.DataShapeKinds;
+import io.syndesis.common.model.DataShapeMetaData;
+import io.syndesis.common.model.action.Action;
+import io.syndesis.common.model.action.ActionDescriptor;
+import io.syndesis.common.model.integration.Step;
+
+/**
+ * @author Christoph Deppisch
+ */
+final class StepMetadataHelper {
```
zregvart: I like the helper class, though I think this could be an interface and then mixed into the `DataShape` interface.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/StepMetadataHelper.java
```diff
@@ -0,0 +1,135 @@
...
+        return !getOutputDataShape(step).getKind().equals(DataShapeKinds.NONE);
+    }
+
+    /**
+     * Safely get the input data shape of given step. If shape is not present this method returns
+     * a dummy data shape of kind {@link DataShapeKinds#NONE}.
+     * @param step
+     * @return
+     */
+    static DataShape getInputDataShape(Step step) {
```
zregvart: These methods could be on the `WithDataShapes` interface thinggy, and `Step` could pull them in that way, so we'd have `Step::inputDataShape()` method or something.

If you decide to go that way, make sure that you place `@Value.Auxiliary` or `@Value.Derived` and `@JsonIgnore` annotations on those methods.

zregvart: Hmm, just saw that `Step` indeed has a `inputDataShape` and that I added it, could it be used instead of this method?

christophd:
@zregvart Thanks for your comments. I agree to having the helpers on the domain model making it part of the API and more readable and maintainable. I will have a look and provide a separate follow up PR on this. Thanks!

PR #5254, opened 2019-04-18
Fixes #5157 remove remaining spring-boot dependency from connectors

### LINKED ISSUE #5157: spring-boot : remove remaining sb dependency from connectors
## This is a...

<pre><code>
[X] Feature request
[ ] Bug report  
</code></pre>

## Description

Connectors still  have some spring boot dependencies  mostly because of meta and verifier service. 

We should get rid of them and use camel's factory finder which is already supported by meta [2][3] but not implemented by connector:

[1] https://github.com/syndesisio/syndesis/blob/master/app/meta/src/main/java/io/syndesis/connector/meta/v1/VerifierEndpoint.java#L72-L75
[2] https://github.com/syndesisio/syndesis/blob/master/app/meta/src/main/java/io/syndesis/connector/meta/v1/ConnectorEndpoint.java#L66-L69

rkorytkowski:
Aside from tests, there's one remaining spring-boot dependency at https://github.com/syndesisio/syndesis/blob/master/app/connector/webhook/src/main/java/io/syndesis/connector/webhook/WebhookServletAutoConfiguration.java, any idea how to get rid of that?

I did not test all modified connectors from UI, but tested some random ones and things seem to be working as expected.


lburgazzoli:
> Aside from tests, there's one remaining spring-boot dependency at https://github.com/syndesisio/syndesis/blob/master/app/connector/webhook/src/main/java/io/syndesis/connector/webhook/WebhookServletAutoConfiguration.java, any idea how to get rid of that?
> 

I think it can be moved to `integration/runtime-springboot` and activated with a condition on the presence of a web-hook connector related class, i.e. `WebhookConnectorCustomizer`

rkorytkowski:
Thanks @oscerd and @lburgazzoli. I'll do that.

rkorytkowski:
Addressed review comments. Tested webhook integration after changes and it worked.

app/server/runtime/src/test/java/io/syndesis/server/runtime/ExtensionsITCase.java
```diff
@@ -346,7 +346,7 @@ public void testListExtensionDetails() throws IOException {
                     .build())
                 .build());
 
-        await().atMost(30, TimeUnit.SECONDS).pollInterval(250, TimeUnit.MILLISECONDS).untilAsserted(() -> {
+        await().atMost(60, TimeUnit.SECONDS).pollInterval(250, TimeUnit.MILLISECONDS).untilAsserted(() -> {
```
zregvart: This is really annoying I need to fix that.

app/pom.xml
```diff
@@ -638,6 +638,7 @@
               <exclude>**/*.webapp</exclude>
               <exclude>**/browserconfig.xml</exclude>
               <exclude>connector/**/META-INF/syndesis/connector/verifier/*</exclude>
+              <exclude>connector/**/META-INF/syndesis/connector/meta/*</exclude>
```
zregvart: Bit confusing as you did add license headers to some of these files. We should also take a look at why this wasn't caught by the license check we do on CI. I think it might be because if the files there not having an extension.

PR #5258, opened 2019-04-19
fix(build docs): Build: specify --camel-snapshot without an explicit …

### LINKED ISSUE #3424: Build: specify --camel-snapshot without an explicit environment variable doesn't work
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## The problem
If we use the --camel-snapshot option from build and we specify a version, this is not taken into account because of the code here: https://github.com/syndesisio/syndesis/blob/master/app/s2i/copy_mvn_repo.sh#L17

so we need to fix the documentation or find a way to pass the option to copy_mvn_repo.sh

## Expected behavior
camel-snapshot option should work also without an explicit env variable set

oscerd:
…environment variable doesn't work

Fixes #3424

### DIFF
--- a/tools/bin/commands/build
+++ b/tools/bin/commands/build
@@ -31,8 +31,8 @@ build::usage() {
 
         --resume <m1>              Resume build from maven module (mvn -rf m1)
         --batch-mode               Run mvn in batch mode
-        --camel-snapshot <version> Run a build with a specific Camel snapshot. If no argument is given the
-                                   environment variable CAMEL_SNAPSHOT_VERSION is used or an error thrown.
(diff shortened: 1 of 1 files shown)


oscerd:
The failing test seems to be related to the odata connector and the refactoring by @rkorytkowski 

/workspace/app/connector/odata/src/test/java/io/syndesis/connector/odata/verifier/ODataVerifierTest.java:[87,33] cannot find symbol
  symbol:   class ODataVerifierAutoConfiguration
  location: class io.syndesis.connector.odata.verifier.ODataVerifierTest

PR #5270, opened 2019-04-23
fix(5057): Correct determination of when to show Next or Done buttons

phantomjinx:
* Incorporates check for what step is being configured and differentiate
  between input and output shapelessness since
  a) input shape doesn't matter to start step
  b) output shape doesn't matter to finish step


phantomjinx:
@gashcrumb 
WDYT of this as possible solution?

gashcrumb:
@phantomjinx if you fancy updating this document it'd be much appreciated :1st_place_medal: , I've added a highlight comment with a link to this PR if you don't have time so I can remember to do it later.

Issue #5193, opened 2019-04-12
[conditional-flows] Add CBR step handler in integration-runtime


christophd:
Introduce new step and step handler in integration runtime. Integration route builder must be able to handle content-based-router step with multiple flows (choice-when).

tplevko:
@mcada can this be closed?

PR #5295, opened 2019-05-02
feat: Add choice step meta data handler

christophd:
Content based router (=choice) step adapts input/output data shapes from previous and subsequent steps during step descriptor meta data lookup.

Also moved some utility methods to step model class

Belongs to #5193

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
@@ -52,11 +52,11 @@ class AggregateMetadataHandler implements StepMetadataHandler {
     @Override
     public DynamicActionMetadata createMetadata(Step step, List<Step> previousSteps, List<Step> subsequentSteps) {
         DataShape inputShape = StepMetadataHelper.getFirstWithInputShape(subsequentSteps)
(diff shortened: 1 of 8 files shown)


app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/meta/AggregateMetadataHandler.java
```diff
@@ -51,16 +51,12 @@ public boolean canHandle(StepKind kind) {
...
+        DataShape inputShape = StepMetadataHelper.getFirstWithInputShape(subsequentSteps)
+                                            .map(s -> s.inputDataShape().orElse(StepMetadataHelper.NO_SHAPE))
                                             .orElse(StepMetadataHelper.NO_SHAPE);
 
-        DataShape outputShape = previousSteps.stream()
-                                            .filter(StepMetadataHelper::hasOutputDataShape)
-                                            .reduce((first, second) -> second)
-                                            .map(StepMetadataHelper::getOutputDataShape)
+        DataShape outputShape = StepMetadataHelper.getLastWithOutputShape(previousSteps)
+                                            .map(s -> s.outputDataShape().orElse(StepMetadataHelper.NO_SHAPE))
```
lburgazzoli: maybe use flatMap

christophd: yes, that is a good idea thx

PR #5329, opened 2019-05-08
feat(4986): Adds a generic email connector - IMAP(s)/POP3(s)/SMTP(s)

phantomjinx:
* Since camel-mail divides up component into 3 component-defn, a proxy
  'email' defn is used for the sake of avoiding duplication.

* 2 connection types - Receive:IMAP/POP3 (Consumer) & Send:SMTP (Producer)

* Connection types have 'secure' option for enabling SSL

* EMailComponent
 * Handles fetch component defn based on protocol from underlying camel-mail
 * Create MailConfiguration based on connection configuration

* EMailReceiveCustomizer
 * Converts MimeMessage into EMailMessageModel
 * Parses email text to ensure it's semi-readable (maybe further work
   required if requested)

* EMailSendCustomizer
 * Configures email for sending
 * Allows for either input-parameters or consumed-data to be preferred
   choice for certain fields. Example, useful for overwriting From-field
   with single email address

* EMailMetaData[...]
 * Serves primarily as checks that the connection being offered for mail
   transport is the correct one for the route being configured

* EMailMessageModel
 * Differs from GMail connector model by including a FROM field which needs
   to be valid email address rather than default camel@localhost
  * SMTP servers have domain checking validation that ensures FROM is not
    spoofed.

* catalog/.../email-[receive|send]/.json
 * Catalog placeholder definitions required for valid 'email' component defn
   but are superseded by the real protocol-based component defns in
   EMailComponent

* ComponentProxyComponent
 * Makes definition getter protected so sub-classes can override it; allows
   for EMailComponent to delegate defn to protocol-defined versions.

* Lots n lots of tests!
 * EMailTestServer using Greenmail

* OData refactoring
 * Moves PropertyBuilder to connector-support/util
 * Moves general certificate functions to connector-support/util

### DIFF
--- a/app/connector/email/pom.xml
+++ b/app/connector/email/pom.xml
@@ -1,0 +1,245 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+   Copyright (C) 2016 Red Hat, Inc.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>io.syndesis.connector</groupId>
+    <artifactId>connector-parent</artifactId>
+    <version>1.7-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>connector-email</artifactId>
+  <name>Connector :: EMail</name>
+  <packaging>jar</packaging>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.apache.camel</groupId>
+        <artifactId>camel-mail</artifactId>
+        <version>${camel.version}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>org.mortbay.jetty</groupId>
+            <artifactId>servlet-api</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava-jdk5</artifactId>
+          </exclusion>
+        </exclusions>
+      </dependency>
+      <dependency>
+        <groupId>org.jsoup</groupId>
+        <artifactId>jsoup</artifactId>
+        <version>1.11.3</version>
+    </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.syndesis.connector</groupId>
+      <artifactId>connector-support-util</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>io.syndesis.common</groupId>
+      <artifactId>common-model</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.syndesis.common</groupId>
+      <artifactId>common-util</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.syndesis.integration</groupId>
+      <artifactId>integration-component-proxy</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>connector-support-verifier</artifactId>
+      <version>${project.version}</version>
+      <optional>true</optional>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.camel</groupId>
+      <artifactId>camel-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.camel</groupId>
+      <artifactId>camel-mail</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.sun.mail</groupId>
+      <artifactId>javax.mail</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jsoup</groupId>
+      <artifactId>jsoup</artifactId>
+    </dependency>
+
+    <!-- test -->
+    <dependency>
+      <groupId>io.syndesis.integration</groupId>
+      <artifactId>integration-runtime</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.github.spotbugs</groupId>
+      <artifactId>spotbugs-annotations</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.glassfish</groupId>
+      <artifactId>javax.el</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-validator</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.camel</groupId>
+      <artifactId>camel-spring-boot</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.camel</groupId>
+      <artifactId>camel-spring</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-beans</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-context</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-autoconfigure</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.icegreen</groupId>
+      <artifactId>greenmail</artifactId>
+      <version>1.5.9</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.camel</groupId>
+      <artifactId>camel-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <resources>
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>true</filtering>
+      </resource>
+    </resources>
+
+    <plugins>
+      <plugin>
+        <artifactId>maven-resources-plugin</artifactId>
+        <configuration>
+          <delimiters>
+            <delimiter>@</delimiter>
+          </delimiters>
+        </configuration>
+      </plugin>
+
+      <plugin>
+        <groupId>io.syndesis.connector</groupId>
+        <artifactId>connector-support-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>inspections</id>
+            <goals>
+              <goal>generate-connector-inspections</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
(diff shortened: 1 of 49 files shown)


oscerd:
Wasn't it better to create a PR for the odata test separated from the mail connector?

phantomjinx:
> Wasn't it better to create a PR for the odata test separated from the mail connector?

The odata changes are made to allow me to utilise the PropertyBuilder and ssl-related bits n pieces in the email tests.

app/connector/email/src/main/java/io/syndesis/connector/email/component/EMailComponent.java
```diff
@@ -0,0 +1,236 @@
...
+    public int getMaxResults() {
+        return maxResults;
+    }
+
+    public void setMaxResults(int maxResults) {
+        this.maxResults = maxResults;
+    }
+
+    private Map<String, Object> bundleOptions() {
+        PropertyBuilder<Object> builder = new PropertyBuilder<Object>();
```
zregvart: Might eliminate the need for `PropertyBuilder` if `Map::compute` was used:

```java
Map<String, Object> options = new HashMap<>();
options.compute(PROTOCOL, val(this::getProtocol));
options.compute(HOST, val(this::getHost));

return options;
```

Where `val` is:
```java
private static <T, U, R> BiFunction<T, U, R> val(final Supplier<R> fn) {
    return (t, u) -> fn.get();
}
```

If we want to be fancy that is :smile:

app/connector/email/src/main/java/io/syndesis/connector/email/meta/EMailMetaDataRetrieval.java
```diff
@@ -0,0 +1,66 @@
...
+        if (actionId.endsWith(EMailFunction.READ.connectorId())) {
+            //
+            // Read action NOT applicable to SMTP connectors
+            //
+            if (protocol.isProducer()) {
+                String msg = "The protocol for the selection connection is set to '" + protocol.id() +
+                    "'. This is not appropriate for a consuming email connector. Please amend the connector and try again.";
+                throw new IllegalStateException(msg);
+            }
+        } else if(actionId.endsWith(EMailFunction.SEND.connectorId())) {
```
zregvart:
```suggestion
        } else if (actionId.endsWith(EMailFunction.SEND.connectorId())) {
```

app/connector/email/src/main/java/io/syndesis/connector/email/meta/EMailMetaDataRetrieval.java
```diff
@@ -0,0 +1,66 @@
...
+            //
+            // Send action is only applicable to SMTP connectors
+            //
+            if (protocol.isReceiver()) {
+                String msg = "The protocol of the selected connection is set to '" + protocol.id() +
+                    "'. Only 'smtp' is appropriate for a producing email connector. Please amend the connector and try again.";
+                throw new IllegalStateException(msg);
```
zregvart:
```suggestion
                throw new IllegalArgumentException(msg);
```

might be more appropriate from what I gather in this context

app/connector/email/src/main/java/io/syndesis/connector/email/meta/EMailMetaDataRetrieval.java
```diff
@@ -0,0 +1,66 @@
...
+        Protocols protocol = emailMetadata.getProtocol();
+
+        if (actionId.endsWith(EMailFunction.READ.connectorId())) {
+            //
+            // Read action NOT applicable to SMTP connectors
+            //
+            if (protocol.isProducer()) {
+                String msg = "The protocol for the selection connection is set to '" + protocol.id() +
+                    "'. This is not appropriate for a consuming email connector. Please amend the connector and try again.";
+                throw new IllegalStateException(msg);
```
zregvart:
```suggestion
                throw new IllegalArgumentException(msg);
```

might be more appropriate from what I gather in this context

app/connector/email/src/main/java/io/syndesis/connector/email/meta/EMailMetadata.java
```diff
@@ -0,0 +1,32 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.connector.email.meta;
+
+import io.syndesis.connector.email.EMailConstants;
+
+public class EMailMetadata implements EMailConstants {
```
zregvart:
```suggestion
public class EMailMetadata {
```

nit: doesn't seem to be used

app/connector/email/src/main/java/io/syndesis/connector/email/model/EMailMessageModel.java
```diff
@@ -0,0 +1,165 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.syndesis.connector.email.model;
+
+import io.syndesis.connector.email.EMailConstants;
+
+public class EMailMessageModel implements EMailConstants{
```
zregvart:
```suggestion
public class EMailMessageModel implements EMailConstants {
```

app/connector/email/src/test/java/io/syndesis/connector/email/AbstractEMailTest.java
```diff
@@ -0,0 +1,162 @@
...
+        public PropertiesParser propertiesParser(PropertyResolver propertyResolver) {
+            return new DefaultPropertiesParser() {
+                @Override
+                public String parseProperty(String key, String value, Properties properties) {
+                    return propertyResolver.getProperty(key);
+                }
+            };
+        }
+
+        @Bean(destroyMethod = "")
```
zregvart:
```suggestion
        @Bean 
```

probably not needed

app/connector/email/src/test/java/io/syndesis/connector/email/consumer/EMailReadRouteTest.java
```diff
@@ -0,0 +1,427 @@
...
+        "logging.level.io.syndesis.integration.runtime = DEBUG"
+    }
+)
+@TestExecutionListeners(
+    listeners = {
+        DependencyInjectionTestExecutionListener.class,
+        DirtiesContextTestExecutionListener.class
+    }
+)
+@SuppressFBWarnings
```
zregvart: :crying_cat_face:

app/connector/email/src/test/java/io/syndesis/connector/email/consumer/EMailReadRouteTest.java
```diff
@@ -0,0 +1,427 @@
...
+public class EMailReadRouteTest extends AbstractEMailRouteTest {
+
+    private EMailTestServer server;
+
+    public EMailReadRouteTest() throws Exception {
```
zregvart: needed, just calls `super()` ?

app/connector/email/src/test/java/io/syndesis/connector/email/server/EMailTestServer.java
```diff
@@ -0,0 +1,225 @@
...
+        } else {
+            greenMail = new GreenMail(type);
+        }
+
+        server = supplier.get();
+    }
+
+    public void start() {
+        if (greenMail == null) {
+            throw new IllegalStateException();
```
zregvart:
```suggestion
            throw new IllegalStateException("Green mail server not initialized");
```

app/connector/email/src/test/java/io/syndesis/connector/email/verifier/EMailVerifierTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+import java.util.Map;
+import org.junit.Test;
+import io.syndesis.connector.email.AbstractEMailTest;
+import io.syndesis.connector.support.verifier.api.Verifier;
+import io.syndesis.connector.support.verifier.api.VerifierResponse;
+
+public class EMailVerifierTest extends AbstractEMailTest {
+
+    public EMailVerifierTest() throws Exception {
```
zregvart: needed?

app/connector/support/util/src/main/java/io/syndesis/connector/support/util/KeyStoreHelper.java
```diff
@@ -78,4 +91,24 @@ private static String generatePassword() {
         return new String(passwordChars, 0, passwordChars.length);
     }
 
+    public static KeyStore defaultKeyStore()
```
zregvart: To make this code portable and to take into account any [customizations configured](https://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/JSSERefGuide.html#Customization) for the runtime.

I'd do something like:

```java
final KeyStore defaultKeystore = KeyStore.getInstance(KeyStore.getDefaultType());
defaultKeystore.load(null);

final TrustManagerFactory factory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
factory.init((KeyStore) null);

for (final TrustManager manager : factory.getTrustManagers()) {
    final X509TrustManager x509Manager = (X509TrustManager) manager;

    final X509Certificate[] issuers = x509Manager.getAcceptedIssuers();
    for (final X509Certificate issuer : issuers) {
        final String alias = issuer.getSerialNumber().toString();
        final TrustedCertificateEntry entry = new TrustedCertificateEntry(issuer);
        defaultKeystore.setEntry(alias, entry, null);
    }
}
```

and remove `CertificateUtil::defaultKeyStore` method.

app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/IntegrationUpdateHandler.java
```diff
@@ -319,7 +319,7 @@ private Connection includeConnector(Connection connection) {
                             dataManager.update(dbConnection);
                         }
 
-                        if (connection.getId().isPresent()) {
+                        if (connection.getId().isPresent() && ! integration.isDeleted()) {
```
zregvart: Put this change in another pull request.

app/integration/component-proxy/src/main/java/io/syndesis/integration/component/proxy/ComponentProxyComponent.java
```diff
@@ -359,7 +369,7 @@ protected void configureDelegateEndpoint(ComponentDefinition definition, Endpoin
         // Extract options from options that are supposed to be set at the endpoint
         // level, those options can be overridden and extended using by the query
         // parameters.
-        Collection<String> endpointProperties = definition.getEndpointProperties().keySet();
+        Collection<String> endpointProperties =getDefinition().getEndpointProperties().keySet();
```
zregvart:
```suggestion
        Collection<String> endpointProperties = getDefinition().getEndpointProperties().keySet();
```

app/connector/support/util/src/main/java/io/syndesis/connector/support/util/KeyStoreHelper.java
```diff
@@ -32,6 +40,11 @@
  */
 public class KeyStoreHelper {
 
+    /**
+     * The password for the java runtime's default keystore
+     */
+    public static final String DEFAULT_KEYSTORE_PASSWD = "changeit";
```
zregvart: This will blow up as soon as someone uses custom `cacerts` keystore

app/connector/support/util/src/main/java/io/syndesis/connector/support/util/KeyStoreHelper.java
```diff
@@ -32,6 +40,11 @@
...
+
     private static final SecureRandom SECURE_RANDOM = new SecureRandom();
```
zregvart: This optimization is not needed, usages of this field should be replaced with `java.util.concurrent.ThreadLocalRandom`

zregvart:
Thanks `pure-bot` really helpful

phantomjinx:
> Thanks `pure-bot` really helpful

Hey @zregvart,
Thanks for the thorough review. Since its already merged, I'll go through your suggestions and create a follow-on PR.

app/connector/email/pom.xml
```diff
@@ -0,0 +1,245 @@
...
+    <!-- test -->
+    <dependency>
+      <groupId>io.syndesis.integration</groupId>
+      <artifactId>integration-runtime</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.github.spotbugs</groupId>
+      <artifactId>spotbugs-annotations</artifactId>
+      <scope>test</scope>
```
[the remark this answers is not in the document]
phantomjinx: Seems to be little pattern to this in the connectors. For example
- [fhir](https://github.com/syndesisio/syndesis/blob/master/app/connector/fhir/pom.xml#L112) uses provided scope
- [aws-s3](https://github.com/syndesisio/syndesis/blob/master/app/connector/aws-s3/pom.xml#L91) & [telegram](https://github.com/syndesisio/syndesis/blob/master/app/connector/telegram/pom.xml#L36) have no scope

None of these seem to present any compile errors so will update to 'provided'

app/connector/email/src/main/java/io/syndesis/connector/email/EMailConstants.java
```diff
@@ -0,0 +1,147 @@
...
+     */
+    String TEXT_HTML = "text/html";
+    String TEXT_PLAIN = "text/plain";
+
+    enum Protocols {
+        IMAP, IMAPS,
+        POP3, POP3S,
+        SMTP, SMTPS;
+
+        public static Protocols getValueOf(String name) {
```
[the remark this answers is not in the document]
phantomjinx: Habit to protect against throwing an exception and returning a default value. However, given these method should always be given a viable name parameter then should be ok to change.

app/connector/email/src/main/java/io/syndesis/connector/email/EMailConstants.java
```diff
@@ -0,0 +1,147 @@
...
+        Priorities(String id) {
+            this.id = id;
+        }
+
+        public static Priorities getValueOf(String name) {
+            if (name == null) {
+                return CONSUMED_DATA;
+            }
+
+            for (Priorities priority : Priorities.values()) {
```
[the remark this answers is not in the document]
phantomjinx: Ditto.

app/connector/email/src/main/java/io/syndesis/connector/email/customizer/EMailReceiveCustomizer.java
```diff
@@ -0,0 +1,131 @@
...
+            mail.setCc(in.getHeader(MAIL_CC, String.class));
+        }
+        if (ObjectHelper.isNotEmpty(in.getHeader(MAIL_BCC))) {
+            mail.setBcc(in.getHeader(MAIL_BCC, String.class));
+        }
+        exchange.getIn().setBody(mail);
+    }
+
+    private String getPlainTextFromMultipart(Multipart multipart)  throws MessagingException, IOException{
+        StringBuilder result = new StringBuilder();
```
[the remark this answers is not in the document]
phantomjinx: That's true if the email chooses to provide the same text in plaintext & html. However, a lot of email I've looked at puts a "please view this email in a web browser" or "this message requires an html reader" type message in the plaintext part.
The plaintext if-statement includes a break to avoid duplication so will add the same to the html if-statement.

app/connector/email/src/main/java/io/syndesis/connector/email/meta/EMailMetadata.java
phantomjinx: Protocol comes from EMailConstants

app/connector/email/src/test/java/io/syndesis/connector/email/consumer/EMailReadRouteTest.java
phantomjinx: Was due to a 'throws Exception" in super constructor. Now removed.

app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/IntegrationUpdateHandler.java
phantomjinx: Sneaked in ... bit late to change now.

app/connector/support/util/src/main/java/io/syndesis/connector/support/util/KeyStoreHelper.java
phantomjinx: That's really cool and a lot more robust. Thx!

PR #5335, opened 2019-05-10
chore: upgrade (Fuse) Camel to 2.21.0.fuse-740010

zregvart:


oscerd:
This should fix #5304

Issue #5214, opened 2019-04-15
[conditional-flows] Base UI adjustments for CBR


christophd:
- A CBR step could be inserted at any point in an existing flow (other than start connection)
- Users can configure one-to-many matching conditions (simple language expressions like in advanced filter) 
- Each condition links to a single sub-flow
- Users can optionally add a single default flow that automatically matches when no other content based sub-flow matches (aka "otherwise" branch). This links to a sub-flow that has no condition.
- Users can navigate to respective sub-flows and back to parent flow
  - The flow the user navigates to is loaded into the editor, replacing the parent flow display
- A CBR created flow has a special start and finish connection, users can not change the configuration properties or data shapes of these connections. The user would have to go to the parent integration and apply changes there
- UI indicates the amount of CBR flows on the parent flow (at the CBR step as overlay icon?)
- We auto-name flows created by the CBR (e.g. based on the rules or via key generator)
- We mark parent flows and sub-flows via meta data (e.g. "primary" or "main" flow vs. "sub-flow")
- A quick selector (e.g. drop down) would be nice as well, perhaps that quick selector could indicate flows that start with flow connectors so that the user can easily distinguish between the "main" flow and these kind of "internal" sub-flows

christophd:
The backend server part works with the `stepKind='choice'` and the following configured step properties:
-  `flows`: Json array of flow configurations holding the condition expression and the flow identifier (e.g. auto generated key or name)
- `default`: a default flow identifier that should be used when no other condition matches (optional)
- `routingScheme`: Camel scheme to use when delegating to flows (optional, default=`direct`)

The `flows` configured property is a Json array String representing one-to-many conditional flow mappings as Json objects that define the condition as simple expression language (like in advanced filter) and the flow name/id.

```json
[
  {"condition": "${body} contains 'Hello'", "flow": "hello-flow"},
  {"condition": "${body} contains 'Bye'", "flow": "bye-flow"}
]
```

This is copied from how basic filter adds the condition rules to the step as configured property.

Besides using a simple expression language the step can also use rule based approach like in basic filter:

```json
[
  {"path": "text", "op": "contains", "value": "Hello", "flow": "hello-flow"},
  {"path": "text", "op": "contains", "value": "Bye", "flow": "bye-flow"}
]
```

Where `text` is a property on the message body and results in `${body.text}`.

PR #5339, opened 2019-05-10
[feat] UI flow navigation

christophd:
Ref #5214

When conditional flows introduce additional sub-flows to the integration the flow toolbar now provides a navigation as dropdown. Users can select the available flows in the dropdown in order to open the flow for editing.

Bildschirmfoto 2019-05-10 um 16 27 33

When a sub-flow is opened the toolbar shows the flow condition.

Bildschirmfoto 2019-05-10 um 16 27 58

Also in this PR the step overlay notification indicating the number of flows this step introduces to the integration.

Bildschirmfoto 2019-05-10 um 16 28 29


oscerd:
Love it!

PR #5345, opened 2019-05-13
Knative connector: initial implementation

nicolaferraro:
Tasks to complete it:

- [x] Porting implementation
- [x] Removing 0.3 compat steps
- [x] Testing


app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGeneratorHelper.java
```diff
@@ -89,6 +90,11 @@ public static boolean filterDefaultDependencies(MavenGav gav) {
             answer = false;
         }
 
+        if ("org.apache.camel.k".equals(gav.getGroupId())) {
+            // Camel K does not work with s2i runtime
+            answer = false;
```
valdar: not sure what this should do? Since knative is not working without camel-k (i.e. using old plain s2i) should we throw errors or maybe handle it at other levels?

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/KubernetesSupport.java
```diff
@@ -66,9 +73,17 @@ public KubernetesSupport(KubernetesClient client) {
     protected void watchLog(String podName, Consumer<InputStream> handler, String sinceTime, Executor executor) throws IOException {
```
valdar: is this used somewhere else where the changes can alter current behavior?

app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGeneratorHelper.java
nicolaferraro: The S2I image builder takes all connectors and embed them into the image, but it fails with camel-k runtime libs because version is not in the standard s2i BOM. Instead of adding camel k to the s2i BOM, I thought it was better to exclude camel k libs from the S2I image (as you pointed out, they are not needed).

app/server/logging/jsondb/src/main/java/io/syndesis/server/logging/jsondb/controller/KubernetesSupport.java
nicolaferraro: This should add a rest call to get the pod details, but not alter the behavior. In fact, if you try to stream the logs from a pod with multiple containers, without specifying the container name, you get http error 4xx.

I've left the standard behavior when the number of containers is <= 1.

KubernetesSupport is used only in ActivityTracking, so to stream logs from integration pods. Integration pods in standard s2i mode are composed of a single container, so nothing should change.

nicolaferraro: I stream the logs of the first container also because sidecars are usually appended at the end of the list.

valdar: Cool thanks.

app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGeneratorHelper.java
valdar: I see, makes perfect sense.

app/integration/runtime-camelk/pom.xml
```diff
@@ -76,6 +76,12 @@
       <artifactId>camel-k-runtime-core</artifactId>
       <version>${camel-k-runtime.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.camel.k</groupId>
+      <artifactId>camel-k-adapter-camel-2</artifactId>
+      <version>${camel-k-runtime.version}</version>
+      <scope>runtime</scope>
```
valdar: this should be in scope compile and you should exclude it from checks like:
```     <plugin>
        <artifactId>maven-dependency-plugin</artifactId>
        <configuration>
          <ignoredUnusedDeclaredDependencies>
            <!-- false positive, doesn't compile without it -->
            <ignoredUnusedDeclaredDependency>org.apache.camel.k:camel-k-adapter-camel-2</ignoredUnusedDeclaredDependency>
          </ignoredUnusedDeclaredDependencies>
        </configuration>
      </plugin>```

**Comment by nicolaferraro** [comment]
Not a problem.. If I set scope to `runtime`, compilation fails on CI (but seems to work on my machine). If I set `compile` the plugin complains because it should be `runtime`.

**Comment by oscerd** [comment]
> Not a problem.. If I set scope to `runtime`, compilation fails on CI (but seems to work on my machine). If I set `compile` the plugin complains because it should be `runtime`.

Deadlock.

**Comment by valdar** [comment]
@nicolaferraro I had the same problem yesterday and after a quick assessment with @lburgazzoli we decided it was plugin check fault and exclude the artifact from checks leaving the scope to compile (i.e. remove the <scope> tag)

File: app/integration/runtime-camelk/pom.xml (hunk 3 above)
Comment (zregvart) [reply app/integration/runtime-camelk/pom.xml]: Having it in scope runtime works on my end, and `grep -r 'org.apache.camel.k.adapter'` doesn't give me anything in `src/main/java`.

@valdar why do you think it needs to be in compile scope?

Comment (zregvart) [reply app/integration/runtime-camelk/pom.xml]: This bit from the [CI build](https://circleci.com/gh/syndesisio/syndesis/71486)?

```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /workspace/app/integration/runtime-camelk/src/main/java/io/syndesis/integration/runtime/camelk/logging/IntegrationLoggingContextCustomizer.java:[44,24] cannot access org.apache.camel.k.adapter.Registry
  class file for org.apache.camel.k.adapter.Registry not found
[ERROR] /workspace/app/integration/runtime-camelk/src/main/java/io/syndesis/integration/runtime/camelk/logging/IntegrationLoggingContextCustomizer.java:[45,24] cannot find symbol
  symbol:   method bind(java.lang.String,io.syndesis.integration.runtime.logging.BodyLogger.Default)
  location: variable runtimeRegistry of type org.apache.camel.k.Runtime.Registry
[ERROR] /workspace/app/integration/runtime-camelk/src/main/java/io/syndesis/integration/runtime/camelk/logging/IntegrationLoggingContextCustomizer.java:[47,24] cannot find symbol
  symbol:   method bind(java.lang.String,io.syndesis.integration.runtime.logging.ActivityTrackingInterceptStrategy)
  location: variable runtimeRegistry of type org.apache.camel.k.Runtime.Registry
```

Hmm, odd.

zregvart: I think the issue is that the `org.apache.camel.k.adapter.Registry` version the `org.apache.camel.k.Runtime` is linked to is not available on classpath.

I don't think you should do stuff like this:

https://github.com/apache/camel-k-runtime/blob/3e9d8f4943a7f89328aec43f9d37cfc59de467cc/camel-k-runtime-core/pom.xml#L120-L153

In `camel-k`, perhaps a `Registry` should be an interface placed in `camel-k-runtime-core` and have two implementations in `camel-k-adapter-camel-2` and `camel-k-adapter-camel-3`. Would that cause circular dependencies?

valdar: you have a good point there, we must investigate further; should we merge this and rise another issue? Since might take time if a change in camel-k is needed and need to be released. wdyt @zregvart ?

nicolaferraro: Agree, we should put that stuff in a common layer.

lburgazzoli: it is a little bit more complicated than adding a common layer as the registry interface has changed in camel 3, so either we create yet another registry interface or we just wait to love to camel 3

lburgazzoli: it is a little bit more complicated than adding a common layer as the registry interface has changed in camel 3, so either we create yet another registry interface or we just wait to love to camel 3

valdar: never wait to love :P

Issue #4203, opened 2018-12-04
Data Virtualisation in Syndesis


heiko-braun:
The bulk of the work (Teiid part) is tracked in `DINT`: https://red.ht/2G26b1q 

Use this issue to track all syndesis local changes needed to integrate data virt.

rareddy:
@zregvart I have the changes related to bringing in the Data Integration engine into syndesis build are here on this branch [1]. Can you help further refine to suit syndesis if needs any? otherwise, I can also submit a PR then go from there?

[1] https://github.com/rareddy/syndesis/tree/teiid-syndesis

zregvart:
@rareddy the bits in on that branch look good to me, open a PR so we can get it merged, one thing I think would make sense is to have those bits behind a feature toggle, so we users can decide if they wish to deploy them or not...

rareddy:
@zregvart I wanted to add Data Integration to include with a --include an option, but I failed to figure out, as I think this needs to be done in the Operator code, any suggestions?

The next question is, where the target image for the Data Integration needs to be hosted, right now I have it Teiid docker hub, is that good? or should we consider something under "syndesis" in your opinion?

gaughan:
We need the image in the product. Ill start a thread. Once I get image details I can kick off and image request.

zregvart:
> @zregvart I wanted to add Data Integration to include with a --include an option, but I failed to figure out, as I think this needs to be done in the Operator code, any suggestions?

`--include` option in the `tools/bin/syndesis` script or in the template generation stuff in `syndesis/install/generator`?

I think it would be best if we used some sort of service discovery mechanism, I'm assuming that the Komodo server will be running as a separate pod, then from Syndesis operator we can discover that it exists (by reacting on a label for example) and include the necessary configuration.

Let me rope in @astefanutti, @nicolaferraro and @lgarciaaco as our resident Operator gurus.

In the template based installation we can add more code to the template generator to include/exclude it (via a flag I think would be the easiest).

We need to rope in @djcoleman, @cunningt and @orpiske to refine this for the product builds.

> The next question is, where the target image for the Data Integration needs to be hosted, right now I have it Teiid docker hub, is that good? or should we consider something under "syndesis" in your opinion?

I think hosting it on Docker Hub is OK for upstream, as @gaughan noted we need another image build in the product, that will have it's own coordinates.

zregvart:
[...]

PR #5358, opened 2019-05-14
#4203: adding the data integration image to the syndesis build

rareddy:


zregvart:
This looks good to me on a technical level, just needs to add changes from the generator -- to include the changed output YAML files.

I'm trying to think of ways this can backfire, and perhaps the only two things would be: the day-to-day of the developer and if we need to increase the default 4GB memory setting for the minishift (via `syndesis minishift --install`) and any memory limit issues we might encounter on the trial cluster.

I can't really tell how big of an issue those two are, care to chip in @rhuss @heiko-braun @gaughan ?

Also heads up @djcoleman @orpiske and @cunningt this will change the Syndesis template and introduce a new image stream.

heiko-braun:
@rareddy How is `IncludeDint` passed into the build? Is it an optional flag to the outermost build scripts? 

On a related note: Please although think through and supply the changes need to the product installation scripts: https://github.com/syndesisio/fuse-online-install

rareddy:
@heiko-braun These are either called manually or using pre-commit hook per @zregvart  See the comment on #4203 

As per the size, I put in size for 1GB requirement for this image, I am sure this will push the limits if they are set very conservatively.

gaughan:
Mini shift and eval cluster might be an issue. Both are short term tho with Openshift 4.1 and intergreatly trial accounts on the horizon. It might be nice to have pods scale when not in use tho

heiko-braun:
Please hold thi suntil we have clarified if we need templates at all for data virt

Issue #5299, opened 2019-05-02
Create an AWS-SNS connector


oscerd:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
As a citizen user of fuse online, I want to be able to connect to an AWS-SNS connector
I want to be able to provide my connection url, authentication details and any additional details such as cache settings.
DoD
I can create an integration with a standard timer writing to an SNS.
I can connect to an AWS-SNS and consume messages
I want to be able to provide a filter policy to restrict the notifications I receive.
I want to be able to use AWS-KMS for encrypt/decrypt

--There is an initial draft of the user doc for this feature.
--The writer acknowledges receiving enough information from engineering to write the initial draft of the user doc for this feature.

heiko-braun:
@gaughan Can you provide a user story and acceptance criteria for this one?

oscerd:
Please, note that there is no consumer on the aws-sns component. This is because of this discussion https://issues.apache.org/jira/browse/CAMEL-3476

oscerd:
In introduced the bridge between aws-sns topic and aws-sqs to be able to consume from sns in Camel. I think we can try to use that, but it won't be completely free (we need to backport the functionality on 2.21.x)

gaughan:
@heiko-braun updated above

oscerd:
To consume you'll need to pass through an SQS connection. This will only need to be set on the aws console. There is no consumer in the component.. because the Java sdk provides the feature of bridging SQS to SNS. It makes sense to provide docs on how to do this on the console and consuming from an SQS endpoint. For KMS I have to check.

oscerd:
For KMS, you need to enable it when you create the topic, so it makes sense to leave this out. If you need to use KMS, you can do that through the AWS console, as for consuming through the SQS queue. For a citizen integrator, creating the bridge, or setup the KMS master key to use on the topic created, seems to be too much. I will show how to do this on a demo and we can document everything.

oscerd:
For the filter policy, you can use the a camel filter as we do for SQS, since you'll use an SQS queue to consume.

oscerd:
I think I have enough material to create the connector which will have only a to action.

oscerd:
Ok, I tested the situation. The connector will have only the possibility to send something to topic.
The encryption can be enabled during the topic creation on the aws console.
[...]

PR #5370, opened 2019-05-15
[Connectors] Polish connector descriptions

avano:
@oscerd I removed the "receive" part from SNS connector based on your comment https://github.com/syndesisio/syndesis/issues/5299#issuecomment-492290347


oscerd:
You're right. Good catch!

PR #5376, opened 2019-05-15
Fix #5343: add knative feature toggle (disabled by default)

### LINKED ISSUE #5343: Allow to toggle Knative features

nicolaferraro:


zregvart:
I think that by branching `deployment.json` we might end up forgetting about updating it as it needs to be updated in two places (even though the addtional one only holds KNative stuff).
I think this can be made even simpler by extending the `ModelData` to include conditional logic, something like adding a `condition` field so that if it holds SpringEl expression it needs to evaluate to `true` for the data to be included in the database.
(2c)

nicolaferraro:
That seems a better @zregvart.
Do you know which target object I can use to evaluate SpringEl conditions against? Can I directly reference application properties in the condition or I need to inject them using `@Value` somewhere?

zregvart:
> Do you know which target object I can use to evaluate SpringEl conditions against?

I'd do it against `org.springframework.core.env.ConfigurableEnvironment`, perhaps you can do `getProperty` on it.

nicolaferraro:
Something like this?

Issue #4181, opened 2018-11-29
Step which is after basic or advanced filter is in the activity log twice


mkralik3:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
When I have some step after the advanced filter or basic filter, this step appears in the activity log twice.
e.g.
**1. Integration:**
![image](https://user-images.githubusercontent.com/16251792/49217560-31da4200-f3ce-11e8-8682-cdb50848483f.png)

![image](https://user-images.githubusercontent.com/16251792/49219408-74eae400-f3d3-11e8-95d3-fda4246dd09d.png)
You can see that DataMaper is there **after** Invoke SQL **again**.
[Webhook to DB before-export.zip](https://github.com/syndesisio/syndesis/files/2628810/Webhook.to.DB.before-export.zip)

**2. Integration:**
![image](https://user-images.githubusercontent.com/16251792/49217316-87fab580-f3cd-11e8-813d-85e70db0a267.png)
In the activity log, you can see that another **unknown** Log step with No output is after Invoke SQL but this step is not **defined**.
![image](https://user-images.githubusercontent.com/16251792/49219453-9ba91a80-f3d3-11e8-95dc-4346a097b499.png)
)
[Webhook to DB advanced after-export.zip](https://github.com/syndesisio/syndesis/files/2628894/Webhook.to.DB.advanced.after-export.zip)

**3. Integration:**
Same situation with Basic filter:
![image](https://user-images.githubusercontent.com/16251792/49219286-16256a80-f3d3-11e8-8126-05c3b07b75c2.png)
![image](https://user-images.githubusercontent.com/16251792/49219486-c1362400-f3d3-11e8-815b-8e3d72606b4e.png)
In the activity log, you can see that another **unknown** Log step with No output is after Invoke SQL but this step is not **defined**.
[Webhook to DB basic after-export.zip](https://github.com/syndesisio/syndesis/files/2628908/Webhook.to.DB.basic.after-export.zip)

**4. Integration:**
 For integration without the advance of the basic filter, **no** extra steps appear after the finish step:
![image](https://user-images.githubusercontent.com/16251792/49219588-09eddd00-f3d4-11e8-93d6-8bf0ce8d1408.png)
[...]

PR #5380, opened 2019-05-16
[feat] Enable multi flow activity logging

christophd:
This PR changes activity tracking in order to support multi flow integrations where the primary integration flow delegates to sub-flows (other Camel routes) during message processing. The step activities within the sub-flows should be logged with the current integration activity. 

In order to make this happen I have introduced the Syndesis activity logging id as exchange header when integration delegates to sub-flows ([ChoiceStepHandler](https://github.com/christophd/syndesis/blob/issue/5213/multi-flow-activity-logging/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/handlers/ChoiceStepHandler.java#L59)). When the header is present the ([ActivityTracker](https://github.com/christophd/syndesis/blob/issue/5213/multi-flow-activity-logging/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/logging/ActivityTracker.java#L48)) reads the header in favor of creating a new id for the exchange. This way the sub-flows are able to use the same activity id when logging events.

Also I introduced a new activity tracking policy for sub-flows that does skip the activity tracking setup and round-up as this should be done only in the parent integration flow when the exchange is started/finished. The policy [applies to](https://github.com/christophd/syndesis/blob/issue/5213/multi-flow-activity-logging/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/ActivityTrackingPolicyFactory.java#L42-L44) subflows only where the existing integration activity tracking policy only applies to primary flows. Flow kinds (primary/subflows) are defined in the [flow metadata](https://github.com/christophd/syndesis/blob/issue/5213/multi-flow-activity-logging/app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java#L110-L121).

UI sets this metadata on the flow when conditional flows are created.

The only thing missing is that multi flow activity logging is facing the same issue as filter step where some steps are logged twice as reported in https://github.com/syndesisio/syndesis/issues/4181.

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/WithMetadata.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/WithMetadata.java
@@ -34,4 +34,10 @@ public interface WithMetadata {
             getMetadata().get(key)
         );
     }
+
+    @JsonIgnore
+    default String getMetadata(String key, String defaultValue) {
+        return Optional.ofNullable(getMetadata().get(key))
+                       .orElse(defaultValue);
+    }
 }
--- a/app/common/model/src/main/java/io/syndesis/common/model/WithMetadata.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/WithMetadata.java
@@ -34,10 +34,4 @@ public interface WithMetadata {
             getMetadata().get(key)
         );
     }
-
-    @JsonIgnore
-    default String getMetadata(String key, String defaultValue) {
-        return Optional.ofNullable(getMetadata().get(key))
-                       .orElse(defaultValue);
-    }
 }
--- a/app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java
@@ -18,22 +18,21 @@ package io.syndesis.common.model.integration;
 import java.io.Serializable;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import io.syndesis.common.model.WithId;
 import io.syndesis.common.model.WithMetadata;
 import io.syndesis.common.model.WithName;
 import io.syndesis.common.model.WithTags;
 import io.syndesis.common.model.connection.Connection;
 import io.syndesis.common.util.json.OptionalStringTrimmingConverter;
-
 import org.immutables.value.Value;
-
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 
 @Value.Immutable
 @JsonDeserialize(builder = Flow.Builder.class)
@@ -42,6 +41,23 @@ public interface Flow extends WithName, WithId<Flow>, WithTags, WithSteps, WithM
 
     class Builder extends ImmutableFlow.Builder {
         // allow access to ImmutableIntegration.Builder
+    }
+
+    enum MetadataKey {
+        TYPE;
+
+        public String value() {
+            return name().toLowerCase(Locale.US);
+        }
+    }
+
+    enum MetadataValue {
+        PRIMARY,
+        SUBFLOW;
+
+        public String value() {
+            return name().toLowerCase(Locale.US);
+        }
     }
 
     @Value.Default
@@ -89,5 +105,18 @@ public interface Flow extends WithName, WithId<Flow>, WithTags, WithSteps, WithM
         return getSteps().stream()
             .flatMap(s -> s.getExtensionIds().stream())
             .collect(Collectors.toSet());
+    }
+
+    @JsonIgnore
+    default boolean isPrimary() {
+        return getMetadata().isEmpty() || getMetadata(MetadataKey.TYPE.value(), MetadataValue.PRIMARY.value())
+                                            .equals(MetadataValue.PRIMARY.value());
+    }
+
+    @JsonIgnore
+    default boolean isSubFlow() {
+        Optional<String> flowType = getMetadata(MetadataKey.TYPE.value());
+        return flowType.isPresent() &&
+                flowType.get().equals(MetadataValue.SUBFLOW.value());
     }
 }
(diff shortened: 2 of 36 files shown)


app/common/model/src/main/java/io/syndesis/common/model/WithMetadata.java
```diff
@@ -34,4 +34,10 @@
             getMetadata().get(key)
         );
     }
+
+    @JsonIgnore
+    default String getMetadata(String key, String defaultValue) {
+        return Optional.ofNullable(getMetadata().get(key))
+                       .orElse(defaultValue);
```
lburgazzoli: I think this can be just `getMetadata().getOrDefault(key, defaultValue)`

app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java
```diff
@@ -44,6 +43,23 @@
         // allow access to ImmutableIntegration.Builder
     }
 
+    enum MetadataKey {
+        TYPE;
+
+        public String value() {
+            return name().toLowerCase(Locale.US);
```
zregvart: Might be a bit dangerous to put `"type"` in metadata as it can easily be overwritten by code trying to store some other kind of _type_, say "async" or "reactive".
Do we need to store this in metadata, can we rather create enum like `FlowAxis` with values `PRMARY` and `ALTERNATE` and store it in property of the `Flow`?

and nit: not a big fan of doing computation over constant values, I'd cache this value in the constructor

app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java
```diff
@@ -90,4 +106,17 @@
             .flatMap(s -> s.getExtensionIds().stream())
             .collect(Collectors.toSet());
     }
+
+    @JsonIgnore
+    default boolean isPrimary() {
+        return getMetadata().isEmpty() || getMetadata(MetadataKey.TYPE.value(), MetadataValue.PRIMARY.value())
```
zregvart:
```suggestion
        return MetadataValue.PRIMARY == getMetadata(MetadataKey.TYPE.value()).map(MetadataValue::valueOf).orElse(MetadataValue.PRIMARY)
```

app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java
```diff
@@ -90,4 +106,17 @@
...
+                                            .equals(MetadataValue.PRIMARY.value());
+    }
+
+    @JsonIgnore
+    default boolean isSubFlow() {
+        Optional<String> flowType = getMetadata(MetadataKey.TYPE.value());
+        return flowType.isPresent() &&
```
zregvart:
```suggestion
        return !isPrimary();
```

app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java
lburgazzoli: btw do we really need distinction between lower and upper case ?

app/common/model/src/main/java/io/syndesis/common/model/WithMetadata.java
christophd: yes true, this helper method is pretty much not needed. I will rmeove

app/common/model/src/main/java/io/syndesis/common/model/WithMetadata.java
```diff
@@ -34,4 +34,10 @@
             getMetadata().get(key)
         );
     }
+
+    @JsonIgnore
+    default String getMetadata(String key, String defaultValue) {
+        return Optional.ofNullable(getMetadata().get(key))
```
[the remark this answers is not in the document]
christophd: yes true, this helper method is pretty much not needed. I will remove that

app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java
christophd: I like the idea of having this as a property on the flow. Let me try this ...

lburgazzoli: maybe we can also move the meta key to the metadata itself and list all the know values

lburgazzoli: and maybe the meta can also have a method that leverages Enum’s static method:

    public static <T extends Enum<T>> T valueOf(Class<T> enumType,
                            String name)

app/common/model/src/main/java/io/syndesis/common/model/integration/Flow.java
christophd: I used flow type property enum instead

christophd:
@zregvart @lburgazzoli thanks guys, I have used a property on the flow model instead of using the metadata. can you please have another look?

app/ui/src/app/platform/types/integration/integration.models.ts
```diff
@@ -91,6 +92,13 @@ export interface Flow extends WithId {
   connections: Array<Connection>;
 }
 
+export const PRIMARY = 'PRIMARY';
+export const ALTERNATE = 'ALTERNATE';
+
+export type FlowType =
+  | 'PRIMARY'
+  | 'ALTERNATE';
```
zregvart: Might want to use the `const` variables from above instead?

christophd: that was my 1st thought, too. but for some reason TypeScript won't let me use the const values there 🤷‍♂

christophd:
> does it make sense to also pull the kind from the metadata to the property of the Flow?

At the moment we introduced kinds only for alternate flows (`conditional` or `default`). We would have to find another kind used for primary flows.

For now the kind is only used on the UI side to mark that an alternate flow is used as default when no condition matches. We could also just use a boolean metadata key `default=true/false` instead

christophd:
@zregvart do you think having default=true/false in metadata is the better option?

app/ui/src/app/platform/types/integration/integration.models.ts
zregvart: Would [string enum](https://www.typescriptlang.org/docs/handbook/enums.html) be a better choice here?

zregvart:
> @zregvart do you think having default=true/false in metadata is the better option?

The reason I'm being such a pain is that metadata is untyped (`Map<String, String>`) and I prefer typed variants, it allows us to evolve the code much easier (IMHO). This is commonly referred as [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)). Having typesafe enum values would be my preferred choice.

Though this is just nitpicking/bikeshedding, so if you feel otherwise there's no reason not to merge this.

christophd:
I see your point! Though I am not happy with introducing the property on the flow model when this is only used for alternate flow types. Alternate flows would use enum values `DEFAULT` or `CONDITIONAL` but primary flows would use some kind like 'UNSPECIFIED' or 'NONE' and I am not happy with that.

I will merge this PR and give that some more thought.

Issue #5213, opened 2019-04-15
[conditional-flows] CBR specific activity logging


christophd:
Content based router step should provide activity logging that gives some information on the matched condition and its sub-flow outcome.

Would be good to use the same activity logging in the CBR sub-flows so we have parent flow and sub-flow in one single activity log.

mcada:
How do I know which condition passed from activity log?

![screenshot_20190627_153941](https://user-images.githubusercontent.com/14313995/60271115-0bdc3800-98f2-11e9-95f8-2f209470ce54.png)

christophd:
@mcada is that an integration with multiple conditional flows steps that you are using?

mcada:
Yes it is. The logging is good and it works. We just can't determine which condition was used just from activity logs.

christophd:
@mcada agreed, the activity logs should be enhanced to reflect the matching condition (maybe as activity log output) if possible. I would like to track this in a separate issue if this is OK with you

mcada:
Sure. In that case I am closing this one as verified.

christophd:
thanks!

PR #5395, opened 2019-05-17
[fix] Avoid duplicate activity log entries

### LINKED ISSUE #4181: Step which is after basic or advanced filter is in the activity log twice

christophd:
Ref #5213 
Fixes #4181 

As soon as the integration route uses blocks like `filter`, `choice` the activity logging is not working as expected. We see duplicate step activity entries and inconsistent activities as reported in #4181 

This PR fixes that activity logging by not reading the step id from exchange header anymore ([ActivityTrackingInterceptStrategy](https://github.com/syndesisio/syndesis/blob/13a5c8e8afe7de7042952707cd21d6293321cc21/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/logging/ActivityTrackingInterceptStrategy.java)). The stepId header is overwritten by nested steps that live inside blocks like filter and choice and this is why the logs get inconsistent. 

Instead use the static step id that the surrounding pipeline was created with. This ensures that the async callback (executed when the pipeline is done) always uses the proper step id when doing the activity logging.

### DIFF
--- a/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/logging/ActivityTrackingInterceptStrategy.java
+++ b/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/logging/ActivityTrackingInterceptStrategy.java
@@ -58,15 +58,18 @@ public class ActivityTrackingInterceptStrategy implements InterceptStrategy {
                 return target;
             }
 
-            return new EventProcessor(target);
+            return new EventProcessor(target, stepId);
         }
 
         return target;
     }
 
     private class EventProcessor extends DelegateAsyncProcessor {
-        EventProcessor(Processor processor) {
+        private final String stepId;
+
+        EventProcessor(Processor processor, String stepId) {
             super(processor);
+            this.stepId = stepId;
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


christophd:
@KurtStam this fixes the issue #4181 that you were working on recently

PR #5441, opened 2019-05-22
feat: Include first integration step in activity logs

christophd:
This PR adds the first integration step to the activity logs. We need this for conditional flows where we delegate to other flows (Camel routes) and we want the first step in that sub-flow to be part of the activity logs.

Example integration activity with conditional flows:

Bildschirmfoto 2019-05-22 um 11 42 29

Also this PR improves activity logs for steps that have nested steps such as filter and choice. These steps do not log the done event at the very end but the start event in order to maintain the step sequence in the activity logs.

Before:
Bildschirmfoto 2019-05-22 um 11 38 03

The AdvancedFilter is the last activity even though it is in the middle of the integration.

After:
Bildschirmfoto 2019-05-22 um 11 38 19

The AdvancedFilter now logs its start and therefore is in place as the sequence describes the integration. Duration is not given for filter steps anymore.


christophd:
Almost forgot to mention that the out message capture processor is not added for log steps anymore. This improves memory usage of the integration a bit as we minimize the amount of saved out messages in the exchange.

PR #5457, opened 2019-05-23
chore: upgrade Spring Security to 4.2.12.RELEASE

zregvart:


zregvart:
> I think the commit message is wrong it's 4.2.12

Good catch!

PR #5459, opened 2019-05-23
Merge syndesis-react

paoloantinori:
in this PR:
- `ui` folder is renamed `ui-angular` folder
- images and companion script have been adapted
- work coming from `syndesis/syndesis-react` is now present under module `ui-react`
- `ui-react` builds in an image called `syndesis/syndesis-ui:latest-react` so with just this PR the react code base is not deployed, just built in parallel


paoloantinori:
looking better now.
added licence headers rules

paoloantinori:
fixes #/5426

PR #5478, opened 2019-05-27
Application dashboards

### LINKED ISSUE #3918: Application dashboards

alexkieling:
First version of a Grafana dashboard for integration apps. It will need to be integrated into a 'home' dashboard that lists all integrations.
fix #3918

### DIFF
--- a/install/addons/syndesis-integration-dashboard.yml
+++ b/install/addons/syndesis-integration-dashboard.yml
@@ -1,0 +1,827 @@
+apiVersion: integreatly.org/v1alpha1
+kind: GrafanaDashboard
+metadata:
+  name: syndesis-integration
+  labels:
+    app: syndesis
+    syndesis.io/app: syndesis
+    syndesis.io/type: infrastructure
+spec:
+  name: syndesis-integration.json
+  json: |
+    {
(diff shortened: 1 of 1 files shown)


alexkieling:
The build seems to be broken. I'll send another PR later.

PR #5495, opened 2019-05-28
syndesis-ui, add entry to ConfigMap to support DataVirt

paoloantinori:
Am I following the correct workflow?
Doesn't feel right to modify the sort-of same file in 4 different places.


paoloantinori:
thanks

PR #5510, opened 2019-05-29
fix(connector): set parameters from HTTP request

### LINKED ISSUE #5152: API provider ignoring query parameters
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Query parameters on API provider operations are always `null`. The query parameter is passed to the unified data shape but the value is not set correctly at runtime. Instead the parameter value is always `null`.

You can reproduce this with a simple GET operation that uses a query parameter and log the body right after entering the operation flow.

zregvart:
When Camel Servlet component binds query parameters to the Camel message headers it uses the configured `HttpHeaderFilterStrategy`, and in Syndesis this is the `SyndesisHeaderStrategy` which will bind only the `Content-Type` header and will not bind the query parameters to Camel message headers.

With this in addition to Camel message headers the query parameter value is fetched from the `HttpServletRequest` if it is present in the Camel message headers, and when the Servlet component is used this is usually the case.

This also refines the support for query parameters allowing multiple query parameter values to be set for the parameter value making the parameter value a JSON array.

Fixes #5152

### DIFF
--- a/app/connector/support/processor/pom.xml
+++ b/app/connector/support/processor/pom.xml
@@ -56,6 +56,11 @@
     </dependency>
 
     <dependency>
+      <groupId>javax.servlet</groupId>
+      <artifactId>javax.servlet-api</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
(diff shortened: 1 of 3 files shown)


zregvart:
cc @oscerd this was the fix I was working on when I found out about CAMEL-12785

PR #5511, opened 2019-05-29
build, fix build error with react

paoloantinori:


heiko-braun:
What issue is this addressing?

paoloantinori:
My bad, I was reducing memory constraints on the wrong build!

@zregvart pointed me to the right direct, but at the same time this successful test proved a suggestion tht we could decide to apply to the other environement.

PR #5520, opened 2019-05-30
Re-sync Syndesis with most recent Syndesis-React

paoloantinori:
Note:
it should affect only `app/ui-react` module


paoloantinori:
manually merged. I suspect there is a bug with the bot that was waiting for some notification although it built fine

PR #5554, opened 2019-06-03
fix: Support conditional flows in API provider integrations

### LINKED ISSUE #5515: [Conditional Flow] conditional flow step is broken when used inside of API provider
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Using conditional step breaks API provider - there may be more issues but I will leave it as one as this may be a complex problem.

out

When I click on the cancel button at the end of the gif and edit the integration again, there is one more operation in operation count (it probably thinks that the condition == API provider operation), but all the operations are gone:

screenshot_20190529_144234

christophd:
Fixes #5515

PR separating conditional flows from api provider operation flows in top level navigation and makes sure that navigation to/from primary flow is working for api provider integrations.

This is done in Angular code base but it may give some hints and ideas to solve this for the react code base, too.

### DIFF
--- a/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/logging/IntegrationActivityTrackingPolicyFactory.java
+++ b/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/logging/IntegrationActivityTrackingPolicyFactory.java
@@ -42,6 +42,6 @@ public class IntegrationActivityTrackingPolicyFactory implements ActivityTrackin
 
     @Override
     public boolean appliesTo(Flow flow) {
-        return flow.isPrimary();
+        return flow.isPrimary() || flow.isApiProvider();
     }
 }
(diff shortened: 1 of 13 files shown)


heiko-braun:
@gashcrumb FYI

christophd:
@zregvart can you please have another look? I just realized that this PR causes trouble for activity logging in API provider using conditional flows (see #5555). I just pushed the fix to this PR.

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/logging/IntegrationActivityTrackingPolicyFactory.java
```diff
@@ -42,6 +42,6 @@ public boolean isInstance(RoutePolicy routePolicy) {
 
     @Override
     public boolean appliesTo(Flow flow) {
-        return flow.isPrimary();
+        return flow.isPrimary() || flow.isApiProvider();
```
zregvart: Perhaps we should define the exception rather than add refinements to the condition, something like:

```suggestion
        return !flow.isAlternate();
```

zregvart: @christophd just nitpicking, looks good, feel free to merge if there's no need to change this...

PR #5562, opened 2019-06-03
fix(ui): make sure `Integration.flows` is never null

### LINKED ISSUE #4977: Deleting all operations from an API Provider integration breaks Apicurito
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
When editing an existing API Provider integration's specification and deleting all defined operations, clicking the Save button in Apicurito does nothing. After that, it's impossible to navigate away from Apicurito, either by clicking Save, Cancel button or the Back link.

This is related to #4976 , inlcuding the error message in the console, but in this case, it's not possible to deploy the resulting integration.

zregvart:
Transforms the integration so that `Integration.flows` always returns a value. In case when the `flows` is `undefined` or `null` the value returned will be an empty array.

Fixes #4977

I'm not sure about the failure cases for this, i.e. when `Object.getOwnPropertyDescriptor` will return something unexpected (i.e. something that doesn't have `.value`) or the reference received from `Object.getOwnPropertyDescriptor` can change over time, so we end up with consistency errors between what was set and what this getter returns.

I've also found an inconsistency with the display of the start icon in the integration list vs integration details pages.

Can you take a look @gashcrumb @kahboom @riccardo-forina

### DIFF
--- a/app/ui-angular/src/app/store/integration/integration.service.ts
+++ b/app/ui-angular/src/app/store/integration/integration.service.ts
@@ -23,6 +23,7 @@ function getFlowsCount(integration: Integration) {
 }
 
 function transform(integration: Integration): Integration {
+  integration.flows = integration.flows || [];
   if (integration.flows.length > 1) {
     // Multiflow isn't a thing we support but may as well flag it
     integration.type = IntegrationType.MultiFlow;
--- a/app/ui-angular/src/app/integration/list/list.component.html
+++ b/app/ui-angular/src/app/integration/list/list.component.html
@@ -21,54 +21,46 @@
   >
     <ng-template #itemTemplate let-integration="item" let-index="index">
       <div class="list-view-pf-left integration-icons">
-        <ng-container
-          *ngIf="
-            integration.type === IntegrationType.SingleFlow;
-            else multiflowIntegration
-          "
-        >
-          <span
-            *ngIf="integrationActionsService.getStart(integration); let step"
-          >
-            <img
-              class="syn-icon-small"
-              [src]="step | synIconPath"
-              aria-hidden="true"
-            />
-          </span>
-          <span class="fa fa-angle-right"></span>
-          <span
-            *ngIf="integrationActionsService.getFinish(integration); let step"
-          >
-            <img
-              class="syn-icon-small"
-              [src]="step | synIconPath"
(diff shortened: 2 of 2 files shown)


app/ui-angular/src/app/store/integration/integration.service.ts
```diff
@@ -23,6 +23,10 @@ function getFlowsCount(integration: Integration) {
 }
 
 function transform(integration: Integration): Integration {
+  const flows = Object.getOwnPropertyDescriptor(integration, 'flows');
```
gashcrumb: Knowing this codebase it's probably a better and more simple fix fix to just look and see if the value is in there or not and just add it, i.e.:

```
if (!'flows' in integration) {
   integration.flows = []
}
```

Same goes for `tags` BTW if that's a problem, since there's another blind check further down in this file.

app/ui-angular/src/app/store/integration/integration.service.ts
```diff
@@ -23,6 +23,9 @@ function getFlowsCount(integration: Integration) {
 }
 
 function transform(integration: Integration): Integration {
+  if (typeof(integration.flows) === 'undefined' || integration.flows === null) {
```
gashcrumb: `typeof integration.flows === 'undefined'` will cover both `null` and `undefined` for ya but it's just a nitpick :-)

zregvart: Is that a TypeScript thing? `typeof(null) === "object"` in JavaScript...

riccardo-forina: `!integrations.flow` and call it a day?

zregvart: I was previously warned not to do those :smirk_cat:, thought I'd do whatever you folk think is the best practice, no worries :+1:

zregvart: @gashcrumb @riccardo-forina `!integration.flows` then?

riccardo-forina: I'll actually be that guy and propose yet another option: `integration.flows = integration.flows || []`

Issue #4828, opened 2019-03-11
Jira Connector


heiko-braun:
See also **https://issues.jboss.org/browse/ENTESB-11513**

Jira Connector
As a citizen user I want to be able to connect to Jira. Jira details such as url and credentials should be provided in a similar to existing connectors.

DoD

I can **create** Jira issues including filed data such as
- component
- assignee
- watchee 
- issue type
- priority
- attachment

I can **read** Jira issues.

I can **update** Jira issues including filed data such as

- component
- assignee
- watchee 
- issue type
- priority
- attachment

I can **delete** Jira issues.

I want to be able to **comment** on an issue.

I want to be able to add my own **Jira query**.

I want to be able to transition to status Done or other available status.

If possible, we can look to get a list of required fields (or any other dynamic data) in the UI

The feature owner has confirmed one of the following:
--There is an initial draft of the user doc for this feature.
--The writer acknowledges receiving enough information from engineering to write the initial draft of the user doc for this feature.

Sub-Tasks
camel-jira component **Done** Waiting PR review
- [x] Add OAuth authentication
- [x] Add features as described above (edit issue, delete issue, update issue, add comment) 
- [x] Add tests
- [x] Write documentation
- [x] Write camel-jira example

jboss-fuse/camel jira component
- [x] Port camel-jira to jboss-fuse/camel fork

syndesis jira connector
- [x] Write jira.json and customizers
- [ ] Add tests
- [ ] Write documentation
- [ ] Add quickstart

claudio4j:
Started work on the [new camel-jira-rest component](https://issues.apache.org/jira/browse/CAMEL-13390)

heiko-braun:
Migrated to **https://issues.jboss.org/browse/ENTESB-11513**

PR #5564, opened 2019-06-03
Jira Connector 

claudio4j:
https://github.com/syndesisio/syndesis/issues/4828


app/connector/jira/pom.xml
```diff
@@ -0,0 +1,152 @@
...
+
+    <properties>
+        <atlassian-jira-rest-java-client-version>5.1.0</atlassian-jira-rest-java-client-version>
+        <jira-guava-version>20.0</jira-guava-version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot</artifactId>
```
lburgazzoli: I think we removed spring-boot from connectors

app/connector/jira/src/main/java/io/syndesis/connector/jira/JiraComponentProxyFactory.java
```diff
@@ -0,0 +1,47 @@
...
+    public ComponentProxyComponent newInstance(String componentId, String componentScheme) {
+        return new ComponentProxyComponent(componentId, componentScheme) {
+            @Override
+            protected Optional<Component> createDelegateComponent(ComponentDefinition definition, Map<String, Object> options) {
+                for (String k: options.keySet()) {
+                    String value = options.get(k).toString();
+                    // envelope the JQL in RAW() if it isn't
+                    boolean notRaw = value != null && value.length() > 4 && !"RAW(".equals(value.substring(0, 4));
+                    if ("jql".equals(k) && notRaw) {
+                        value = "RAW(" + value + ")";
```
lburgazzoli: You should be able to set the option `raw` from the connector json

app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGenerator.java
```diff
@@ -51,30 +71,7 @@
...
-import java.util.TreeSet;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.stream.Collectors;
-
-import static io.syndesis.integration.project.generator.ProjectGeneratorHelper.addResource;
-import static io.syndesis.integration.project.generator.ProjectGeneratorHelper.addTarEntry;
-import static io.syndesis.integration.project.generator.ProjectGeneratorHelper.compile;
-import static io.syndesis.integration.project.generator.ProjectGeneratorHelper.mandatoryDecrypt;
+import static io.syndesis.integration.project.generator.ProjectGeneratorHelper.*;
```
lburgazzoli: guess this file has been changed by mistake, afaik, we do not use star import.

app/connector/jira/src/main/java/io/syndesis/connector/jira/customizer/AddCommentCustomizer.java
```diff
@@ -0,0 +1,54 @@
...
+        }
+        component.setBeforeProducer(this::beforeProducer);
+    }
+
+    private void beforeProducer(Exchange exchange) {
+        JiraComment jiraComment;
+        if (exchange.getIn().getBody() != null) {
+            jiraComment = (JiraComment) exchange.getIn().getBody();
+        } else {
+            jiraComment = new JiraComment(defaultIissueKey, defaultComment);
```
lburgazzoli: this does not change over time so it can eventually be cached

app/connector/jira/pom.xml
claudio4j: Removed the spring-boot dependencies. Originally I copied from box connector and spring-boot is there.

app/connector/jira/src/main/java/io/syndesis/connector/jira/JiraComponentProxyFactory.java
claudio4j: Fixed that, thanks for the tip.

app/integration/project-generator/src/main/java/io/syndesis/integration/project/generator/ProjectGenerator.java
claudio4j: It was a mistake, fixed, thanks.

app/connector/jira/src/main/java/io/syndesis/connector/jira/customizer/AddCommentCustomizer.java
claudio4j: Do you mean the values from the option's map ? They are cached in the instance attributes.

lburgazzoli: I mean, `new JiraComment(defaultIissueKey, defaultComment);` does not change over invocations so it could eventually be a "lazy sincleton".

app/connector/jira/src/main/resources/META-INF/syndesis/connector/jira.json
```diff
@@ -0,0 +1,651 @@
...
...42OTkxNzgsMTMyLjgxNTk0MyAyNTUuMzI0NDQ0LDEzMi44MzU1NTYgTDI1NS4zMjQ0NDQsMTAuNjY2NjY2NyBDMjU1LjMyNDQ0NCw0Ljc3NTYyOTM0IDI1MC41NDg4MTUsMy42MDcyMjAwMWUtMTYgMjQ0LjY1Nzc3OCwwIFoiIGZpbGw9IiMyNjg0RkYiPjwvcGF0aD4KCQkJCTxwYXRoIGQ9Ik0xODMuODIyMjIyLDYxLjI2MjIyMjIgTDYwLjg3MTExMTEsNjEuMjYyMjIyMiBDNjAuODkwNzIzOCw5MS44ODc0ODg4IDg1LjcxMjUxMTIsMTE2LjcwOTI3NiAxMTYuMzM3Nzc4LDExNi43Mjg4ODkgTDEzOC45ODY2NjcsMTE2LjcyODg4OSBMMTM4Ljk4NjY2NywxMzguNjY2NjY3IEMxMzkuMDI1OTA1LDE2OS4yOTE5MjMgMTYzLjg2MzYwNywxOTQuMDk3ODAzIDE5NC40ODg4ODksMTk0LjA5Nzc3OCBMMTk0LjQ4ODg4OSw3MS45Mjg4ODg5IEMxOTQuNDg4ODg5LDY2LjAzNzg1MTYgMTg5LjcxMzI2LDYxLjI2MjIyMjIgMTgzLjgyMjIyMiw2MS4yNjIyMjIyIFoiIGZpbGw9InVybCgjbGluZWFyR3JhZGllbnQtMSkiPjwvcGF0aD4KCQkJCTxwYXRoIGQ9Ik0xMjIuOTUxMTExLDEyMi40ODg4ODkgTDAsMTIyLjQ4ODg4OSBDMy43NTM5MTM2MmUtMTUsMTUzLjE0MTkyIDI0Ljg0OTE5MTMsMTc3Ljk5MTExMSA1NS41MDIyMjIyLDE3Ny45OTExMTEgTDc4LjIyMjIyMjIsMTc3Ljk5MTExMSBMNzguMjIyMjIyMiwxOTkuODU3Nzc4IEM3OC4yNDE3NjcsMjMwLjQ1NTMyIDEwMy4wMjAyODUsMjU1LjI2NTY0NyAxMzMuNjE3Nzc4LDI1NS4zMjQ0NDQgTDEzMy42MTc3NzgsMTMzLjE1NTU1NiBDMTMzLjYxNzc3OCwxMjcuMjY0NTE4IDEyOC44NDIxNDgsMTIyLjQ4ODg4OSAxMjIuOTUxMTExLDEyMi40ODg4ODkgWiIgZmlsbD0idXJsKCNsaW5lYXJHcmFkaWVudC0yKSI+PC9wYXRoPgoJCTwvZz4KPC9zdmc+Cg==",
```
zregvart: The new thing here is the `assets:` scheme like this:

https://github.com/syndesisio/syndesis/blob/e34d1ff8c287f908f4729d8d70d261773e8735df/app/connector/activemq/src/main/resources/META-INF/syndesis/connector/activemq.json#L305

```suggestion
  "icon": "assets:jira.svg",
```

zregvart:
I've been looking into the dependency issues, some of these can be dealt with exclusions and can be fixed, the big issue though is the use of Guava 20.0, which is not compatible with 27.0.1-jre we're using. This will lead to issues if a connector using Guava is used with Jira connector (e.g. google-sheets). Not sure how to fix this, shading is probably one way though it would require that the `com.atlassian.jira:jira-rest-java-client-api` transitive dependency is modified so it shades the Guava it depends on.

I don't think we can merge this until we find a solution for this.

claudio4j:
> I've been looking into the dependency issues, some of these can be dealt with exclusions and can be 
> fixed, the big issue though is the use of Guava 20.0, which is not compatible with 27.0.1-jre we're 

Unfortunately jira rest java client library depends on guava 20. I did an integration with google-calendar and jira and there are no errors related to the guava version.

zregvart:
@claudio4j I think the fact that you saw no errors was due to classpath entries being sorted alphabetically. From what I understand there will be two Guava jars on the classpath, and the classes from version 20 will be loaded before version 27. If it were the other way around, the Jira connector would fail as some of the methods from `Objects` were moved to `MoreObjects`.
It worries me that we might depend on the order of jars on the classpath.
Perhaps we can merge this (once the CI checks pass) and mark the Jira connector as tech preview and try to address this issue in the next version.

claudio4j:
@zregvart @lburgazzoli @oscerd related to the guava version, updated the JRJC (jira rest java client) to 5.1.2-2bd0a62e, this version uses the newer guava version syndesis declares on master pom. 
Thanks for the review.

oscerd:
We cannot use that version on upstream camel. Seems to be a temporary build. This is the same for Fuse camel.

zregvart:
@oscerd why do you think it's a temporary build, I'm not sure it means anything, they could just be adding git commit ids to the version, from what I understand @claudio4j found out that there are several of these versions present in Atlassian Maven repository for a while now.

oscerd:
In the past we had troubles with external maven repository, upstream. I really don't want to rely on stuff someone could remove tomorrow. We had exactly problem like this with the old camel Jira IIRC

oscerd:
Btw if you feel safe I'm ok.

zregvart:
> Btw if you feel safe I'm ok.

Well, we'll see, I'd like to stay optimistic and hope that there will not be any issues :crossed_fingers:

Issue #5255, opened 2019-04-18
Icon of a step extension not visible


zregvart:
## This is a...

 - [ ] Feature request
 - [ ] Regression (a behavior that used to work and stopped working in a new release)
 - [x] Bug report  
 - [ ] Documentation issue or request

## Description

Icon placed in `src/main/resources/META-INF/syndesis/icon.[svg|png]` is bundled but not presented in the UI. The `icon` property points to `extension:icon.[svg|png]` which is directly rendered for `src` attribute of an `<img>`.

nicolaferraro:
I've seen the same issue with connector extensions. It happens in the react UI with every connector having a custom scheme like 'db:' or 'extension:'. It seems the [`getConnectionConnector`](https://github.com/syndesisio/syndesis/blob/3736a21944dacd254663f493010f2122aa7b7593/app/ui-react/packages/api/src/helpers/connectionFunctions.ts#L28) function doesn't get called in all pages.

![Screenshot from 2019-06-03 12-37-02](https://user-images.githubusercontent.com/4151857/58796086-933cd100-85fc-11e9-8ac5-c74e8d316943.png)

cc: @riccardo-forina

mastepan:
@nicolaferraro @gashcrumb @paoloantinori  just checking. Is there anything that should be checked by QE?

PR #5568, opened 2019-06-03
A pack of performance optimizations

### LINKED ISSUE #5506: [staging] - syndesis server restarts when saving, renaming or deleting an integration
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Maybe it's resource constraints on the backend, but I dunno, I seem to be able to get the backend to restart pretty much on demand.  At first I thought it was just a matter of spamming the save button a bit, but nowadays it seems to just take a save.  Or use the inline-edit control  on the detail page to rename an integration and the backend restarts.  Or deleting an integration also appears to trigger it.  

At the moment there's 9 integrations defined on staging, so it doesn't seem like that should be a lot of stuff.

Unfortunately I don't seem to have access to the openshift console any more so I've no backend logs that correlate with any of this.

nicolaferraro:
Fix #5506

Marking as WIP to avoid auto-merge. Please help me check this because it contains many improvements, some of which we may not want to add now.

Most relevant changes:
- Adding caches based on soft references, that are cleaned up automatically by the garbage collector before the server goes OOM
- Use DB filters instead of Java filters when possible, to avoid retrieving the full dataset of certain objects from the DB
- Move icons outside the connector definition: connector definition is copied in multiple places of an integration definition, so are the icons, that weight ~5-10k each and cause slow queries (this alleviates the problem, full solution should include strong changes to the data model).
- Set request timeouts when contacting Prometheus to avoid cascading failures when Prometheus is unreachable
- Increase memory of the oauth-proxy, because it goes OOM when the amount of data exchanged with the UI increases

For the icons to be displayed properly in all screens of the react app, issue #5255 needs to be solved as well.


app/common/util/src/main/java/io/syndesis/common/util/cache/LRUSoftCache.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private Cache<K, Reference<V>> cache;
+
+    public LRUSoftCache(int maxElements) {
+        this.cache = new LRUDefaultCache<>(maxElements);
+    }
+
+    @Override
+    public V get(K key) {
+        return getOptional(key)
```
lburgazzoli: maybe we can get rid of the optional thing here: the optional object being created by getOptional does not provide any real value

app/common/util/src/main/java/io/syndesis/common/util/cache/LRUSoftCache.java
```diff
@@ -0,0 +1,95 @@
...
+            .stream()
+            .filter(k -> cache.getOptional(k).flatMap(ref -> Optional.ofNullable(ref.get())).isPresent())
+            .collect(Collectors.toSet());
+    }
+
+    @Override
+    public Collection<V> values() {
+        return cache.values()
+            .stream()
+            .map(ref -> Optional.ofNullable(ref.get()))
```
lburgazzoli: can't this be just cahe.values().stream().filter(ref -> ref.get() != null) ?

app/common/util/pom.xml
```diff
@@ -117,4 +122,16 @@
 
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>-Xmx200m</argLine>
```
zregvart: I think this removes the code coverage setup by jacoco, you can change the `basepom.test.memory` property instead

riccardo-forina:
We also noticed that `OptionalInt` was actually returned as a `number`
right away, not as the definition promised by the swagger file. We had to
add a few hard-cast to work around that in the React codebase.

app/server/monitoring/src/main/java/io/syndesis/server/monitoring/DeploymentStateMonitorController.java
```diff
@@ -70,7 +69,7 @@ public DeploymentStateMonitorController(MonitoringConfiguration configuration, D
     @SuppressWarnings("FutureReturnValueIgnored")
     public void open() {
         LOGGER.info("Starting deployment state monitor.");
-        scheduler.scheduleAtFixedRate(this, configuration.getInitialDelay(), configuration.getPeriod(), TimeUnit.SECONDS);
+        scheduler.scheduleWithFixedDelay(this, configuration.getInitialDelay(), configuration.getPeriod(), TimeUnit.SECONDS);
```
zregvart: :+1:

app/common/util/src/main/java/io/syndesis/common/util/cache/LRUSoftCache.java
nicolaferraro: Close.. "getting optional value then filtering" is not equal to "filtering then getting value". In the latter case, the value may disappear..

lburgazzoli: it can be map(ref::get).filter(Objects::notNull)

nicolaferraro: Right, I always fear passing nulls to `map` functions, but I think it only causes troubles in `Optional`, not in `Stream`s. Btw, I'm adding a Guava impl of the Cache. I don't see problems in using it but in any case we can switch back to this one in case of issues.

nicolaferraro:
Syndesis-before-after

I do feel something changed (taken before Guava).

heiko-braun:
Well done!

nicolaferraro:
@zregvart , I've updated the icon generation, can you do a quick check before merging?

PR #5577, opened 2019-06-04
chore(build): bit faster React UI build turnaround

### LINKED ISSUE #5545: React-ui building/testing performance
Obviously, the react-ui is still being heavily developed so this is merely a signpost in the ground as it were.

1. When building dev images for testing in minishift, I like to do a flash build of everything first. This ensures all modules have been recompiled & I'm not wasting time with old code. I then build the relevant images & deploy accordingly. While doing the first compile, the angular-ui used to take quite a short period of time. However, the react-ui seems to take a noticeably longer period of time & appears to run tests too (not sure angular-ui did??). Is there a way the build time could be curtailed?

2. When pushing to a branch ready to generate a PR, I've come across a git hook that seems to re-execute testing for the react-ui. No code in my PR relates to the react-ui so not sure why it should have activated (see screenshot). As can be seen, it unfortunately, failed hence the only way to push to my branch is to disable the git hook. Thus, can this git hook be limited, please?

zregvart:
Removes Maven plugins with no relevance to the UI build and configures the build to operate incrementaly by not removing `node_modules` and `config.json` files. Refines the `flash` profile not to include `lint` and `test` runs.

On my machine:

`mvn clean verify` takes ~6min (without `clean` ~15 sec less)
`mvn clean verify -Pflash` takes ~4:40min (same without `clean`)

I think this is as fast as we can make it for now. Not sure about `husky` it still installs hooks (it's present in `package.json`).

Fixes #5545


app/ui-react/pom.xml
```diff
@@ -452,6 +388,139 @@
         </plugins>
       </build>
     </profile>
+
+    <profile>
+      <id>non-ui-disabled</id>
```
paoloantinori: I don't want to be picky here, but can we either make the id more clear in its intentions or add an xml comment. I cannot really wrap my mind around what `non-ui-disabled` actually means

paoloantinori:
1) what's the logic for having a new profile enabled by default?
2) in the previous days I was fooling around with this maven plugin: https://blog.sebastian-daschner.com/entries/analyze_maven_build_time  Tried on syndesis and was giving out interesting metrics. I was considering suggesting the addition on both syndesis and camel, eventually controlled by a profile

app/ui-react/pom.xml
zregvart: How about `non-ui-related-so-disabled`?

zregvart:
> * what's the logic for having a new profile enabled by default?

The need is that the plugins that are needed for the UI build are visible and easily discoverable just by looking at `<build><plugins>` having a great deal of plugins listed there with no connection to the UI build would make it more difficult (IMHO); and since Maven doesn't support any other notion of grouping configuration than in profiles an activated by default profile suits this need good (IMHO).

> * in the previous days I was fooling around with this maven plugin: [blog.sebastian-daschner.com/entries/analyze_maven_build_time](https://blog.sebastian-daschner.com/entries/analyze_maven_build_time)  Tried on syndesis and was giving out interesting metrics. I was considering suggesting the addition on both syndesis and camel, eventually controlled by a profile

app/ui-react/pom.xml
paoloantinori: definitely better!

Issue #5528, opened 2019-05-31
Wrong schema in the default postgres connection


asmigala:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
The `PostgresDB` connection defined in syndesis by default has the wrong schema set - should be `public` instead of `sampledb` (which is the name of the database, not schema). As far as I can tell, it's not used anywhere in syndesis, since the database steps work with either value - the only problem is that Data Virtualization cannot import the sample db out of the box, because it actually uses the schema.

paoloantinori:
We should not hardcode `null` here:

https://github.com/syndesisio/syndesis/blob/1d1a37099de230b6e0de1a391cb6229d52d10832/app/connector/sql/src/main/java/io/syndesis/connector/sql/customizer/SqlConnectorCustomizer.java#L116

paoloantinori:
https://issues.jboss.org/browse/ENTESB-9996

KurtStam:
@asmigala Can you provide a bit more information. With what credentials and JDBC connection information are you trying to access the sampledb? It currently is grabbing the schema from the connection.

asmigala:
@KurtStam everything works fine in syndesis, the problem is when trying to import the default connection into DV on the Data tab. In fact, the problem appears even just by starting the komodo-server pod, which will try to deploy a vdb internally an fail, since it tries to access the non-existent `sampledb` schema.

@rareddy can probably provide more details on how it gets the schema from syndesis

rareddy:
DV reads the configuration of the data source from syndesis, and uses it create connection and read metadata, issue queries etc. The reason it also happens during the startup of komodo-server pod is, at startup Komodo tries to synchronize the connection configuration between two systems.

@asmigala apart from failing to proceed further, if you see other issues with wrong configuration those we need to capture for DV

KurtStam:
@rareddy can you provide me the JDBC connection info you are using?

rareddy:
we just make rest call to `http://syndesis-server/connection` and read the configuration from it.  This issue saying is on "PostgreDB" connection by default created has default "schema" value is defined as "sampledb". Whereas in Postgres the default schema value is "public". So, this value can be changed to "public" or where ever we are creating the schema in PosgreSQL, create the sampledb schema on it.

KurtStam:
[...]

PR #5591, opened 2019-06-04
fix for 5528 - Wrong schema in the default postgres connection

KurtStam:
Fix for #5528, now creating a sampledb schema - and creating the tables and functions inside of this schema.


asmigala:
Just a question - why create a new schema and not use the `public` one?

KurtStam:
We'd like the sampledb user to be contained to the sampledb schema, so the jdbc url specifies the schema to be `sampledb`. postgresql searches in schemas `sampledb`, `public`. I guess vdb assumes the tables to be in the sampledb schema.

PR #5606, opened 2019-06-05
Excluded tests from -Pflash UIs builds.

valdar:


.gitignore
```diff
@@ -91,3 +91,7 @@ ui-react/syndesis/.env.production.local
 # build cache
 ui-react/*/.cache
 ui-react/*/.rts2_cache*
+
+# package-lock.json
+*/ui-react/package-lock.json
```
riccardo-forina: Good call but one should never ever user `npm` with Syndesis, only `yarn`

app/ui-react/pom.xml
```diff
@@ -386,10 +390,6 @@
                 <id>yarn ng lint</id>
                 <phase />
               </execution>
-              <execution>
```
riccardo-forina: This is confusing, `yarn ng` is for Angular. Where is the stuff to build the React stuff?

.gitignore
valdar: better be safe than sorry ;)

app/ui-react/pom.xml
valdar: it is outside of the `flash` profile i.e. https://github.com/syndesisio/syndesis/blob/2b063db329cbc9a76b0773d0db4a3b62dcf28943/app/ui-react/pom.xml#L94-L162 so if you say that yarn ng is useless I can remove more stuff from that profile and leave just the skip test property to `true` (which is the bit actually disabling the tests).

riccardo-forina: Useless I don't know, surely it's not needed to build the React app. Mind, there are two Angular based projects that we build when building the main app, but that's taken care for by the top-level `yarn build` command. The fact that under the hood there is `ng` involved should not bleed in the POM I think.

I think this is just a leftover from the conf that we used to build the Angular version.

valdar: I think so as well, the `ui-react/pom.xml` has clearly started as a copy/paste from the angular one. So removed all of those.

zregvart:
This is already in #5577, great minds think alike

valdar:
@zregvart judging from the CircleCI progresses you will have to rebase though :grin:

PR #5610, opened 2019-06-05
Build, added maven extension to keep track of build time per module

paoloantinori:


zregvart:
Can we add it to [`.mvn/extension.xml`](https://maven.apache.org/examples/maven-3-lifecycle-extensions.html#Use_your_extension_in_your_build.28s.29) instead? Like [the one](https://github.com/apache/camel/blob/93d66102fbc15e2bd8290113d226049aadf28737/.mvn/extensions.xml) in Camel repository.

paoloantinori:
Sure, I wasn't aware of this feature.

PR #5620, opened 2019-06-06
Incremental build option in Syndesis CLI

zregvart:
This adds the `--incremental` build flag that influences the build to remove any modules that do not have committed or uncomitted changes in them against a git ref (default `master`).

Also configures the use of it on CircleCI.


zregvart:
This surely has some issues with it, for instance it doesn't take into account dependencies of a module. Scream if you think that's really important to get into this also.

oscerd:
It's fine to have the incremental build on circle CI, essential we need to avoid problem while building locally, so this must be disabled as default, as it is.

PR #5621, opened 2019-06-06
fix(ci): don't try to push images that do not exist

zregvart:
Tries to filter out images that haven't been built from so that the build doesn't fail trying to push images that do not exist.

Kinda wrote this one blindly (no way to test really), might need a second pair of eyes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,17 @@ push_images: &push_images
         curl -fsSL https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz | tar xz -C /usr/bin --strip-components 1
       fi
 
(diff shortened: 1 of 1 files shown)


zregvart:
YOLO

paoloantinori:
So this was incomplete. `IMAGES` env var was not set and every image push was silently skipped.

I have set an env var at Circle CI level, but it might be a better idea have it explicit in circle ci config that we version.

On top of that, probably some explicit logging on when a push is skipped bight be useful for eventual future debug.

paoloantinori:
@rkorytkowski unluckily I haven't tracked the list of images built.
I'm pretty sure I have extracted the list I had used from the ImageStreams defined on staging, this might actually miss some.

PR #5623, opened 2019-06-06
fix(5599): Fix handling of options in email and odata connectors

phantomjinx:
* ConnectorOptions
 * Provides utilities for extracting values from options map or return
   null if not found

* More tests to check failure conditions


.gitignore
```diff
@@ -52,7 +52,6 @@ ui/node
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
-syndesis/
```
oscerd: Is this safe?

phantomjinx: Without that I couldn't add the java source files inside a syndesis directory, eg.
`.../syndesis/app/connector/support/util/src/main/java/io/syndesis/connector/support/util/ConnectorOptions.java
`
The 'syndesis' package trips this gitignore and the src file cannot be added.

phantomjinx: @riccardo-forina 
Would you mind considering the above please?

riccardo-forina: `/syndesis` would make it sure that the rule is applied only at the root of the repo I think. Would that work?

riccardo-forina: Anyway, I'm not sure why that ignore is needed, I think it's been added by @paoloantinori because he noticed some temporary stuff being created while building? I dunno

phantomjinx: Gone with your suggestion. If this doesn't work, we can revisit.

Issue #5622, opened 2019-06-06
Atlas Map step fails to load in some cases


asmigala:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
Trying to add a data mapper step between a `Get a specific Event` and `Create Event`, an error message is shown instead of the data mapper ui: 
`Data Mapper UI Initialization Error: Could not load Maven class path: undefined undefined`

![2019-06-06-133925_sel_grim](https://user-images.githubusercontent.com/9480152/59030610-bc11d000-8861-11e9-8474-60d4b7176ae1.png)

## Steps to reproduce:
1. Start step: timer
2. Finish step: Create Event
3. Add a Get a specific Event in between
4. Try to add a data mapper step between the two google calendar steps

mmuzikar:
Happened to me as well with this integration:
1. Db select * from contact
2. split
3. datamapper
4. Gmail

![image](https://user-images.githubusercontent.com/46345469/59031034-e31cd180-8862-11e9-8062-1fdacb1feb8d.png)

riccardo-forina:
What's the content of the `config.json`? It sounds like some misconfiguration of the atlasmap paths there.

asmigala:
@riccardo-forina where do I find the `config.json`?

asmigala:
```
{
  "apiBase": "https://syndesis.my-minishift.syndesis.io",
  "apiEndpoint": "/api/v1",
  "title": "Syndesis",
  "consoleUrl": "",
  "project": "syndesis",
  "datamapper": {
    "baseMappingServiceUrl": "https://syndesis.my-minishift.syndesis.io/api/v1/atlas/",
    "baseJavaInspectionServiceUrl": "https://syndesis.my-minishift.syndesis.io/api/v1/atlas/java/",
    "baseXMLInspectionServiceUrl": "https://syndesis.my-minishift.syndesis.io/api/v1/atlas/xml/",
    "baseJSONInspectionServiceUrl": "https://syndesis.my-minishift.syndesis.io/api/v1/atlas/json/",
    "disableMappingPreviewMode": false
  },
  "datavirt": {
    "dvUrl": "/vdb-builder/v1/"
  },
  "features" : {
    "logging": false
  },
  "branding": {
    "appName": "Syndesis",
    "favicon32": "/favicon-32x32.png",
[...]

PR #5648, opened 2019-06-07
chore: Data Mapper should skip maven classpath resolution in Syndesis…

igarashitm:
… (#5622)

A bit better error message for #5622, show a Step ID for which Document fails to load. atlasmap/atlasmap#971 will add human readable Document name in addition.


igarashitm:
ah gotcha, let me move over there

PR #5655, opened 2019-06-10
fix(5273): Mustache errors showing wrong line number

phantomjinx:
* CodeMirror editor starts at line 0 (due to array index) so need to
  increment in order to display correct line number to user.

* This only affects the mustache lint as both velocity and freemarker
  already deal with this internally.

* Note. This fixes the angular UI only.


phantomjinx:
Failures due to errors related to _server_ and not _ui_. Can this be overridden & merged, please?

Issue #4878, opened 2019-03-13
Inconsistent descriptions for SQL connector


avano:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
When selecting SQL connector action, it states:

```
Periodically invoke SQL to obtain, store, update, or delete data.
```

When I click on it, it brings me to the query configuration where the description is:

```
Enter a SQL statement that starts with SELECT.
```

This doesn't match the previous screen.

Also the validator states more methods when entering some invalid query:

```
Your statement is invalid and should start with INSERT, UPDATE, SELECT or DELETE.
```

zregvart:
I think @TovaCohen might help us with this one.

TovaCohen:
I think I corrected this: 
https://github.com/syndesisio/syndesis/pull/5684

zregvart:
Thanks @TovaCohen, let's mark this as done.

avano:
@TovaCohen @zregvart it seems that I failed to explain the point in the description :-) my main point was the bit that says `Enter a SQL statement that starts with SELECT.` - in the current react ui there is a tooltip that says something similar:

![sql](https://user-images.githubusercontent.com/7081216/59328415-a29ed700-8cec-11e9-8e89-209aaa8c2d3d.png)

The `SELECT` statement isn't the only one that can return something, for example you can have: `DELETE FROM CONTACT WHERE first_name = 'Joe' RETURNING last_name` that results in:

```
{"exchange":"i-Lh9N5TBJdqOxOYWs6Yqz","status":"begin"}
{"exchange":"i-Lh9N5TBJdqOxOYWs6Yqz","step":"-Lh9MltHLTBhZJpO-N2G","id":"i-Lh9N5UFJdqOxOYWs6Yrz","duration":11632212}
{"exchange":"i-Lh9N5TBJdqOxOYWs6Yqz","step":"-Lh9Lu1HLTBhZJpO-N2F","id":"i-Lh9N5hFJdqOxOYWs6Ytz","message":"Body: [[{\"last_name\":\"Jackson\"}]] "}
2019-06-12 06:29:21.678  INFO 1 --- [r://integration] -Lh9LaxnLTBhZJpO-N2F                     : Body: [[{"last_name":"Jackson"}]] 
[...]

PR #5684, opened 2019-06-11
fixes: #4878 Correct description of Periodic SQL invocation action

TovaCohen:
I think this update corrects the description for the Periodic SQL Invocation action, which is available only as a SQL start connection, and supports only specification of a SELECT statement. 

I should have put "fixes: #4878 " in the commit message.


zregvart:
> I should have put "fixes: #4878 " in the commit message.

I usually put them at the end of the commit message, here's [some docs on that](https://help.github.com/en/articles/closing-issues-using-keywords).

heiko-braun:
You can also add keywords to the pull request message i think

PR #5688, opened 2019-06-12
chore: refine description and labelHint

### LINKED ISSUE #4878: Inconsistent descriptions for SQL connector

zregvart:
Refines the description and labelHint for SQL connector.

Fixes #4878

### DIFF
--- a/app/connector/sql/src/main/resources/META-INF/syndesis/connector/sql.json
+++ b/app/connector/sql/src/main/resources/META-INF/syndesis/connector/sql.json
@@ -58,7 +58,7 @@
         },
         "propertyDefinitionSteps": [
           {
-            "description": "Enter a SQL statement that starts with SELECT.",
+            "description": "Enter a SQL statement that returns results.",
             "name": "SQL statement",
             "properties": {
               "query": {
@@ -67,7 +67,7 @@
                 "group": "common",
                 "javaType": "java.lang.String",
                 "kind": "path",
-                "labelHint": "SQL SELECT statement to be executed.",
+                "labelHint": "SQL statement to be executed.",
                 "order": "1",
                 "required": true,
                 "secret": false,


app/connector/sql/src/main/resources/META-INF/syndesis/connector/sql.json
```diff
@@ -58,7 +58,7 @@
         },
         "propertyDefinitionSteps": [
           {
-            "description": "Enter a SQL statement that starts with SELECT.",
+            "description": "Enter a SQL statement that returns results.",
```
TovaCohen: @zregvart  - I think that this changes the description and hint for the Periodic SQL Invocation action, right? The Periodic SQL Invocation action is available only to a start connection, and the user doc says that you must specify a SELECT statement. But you've updated this to imply that a SELECT statement is not required. So I'm wondering, what else can you specify here? It sounds like the doc is wrong. You can see the doc here: [https://access.redhat.com/documentation/en-us/red_hat_fuse/7.3/html-single/connecting_fuse_online_to_applications_and_services/index#adding-db-connection-start_db](url)
WDYT?   Thanks!

TovaCohen: I think your first suggestion is more helpful that what is there now. 
And I'll update the doc to make it clear that the statement just has to return results. The user is not limited to entering a SELECT statement. Thanks!

PR #5706, opened 2019-06-12
Replace ClusterRoles for Roles in syndesis operator

lgarciaaco:
Until now, the operator uses 2 `ClusterRoles`, *edit* which is bind later to syndesis-server `ServiceAccount` and *view* for syndesis-prometheus `ServiceAccount`. I don't think syndesis-prometheus needs any extra privileges but @jamesnetherton correct me if I am mistaken. syndesis-server `ServiceAccount` doesn't need everything what is included in the edit `ClusterRole`, besides there is no reason to use `ClusterRoles` since permissions are namespace required and not cluster wide required.

Lastly, the integration with OLM doesn't support binding an existing `ClusterRole`, which means keeping the `ClusterRole` approach would require to recreate the whole edit and view `ClusterRole` under a different name.

This PR will:
 - Replace the `ClusterRole` *edit* with the `Role` *syndesis-editors* for `ServiceAccount` syndesis-server.
- Remove the RoleBinding for `ClusterRole` *view*
- Add `build_local.sh` which helps locally building and pushing the operator
- Add `operatorsource.yaml` which helps testing locally the OLM integration
- Remove unnecessary yaml files
- Rename the csv file (left over from https://github.com/syndesisio/syndesis/pull/5696)


jamesnetherton:
@lgarciaaco I think you're right. AFAIK `syndesis-prometheus` does not need any extra privileges.

nicolaferraro:
I remember some cluster roles were needed for authentication with oauth proxy. Maybe @zregvart can shed some light here.

zregvart:
> I remember some cluster roles were needed for authentication with oauth proxy. Maybe @zregvart can shed some light here.

I know it needs a [service account](https://github.com/syndesisio/syndesis/blob/master/install/support/serviceaccount-as-oauthclient-restricted.yml), we had at some point required a ClusterRole for creating `OAuthClient` objects, but that was reverted. I think we're good with oauth-proxy.

lgarciaaco:
> > I remember some cluster roles were needed for authentication with oauth proxy. Maybe @zregvart can shed some light here.
> 
> I know it needs a [service account](https://github.com/syndesisio/syndesis/blob/master/install/support/serviceaccount-as-oauthclient-restricted.yml), we had at some point required a ClusterRole for creating `OAuthClient` objects, but that was reverted. I think we're good with oauth-proxy.

Mhh ... I could log onto syndesis without problems so I think oauth proxy is fine. I also published few integrations to make sure `syndesis-server` has all privileges required.

install/operator/build_local.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+#
+# This script will help you to rapidly build and push the operator
```
nicolaferraro: Isn't https://github.com/syndesisio/syndesis/blob/e14a73f690971535a9d009f045e55dae7a2dda73/tools/bin/commands/build#L167 supposed to do that?

install/operator/deploy/operatorsource.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: operators.coreos.com/v1
+kind: OperatorSource
+metadata:
+  name: syndesis
+  namespace: openshift-marketplace
+spec:
+  type: appregistry
+  endpoint: https://quay.io/cnr
+  registryNamespace: lgarciaac
```
nicolaferraro: Maybe not add a personal one or add some comments

install/operator/build_local.sh
lgarciaaco: Kindof ... it doesn't do it locally, it builds the binary in a docker container which takes longer specially for me with a slow internet connection because it has to download all dependencies. I just thought this file could as well be useful for someone else with a slow internet connection and that's why I pushed, but it doesn't really needs to be there

install/operator/deploy/operatorsource.yaml
lgarciaaco: I will remove this file

PR #5709, opened 2019-06-12
Fixes #5263, fixes #5266, adds support for creating unassigned environment names cached in memory, managed across integrations and removed with the delete environment endpoint

### LINKED ISSUE #5263: Add an API to create an environment tag
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description
There's APIs to get, rename and delete environments, unfortunately there isn't an API to create one.

For the moment in [this PR](https://github.com/syndesisio/syndesis-react/pull/161) I'm just sending the request to the rename API so the UI doesn't blow up when you try and add a new environment.

@dhirajsb if you get a chance this'd be nice...

### LINKED ISSUE #5266: /public/integrations/<id>/tags deletes unused tags and we probably don't want that
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description
With the new CI/CD design management of environments is separate from actually tagging integrations.  So right now in the new design implementation, if you uncheck an environment tag and save it and that tag happens to be unchecked for all integrations, it's removed, and then no longer available on the 'manage ci/cd' page, which seems an odd thing to do.

Can test out the new interface via the PR storybook link [in this PR](https://github.com/syndesisio/syndesis-react/pull/161), also as reference the design is [here](https://redhat.invisionapp.com/share/36QTHNFFCUB#/screens/350160534)

dhirajsb:


gashcrumb:
Cool!  Think we'll need any changes in the react app for this?

dhirajsb:
@gashcrumb I don't think so, from what I can tell. I had to rebuild my minishift and now I'm getting the can't resolve db image error. But from the react code it looks like the create call with PUT should work. I added some unit tests to check new use cases as well.

zregvart:
Could we use `io.syndesis.common.model.environment.Environment` or refactor it so it can be used in this case. I'm thinking what happens when the server pod restarts that `unusedEnviroments` gets lost, right?

dhirajsb:
@zregvart yes, unused environments will be lost, but since they haven't been assigned, it's not a catastrophic error. The possibility of a user creating an environment and leaving them unassigned to integrations is low, and recovery is just a matter of creating the environment again. 
We'll do the refactor in 7.5 after discussing how the concept of environments relates to Fuse configuration variables.

PR #5712, opened 2019-06-13
Lots operator of tech debt reduction.

chirino:
* build improvement: You can build outside the GOPATH, detects your best build option, go or docker, builds images using s2i or docker. 
* elimated the use of OpenShift templates, we only use mustache templates now. 
* vendor the operator dependencies: simplifies downstream build and protects us from dependencies disapearing on us. 
* updating the syndesis operator to conform more to the default operator-sdk v0.8.0 file layout and build strategy. 
* you can use the operator-sdk to build this operator. 
* Mustache templates can use CRDs which not registered with the operator.


heiko-braun:
@chirino I've added `pr/wip` to put this on hold. Master is currently used for the 7.4 release and should not receive any changes to we don't itend to ship. Looking at the changes here, I'd assume this was intended for 7.5. 

Next week we will create the branch for the 7.4 release (syndesis 1.7.x) and from that point on we can include this.

zregvart:
It's a pretty big PR so it'll take a while and perhaps more eyes on it to have it reviewed, what I noticed at first glance is:
 - use of `go mod` I've seen [experienced go developers struggle with it](https://twitter.com/search?q=%40chuhnk%20go%20mod&src=typed_query), the operator-sdk is still using Gopkg, so perhaps we should take some cues from that
 - I would prefer a statically typed object creation (no mustache templates)
 - I think this is a good first step towards operator-only installation, I think for the next step we should move the mustache templates over to code
 - we need more tests here, lots and lots of tests

This can be all addressed in the followup PRs.

install/operator/deploy/crds/syndesis_v1alpha1_syndesis_crd.yaml
```diff
@@ -0,0 +1,224 @@
...
+        metadata:
+          type: object
+        spec:
+          properties:
+            components:
+              properties:
+                db:
+                  properties:
+                    database:
+                      type: string
```
lgarciaaco: Is the operator reconciling after this fields are changed? In the current state the operator ignores the CR (I think it is only used for komodo bits)

install/operator/build/Dockerfile
```diff
@@ -1,7 +1,16 @@
-FROM alpine:3.8
+FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
```
lgarciaaco: Is it intended to use `registry.access.redhat.com` for the upstream version of the operator?

install/operator/build.sh
```diff
@@ -1,15 +1,40 @@
...
+  --image-build <auto|docker|s2i>    how to build the image (default: auto)
+  --image-name <name>                docker image name (default: syndesis/syndesis-operator)
+  --s2i-stream-name <name>           s2i image stream name (default: syndesis-operator)
+
+ENDHELP
+	exit 0
+fi
+
+#
+# TODO Could we avoid copying these files by moving them under the build directory
```
lgarciaaco: Big kudos for improving this, I also like think that moving `addons` is the way to go. At least for now we could delete the folder after the operator is built

install/operator/build/Dockerfile
chirino: Yep. And encouraged.

install/operator/deploy/crds/syndesis_v1alpha1_syndesis_crd.yaml
chirino: Nope. Nothing in that area changed.

chirino:
I’ve not had any issues with go mod yet. 
This change can go into 7.4 I think. I’ve already started a prod branch to verify they can build it in osbs. Static types come with their own sets of problems (importing lots of different dependencies increasing your chances of a module conflict).  Agree about the tests. I find the lack of them troubling.

zregvart:
I have no experience with `go mod` and very little experience with Gopkg, I saw those tweets and it raised a red flag for me. It could very well be that we never have any issues with `go mod`.
@chirino if you think this is good for 7.4 I'm fine with that. I do think that making such a big change at the end of the sprint brings about a lot of risk.
Can you make it [pass the CI](https://circleci.com/gh/syndesisio/syndesis/76153)? We seem to be using a go version without `go mod` and I think that's why it fails.

chirino:
Rebased pr.. and trying to get circle ci to use a newer go version.

chirino:
@zregvart hey this finally passed CI.  Let me know if you spot something glaring.

zregvart:
Awesome, I kinda think we should merge this and then improve on it as needed. Any objections @syndesisio/backend ?

lgarciaaco:
@chirino @zregvart Why merge this in code freeze? There is a reason why it is called code freeze. I am spending the whole morning adjusting to the changes this PR introduces, with work to be done that is to be delivered within this sprint, really this could wait two more weeks

zregvart:
@lgarciaaco this was merged to `master`, we're only concerned about changes that go into the `1.7.x` branch that the _code freeze_ applies to. This will not go into `1.7.x` and will be targeted at the next release.

lgarciaaco:
Then I am spending the whole morning in the wrong branch ... :D

PR #5723, opened 2019-06-13
fix(connector): Box - OutputDataShape java.io.OutputStream not needed as it can't be mapped to anything

### LINKED ISSUE #5234: Box - OutputDataShape java.io.OutputStream not needed as it can't be mapped to anything
## This is a...

<pre><code>
[X] Feature request
[X] Bug report  
</code></pre>

## Description
`Download file` action specifies the datashape as `java.io.OutputStream` and that is not needed, as it can't be mapped to anything.

dm

It would be better if the connector was returning some custom structure (if it is possible) with properties like `fileName` `content` `size` `...` that could be mapped to something

tadayosi:
Fixes #5234


tadayosi:
Oops, pushed again with unused imports errors fixed.

PR #5725, opened 2019-06-13
fix: distinguish between differently named security definitions

### LINKED ISSUE #5669: [api client connector] basic authentication does not work
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Basic authentication is selected by default but `Next` button is disabled:

screenshot_20190611_132057

Adding two Basic authentication options is broken - it has the same name, selecting them does not work - the same bug as above:

screenshot_20190611_132220

zregvart:
When several security definitions of the same type are present in the OpenAPI document it's difficult to distinguish between them.

This adds name and description (if present) to the label and type followed by the name separated by colon in the value of the `authenticationType` property `enum` values. This serves as distinction for the user and also for the code handling those `enum` values.

Fixes #5669


zregvart:
You might want to try this fix for #5669 @gashcrumb and @riccardo-forina before introducing any fixes to the UI.

riccardo-forina:
Thanks :)

PR #5728, opened 2019-06-13
Fix image for syndesis operator olm integration

lgarciaaco:
... and exclude my development files ...

### DIFF
--- a/install/operator/deploy/manifests/1.7.0/syndesisoperator.1.7.0.clusterserviceversion.yaml
+++ b/install/operator/deploy/manifests/1.7.0/syndesisoperator.1.7.0.clusterserviceversion.yaml
@@ -240,8 +240,7 @@ spec:
               serviceAccountName: syndesis-operator
               containers:
                 - name: syndesis-operator
-                  image: lgarciaac/syndesis-operator:latest
-                  #image: syndesis/syndesis-operator:1.7.1-20190607
+                  image: syndesis/syndesis-operator:latest
                   imagePullPolicy: Always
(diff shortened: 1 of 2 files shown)


install/operator/deploy/manifests/1.7.0/syndesisoperator.1.7.0.clusterserviceversion.yaml
```diff
@@ -240,8 +240,7 @@ spec:
               serviceAccountName: syndesis-operator
               containers:
                 - name: syndesis-operator
-                  image: lgarciaac/syndesis-operator:latest
-                  #image: syndesis/syndesis-operator:1.7.1-20190607
+                  image: syndesis/syndesis-operator:latest
```
zregvart: Bit odd that we use `latest` on a versioned manifest

lgarciaaco: The final version for 1.7 is not yet build

zregvart: Perhaps we should tag the latest `1.7.*` with `1.7`, I'm a bit afraid that we'll forget to change this from `latest`.

lgarciaaco:
>  @lgarciaaco please make sure that we remember to change from `latest` to a released version when we do release the version; would be good to have this as a check or as a task that `syndesis release` does.

No worries, I wont :)

Issue #5424, opened 2019-05-20
[camel-k] Syndesis integration deployment with Camel-k error 


tplevko:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

I can't use the camel-k runtime in syndesis as integration runtime. 
This error seems to be the root cause: https://github.com/apache/camel-k/issues/466

The camel-k operator is attached:

[log.txt](https://github.com/syndesisio/syndesis/files/3198155/log.txt)

## Steps to reproduce

* install syndesis with camel-k runtime: ```./syndesis minishift --install --project syndesis --camel-k```
* ```oc get configmap syndesis-server-config -o yaml | sed 's/controllers:/controllers:\n      integration: camel-k/' | oc apply -f -```
* ```oc delete pod -l syndesis.io/component=syndesis-server```
* create sample integration - timer -> DB
* observe the error in camel-k operator pod

tplevko:
@valdar can you please provide input - QE is blocked by this, we can't run the new integration runtime. Raising to prio to p0.

heiko-braun:
@tplevko @valdar Does https://github.com/syndesisio/syndesis/pull/5477 address this problem or is it different?

valdar:
@heiko-braun different problem :/

So there are quite some issues to fix before being able to flawlessly install a pipeline syndesis tag with `syndesis install`:

- https://issues.jboss.org/browse/ENTESB-10708
- https://issues.jboss.org/browse/ENTESB-10711
- https://issues.jboss.org/browse/ENTESB-10712

That said @tplevko  these are the steps to have syndesis + camel-k from master installed:

1. clone syndesis git repo
2. build it locally: `cd app && ../mvnw clean install -Pflash`
3. run minishift with enough resources and, before starting it, remember to enable admin-user addon with `minishift addons enable admin-user`
4. oc login as cluster admin: `oc login -u system:admin`

**now there are 2 options in order to provide the necessary artifacts to camel-k operator, THE 2 OPTIONS, AND THEIR STEPS, ARE ALTERNATIVE:**

**FIRST OPTION:** exposing your `.m2/repository` folder through http and instruct camel-k operator to use it:

5. `cd ~/.m2/repository` and `python -m SimpleHTTPServer 9090`
[...]

PR #5738, opened 2019-06-14
chore(dev): improve dev experience with Camel K

nicolaferraro:
While fixing some bugs with Camel K I've realized that configuring syndesis to use camel k and working with libraries is a half-nightmare, so I've added a "syndesis kamel" script with some useful commands to help syndesis developer. Don't know if this approach can be extended to be useful also for QE.

Note: all the commands require a full build of the backend `syndesis build -c -f -b`.

Basically:

- `syndesis kamel --activate` configures syndesis to use camel k as integration runtime (`--deactivate` restores `s2i`)
- `syndesis kamel --configure` makes sure that the camel k integration platform is ready to be used for syndesis development, no matter which version of Camel K you're using (0.3.4+). This includes loading the catalog and configuring maven mirrors
- `syndesis kamel --sync-runtime` loads all syndesis modules in the camel k operator (with `oc rsync` under the hood), so that people can use connector snapshots and see the changes
- `syndesis kamel --sync-lib` can be used to sync generic libraries (e.g. Camel snapshots)

A typical dev workflow is:
- Dev changes the connector
- `syndesis kamel --sync-runtime` to upload the changes
- `kamel reset` to make camel k rebuild all integrations from scratch

Wdyt?
cc: @paoloantinori, @heiko-braun, @squakez, @tplevko

### DIFF
--- a/tools/bin/commands/kamel
+++ b/tools/bin/commands/kamel
@@ -1,0 +1,211 @@
+#!/bin/bash
+
+LOCAL_MAVEN_REPOSITORY=~/.m2/repository
+OPERATOR_MAVEN_REPOSITORY=/tmp/artifacts/m2
+
+kamel::description() {
+    echo "Tools for developing integrations using Camel K"
+}
+
+kamel::usage() {
+    cat <<EOT
+    --sync-runtime          Synchronize runtime snapshot libraries with the camel k operator.
+    --sync-lib              Synchronize snapshot libraries in the camel k operator (e.g. org/apache/camel/camel-activemq).
+
+    --configure             Configure the Camel K platform to be used by Syndesis
+
+    --activate              Activate usage of Camel K as runtime for integrations
+    --deactivate            Deactivate usage of Camel K as runtime for integrations, going back to s2i
+
+EOT
+}
+
+
+kamel::run() {
+    source "$(basedir)/commands/util/openshift_funcs"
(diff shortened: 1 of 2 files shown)


jamesnetherton:
I gave this a quick spin and it seems to be working nicely. Definitely simplifies some of the manual steps outlined in #5424. So to that end it should be useful for both Syndesis devs & QE.

paoloantinori:
`syndesis kamel` because `syndesis camel` would have been super confusing, right?

nicolaferraro:
Yeah, I don't even like the word 'kamel'... But "syndesis camel" is not more confusing? XD

lburgazzoli:
What about: `syndesis camel --with-k` ?

riccardo-forina:
The joke was `syndesis kamel --the-k-is-silent`

valdar:
Good and very useful job @nicolaferraro . 

I have one concern though, should we remove the `--camel-k` option from `syndesis install` command? Is there some functionality not covered by `syndesis kamel` that was presente in `syndesis install --camel-k`? It might be confusing

nicolaferraro:
@valdar the `--camel-k` flag on `syndesis install` allows to install the operator in the cluster/namespace and it's useful for people that don't have the `kamel` binary installed (and it's the same flag used on fuse-online-install for productized operator), while this is for helping during development of connectors. I'm not sure if it's better to reorganize them..

PR #5747, opened 2019-06-14
feat: require OAuth settings for connection creation

### LINKED ISSUE #5582: [react-ui] Cannot validate salesforce connection
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
screenshot_20190604_143143

Error in meta pod:
```
2019-06-04 12:15:54.639  INFO 1 --- [  XNIO-3 task-7] i.s.c.s.verifier.api.ComponentVerifier   : Verify (PARAMETERS): salesforce === ERROR
2019-06-04 12:15:54.639 ERROR 1 --- [  XNIO-3 task-7] i.s.c.s.verifier.api.ComponentVerifier   : salesforce --> 
2019-06-04 12:15:54.639 ERROR 1 --- [  XNIO-3 task-7] i.s.c.s.verifier.api.ComponentVerifier   :    ILLEGAL_PARAMETER_GROUP_COMBINATION : null
2019-06-04 12:15:54.639 ERROR 1 --- [  XNIO-3 task-7] i.s.c.s.verifier.api.ComponentVerifier   :    ILLEGAL_PARAMETER_GROUP_COMBINATION : null
2019-06-04 12:15:54.639 ERROR 1 --- [  XNIO-3 task-7] i.s.c.s.verifier.api.ComponentVerifier   :    ILLEGAL_PARAMETER_GROUP_COMBINATION : null
```

zregvart:
When creating OAuth-capable connections issues can arise when connection properties are configured by manual entry. For instance refresh token that is hidden from the UI will not be filled and the connection will not be valid as the access token for it will expire and could not be renewed.

This introduces a change in the connection creation that if a new OAuth-capable connection creation is attempted OAuth settings must be set. The user is prompted to enter the properties in OAuth settings and when those are set, connection creation via OAuth flow is allowed.

Peek 2019-06-14 15-42

Fixes #5582

### DIFF
--- a/app/server/credential/src/main/java/io/syndesis/server/credential/AcquisitionMethod.java
+++ b/app/server/credential/src/main/java/io/syndesis/server/credential/AcquisitionMethod.java
@@ -15,9 +15,9 @@
  */
 package io.syndesis.server.credential;
 
-import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-
 import org.immutables.value.Value;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 
 /**
  * Information on the method of credential acquisition, needed for UI to present
@@ -33,6 +33,8 @@ public interface AcquisitionMethod {
         // builder implemented by Immutables, access allowed through this
         // subclass
     }
+
+    boolean configured();
 
     String getDescription();
 
--- a/app/server/credential/src/main/java/io/syndesis/server/credential/CredentialProviderRegistry.java
+++ b/app/server/credential/src/main/java/io/syndesis/server/credential/CredentialProviderRegistry.java
@@ -20,8 +20,8 @@ import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
-import io.syndesis.server.dao.manager.DataManager;
 import io.syndesis.common.model.connection.Connector;
+import io.syndesis.server.dao.manager.DataManager;
 
 import org.springframework.boot.autoconfigure.social.SocialProperties;
 import org.springframework.core.io.support.SpringFactoriesLoader;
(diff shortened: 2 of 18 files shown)


app/ui-react/packages/models/src/models-internal.d.ts
```diff
@@ -17,6 +17,7 @@ export interface AcquisitionMethod {
   type?: 'OAUTH1' | 'OAUTH2';
```
riccardo-forina: Wait, wasn’t this file in the gitignore?

heiko-braun: Gitignore doesnt move between repos

zregvart: I'll remove it, must be some magic git workflow I did...

zregvart: Seems that we [keep adding it](https://github.com/syndesisio/syndesis/commits/master/app/ui-react/packages/models/src/models-internal.d.ts), so it must be something with .gitignore

dhirajsb:
@zregvart I think this is a bad idea. Being able to configure distinct Salesforce secrets on distinct connections allows users to connect to Salesforce using multiple users/apps or to multiple Salesforce instances. The way the current OAuth secrets settings is designed, we would be going backwards and limiting users to a single Salesforce app secret for an entire Syndesis instance. 

This doesn't make sense to me. @gaughan, @heiko-braun is it acceptable to limiting users to only one OAuth secret and user for one Salesforce instance (or any other OAuth capable connector) in Syndesis?

dhirajsb:
The description of the problem as a primary requirement is incorrect. When directly configuring credentials in a connection, there is no expectation of saving a `refreshToken` in connector configuration. The Salesforce connector will login and generate one every time the integration is started and maintain it as long  as the app is running. 
The connection can be validated without a `refreshToken` (I verified this during debugging by simply removing the empty `refreshToken` being sent in the verification call) and an `accessToken` will be generated at runtime as I mentioned before.

zregvart:
@dhirajsb for Salesforce and in general other OAuth based access control, you can use the same client (client id/secret) to access multiple instances (Salesforce) or accounts (Twitter/Google *). Your assumption that this limits the use is not a valid one.

What we're trying to do here is to simplify connection creation for the user, and by making it simpler we're simplifying the implementation.

dhirajsb:
@zregvart isn't the clientId and clientSecret generated by a particular Salesforce instance? How will that work with another Salesforce instance? 
Even with a single instance, there are valid use cases where users might want to use different clientId/clientSecret credentials for accessing different Salesforce objects because of Salesforce access control issues.

zregvart:
> @zregvart isn't the clientId and clientSecret generated by a particular Salesforce instance? How will that work with another Salesforce instance?

Connected App is defined within an instance. The same client ID and secret can be used across instances.

> Even with a single instance, there are valid use cases where users might want to use different clientId/clientSecret credentials for accessing different Salesforce objects because of Salesforce access control issues.

Administrators of each instance can control permissions assigned to the connected app separately from the instance the connected app resides.

While I might see reasoning for supporting multiple OAuth settings per connector I do not see any reason for the end users to manually fill in the required configuration for OAuth (besides client ID and secret). This generates problems, as it was demonstrated in several raised issues. We can add support for multiple OAuth settings per connector and still enforce the OAuth workflow.

gaughan:
I agree with cleaning up the manual fill in the required configuration for OAuth 
Secondary but still of value is adding support for multiple OAuth settings per connector.

PR #5750, opened 2019-06-14
feat: add husky as a git hooks framework

riccardo-forina:
The UI workflow needs 2 hooks:

* pre-commit: automatically lint `prettier` to ensure that all the code is styled in the same way regardless of who is contributing
* pre-push: _this is run only if a change in the `app/ui-react` folder is detected_ run the linter, the test suite and the storybook builder for the whole UI. This is useful to catch the most common errors that are done when adding/changing some feature.


package.json
```diff
@@ -1,7 +1,16 @@
...
-    "@commitlint/config-conventional": "^6.1.3"
+    "@commitlint/config-conventional": "^6.1.3",
+    "husky": "^2.4.1"
   },
-  "dependencies": {}
+  "dependencies": {},
+  "husky": {
+    "hooks": {
+      "pre-commit": "cd app/ui-react; yarn precommit",
+      "pre-push": "BRANCH=$(git rev-parse --abbrev-ref HEAD) UICHANGES=$(git diff --name-only --diff-filter=b $(git merge-base HEAD@{1} $BRANCH) | grep app/ui-react); if [[ -z \"${UICHANGES}\" ]]; then echo 'No changes on ui-react detected'; else cd app/ui-react && yarn prepush; fi"
```
riccardo-forina: Now that I think about this, I'm not sure it will work if the latest commit didn't change something in the ui-react folder. Any idea?

zregvart: This is what I did for the incremental build thinggy: https://github.com/syndesisio/syndesis/blob/master/tools/bin/commands/util/common_funcs#L153

zregvart:
I need to add the `pre-commit` hooks we already have to this, we can merge this when you think it works okay (needs rebase) and I can do a follow up PR for the existing `pre-commit` stuff.

package.json
riccardo-forina: Ok so I changed the script to use `syndesis -m ui-react --incremental` instead, and also added `npm-run-all` to pave the way to future additional hooks. Just add a new `[hook]:[module]` script in the `scripts` section with the script that needs to be run and it will be executed automatically.

riccardo-forina:
`wip` removed. I'm not that sure the `pre-push` script is actually detecting changes, but it's the least important of the two. the `pre-commit` one is working fine.

PR #5755, opened 2019-06-14
fix(5321): Actually delete integrations and connections on delete

phantomjinx:
* Rather than updating resources in jsondb with a deleted flag, actually
  remove them instead.

* Deprecate delete flag methods, preserving for use by existing installs

* ConnectionHandler
 * Remove ConectionBulletinBoards when deleting connections

* IntegrationHandler
 * Remove Integrations, IntegationDeployments & IntegrationBulletinBoards

* Adds / updates tests

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/integration/IntegrationBase.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/integration/IntegrationBase.java
@@ -45,6 +45,11 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 
 public interface IntegrationBase extends WithProperties, WithResourceId, WithVersion, WithModificationTimestamps, WithTags, WithName, WithSteps, ToJson, WithResources {
 
+    /**
+     * @deprecated fully deleted from the data manager in 7.4+.
+     *      Retained for filtering in existing installations.
+     */
+    @Deprecated
     @JsonInclude(JsonInclude.Include.NON_DEFAULT)
     @Value.Default
     default boolean isDeleted() {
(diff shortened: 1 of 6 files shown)


phantomjinx:
@zregvart 
Turns out not as extensive as I thought, _mea culpa!_

KurtStam:
Are we sure we want to do this? Initially this was done by design. Did the design change?

phantomjinx:
> Are we sure we want to do this? Initially this was done by design. Did the design change?

For my part, I am not in a position to question the original design. That said, I would be interested to understand the rationale for the design choice.

While developing/testing this, I obviously looked at my live jsondb.
My observations:
- 6 integrations, 4 connections created & visible in UI
- 27069 records in jsondb
- 8749 connector records (standard for install?)
- 318 connector-templates (standard for install?)
- 13119 integration-deployments records
- 3355 integrations records
- 195 activity records (standard for install?)
- 80 connection-bulletin-board records
- 1199 connection records
- 54 integration-bulletin-board records
- New connection generates 260 records
- New integration generates 411 records
- Deleting resource before change: ~2secs
- Deleting resource after change: immediate

My questions:
Obviously, performance improves if few-no records in jsondb. However, if a time lag becomes evident after only 6 integrations then with 'deleted' resources remaining & total number of records only increasing then how will the performance of the jsondb ever not continually degrade?

The isDeleted() function of integrations is almost exclusively for filtering out deleted resources. Thus, everytime records are fetched this filtering takes place, adding a performance cost. Is this performance cost worth it?

zregvart:
I've never heard the rationale for keeping the deleted integrations and dependencies around, we certainly have had performance issues with JSONDB tied to the number of rows we store in it, so this looks good to me.

@chirino do you remember why we don't delete the integrations, i.e. only mark them as deleted and keep them around?

zregvart:
I put a WIP label and approved, once we are confidant that this change is okay we can simply remove the WIP label to get it merged.

PR #5756, opened 2019-06-14
Auto form adjustments

seanforyou23:
This PR removes the green text highlighting and border from auto-form inputs in success states and replaces the pf3 style field level help with a pf4 variation. I've updated the shared widgets in auto-form as well as the describe data shape step.

What do we think about the new info icon. That was the closest match to what we had previously, can update if need be.

The only other thing is that I find it difficult to close the popover, my cursor is in within bounds of the button element but seem to only be able to close it when hovered over the svg inside. It's pretty annoying now that we have more of these popovers throughout the app. Does anyone else see this? I can investigate this on the pf side if so.

remove green input border for success state in auto-form
replace pf3 field level help with pf4 popover in auto-form components
update styles for auto-form storybook
refactor checkbox to not nest label

Current field level help

Screen Shot 2019-06-14 at 1 37 06 PM

New field level help

Screen Shot 2019-06-14 at 5 30 43 PM

### DIFF
--- a/app/ui-react/packages/.storybook/vendor.css
+++ b/app/ui-react/packages/.storybook/vendor.css
@@ -1,6 +1,11 @@
 @import url('~patternfly/dist/css/patternfly.css');
 @import url('~patternfly/dist/css/patternfly-additions.css');
 @import url('~patternfly-react/dist/css/patternfly-react.css');
+@import url('~@patternfly/patternfly/patternfly-variables.css');
+@import url('~@patternfly/patternfly/patternfly-shield-noninheritable.css');
+@import url('~@patternfly/patternfly/patternfly-common.css');
+@import url('~@patternfly/patternfly/utilities/Spacing/spacing.css');
+@import url('~@patternfly/patternfly/utilities/Display/display.css');
 
 html,
 body {
--- a/app/ui-react/packages/auto-form/package.json
+++ b/app/ui-react/packages/auto-form/package.json
@@ -77,6 +77,9 @@
     }
   },
   "dependencies": {
+    "@patternfly/react-core": "^3.38.1",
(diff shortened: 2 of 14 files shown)


gashcrumb:
The array storybook looks a bit off.  This one's tough because the consumer of auto-form kinda suggests what layout to use via classes passed in as part of the form config, for this particular story those 3 controls should be in columns, like the basic filter page for example:

image

gashcrumb:
I may actually try and refactor the way arrays are set up, but we'll still kinda want to be able to organize fields into nice rows.

gashcrumb:
I feel like there isn't enough of a border or it's not enough contrast or something:

image

also the close icon looks odd, and as you say the behavior is a bit odd.  I suspect though the usual way to close these is to just click outside of the popover and that seems to work alright.  Maybe just don't have a close button?

dongniwang:
The "i" icon feels pretty heavy, and just found out that PF uses "?" icon https://www.patternfly.org/v4/design-guidelines/usage-and-behavior/forms...

In terms of the pop-over closing behavior, I would say it's worth reaching out to PF.

seanforyou23:
Ok I've pushed some updates. Looks like I was hi-jacking the hook for adding classes to the checkbox control container which was causing the layout issues in the form array. Using a more generic selector now so we can keep that previous hook for classes undisrupted. Looking normal again.

Screen Shot 2019-06-14 at 5 16 01 PM

Noted issue about close behavior in pf-react https://github.com/patternfly/patternfly-react/issues/2275. @gashcrumb I don't see any options for excluding the close button - will look into this.

gashcrumb:
@paoloantinori so...  can we let this one in?

PR #5767, opened 2019-06-15
user doc: Updated doc for new Twitter actions

TovaCohen:
Create a new file for just the Send action, since it does not trigger execution. 
Updated file for triggering an integration with a Twitter connection. 
Asked Claudio to review.


oscerd:
Looks good so far

PR #5768, opened 2019-06-15
user doc: applying conditional expressions to mappings

TovaCohen:
This is still a work in progress. Need to add content about exactly what you can specify in the expression input field.


zregvart:
If I'm not mistaking conditional expressions in the mapper are a tech preview item, should we note that?

cc @igarashitm

TovaCohen:
Yes, Zoran, you are of course right - conditional mapping is Tech Preview. I just confirmed that with Gary. 
I'll add more to the doc for the conditional expressions you can specify, and will add the Tech Preview note then.

PR #5783, opened 2019-06-17
updates Twitter search Keyword field to Query field

TovaCohen:
This updates the twitter.json file to change the name of the Keyword input field to Query. 
It also updates the hint to: 
Return tweets that match this Twitter-format search expression.

### DIFF
--- a/app/connector/twitter/src/main/resources/META-INF/syndesis/connector/twitter.json
+++ b/app/connector/twitter/src/main/resources/META-INF/syndesis/connector/twitter.json
@@ -81,7 +81,7 @@
                 "group": "common",
                 "javaType": "java.lang.String",
                 "kind": "path",
-                "labelHint": "Return tweets that match this Twitter-format search expression.",
+                "labelHint": "Twittter-format search expression, for example: \"My Product\" returns tweets that contain exactly My Product.",
                 "required": true,
                 "secret": false,
                 "tags": [],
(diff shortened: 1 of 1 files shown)


app/connector/twitter/src/main/resources/META-INF/syndesis/connector/twitter.json
```diff
@@ -77,11 +77,11 @@
               "keywords": {
                 "componentProperty": false,
                 "deprecated": false,
-                "displayName": "Keywords",
+                "displayName": "Query",
                 "group": "common",
                 "javaType": "java.lang.String",
                 "kind": "path",
-                "labelHint": "Multiple search values can be separated with comma.",
+                "labelHint": "Return tweets that match this Twitter-format search expression.",
```
claudio4j: How can the user know what is the "Twitter-format" ? WDYT to add an example: camel OR knative ?

TovaCohen: I did a web search for "Twitter-format search operator" and easily found the info. I don't want to write something that implies that that is the only operator you can specify. But you're right, examples are always helpful. How about: 
Twittter-format search expression, for example: "My Product" returns tweets that contain exactly My Product.

claudio4j: That sounds good, thanks Tova.

TovaCohen:
@claudio4j  I committed and pushed that update to this PR. Can you review and merge if correct? Thanks!

PR #5786, opened 2019-06-17
Fixes #5777 Operator image is not being pushed to dockerhub (3rd atte…

### LINKED ISSUE #5777: Operator image is not being pushed to dockerhub
## This is a...

<pre><code>
[ ] Feature request
[x ] Bug report  
</code></pre>

## Description

Fix for https://github.com/syndesisio/syndesis/pull/5621 to explicitly list docker images to push and include operator and upgrade images.

rkorytkowski:
…mpt)

https://github.com/syndesisio/syndesis/issues/5777

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,8 +27,6 @@ job_default: &job_defaults
 push_images: &push_images
   deploy:
     command: |
-      DOCKER_IMAGES_TO_PUSH="syndesis-meta syndesis-operator syndesis-s2i syndesis-server syndesis-ui syndesis-ui-legacy syndesis-upgrade"
-
       if [ "${CIRCLE_PROJECT_USERNAME}" != "syndesisio" ]; then
         exit 0
       fi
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -37,9 +35,10 @@ push_images: &push_images
         curl -fsSL https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz | tar xz -C /usr/bin --strip-components 1
       fi
 
+      # Check, if docker images declared by jobs have actually been built
       IMAGES_TO_PUSH=""
-      for image in ${DOCKER_IMAGES_TO_PUSH}; do
-        if [ ! -z "$(docker images -q $image)" ]; then
+      for image in ${DOCKER_IMAGES}; do
```
rkorytkowski: Note that I'm using the DOCKER_IMAGES variable, which is set in individual jobs to the docker image name produced by that job see e.g. https://github.com/syndesisio/syndesis/pull/5786/files#diff-1d37e48f9ceff6d8030570cd36286a61R88 There's no need to maintain a separate list of images in such a case. I'm still checking, if images have been built, but not sure, if there's a need for that and maybe it's better to fail the build, if an image declared by a job is missing.

Issue #5754, opened 2019-06-14
Trailing right arrow at end of breadcrumbs


mcoker:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
There is an extra `>` at the end of the breadcrumbs

## Expected behavior
Should be no `>` after the last breadcrumb item

## Screenshot
<img width="343" alt="Screen Shot 2019-06-14 at 12 09 24 PM" src="https://user-images.githubusercontent.com/35148959/59525985-6b235c80-8e9d-11e9-8254-6b394fe382db.png">
<img width="397" alt="Screen Shot 2019-06-14 at 12 09 17 PM" src="https://user-images.githubusercontent.com/35148959/59525986-6b235c80-8e9d-11e9-8b6e-fea8bd174f86.png">
<img width="396" alt="Screen Shot 2019-06-14 at 12 09 11 PM" src="https://user-images.githubusercontent.com/35148959/59525987-6b235c80-8e9d-11e9-9d33-37f620fc0f9e.png">

cc @syndesisio/uxd

PR #5799, opened 2019-06-18
fix: do not show a trailing arrow in the breadcrumb

riccardo-forina:
Fixes https://github.com/syndesisio/syndesis/issues/5754


riccardo-forina:
Not sure if we want to backport this. CC @paoloantinori

Issue #5697, opened 2019-06-12
API provider definition can't be edited after creating the integration


mmuzikar:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
After creating an API provider integration, the definition cannot be changed. Any changes made are just ignored. This blocks me to verify #5332 because nothing can be reflected in the UI since nothing can be changed.

![Peek 2019-06-12 14-31](https://user-images.githubusercontent.com/46345469/59351570-fcb89000-8d1e-11e9-8b51-0432ddebddc9.gif)

mmuzikar:
@riccardo-forina There are still problems with editing the definition, as it is shown in this GIF:

![Peek 2019-06-18 13-58](https://user-images.githubusercontent.com/46345469/59680297-426ad200-91d1-11e9-8f0f-2ac567309996.gif)

riccardo-forina:
Right, so this is an edge case that I didn't consider. 

Before showing Apicurio we call an API- `/v1/api/integrations/[integration id]/specification`-to get its specification back; the problem is that in this case the integration has not been saved yet, so the call to the API fails because we pass an ID which isn't a valid one.

Without the specification, we can't show the editor. In addition to that, we are missing the UI part of this PR here https://github.com/syndesisio/syndesis/pull/5724; so what happens right now is that if you change the specification for an existing integration, any customization done on that integration-eg, adding steps or configuring return codes-is effectively lost.

@zregvart any chance we can have a version of `/v1/api/integrations/[integration id]/specification` that instead of asking for an id asks for the actual integration object?

mmuzikar:
It is now possible to edit the definition, but it would probably be easier and more convenient to write it by hand right now.
Integration has to be saved everytime user changes the definition, which shows the save dialog on each save, meaning each save the user has to input an integration name.
On each save of the definition user is navigate to an operation flow, not list of all operations.

Attaching GIF of what it currently behaves like:
![Peek 2019-06-25 09-32](https://user-images.githubusercontent.com/46345469/60079196-b6efc480-972d-11e9-8cee-9699b3b6a97c.gif)

I have zero clues about what each of the buttons **actually** do, but shouldn't the save in API provider definition save the definition, not after the integration is saved? Or is it that the definition gets saved but the integration implementing the definition has to be saved in order for the definition to get saved?

[...]

PR #5805, opened 2019-06-18
fix: get an updated integration when editing its definition using the new API introduced in #5724

riccardo-forina:
Fixes https://github.com/syndesisio/syndesis/issues/5697

### DIFF
--- a/app/ui-react/packages/api/src/helpers/integrationFunctions.ts
+++ b/app/ui-react/packages/api/src/helpers/integrationFunctions.ts
@@ -1503,6 +1503,12 @@ export function getMetadataValue<T>(
   return typeof metadata !== 'undefined' ? metadata[mapKey] : defaultValue;
 }
 
+export function isIntegrationEmpty(integration: IntegrationOverview) {
+  return (
+    integration.flows!.length === 1 && integration.flows![0].steps!.length === 0
+  );
+}
+
 /**
  * Returns true if the given integration is an API provider integration
  * @param integration
--- a/app/ui-react/packages/api/src/useApiProviderIntegration.tsx
+++ b/app/ui-react/packages/api/src/useApiProviderIntegration.tsx
@@ -1,11 +1,13 @@
+import { Integration } from '@syndesis/models';
 import * as React from 'react';
 import { ApiContext } from './ApiContext';
 import { callFetch } from './callFetch';
+import { isIntegrationEmpty } from './helpers';
 
 export function useApiProviderIntegration() {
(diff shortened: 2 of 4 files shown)


riccardo-forina:
@zregvart I'm not sure if I'm doing something wrong, but I'm not getting the properly updated integration back from the API you added in #5724. Check the gif:

lost-integration-customization

So basically I have an existing API Provider based integration saved, I customize one of the operations - the GET /beers one - to add a Log step. Then I edit the definition, remove the DELETE operation, and call the new API to get the updated integration object. Here is a sample cURL copied from the dev tools: https://gist.github.com/riccardo-forina/e9ec2bef2bdb61a840b09880443f05e7
If you check the raw integration object, the Log step is there, but I get back the integration without the DELETE operation - correct - but also without the Log step - wrong.

zregvart:
@riccardo-forina that should be `PUT` instead of `POST` for updating. I don't see those paths in the OpenAPI document so I guess that's the source of the confusion. We have `POST /api/v1/apis/generator` to create a _blank_ integration from the OpenAPI document and `PUT /api/v1/apis/generator` to update a  integration from the OpenAPI document. Both are multipart and accept the OpenAPI document in the `specification` part and the `PUT` one accepts `integration` part also.

riccardo-forina:
It is a PUT in the attached gist

zregvart:
I've found that the flow ID is set to `new-integration:flows...` it needs to be same as the value returned from `POST /api/v1/apis/generator` otherwise the matching of the flows from the given integration and the uploaded specification fails. So the end result is to remove the two flows generated from the GET and POST operations and add a new flow for the GET operation other than matching the two GET flows and performing a single update. Trying to figure out where this happens.

gashcrumb:
Ah yeah, ideally when we create flow IDs they probably shouldn't make use of the integration ID at all or rely on it, it should just be a unique key like other ids.

zregvart:
There's a `PUT /api/v1/apis/generator` instead of `POST /api/v1/apis/generator` when clicking `Next` on new API provider integration from scratch and the `integration` has the `id` of `new-integration` at that point.

gashcrumb:
Yeah, at that point we're using that `new-integration` id to ensure that there's some Id field set for the integration as it's used in the routing.  That'll get deleted when we eventually create the integration, as at that point we do a POST with no ID if I recall correctly.

riccardo-forina:
Correct.

riccardo-forina:
Tested against latest BE, I think this is good to go.

Issue #5174, opened 2019-04-10
Unable to start integration from box connector


avano:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
There are currently no actions to start the integration in box connector and 2 actions to finish (upload, download).

This should be "From" I assume:
https://github.com/syndesisio/syndesis/blob/28cb62a0c74be5c58792d8aca83a2a194fdfb22a/app/connector/box/src/main/resources/META-INF/syndesis/connector/box.json#L87-L88

heiko-braun:
//cc @tadayosi

avano:
@oscerd mentioned that the download action should be a "To" pattern, so I'm closing this

tadayosi:
Yes, actually this is intentionally set to `To` as unlike camel-dropbox, camel-box doesn't support consumer endpoint for downloading at this moment.

PR #5811, opened 2019-06-18
ENTESB-10855 Download action of box connector doesn't show as Start Action

claudio4j:
https://issues.jboss.org/browse/ENTESB-10855


avano:
@claudio4j that caught my eyes as well, but tadayosi explained it here - https://github.com/syndesisio/syndesis/issues/5174#issuecomment-481609818

claudio4j:
Ok, thanks for the reference. Closing it.

PR #5812, opened 2019-06-18
TT-660 Fixed preview to use view defined sql

blafond:


blafond:
rebasing from master seems wrong. closing PR and re-submitting

PR #5862, opened 2019-06-21
fix: extensions can have no actions

### LINKED ISSUE #5835: Import extension: "Something is wrong" dialog appears, extension not imported
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Customizations - Extensions - Import Extension
Select an extension and this happens:
importextension

No extension is imported when I check it on Extensions page.

gashcrumb:
fixes #5835

### DIFF
--- a/app/ui-react/syndesis/src/modules/extensions/pages/ExtensionImportPage.tsx
+++ b/app/ui-react/syndesis/src/modules/extensions/pages/ExtensionImportPage.tsx
@@ -242,7 +242,7 @@ export default class ExtensionImportPage extends React.Component<
                               this.state.extension.id &&
                               this.state.extension.extensionId ? (
                                 <ExtensionImportReview
-                                  actions={this.state.extension.actions.map(
+                                  actions={(this.state.extension.actions || []).map(
                                     (action: Action) =>
                                       ({
                                         description: action.description,


rareddy:
Testing this on the staging still shows the error that I can not add an import extension.

gashcrumb:
@rareddy this works on staging with the extension that I was supplied with, perhaps you have an extension that shows a different problem.  Can you raise a new issue and also provide whatever extension you were trying to upload?

rareddy:
@gashcrumb I was trying to upload a `.jar` file, do I need to register that extension before I can upload?

gashcrumb:
No, that registration happens after the upload completes and you click "Import".  What I noticed here on staging is currently clicking "Import" returns a HTTP 500 from the backend, at least for the original extension that I used when creating.  Can you attach the file here then?  Easier if I can see what's going on.

rareddy:
@gashcrumb I am dragging and dropping the file. It is teiid's jdbc library. I can not attach here looks like you can access it [here](https://oss.sonatype.org/service/local/repositories/releases/content/org/teiid/teiid/12.2.1/teiid-12.2.1-jdbc.jar)

gashcrumb:
@rareddy That doesn't appear to be a Syndesis extension jar which is what that particular page is expecting.  You'd want to wrap that driver into a library extension [like this one](https://github.com/syndesisio/syndesis-extensions/tree/master/syndesis-library-jdbc-driver)

PR #5896, opened 2019-06-24
Fix jira connector dependencies

claudio4j:


zregvart:
@claudio4j is this a blocking issue, i.e. is the JIRA connector not working because of it? We need this to be a blocking issue to backport it.

app/connector/jira/pom.xml
```diff
@@ -128,7 +120,6 @@
         <dependency>
             <groupId>com.atlassian.jira</groupId>
             <artifactId>jira-rest-java-client-core</artifactId>
-            <scope>provided</scope>
```
zregvart:
```suggestion
            <scope>runtime</scope>
```

app/connector/jira/pom.xml
```diff
@@ -175,6 +166,19 @@
...
             </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <configuration>
+                    <!-- jira-rest-java-client-core is a hard transitive dependency on jira-rest-java-client-api
+                        should be at compile scope, but maven-dependency-plugin will throw a WARN when validating it -->
+                    <usedDependencies>
+                        <dependency>com.atlassian.jira:jira-rest-java-client-core</dependency>
```
zregvart: if `com.atlassian.jira:jira-rest-java-client-core` is declared at `runtime` scope this would not be needed.

claudio4j:
> Would changing `com.atlassian.jira:jira-rest-java-client-core` to `runtime` scope work?

runtime scope works when integration starts, but fails when open JiraVerifierExtension in IDE as it can't resolve AsynchronousJiraRestClientFactory from jira-rest-java-cliet-core

app/connector/jira/pom.xml
claudio4j: runtime scope works when integration starts, but fails when open JiraVerifierExtension in IDE as it can't resolve AsynchronousJiraRestClientFactory from jira-rest-java-cliet-core

claudio4j:
> @claudio4j is this a blocking issue, i.e. is the JIRA connector not working because of it? 

It is a blocker, changed to p0. Thanks.

claudio4j:
@zregvart I changed to runtime, it worked. When I said it didn't work previously, was because jira-rest-java-cliet-core was excluded from camel-jira component as I was overriding jira-rest-java-cliet-core version. Later I found it, the exclusion was unnecessary and removed it, but didn't test the runtime scope for it, which I did now. Thanks.

PR #5902, opened 2019-06-24
chore(dev): add a update-manifest dev command

nicolaferraro:
This is the `syndesis dev --update-manifest <connector>` command I've shown.

It uses spring-boot actuator under the hood to pass the new connector. Don't know if we should use a different strategy or we need to protect more the endpoint (management port is only exposed locally to the container and cannot be called from outside.. I port-forward in the pod to call it).

Wdyt?

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/actuator/ConnectorsEndpoint.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/actuator/ConnectorsEndpoint.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


app/server/endpoint/src/main/java/io/syndesis/server/endpoint/actuator/ConnectorsEndpoint.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.context.annotation.Configuration;
+
+import java.util.List;
+
+/**
+ * Defines methods for refreshing connectors without restarting the server.
+ */
+@Configuration
+@ConditionalOnProperty(value = "endpoints.connectors.enabled", havingValue = "true", matchIfMissing = true)
+public class ConnectorsEndpoint extends AbstractEndpoint<List<Connector>> {
```
zregvart: Could we use the existing `PUT /api/v1/connector` endpoint?

https://github.com/syndesisio/syndesis/blob/f29e90402c330e4e67c6103530725acec1eb739b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectorHandler.java#L214

nicolaferraro:
This doesn't add spring-mvc, it's the way [actuator 1.5.x custom endpoints should be configured](https://docs.spring.io/spring-boot/docs/1.5.21.RELEASE/reference/htmlsingle/#production-ready-customizing-endpoints-programmatically). I'd like using our standard REST API but then I'd use the public interface instead of the management interface and I'd need authentication via session cookie (and the token lookup function from bash doesn't work anymore in latest version afaik).

Issue #5833, opened 2019-06-20
Route for API Provider integration not created


asmigala:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
After deploying an API Provider integration, a route is not created and the following is logged in syndesis-server log:

```
io.fabric8.kubernetes.client.KubernetesClientException: Failure executing: GET at: https://kubernetes.default.svc/apis/route.openshift.io/v1/namespaces/syndesis/routes/i-qwer. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. routes.route.openshift.io "i-qwer" is forbidden: User "system:serviceaccount:syndesis:syndesis-server" cannot get routes.route.openshift.io in the namespace "syndesis": no RBAC policy matched.
```

Happens on 1.7.1 and master

heiko-braun:
@lgarciaaco Is this related to your recent changes?

heiko-braun:
Related to #5892

lgarciaaco:
> @lgarciaaco Is this related to your recent changes?

yes

heiko-braun:
@lgarciaaco Are the commits supposed to fix the issue? In this case please move the issue to `Done`

heiko-braun:
@nicolaferraro Can you verify if it's fixed?

nicolaferraro:
Sure

nicolaferraro:
I confirm it works on 1.7.7.

asmigala:
Verified with 1.7.8

Issue #5892, opened 2019-06-24
A route for WebHook is not created automatically


mkralik3:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
After deployed the integration with WebHook, the route is not created automatically as before. Because of that, the UI doesn't contain the URL either.

Version: (`syndesis minishift --install --tag 1.7`)
Syndesis: 1.7.3
Commit ID: ef7b10dfca7246c95ab685eeddc16d61f73e6d34

heiko-braun:
@lgarciaaco Are the commits supposed to fix the issue? In this case please move the issue to `Done`

PR #5915, opened 2019-06-25
Allow syndesis-server to manage routes

lgarciaaco:
Fixes https://github.com/syndesisio/syndesis/issues/5833


lgarciaaco:
it might also fix https://github.com/syndesisio/syndesis/issues/5892

Issue #5925, opened 2019-06-25
Failed to install from operatorhub


lgarciaaco:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

Currently fuse-online is installed via https://github.com/syndesisio/fuse-online-install, which takes care that the ImageStreams are present with the correct version: https://github.com/syndesisio/fuse-online-install/blob/d0190189b85ad0a43e93cea0beb48326e14e8867/install_ocp.sh#L925.

For the operatorhub integration, we have to move this logic (creating ImageStreams) to the operator.

After discussing with @nicolaferraro and @heiko-braun, it seems that the best solution for this release would be to give the user the possibility to download the complete `syndesis-template.yaml` from the fuse-online repository.

heiko-braun:
@lgarciaaco @nicolaferraro is this in progress, or fixed?

lgarciaaco:
Fixed by https://github.com/syndesisio/syndesis/pull/5926

lgarciaaco:
@pure-bot pls let me close this issue :)

tmielke:
If you resolve this one, can you please also resolve https://issues.jboss.org/browse/ENTESB-10903?

heiko-braun:
@lgarciaaco Items that require QE verification need to stay `Open` and will be move to the `Done` column.

avano:
@lgarciaaco @nicolaferraro @heiko-braun adding a new parameter to the operator doesn't solve the issue that there are no image streams, as the new parameter points to the same teplate that is contained in the operator - with no image streams for syndesis infrastructure: https://raw.githubusercontent.com/syndesisio/fuse-online-install/1.7.4/resources/fuse-online-template.yml

Image streams are by design not present for productized templates: https://github.com/syndesisio/syndesis/blob/master/install/generator/02-syndesis-image-streams.yml.mustache#L1


We don't have the template with image streams, (I assume, didn't inspect it closely) the images are feeded from `install_ocp.sh` into https://github.com/syndesisio/fuse-online-install/blob/1.7.4/templates/fuse-online-image-streams.yml (as you can see, here are only the placeholders)

I followed the guide from Luis and I did this change to the upstream CSV file: https://github.com/avano/syndesis/commit/c7e8f7d7ef3970a4ce0ff566f4662f2d579ddeb3

[...]

PR #5926, opened 2019-06-25
Allow `syndesis-template.yml` to be downloaded from fuse-online-install

lgarciaaco:
Fixes https://github.com/syndesisio/syndesis/issues/5925

### DIFF
--- a/install/operator/pkg/syndesis/template/install.go
+++ b/install/operator/pkg/syndesis/template/install.go
@@ -1,7 +1,7 @@
 package template
 
 import (
-	"bytes"
+	"fmt"
 
 	v1 "github.com/openshift/api/template/v1"
 	"github.com/syndesisio/syndesis/install/operator/pkg/apis/syndesis/v1alpha1"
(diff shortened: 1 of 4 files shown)


install/operator/pkg/syndesis/template/install.go
```diff
@@ -44,7 +46,25 @@ func GetInstallResourcesAsRuntimeObjects(scheme *runtime.Scheme, syndesis *v1alp
 }
 
 func GetInstallResources(scheme *runtime.Scheme, syndesis *v1alpha1.Syndesis, params InstallParams) ([]runtime.RawExtension, error) {
-	res, err := util.LoadResourceFromFile(scheme, *configuration.TemplateLocation)
+	var templateLocation string
+
+	if releaseVersion := *configuration.ReleaseVersion; len(releaseVersion) > 0 {
+		var fileUrl bytes.Buffer
```
zregvart: I'd go with `fmt.Sprintf` here instead, something like:
```suggestion
		fileUrl := fmt.Sprintf("https://raw.githubusercontent.com/syndesisio/fuse-online-install/%s/resources/fuse-online-template.yml", releaseVersion)
```

Would also be good to put `"https://raw.githubusercontent.com/syndesisio/fuse-online-install/%s/resources/fuse-online-template.yml"` into a constant.

lgarciaaco:
Yea I agree it is more readable that way

lgarciaaco:
> This is a good first iteration, we need to track this for the next release and make sure that the CRD in the newly re-factored operator has this `releaseVersion` property. Also it might bestow us to serve this file from a syndesis/Fuse/Red Hat related domain instead from raw.githubusercontent.com.

We have discussed about the CRD idea. Having the release version in the CR might be problematic the way it is right now, where we deliver fuse/syndesis as one (all images for different components are tested together). The upgrade process makes sure to upgrade the operator to match the version that is targeted. If we build the version in the CR, we have giving customers the option to blow up the stack by easily changing the version to something the operator doesn't support. We would have to take this into consideration.

lgarciaaco:
> I do wonder what issues we'd encounter if the download fails for instance in firewalled installation of OpenShift without network access. And in this case how is that communicated to the end user.

But isn't this the same constrain we have now? ATM https://raw.githubusercontent.com/syndesisio/fuse-online-install/1.6.21/resources/fuse-online-image-streams.yml is downloaded by `install_ocp.sh`, so an internet connection is required.

Either way, I think the operator should fail the installation if the file can't be downloaded, which is the current behavior

nicolaferraro:
Yeah, so I agree that the `--release-version` flag should only go in the operator, while the CRD already tracks the current version that is installed. The operator makes sure they match at the end of the upgrade process.

In case of download error, the operator should retry (I expect this to happen looking at the code).

To solve the firewall issue, I'd not encode the base URL in the code but use e.g. a `--release-url` parameter that allows to customize it (or something like that, even a template url). This way, the template can be provided in the local network in case of issues.

For 1.8 I think we should change the logic from "downloading the template", to "downloading a tar".. but I haven't looked carefully at the changes in the operator.

lgarciaaco:
@zregvart I didn't change the `fileUrl`, I don't think it is necessary since it is used in only one place, and it will eventually be modified by adding the release version ...

lgarciaaco:
fixes https://github.com/syndesisio/syndesis/issues/5925

PR #5931, opened 2019-06-25
Use maven-deploy-plugin 2.8.2, 3.0-M1 not backwards compatible…

cunningt:
Use maven-deploy-plugin 2.8.2, 3.0-M1 not backwards compatible for altDeploymentRepository


cunningt:
@zregvart Applied changes so that we're only setting dep.plugin.deploy.version to override basepom

PR #5932, opened 2019-06-25
Use maven-deploy-plugin 2.8.2, 3.0-M1 not backwards compatible for altDeploymentRepository

cunningt:
Use maven-deploy-plugin 2.8.2, 3.0-M1 not backwards compatible for altDeploymentRepository


cunningt:
@zregvart We're hitting https://projects.engineering.redhat.com/browse/NCL-4366 - the layout is being set for us in production and we don't have control of it.     We're forced to use maven-deploy-plugin < 3.

PR #5952, opened 2019-06-27
Fixes #5917, handle environment tags appropriately on integration delete

### LINKED ISSUE #5917: Delete integration causes that associated tag is deleted too
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
When I have a tag which is associated only to the one integration and I delete that integration, the tag is automatically deleted too. Shouldn't the tag stay in the Syndesis when the tag lifecycle is managed in the Manage CI/CD view?  A similar case as with https://github.com/syndesisio/syndesis/issues/5266 In my POV, for consistency, the deletion of the tag should be able only in the Manage CI/CD view.

dhirajsb:


zregvart:
This triggers a test failure:
```
java.lang.NullPointerException
	at io.syndesis.server.endpoint.v1.handler.integration.IntegrationHandlerTest.shouldDeleteIntegrationsAndDeletingAssociatedResources(IntegrationHandlerTest.java:152)
```

dhirajsb:
@zregvart yeah, there's a circular dependency issue among the beans now. Working on resolving it now.

dhirajsb:
@zregvart all tests pass now

dhirajsb:
Refactoring this dependency would require moving out the state managed inside the PublicApiHandler. There are lot of instances where handlers are calling each other. See `IntegrationSupportHandler` and `IntegrationHandler`. 
Considering that we are only fixing blockers in this sprint, I am not going to do any major refactoring to satisfy a new need.

gashcrumb:
Yeah, I'd suggest raising an issue that points to those spots to address in the next release for sure, ideally with links to the lines of offending code, that way we don't forget what needs to be done.

gashcrumb:
Mostly because I agree, it's best to focus on fixing the specific issue without introducing too many changes to help minimize any unpleasant side-effects.  I know I've bitten my own ass several times in the past with that kind of thing.  

Tastes like you think it would.

dhirajsb:
@gashcrumb yeah, and especially since in 7.5 I'm going to refactor and remove the state being managed inside the `PublicAPIHandler` anyway. So that'll be the right time and place to break the design out into more reusable parts.

gashcrumb:
Perfect!  Maybe create that issue now and link it to this PR @dhirajsb?

Does that work for you @zregvart?

dhirajsb:
Created issue #5995 to address this in 7.5.

zregvart:
> Does that work for you @zregvart?

Sure, though in broad terms we should not have two APIs doing the same thing. I feel as though we haven't tried even designing this functionality and we're just piling on technical debt.

dhirajsb:
It's not two APIs doing the same thing, rather invocation of one API having a side effect of affecting state "primarily" managed by another. 
This is quite common and is resolved on the implementation side by having a separate entity that's managing the shared state. Either in the form of a common shared helper class or having the DB manage relationships between entities. 
Even with that refactor, the end result for the user is the same. Calling an API in one interface affects some state returned by another API. The only way to never run into such a situation would be to put all operations in a single API. 
We do have a design for addressing this in the future in the new issue mentioned earlier. If you have ideas about how to make it an ideal solution, please comment on the same. 
I think we are all in agreement in that this needs to be done better. I'm just saying that we can defer this change and address it immediately in the next release and get rid of this hack before it becomes technical debt.

dhirajsb:
Created https://github.com/syndesisio/syndesis/pull/6022 to backport to 1.7.x.

rkorytkowski:
@dhirajsb, the change is suspected to cause test failures. See #6111. Will you look into that?

heiko-braun:
@rkorytkowski @dhirajsb is on PTO

Why do you suspect this change? Is has been on 1.7.x for quite some time and afaik there have not been test suite errors

rkorytkowski:
The error suggests it's the same issue, which was raised by @zregvart in comments above. I needed to rerun tests for my previous PR this week for master as well due to the same reason. Not sure why it wouldn't manifest in 1.7.x at all, or maybe it was, but someone reran tests?

rkorytkowski:
I'll double check and rebase my PR against master...

zregvart:
> Why do you suspect this change? Is has been on 1.7.x for quite some time and afaik there have not been test suite errors

This change is only on `master`. The error is in the integration tests and caused by a failure to launch the Spring application context due to circular bean reference that this PR introduces and crudely tries to circumvent.

rkorytkowski:
Ahh right https://github.com/syndesisio/syndesis/pull/6022 was never merged to 1.7.x.

Issue #5934, opened 2019-06-26
Unable to deploy from master or 1.7.5


avano:
## This is a...

<pre><code>
[ ] Feature request
[X] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
```
syndesis install --project syndesis -y --test-support
```

or 

```
syndesis install --project syndesis -y --test-support --tag 1.7.5
```

results in:

```
{"level":"info","ts":1561527728.6437426,"logger":"action","msg":"Installing Syndesis resource","type":"install","name":"app"}
[...]

PR #5954, opened 2019-06-27
Port permission for operator

lgarciaaco:
When  `install/operator/pkg/generator/assets/install/05-syndesis-security.yml.mustache` was moved to inside the operator, apparently an old version was copied missing most of the required permissions.

Delete `patch_for_using_template_from_configmap` since the `syndesis.yml` doesn't contain the resources, not need to debug this file

Some generated stuff ...


lgarciaaco:
@chirino, @zregvart what about I delete `install/generator/` since it is not used anymore?

lgarciaaco:
Fixes https://github.com/syndesisio/syndesis/issues/5934

zregvart:
> @chirino, @zregvart what about I delete `install/generator/` since it is not used anymore?

That sets us on a road where we need to have other environments (developer staging, product staging, evaluation) supporting operator, so we need to coordinate with those teams.

lgarciaaco:
> > @chirino, @zregvart what about I delete `install/generator/` since it is not used anymore?
> 
> That sets us on a road where we need to have other environments (developer staging, product staging, evaluation) supporting operator, so we need to coordinate with those teams.

I don't get it. Files aren't used by the operator, if some team or tool is using them, they should be using the new location

zregvart:
> I don't get it. Files aren't used by the operator, if some team or tool is using them, they should be using the new location

Ah, I thought we lost the template entirely. Then we need to notify the other teams of the location change with and we need update our tooling with the new location.

lgarciaaco:
> > I don't get it. Files aren't used by the operator, if some team or tool is using them, they should be using the new location
> 
> Ah, I thought we lost the template entirely. Then we need to notify the other teams of the location change with and we need update our tooling with the new location.

You are right, there is no template anymore with the latest version. But the .mustache files still and they are used from within the operator. @chirino correct me if I am not telling the truth.

Nevertheless we can have a separate discussion for that topic, in this PR I am not deleting the files *yet*

lgarciaaco:
fixes https://github.com/syndesisio/syndesis/pull/5954

PR #5958, opened 2019-06-27
fix: point to the repo syndesis bin

### LINKED ISSUE #5957: git hooks assumes `syndesis` binary on PATH
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Git hooks assume to find `syndesis` binary on path, which is not a strict requirement.
Hooks should not make this assumption or degrade gracefully otherwise.

riccardo-forina:
Fixes #5957

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   },
   "scripts": {
     "pre-commit:ui-react": "cd app/ui-react && yarn pre-commit",
-    "pre-push:ui-react": "syndesis -m ui-react --incremental"
+    "pre-push:ui-react": "./tools/bin/syndesis -m ui-react --incremental"
   }
 }
 


package.json
```diff
@@ -15,7 +15,7 @@
   },
   "scripts": {
     "pre-commit:ui-react": "cd app/ui-react && yarn pre-commit",
-    "pre-push:ui-react": "syndesis -m ui-react --incremental"
+    "pre-push:ui-react": "./tools/bin/syndesis -m ui-react --incremental"
```
paoloantinori: what's the assumed working folder? project root? is that always the case?
Isn't there a way to express a location based on the position of `package.json` itself?

paoloantinori: I see also the other scripts make this assumption, so I assume it's fine

PR #5972, opened 2019-06-27
refactor: don't use flow id to correlate API operations

### LINKED ISSUE #5963: Drop integration id and operation id from the flow id
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description

Currently we're creating flow id of the API provider generated flows by concatenating `integration id`, `:flows:` and `operation id` (of the OpenAPI operation).

We should store the operation id in the flow metadata, and modify the logic to match the operation with the flow based on the metadata.

The existing flows need to be migrated such that flow ids are regenerated and the operation id from the flow id is placed into the metadata of the flow.

zregvart:
Before this we correlated API operations from OpenAPI document by embedding the integration id and the operation id in the flow id. This stores the operation id in the flow metadata and uses key generator generated keys for flow ids.

List of endpoint changes:

 - `POST /api/v1/apis/generator` doesn't set integration id, flow ids are key-generated, operation ids are placed in metadata
 - `PUT /api/v1/apis/generator` and `PUT /api/v1/integrations/{id}/specification` correlate using operation id stored in the metadata

Schema was upgrated to 30 and a migration was added to change the flow ids and put the operation id in the flow metadata.

Fixes #5963

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/api/ApiGeneratorHelper.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/api/ApiGeneratorHelper.java
@@ -19,6 +19,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Predicate;
 
 import io.syndesis.common.model.DataShape;
 import io.syndesis.common.model.Kind;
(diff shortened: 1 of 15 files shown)


app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/api/ApiGeneratorHelper.java
```diff
@@ -145,6 +156,10 @@ public static Integration updateFlowsAndStartAndEndDataShapes(final Integration
             .resources(given.getResources()).build();
     }
 
+    private static Predicate<Flow> operationIdIs(String operationId) {
```
christophd: how about renaming to `operationIdEquals` I find the current method name hard to read (because of `IdIs`) and initially thought that this is a typo 😄

app/server/runtime/src/test/java/io/syndesis/server/runtime/integration/ApiIntegrationUpdateITCase.java
```diff
@@ -187,4 +188,8 @@ public void shouldUpdateApiIntegrationViaApiGenerator() throws IOException, JSON
         return ret;
     }
 
+    private static Predicate<Flow> operationId(final String operationId) {
```
christophd: maybe rename to `operationIdEquals`

Issue #5906, opened 2019-06-24
Tagging process misses changing versions in app/extension


cunningt:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

The syndesis tagging process misses changing the versions of these two files- 

app/extension/example/log-step/pom.xml
app/extension/example/simple-action-step/pom.xml

The result is a tag that does not build correctly.

heiko-braun:
@KurtStam would you mind taking a look?

zregvart:
I think we should either remove those completely as they're not part of the build or we move them to another `*-examples` repository. Those examples don't get any attention and I'm not sure what purpose they serve if hardly anyone can discover them.

heiko-braun:
+1 for removing them if not used.

cunningt:
Additionally, the tests in app/extensions also are killing the build after a versions:set.     

extension/maven-plugin/src/it/with-data-shape-variants/pom.xml:    <syndesis.version>1.7.4.fuse-740002</syndesis.version>
extension/maven-plugin/src/it/with-data-shape/pom.xml:    <syndesis.version>1.7.4.fuse-740002</syndesis.version>
extension/maven-plugin/target/it/with-data-shape-variants/pom.xml:    <syndesis.version>1.7.4.fuse-740002</syndesis.version>

If we do a mvn:versions-set on the project, these syndesis-version do not get set as well.    It seems to me like they should be using project.version rather than syndesis.version.

zregvart:
I think the issue in https://github.com/syndesisio/syndesis/issues/5906#issuecomment-505654265 is not related to this and is not related to setting version via maven-versions-plugin. I think it's related to how PME is replacing versions at runtime via Maven extension.

The `${syndesis.version}` is based on `${project.version}`:

https://github.com/syndesisio/syndesis/blob/9111b1227a5a48d0b976cdfc7dcdb1adb7c61d45/app/extension/maven-plugin/src/it/with-data-shape/pom.xml#L76

zregvart:
Fixed via #5976

PR #5976, opened 2019-06-27
Remove example/log-step and example/simple-action (#5906)

cunningt:
example/log-step and example/simple-action are not being tagged correctly, and according to the comments in #5906 they are not really useful in their current location.    Removing.


zregvart:
Ref #5991

PR #5980, opened 2019-06-28
fix: don't require authentication for /config.json

### LINKED ISSUE #5602: Trouble logging into a redeployed syndesis
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
When I have a syndesis instance deployed in minishift, log in to it, then delete the whole project and create a new one, I get a bunch of different errors when trying to log into syndesis. Not sure what exactly fixes it, some combination of ctrl+f5, clearing the site data and opening a new tab.

## Steps to reproduce
1. create a new project in minishift
2. deploy syndesis into it
3. log into syndesis
4. delete the project
5. create the project again
6. deploy syndesis
7. try logging in again

Please note that I'm NOT clicking on the `Report the issue` button in the following video, I just happen to move my mouse there when hitting ctrl+f5.

syndesis-login

zregvart:
Seems that the request for `/config.json` is resulting in a 302 redirect to the HTML page served from oauth-proxy to authenticate. Parsing this response as JSON fails and we get an error in the UI.

This only seems to happen on a particular installation, and I think this should fix this issue. Having public `/config.json` doesn't seem like a security issue to me.

Fixes #5602

### DIFF
--- a/install/generator/04-syndesis-oauth-proxy.yml.mustache
+++ b/install/generator/04-syndesis-oauth-proxy.yml.mustache
@@ -99,6 +99,7 @@
             - '--skip-auth-regex=/icons[^/]*'
             - '--skip-auth-regex=/static[^/]*'
             - '--skip-auth-regex=/precache-manifest.*.js'
+            - '--skip-auth-regex=/config.json'
             - --skip-auth-preflight
             - --openshift-ca=/etc/pki/tls/certs/ca-bundle.crt
             - --openshift-ca=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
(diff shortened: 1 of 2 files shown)


zregvart:
When merged this cannot be backported as is, we need to take the change in `04-syndesis-oauth-proxy.yml.mustache` and run the template generator on `1.7.x` branch to create a backport PR.

zregvart:
I found that the diff for 1.7.x of syndesis.yml is the same, those bits for imagestreams resource verbs are not part of the template either. So doing this via backport label so we can keep everything linked properly.

PR #5994, opened 2019-06-28
fix: remove karma-jasmine-jquery and therefore the transitive bower dependencies

gashcrumb:
@cunningt FYI, also am I correct in thinking this needs to go into the 1.7.x branch?


zregvart:
Backport in `1.7.x` in #5997

PR #6016, opened 2019-07-01
Remove cypress from dev-dependencies

cunningt:
Remove cypress from dev-dependencies - it requires external binaries that are unavailable inside the proxy.  

Talked to @riccardo-forina about this and he said it was not used currently but would be in the future and that it would be okay in 1.7.x to build without the dependency on cypress.    I'll figure out how to get cypress installed for 1.8, but if we could remove it for 1.7.x that would be helpful.


zregvart:
Should this be forward-ported to `master`?

cunningt:
@zregvart No, this should stay only on 1.7.x.     My understanding is that cypress will be needed for 1.8 and there's plenty of time to work out the prod issues of getting it installed behind the firewall.

PR #6019, opened 2019-07-01
fix(operator): Fixes to operator compilation

phantomjinx:
- go variables must match json naming of RouteHostname

- zz_generated.openapi.go
 - regeneration for RouteHostname & inclusion of addons

- .lib.sh
 - mktemp command requires a template of 'XXX's on linux in order to
   properly name the unique tar archive


install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
```diff
@@ -11,7 +11,7 @@ import (
 // SyndesisSpec defines the desired state of Syndesis
 // +k8s:openapi-gen=true
 type SyndesisSpec struct {
-	RouteHostName        string          `json:"routeHostname,omitempty"`
+	RouteHostname        string          `json:"routeHostname,omitempty"`
```
chirino: Perfect. We need to prefer json encoding compat instead of go type compat.

PR #6020, opened 2019-07-01
Changes to ui poms (phases, node-sass/iltorb) for proxies

cunningt:
Adding a number of changes here to support the build of the ui behind a proxy.      The execution phases needed to be changed so that yarn/npm install happened first, then downloads / config changes, and then the build.

yarnInheritsProxyConfigFromMaven needed to be added to a number of executions so that maven does not add in --proxy to yarn commands like lint where that option is not supported.

Some additional changes here to support the required downloads of node-sass and iltorb.

Similar changes need to go into master, but I will do them separately because I think there need to be other additions (cypress, etc).

### DIFF
--- a/app/ui-angular/pom.xml
+++ b/app/ui-angular/pom.xml
@@ -93,6 +93,7 @@
         <executions>
           <execution>
             <id>install node and npm</id>
+            <phase>initialize</phase>
             <goals>
               <goal>install-node-and-npm</goal>
             </goals>
@@ -102,6 +103,7 @@
           </execution>
           <execution>
             <id>install yarn</id>
+            <phase>initialize</phase>
             <goals>
               <goal>install-node-and-yarn</goal>
             </goals>
@@ -112,6 +114,7 @@
           </execution>
           <execution>
             <id>yarn install</id>
+            <phase>compile</phase>
             <goals>
               <goal>yarn</goal>
             </goals>
@@ -121,30 +124,36 @@
           </execution>
           <execution>
             <id>yarn ng lint</id>
+            <phase>compile</phase>
             <goals>
               <goal>yarn</goal>
             </goals>
             <configuration>
               <arguments>lint</arguments>
+              <yarnInheritsProxyConfigFromMaven>false</yarnInheritsProxyConfigFromMaven>
             </configuration>
           </execution>
           <execution>
             <id>yarn ng test</id>
+            <phase>compile</phase>
             <goals>
               <goal>yarn</goal>
             </goals>
             <configuration>
               <arguments>test:ci</arguments>
               <skip>${skipTests}</skip>
+              <yarnInheritsProxyConfigFromMaven>false</yarnInheritsProxyConfigFromMaven>
             </configuration>
           </execution>
           <execution>
             <id>yarn cleanup</id>
+            <phase>compile</phase>
             <goals>
               <goal>yarn</goal>
             </goals>
             <configuration>
               <arguments>cleanup</arguments>
+              <yarnInheritsProxyConfigFromMaven>false</yarnInheritsProxyConfigFromMaven>
             </configuration>
           </execution>
           <execution>
@@ -152,8 +161,10 @@
             <goals>
               <goal>yarn</goal>
             </goals>
+            <phase>compile</phase>
             <configuration>
               <arguments>build:ci</arguments>
+              <yarnInheritsProxyConfigFromMaven>false</yarnInheritsProxyConfigFromMaven>
             </configuration>
           </execution>
         </executions>
@@ -205,9 +216,9 @@
             <executions>
               <execution>
                 <id>set https proxy</id>
-                <phase>initialize</phase>
+                <phase>generate-sources</phase>
                 <goals>
-                  <goal>yarn</goal>
+                  <goal>npm</goal>
                 </goals>
                 <configuration>
                   <arguments>config set https-proxy http://${proxy-user}:${proxy-password}@${proxy-server}:${proxy-port}</arguments>
@@ -215,9 +226,9 @@
               </execution>
               <execution>
                 <id>set http proxy</id>
-                <phase>initialize</phase>
+                <phase>generate-sources</phase>
                 <goals>
-                  <goal>yarn</goal>
+                  <goal>npm</goal>
                 </goals>
                 <configuration>
                   <arguments>config set proxy http://${proxy-user}:${proxy-password}@${proxy-server}:${proxy-port}</arguments>
@@ -225,9 +236,9 @@
               </execution>
               <execution>
                 <id>set maxconn</id>
-                <phase>initialize</phase>
+                <phase>generate-sources</phase>
                 <goals>
-                  <goal>yarn</goal>
+                  <goal>npm</goal>
                 </goals>
                 <configuration>
                   <arguments>config set maxsockets 30</arguments>
@@ -253,7 +264,7 @@
             <executions>
               <execution>
                 <id>set registry</id>
-                <phase>initialize</phase>
+                <phase>generate-sources</phase>
                 <goals>
                   <goal>npm</goal>
                 </goals>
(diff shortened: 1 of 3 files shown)


app/ui-angular/pom.xml
```diff
@@ -93,6 +93,7 @@
         <executions>
           <execution>
             <id>install node and npm</id>
+            <phase>initialize</phase>
```
zregvart:
```suggestion
            <phase>pre-compile</phase>
```initialize

Perhaps `pre-compile` here would be better, this would trigger installing node/npm on just running the licence check (I think?)

Do we need `npm` I thought we use only `yarn`, perhaps this whole execution could be removed?

File: app/ui-angular/pom.xml [hunk 2]
```diff
@@ -102,6 +103,7 @@
           </execution>
           <execution>
             <id>install yarn</id>
+            <phase>initialize</phase>
```
zregvart:
```suggestion
            <phase>pre-compile</phase>
```

app/ui-angular/pom.xml
```diff
@@ -121,39 +124,47 @@
           </execution>
           <execution>
             <id>yarn ng lint</id>
+            <phase>compile</phase>
```
zregvart:
```suggestion
            <phase>verify</phase>
```

So not to trigger lint on `mvn compile`

app/ui-angular/pom.xml
```diff
@@ -121,39 +124,47 @@
...
               <goal>yarn</goal>
             </goals>
             <configuration>
               <arguments>lint</arguments>
+              <yarnInheritsProxyConfigFromMaven>false</yarnInheritsProxyConfigFromMaven>
             </configuration>
           </execution>
           <execution>
             <id>yarn ng test</id>
+            <phase>compile</phase>
```
zregvart:
```suggestion
            <phase>test</phase>
```

Probably makes more sense than `compile`

app/ui-angular/pom.xml
```diff
@@ -121,39 +124,47 @@
...
             </goals>
             <configuration>
               <arguments>test:ci</arguments>
               <skip>${skipTests}</skip>
+              <yarnInheritsProxyConfigFromMaven>false</yarnInheritsProxyConfigFromMaven>
             </configuration>
           </execution>
           <execution>
             <id>yarn cleanup</id>
+            <phase>compile</phase>
```
zregvart: I think this is used to remove any customized `config.json` files so they don't end up being packaged in the app, so doing this before probably makes more sense.

```suggestion
            <phase>generate-sources</phase>
```

app/ui-angular/pom.xml
```diff
@@ -205,29 +216,29 @@
             <executions>
               <execution>
                 <id>set https proxy</id>
-                <phase>initialize</phase>
+                <phase>generate-sources</phase>
                 <goals>
-                  <goal>yarn</goal>
+                  <goal>npm</goal>
```
zregvart: I'm curious as to why this needs to be npm instead of yarn as we use yarn exclusively.

app/ui-angular/pom.xml
```diff
@@ -205,29 +216,29 @@
...
                   <arguments>config set https-proxy http://${proxy-user}:${proxy-password}@${proxy-server}:${proxy-port}</arguments>
                 </configuration>
               </execution>
               <execution>
                 <id>set http proxy</id>
-                <phase>initialize</phase>
+                <phase>generate-sources</phase>
                 <goals>
-                  <goal>yarn</goal>
+                  <goal>npm</goal>
```
zregvart: I'm curious as to why this needs to be npm instead of yarn as we use yarn exclusively.

app/ui-angular/pom.xml
```diff
@@ -205,29 +216,29 @@
...
                   <arguments>config set proxy http://${proxy-user}:${proxy-password}@${proxy-server}:${proxy-port}</arguments>
                 </configuration>
               </execution>
               <execution>
                 <id>set maxconn</id>
-                <phase>initialize</phase>
+                <phase>generate-sources</phase>
                 <goals>
-                  <goal>yarn</goal>
+                  <goal>npm</goal>
```
zregvart: I'm curious as to why this needs to be npm instead of yarn as we use yarn exclusively.

app/ui-react/pom.xml
```diff
@@ -283,27 +281,58 @@
       <build>
         <plugins>
           <plugin>
-            <groupId>com.github.eirslett</groupId>
-            <artifactId>frontend-maven-plugin</artifactId>
+            <groupId>com.googlecode.maven-download-plugin</groupId>
+            <artifactId>download-maven-plugin</artifactId>
+            <version>1.3.0</version>
```
zregvart:
```suggestion
            <version>${maven-download.version}</version>
```

Or better yet we could put this in the parent POM.

app/ui-react/pom.xml
```diff
@@ -81,6 +81,7 @@
 
           <execution>
             <id>install-node-and-yarn</id>
+            <phase>initialize</phase>
```
zregvart:
```suggestion
            <phase>pre-compile</phase>
```

app/ui-react/pom.xml
```diff
@@ -92,6 +93,7 @@
 
           <execution>
             <id>yarn-install</id>
+            <phase>compile</phase>
```
zregvart:
```suggestion
            <phase>pre-compile</phase>
```

app/ui-react/pom.xml
```diff
@@ -102,11 +104,13 @@
 
           <execution>
             <id>yarn-lint</id>
+            <phase>compile</phase>
```
zregvart:
```suggestion
            <phase>verify</phase>
```

app/ui-angular/pom.xml
cunningt: My preference here would be to keep the installation to initialize, which is designed to initialize the build state (https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html).

app/ui-angular/pom.xml
cunningt: Changed

app/ui-angular/pom.xml
cunningt: Changed

app/ui-angular/pom.xml
cunningt: Changed

app/ui-angular/pom.xml
cunningt: Changed, I'm not sure it solely with yarn, but once this goes in I can test it out on staging.

app/ui-angular/pom.xml
cunningt: Changed, I'm not sure it solely with yarn, but once this goes in I can test it out on staging.

app/ui-angular/pom.xml
cunningt: Changed, I'm not sure it solely with yarn, but once this goes in I can test it out on staging.

app/ui-react/pom.xml
cunningt: Changed

app/ui-react/pom.xml
cunningt: Changed

app/ui-react/pom.xml
cunningt: Changed

cunningt:
@zregvart I went ahead and made the modifications requested (with the exception of changing the phase of installation of yarn, which I believe belongs in initialize).    I'm not sure that ui-angular will work solely with yarn goals (vs. mixed yarn/npm), but I made those changes and once this goes in I will give it a try and submit a further PR if we need changes.

cunningt:
After testing, also had to change a few of the angular phases back.

Added platform detection.

zregvart:
Should we forward-port this to `master`?

cunningt:
@zregvart If it'll go cleanly, sure.

PR #6022, opened 2019-07-01
Fixes #5917, handle environment tags appropriately on integration delete

### LINKED ISSUE #5917: Delete integration causes that associated tag is deleted too

dhirajsb:


heiko-braun:
#5917 is not blocking. @dhirajsb why do you think this should go into 1.7.x!

dhirajsb:
@heiko-braun sorry, for some reason I thought that issue was a p0. Closing this PR.

PR #6058, opened 2019-07-03
Upgrade spring-security to 4.2.13

cunningt:
Need to update spring-security on master


zregvart:
I assume it's also OK to backport this to `1.7.x`.

PR #6071, opened 2019-07-05
Phantomjinx/template change

riccardo-forina:
So basically the error was that `language` in the `handleTemplateTypeChange` function was referring to the _old_ value. Using `newType` fixes the issue. So no need for the added `useEffect`.


phantomjinx:
@riccardo-forina 
This seems to be failing tests & as its a simple change, I've modified my commit and force-pushed to the original PR so closing this one.

riccardo-forina:
Uh thanks, I meant to open this on your branch but I must have fat fingered it.

PR #6074, opened 2019-07-05
Fix syndesis permissions and legacy mode

nicolaferraro:
Fix #6028
Fix #5914

When installing upstream version from template (or legacy mode), some errors are thrown but they are related to camel k and knative and you can ignore them. This is the same behavior it was present before on upstream template.
Those errors will not be present in the final product release.

I've added back the viewer role to let prometheus scrape pods.

This fixes the problem with previous version, that only worked  with full cluster-admin permissions, because it required permissions that are normally not allowed to standard users.

PR #6075, opened 2019-07-05
Master security fix

nicolaferraro:
Same as #6074 for master


zregvart:
On `master` we also need to apply this to the handlebar templates that [are in the operator](https://github.com/syndesisio/syndesis/tree/master/install/operator/pkg/generator/assets/install).

Issue #5955, opened 2019-06-27
Template type change results in errors 


mmuzikar:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
When I input a correct template in a specific template type, the Done button results in errors in the browsers console. Only time I am allowed to continue with the integration creation process is if I stick with Mustache and don't try to change the type at all, after the type is changed once, every click **even with correct template type** produces error.

Here you can see it:
![Peek 2019-06-27 13-31](https://user-images.githubusercontent.com/46345469/60263100-17266800-98e0-11e9-8e16-f58615975d47.gif)

What is strange is that the Freemarker template was valid even as a Velocity template... and was valid only after clicking from Velocity.

The error message in console is:
```
Error: Bad Request api.js:3285
```

mmuzikar:
@heiko-braun I didn't consider this blocking issue since template step works BUT there are some workarounds needed for that. ie. if you want to use the mustache type you cannot change the type at all, and with changing from velocity to freemarker it's a matter of clicking through random types.

Would you consider this a blocker or do we need to add a mention of this known issue to the documentation?

mmuzikar:
Verified by QE

Issue #5607, opened 2019-06-05
Unify name of template


mmuzikar:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Documentation issue or request
</code></pre>

## Description
The template step currently is currently referred to as `template` and `templater`. "Normal user" might not even notice it, but it could be a problem while dealing with localization.

## Screenshots
In one picture you can see that in `Add to integration` it's templater, in `Upload template` it's template. 
![image](https://user-images.githubusercontent.com/46345469/58961124-cf149980-87a8-11e9-9a0a-5282c3817cf5.png)
![image](https://user-images.githubusercontent.com/46345469/58961493-87dad880-87a9-11e9-9aa4-68002e1ff6cd.png)

mmuzikar:
Verified by QE.

PR #6076, opened 2019-07-05
fix(5607): Unify template step naming

phantomjinx:
* Removes templater codename as no longer required and confuses with template.


gashcrumb:
"template" here feels too generic.  But I can't think of a more precise term yet to suggest yet.

phantomjinx:
> "template" here feels too generic. But I can't think of a more precise term yet to suggest yet.

Problem is that there is an awful lot of "template" and "template step" labels and names to change if we go that way. It made sense to remove "templater" as the lesser contributor.

phantomjinx:
Same problem as #5955, the Camel-K compiling in test-builds seems to be failing to find dependencies. Is someone able to help with this please?

riccardo-forina:
`useTemplateEditor` maybe?

gashcrumb:
+1

phantomjinx:
> `useTemplateEditor` maybe?

1. Do you want all BE java classes updated to "TemplateEditor"?
2. Do you want the labels seen by users to say "Template Editor" rather than than "Template"?

@riccardo-forina 
@gashcrumb

riccardo-forina:
For the user, I dunno. For the BE classes, is there value in using the same name?

phantomjinx:
So have being mulling over this for the last couple of days...

The objective of the issue is to unify the template naming. That is achieved with the current PR.

To change 'template' to something more original requires a number of further changes, including
- Labels in the UI
- Package & class names in ui-react code
- Package & class names in BE
- The stepKind which bridges both FE & BE

Now, the point was raised why change stuff not seen by the user. The reason, to my mind, is that it should be obvious in the code where a feature resides so, for example, the 'template' feature should be in the 'step/template' package & the main class should be 'TemplateStepHandler'.

At this juncture, I feel that the breadth of change & possibility of breaking things is too high to justify a renaming without a more definitive driver for doing it.

Interested to hear your views, @gashcrumb, @riccardo-forina.

gashcrumb:
I was just thinking in terms of maintenance, non generic self describing
class names are nicer.   I wasn't talking about the UI or user visible bits
at all :-)

phantomjinx:
> I was just thinking in terms of maintenance, non generic self describing class names are nicer. I wasn't talking about the UI or user visible bits at all :-)

No problem. Its just the original issue #5607 was more concerned about what the user saw on the screen so am trying to straddle 2 related concerns.

phantomjinx:
Any chance that this PR could be approved so at least the "templater" name is now removed? Then we can move on to refactoring a different name for the 'template' feature if required.
Thx.
@gashcrumb, @riccardo-forina

Issue #5883, opened 2019-06-24
Replace of start/finish connection in integration - the old connection still showing up


tplevko:
See also **https://issues.jboss.org/browse/ENTESB-11411**

## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem

When editing the start or finish connection in integration - the old connection, which is being updated is still showing, as you can see in the recording. The "clean" start/stop icon there would be much more intuitive (image attached).

## Screenshot

![out](https://user-images.githubusercontent.com/4180208/59997420-5729e880-965d-11e9-8a77-3820c25f0e50.gif)

![Screenshot from 2019-06-24 08-51-00](https://user-images.githubusercontent.com/4180208/59997480-7aed2e80-965d-11e9-95e4-03c592c4e50f.png)

## Tasks involved / Steps to Reproduce

1. Create integration with start and stop connection
2. Select delete one of these connections
3. See that the old connection is still showing on the place of connection which is updated

heiko-braun:
Migrated to **https://issues.jboss.org/browse/ENTESB-11411**

PR #6079, opened 2019-07-05
fix(5883) - Replace of start/finish connection in integration - the old connection still showing up

elvisisking:
- See issue #5883
- `AddStepPage` was  not deleting the start or finish endpoint step (now it does)
- modified `EditorSidebar` to display a placeholder component when the start or finish step is missing

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/AddStepPage.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/AddStepPage.tsx
@@ -160,6 +160,14 @@ export const AddStepPage: React.FunctionComponent<
                     getFirstPosition(state.integration, params.flowId) ||
                   position === getLastPosition(state.integration, params.flowId)
                 ) {
+                  setIsDeleting(true);
+                  const newInt = await removeStep(
+                    state.integration,
+                    params.flowId,
+                    position!
+                  );
+                  state.integration = newInt;
+                  setIsDeleting(false);
                   history.push(getDeleteEdgeStepHref(position!, params, state));
                 } else {
                   /**
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSidebar.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSidebar.tsx
@@ -7,7 +7,11 @@ import {
 import * as React from 'react';
 import { EntityIcon } from '../../../../shared';
 import { IUIStep } from './interfaces';
-import { getDataShapeText, toUIStepCollection } from './utils';
+import {
+  getDataShapeText,
+  isPlaceholderEndpointStep,
+  toUIStepCollection,
+} from './utils';
 
 function makeActiveStep(
   position: number,
@@ -18,6 +22,7 @@ function makeActiveStep(
 ) {
   return activeStep ? (
     <IntegrationFlowStepWithOverview
+      key={position - 1}
       icon={
         <EntityIcon
           alt={activeStep.name}
@@ -41,6 +46,7 @@ function makeActiveStep(
     />
   ) : (
     <IntegrationFlowStepGeneric
+      key={position - 1}
       icon={<i className={'fa fa-plus'} />}
       i18nTitle={`${position}. ${title}`}
       i18nTooltip={tooltip}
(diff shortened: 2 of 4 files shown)


elvisisking:
@riccardo-forina These changes aren't exactly what we talked about but seem to work. I'm not that familiar with the integration editor so I could be missing some things. Can you test drive and let me know? Thanks!

app/ui-react/syndesis/src/modules/integrations/components/editor/AddStepPage.tsx
```diff
@@ -160,6 +160,14 @@ export const AddStepPage: React.FunctionComponent<
                     getFirstPosition(state.integration, params.flowId) ||
                   position === getLastPosition(state.integration, params.flowId)
                 ) {
+                  setIsDeleting(true);
+                  const newInt = await removeStep(
+                    state.integration,
+                    params.flowId,
+                    position!
+                  );
+                  state.integration = newInt;
```
riccardo-forina: Sorry, the state object shouldn’t be mutated, ever.

app/ui-react/syndesis/src/modules/integrations/components/editor/EditorSidebar.tsx
```diff
@@ -51,6 +57,24 @@ function makeActiveStep(
   );
 }
 
+function getPlaceholderActiveStepComponent(
```
riccardo-forina: These functions would be better as function components.

app/ui-react/syndesis/src/modules/integrations/components/editor/AddStepPage.tsx
elvisisking: Would this work:
```
                  history.push(
                    getDeleteEdgeStepHref(
                      position!,
                      params, {
                      ...state,
                      integration: newInt,
                    })
                  );
```
Something like that is done when a non-endpoint is deleted. Is that what you had in mind @riccardo-forina?

elvisisking:
Testing further I've found that canceling out does not reset the values @riccardo-forina. For example, if you:
1. Have an integration with both a start and finish connection
1. then delete the start connection
1. the *Choose connection* page displays correctly (focus is on start connection which is undefined)
1. then cancel out of this page
1. the *Add to Integration* page is displayed but the start connection is undefined. So the original value is lost.

Could local state be used to save off the previous integration? And then when cancel is clicked put back the saved integration some way? Maybe like this:
```
const [lastVersion, setLastVersion] = React.useState(state.integration);
...
setLastVersion(state.integration); // set before deleting an endpoint
...
cancelHref={cancelHref(params, {...state, integration: lastVersion})} // cancel callback
```

Thanks for any ideas/help @riccardo-forina.

app/ui-react/syndesis/src/modules/integrations/components/editor/AddStepPage.tsx
riccardo-forina: Well that's not mutating the `state` object, it's creating a new object that holds also the values from the `state`. But the thing is, you're trying to modify the `integration` object that's describing the _current_ value, but you can't be sure the user will conclude the delete process.

In the following pages, links to go back and abort the action will use the integration object in the route state to say "this is the integration, unchanged"; but with this change, this assertion will not be true anymore.

gashcrumb:
I'm kinda thinking if we actually try and isolate all the editor pages from the actual step array we could kinda change how this whole interaction works.  To be honest, the delete confirmation dialog doesn't exactly make sense any more, as we don't actually delete/overwrite the step when the user clicks "ok", rather we introduce a flow of pages that builds a new step and at the end (kinda) we replace the old step.  The delete prompt totally made sense in the angular version of the editor, because upon accepting the editor would overwrite the step with a "blank" step, then validation would kick in and the user would be looking at the step selection page.  

If the sub-pages in the editor didn't direclty have access to the step array, position, flow etc. we could probably not even bother with the delete prompt until the very end, when the user is actually going to change the integration object.  This change kinda also means the trash icon doesn't make sense either, because we're not actually deleting the step, we're reconfiguring it.

riccardo-forina:
+1 for dumping the delete logic and clear out that we are just reconfiguring

gashcrumb:
@elvisisking why don't we park this PR and revisit the overall approach for 7.5?

elvisisking:
Closing. Revisit #5883 in 7.5.

PR #6082, opened 2019-07-07
user doc: add xref from general json def file info to ui details

TovaCohen:
This just adds a cross-reference from the topic that talks about the requirements in the extension definition json file to the topic that describes the user interface properties that the json file needs to specify.


zregvart:
Backport in #6134

Issue #6014, opened 2019-07-01
SQL connector cannot access the sampledb tables in 1.8-SNAPSHOT


jsafarik:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

Same issue as https://github.com/syndesisio/syndesis/issues/5603, the tables are in public schema and syndesis can't find them. When I tried to do `SELECT * FROM todo` and click the next button, it says `Table 'TODO' does not exist in schema 'sampledb'. Unable to fetch and process metadata`.

The database shows this
```
sampledb=> \dt
          List of relations
 Schema |  Name   | Type  |  Owner
--------+---------+-------+----------
 public | contact | table | sampledb
 public | todo    | table | sampledb
(2 rows)
```

after changing the schema to `sampledb`, it works as supposed.

Reporting this again because it only occurs in 1.8-SNAPSHOT,

KurtStam:
Weird.. they should be in the sampled schema. Can you verify that `syndesis-sampledb-config` configmap has the line:

https://github.com/syndesisio/syndesis/blob/master/install/syndesis.yml#L565

KurtStam:
I seems to be using

operator/pkg/generator/assets/install/04-syndesis-db.yml.mustache 

now, which is an older version, hence the regression.

jsafarik:
@KurtStam I just looked to the `syndesis-sampledb-config` configmap and see no line `CREATE SCHEMA AUTHORIZATION sampledb;` there. The config goes as
```
...
psql -d sampledb -U sampledb <<'EOF'
  CREATE TABLE IF NOT EXISTS contact (first_name VARCHAR, last_name VARCHAR, company VARCHAR, lead_source VARCHAR, create_date DATE);
...
```
the other file you posted looks like the one it uses.

Issue #5690, opened 2019-06-12
DB procedure name changed to text input instead of dropdown


mmuzikar:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
In angular the procedure name input was a dropdown that offered all available procedures within the DB. Now it's just a text input, which is more inconvenient for users to fill in and is quite error prone.

![image](https://user-images.githubusercontent.com/46345469/59343208-9bd38c80-8d0b-11e9-9134-63d859e6f5a0.png)

gashcrumb:
This seems suspiciously similar to https://github.com/syndesisio/syndesis-react/issues/382

riccardo-forina:
I think this is a BE related issue.

This is the-redacted-call to the API to get the action descriptor back:

```curl 'http://localhost:3000/api/v1/connections/5/actions/sql-stored-start-connector' --data-binary '{}'```

We are passing an empty object in the body because we're starting from scratch and we don't have any pre-configured value for this action.

We get back this object:
```
{
    "inputDataShape":
    {
        "kind": "none"
    },
    "outputDataShape":
    {
        "kind": "any"
    },
    "propertyDefinitionSteps":
    [
        {
            "name": "Procedure name",
            "properties":
            {
                "procedureName":
                {
                    "componentProperty": true,
                    "deprecated": false,
                    "displayName": "Procedure name",
                    "group": "producer",
                    "javaType": "java.lang.String",
                    "kind": "path",
                    "labelHint": "Name of the stored procedure.",
                    "required": true,
                    "secret": false,
                    "type": "string"
                },
                "schedulerExpression":
                {
                    "defaultValue": "60000",
                    "deprecated": false,
                    "displayName": "Period",
                    "group": "consumer",
                    "javaType": "long",
                    "kind": "parameter",
                    "labelHint": "Delay between scheduling (executing).",
                    "required": false,
                    "secret": false,
                    "type": "duration"
                },
                "template":
                {
                    "componentProperty": false,
                    "deprecated": false,
                    "displayName": "Template",
                    "group": "producer",
                    "javaType": "java.lang.String",
                    "kind": "path",
                    "labelHint": "Stored Procedure template to perform.",
                    "required": true,
                    "secret": false,
                    "type": "hidden"
                }
            },
            "description": "Select the stored procedure."
        }
    ],
[...]

PR #6095, opened 2019-07-09
fix: DB procedure name changed to text input instead of dropdown #5690

KurtStam:
We have two copies of this file now, and this PR is trying to manually resolve the differences.

fixes regressions: #5690 and #6014


zregvart:
We, for the time being, have duplicate definitions for template and the operator, so same change needs to go [here](https://github.com/syndesisio/syndesis/blob/master/install/operator/pkg/generator/assets/install/04-syndesis-db.yml.mustache) also.

zregvart:
Perhaps as a follow-up we can move this to an addon?

zregvart:
Doesn't need a backport, the operator on `1.7.x` doesn't use this copy of the template and the template it uses already has this changes:

https://github.com/syndesisio/syndesis/blob/88c5d69775b6d6f4abcc5e524b3fb585a9823b7b/install/syndesis.yml#L565-L599

PR #6105, opened 2019-07-09
user doc: Add doc for specifying ui properties in Java class files

TovaCohen:
Plus small updates to the new topic about defining user interface properties. 
Plus cross-references from the step extension example code to the new topic about defining ui properties. Luca reviewed the new content.


zregvart:
Backport in #6134

PR #6137, opened 2019-07-11
Changes the pod names in the templates to match the actual container names

orpiske:
Sending this as a proposed fix for https://issues.jboss.org/browse/ENTESB-11030. 

Doing it on the template generator is, IMHO, easier to maintain than implementing a bunch of regex on the productization scripts.

### DIFF
--- a/install/generator/07-syndesis-upgrade.yml.mustache
+++ b/install/generator/07-syndesis-upgrade.yml.mustache
@@ -19,7 +19,7 @@
       app: syndesis
       syndesis.io/app: syndesis
       syndesis.io/type: infrastructure
-    name: syndesis-upgrade
+    name: {{ProductName}}-upgrade
     annotations:
       openshift.io/display-name: "Syndesis Upgrade"
       description: |-
@@ -54,7 +54,7 @@
   - apiVersion: v1
     kind: Pod
     metadata:
-      name: syndesis-upgrade-{{ Tags.Upgrade }}
+      name: {{ProductName}}-upgrade-{{ Tags.Upgrade }}
     spec:
       serviceAccountName: syndesis-operator
       containers:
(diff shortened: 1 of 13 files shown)


zregvart:
Thanks @orpiske on `master` (1.8 onward) we have changed how the operator to work with another set of templates in `install/operator/pkg/generator/assets` so we need a corresponding change in there also and a change to the operator code to include the new `ProductName` parameter.

We can either merge this and do a follow-up issue/pull request or if you feel up to it you can add this to this pull request.

orpiske:
Thanks @zregvart! I will look into changing that as well and I'll let you know if I can do it in a timely manner.

install/generator/07-syndesis-upgrade.yml.mustache
```diff
@@ -1,7 +1,7 @@
 - apiVersion: v1
   kind: PersistentVolumeClaim
   metadata:
-    name: syndesis-upgrade
+    name: {{ProductName}}-upgrade
```
orpiske: This is a mistake on my part. The intention is to change only the pod names. I will fix this (and other occurrences if any).

orpiske:
@zregvart I included the changes in the other templates you mentioned and included the ProductName in the Context struct. 

I don't think this is complete yet since I couldn't test it. It's my very first PR in Go, so I may ask you or @nicolaferraro a few questions about that next week :)

orpiske:
So, this does seem to do the trick. I was hoping to do more testing and I looked into modifying generator_test.go for that, but it looks like I cannot descend to the part that contains the actual container names so I can write a test for it.

zregvart:
@orpiske I think this is fine, we're severely lacking tests in the operator code so this is not an exception. If you think this is as far as you can run with this we can merge it and improve upon it by adding tests and whatnot in subsequent PRs.

Thanks!

PR #6156, opened 2019-07-12
fix: Fuse online documentation links to 7.4

### LINKED ISSUE #6127: Links to user doc go to 7.3 doc when they should target 7.4 URLs
## This is a...

<pre><code>
[ ] Feature request
[X] Bug report  
</code></pre>

## Description

I noticed this behavior in the staging site, and Christoph just confirmed that this is the situation in the latest 1.7.x.  

When I click the ? (Help) icon and then click User Guide, I get the 7.3 user guide at this link: 

https://access.redhat.com/documentation/en-us/red_hat_fuse/7.3/html-single/integrating_applications_with_fuse_online/index

I expected the help icon links would now point to 7.4 titles, with links like this: 

https://access.redhat.com/documentation/en-us/red_hat_fuse/7.4/html-single/integrating_applications_with_fuse_online/index

which would display "404 Not Found" until we publish the 7.4 doc. 

I updated the release number in app/ui-react/syndesis/src/i18n/locales/shared-translations.en.json to 7.4: 

 "project": {
        "name": "Syndesis",
        "version": "7.4"

Also in that file, there are user doc links that contain references to $t(project.version)

My update to the shared-translations.en.json file is in this commit:
https://github.com/syndesisio/syndesis/commit/214b64925d99a20c8b26426f784c0f843eded256

So it looks like I did not update the right file. 
What file are the help links coming from?
And just to be clear, there are three links to user doc and they all have the same problem.

christophd:
Fixes #6127

### DIFF
--- a/app/ui-react/syndesis/src/app/UI.tsx
+++ b/app/ui-react/syndesis/src/app/UI.tsx
@@ -228,19 +228,19 @@ export const UI: React.FunctionComponent<IAppUIProps> = ({ routes }) => {
                                   }}
                                   onSelectSampleIntegrationTutorials={() => {
                                     window.open(
-                                      'https://access.redhat.com/documentation/en-us/red_hat_fuse/7.3/html-single/fuse_online_sample_integration_tutorials/',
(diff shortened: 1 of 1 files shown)


TovaCohen:
Thanks!!!!!!!

Issue #5215, opened 2019-04-15
[conditional-flows] UI data shape handling of CBR step


christophd:
- A CBR step can be inserted at any point in an existing flow. When added the step adopts the output data shape of the previous step and this is the fixed starting shape of all conditional flows added (= first step in conditional flow provides this shape as fixed output shape).
- By default the CBR step has no output shape defined so the conditional flows can produce any output and the primary flow is not able to access the shapes created within the conditional flow.
- Changes to the input/output data shapes of the CBR step must be propagated to all sub-flows

christophd:
Could be also done with #5181

lburgazzoli:
I think that a CBR step could be handled in a way similar to the "shapeless connectors" so the user is prompted to provide a shape when the CBR step is added to the flow.

In case a step already exists after the CBR one, the UI should offers also an option to set the output shape to the one of the subsequent step.

lburgazzoli:
We eventually need to propagate warning on a flow up to the source so i.e. if the data shape of a CBR is changed, all the flow could potentially need to be updated thus the CBR step should report a warning till all the flows have been fixed.

christophd:
I have to move the user defined data shape handling to a separate issue. Most likely this particular feature will not make it to 7.4.

christophd:
@gashcrumb @mcada FYI this seems to be broken with react UI. The conditional flows step adopts the shape successfully but the shape is not set as fixed output shape to the 1st step of each conditional flow.

gashcrumb:
Yep, you're right it's totally not doing that at the moment.  I'll get that implemented tomorrow.

mastepan:
@gashcrumb please change pipeline to Done if the issue's been fixed. Ignore this otherwise. Thanks.

christophd:
@gashcrumb has this fix https://github.com/syndesisio/syndesis-react/pull/470 been backported to 1.7.x already? I still see the error on 1.7.x branch

gashcrumb:
Hm, it should be that was done before we branched for release.  I'll check
in a few, what integration did you test with?

christophd:
I wrote a new integration `SQL => Split => Conditional Flow => Log` The sub flow should be provided with the single element SQL shape as start connection

gashcrumb:
This is where it's done on master at least => https://github.com/syndesisio/syndesis/blob/master/app/ui-react/packages/api/src/helpers/integrationFunctions.ts#L702-L709
[...]

PR #6167, opened 2019-07-15
fix: use band-aid to discard old action definition when a new connection is set  

### LINKED ISSUE #6158: [Output data shape] Data shape change is not propagated to following step
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Changing step inside of an integration does not change output data shape. This leads to following:

Original integration:
screenshot_20190715_095535

Changing first step to webhook:
screenshot_20190715_095515

Note that output data shape is still SQL Result. There was also no option to define webhook output data shape. Following data mapper still contains previous setting which will obviously not work:

screenshot_20190715_095558

gashcrumb:
fixes #6158


christophd:
@gashcrumb is this also fixing #5215?

Issue #6100, opened 2019-07-09
Base docker S2I image with camel-fuse.x.y.z dependencies


balejos:
Coming from a Retro: The creation of a docker image with dependencies could help as a way to facilitate the development environment set-up.

balejos:
Is still valid? Please, could you follow up on yesterdays discussion? @squakez @zregvart @KurtStam

squakez:
Just a comment to frame properly the discussion. The issue I've been stumbled upon is about the need to recreate the S2I docker image that is used by runtime for building integration when there is any change that is needed by the integration (ie, changing a connector business logic).

This is the dockerfile I was referring: https://github.com/syndesisio/syndesis/blob/master/app/s2i/src/main/docker/Dockerfile

If instead the actual FROM we have a specific FROM of an image with pre-cached camel-fuse/syndesis dependencies in its local maven we'd spare all the actual download.

Another approach proposed was to install a local Nexus to proxy all requests locally instead, that could be a valid alternative.

I am not sure if however the effort is justified since we should push on Camel-K. I don't have yet enough info about the latter, neither enough experience on the project to take a proper decision.

lburgazzoli:
Another option would be to take into account camel bits in any case, not only for [snapshots](https://github.com/syndesisio/syndesis/blob/master/app/s2i/copy_mvn_repo.sh#L17-L24)

squakez:
Thanks @lburgazzoli. In any case I understand we should better dedicate efforts to move towards Camel-K instead. I've just seen @nicolaferraro is exactly working on improvements to make connectors development smoother. What do you think?

lburgazzoli:
IMHO it make sense to spend some time on this issue as it could be useful even in the context of camel-k where we could use a pre configured s2i image with all the dependencies for the camel-k operator itself, see https://github.com/syndesisio/syndesis/issues/4519

squakez:
Good to know then. I'll try to dedicate some time and ping you and the rest of the team if need some help/advice.

squakez:
The problem of the Docker file is that it cannot make any use of intermediate layer caching because the [presence of the local copy of SNAPSHOT dependencies](https://github.com/syndesisio/syndesis/blob/master/app/s2i/src/main/docker/Dockerfile#L10).

In order to prevent this we can add a stable previous version generated project whose goal is to warm up and cache all integration dependencies. This project won't change during the rest of development, so the intermediate docker layer will be cached and always available at image build time.

[...]

PR #6169, opened 2019-07-16
#6100 - create cacheable intermediate layer for S2I image

squakez:
#6100 The copy script would create a local cache with syndesis dependencies. This is filled in only the first time or every time a `mvn clean` is performed.

The docker image execute a `mvn verify` step to download dependencies (using cached syndesis jars, it would be the cached intermediate layer), and a `mvn package` that would take care to refresh any further dependency (or out of synch).


app/s2i/copy_mvn_repo.sh
```diff
@@ -12,6 +12,19 @@ fi
 
 mkdir -p "${target_dir}/m2"
 cd ${local_repo}
+
+if [[ ! -d "${target_dir}/m2.cache" ]]; then
+    mkdir -p "${target_dir}/m2.cache"
```
nicolaferraro: This should be idempotent, no need to check ;)

app/s2i/src/main/docker/Dockerfile
```diff
@@ -34,4 +45,4 @@ RUN cd /tmp/artifacts/m2/project \
 RUN mkdir -p /opt/ipaas/
 COPY lic* /opt/ipaas/
 
-USER 1000
+USER 1000
```
nicolaferraro: Github likes newlines

app/s2i/copy_mvn_repo.sh
```diff
@@ -12,6 +12,19 @@ fi
...
+fi
+
+if [[ ! "$(ls -A ${target_dir}/m2.cache)" ]]; then
+     echo "Going to cache locally Syndesis ${version} dependencies. Clean target directory to refresh the cache."
+     find ./io/syndesis | grep -F -- "/${version}/" | grep -v -- "-sources\." | grep -v -- "-tests\." | xargs tar -c | tar -vx -C "${target_dir}/m2.cache"
```
nicolaferraro: Here if you initialize the `m2/.cache` with dependencies from your local maven (`mvn copy:dependencies`) you add the benefit of avoiding a "download the maven world", and will keep the benefits of layering.

zregvart:
From what I see, this adds `target/m2.cache` and copies over Syndesis dependencies from `~/.m2/repository` and then within the Docker container runs a Maven build that uses this `m2.cache` directory copied over.
What I'm failing to understand is, how is this different from the build we already have, since it's doing much the same thing and the old bits are kept with this change?

squakez:
> From what I see, this adds `target/m2.cache` and copies over Syndesis dependencies from `~/.m2/repository` and then within the Docker container runs a Maven build that uses this `m2.cache` directory copied over.
> What I'm failing to understand is, how is this different from the build we already have, since it's doing much the same thing and the old bits are kept with this change?

The problem reside in this statement https://github.com/syndesisio/syndesis/blob/master/app/s2i/src/main/docker/Dockerfile#L10
As the `copy_mvn_repo.sh` script always refresh snapshot dependencies, then Docker think it must recreate the below layers from scratch.

> From what I see, this adds `target/m2.cache` and copies over Syndesis dependencies from `~/.m2/repository` and then within the Docker container runs a Maven build that uses this `m2.cache` directory copied over.
> What I'm failing to understand is, how is this different from the build we already have, since it's doing much the same thing and the old bits are kept with this change?

The idea is that the `mvn verify` will be executed only the first time or at any target clean because we fix locally a copy of dependencies that are not going to be refreshed at every build.

zregvart:
And as I understand I would need to run a clean build to get the S2I image updated?

If `-SNAPSHOT` dependency changed then I would expect that the docker image also needs to change. Perhaps we could refactor this to copy over the `-SNAPSHOT` dependencies to the last layer added instead?

I would suggest that we create the cached Maven repository outside of the docker build so it can take advantage of cached artefacts in the local Maven repository and copy the resulting repository to the image.

nicolaferraro:
> And as I understand I would need to run a clean build to get the S2I image updated?
> 
> If `-SNAPSHOT` dependency changed then I would expect that the docker image also needs to change. Perhaps we could refactor this to copy over the `-SNAPSHOT` dependencies to the last layer added instead?
> 
> I would suggest that we create the cached Maven repository outside of the docker build so it can take advantage of cached artefacts in the local Maven repository and copy the resulting repository to the image.

I think that's almost what he's doing:
- phase 1: copy snapshots, download dependencies
- phase 2: override snapshots, download dependencies (if changed)

Phase 1 is skipped from second time you build.

As you said, if phase 1, instead of downloading dependencies, could copy them from local maven, this would be awesome, because it would improve build time, but it also allows to work with external (read Camel) snapshots.

squakez:
> And as I understand I would need to run a clean build to get the S2I image updated?
> 

Not really, as the second build would bring the refreshed ones and would update accordingly. Let's say that the second build would take care of the incremental changes that there are between the 1.x-cached and the 1.x-refreshed. The first would bring majority of dependencies that you won't be touching. As in my case, working on a connector, I only refresh that connector dependencies, the rest won't be touched.

> If `-SNAPSHOT` dependency changed then I would expect that the docker image also needs to change. Perhaps we could refactor this to copy over the `-SNAPSHOT` dependencies to the last layer added instead?
> 
> I would suggest that we create the cached Maven repository outside of the docker build so it can take advantage of cached artefacts in the local Maven repository and copy the resulting repository to the image.

Yep, that's what @nicolaferraro was mentioning. I'm working right now on that part making sure that by running `mvn dependency:copy-dependencies` is still valid as a caching layer.

squakez:
Just an head up after several experiments I've run. Actually there is a little limitation with the BOM dependencies, as they are not imported through the `mvn copy` plugin.

```
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-dependencies</artifactId>
        <version>${spring-boot.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
      <dependency>
        <groupId>org.apache.camel</groupId>
        <artifactId>camel-spring-boot-dependencies</artifactId>
        <version>${camel.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
      <dependency>
        <groupId>io.syndesis.integration</groupId>
        <artifactId>integration-bom</artifactId>
        <version>${syndesis.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
```
are not imported, so, unless copied through script (as we're doing for syndesis) the plugin won't work. I can issue a new request and work on that one separately from the issue I was solving here (docker caching layer).

lburgazzoli:
@squakez what about using a custom maven settings like:

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<settings 
  xmlns="http://maven.apache.org/SETTINGS/1.0.0" 
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
    
  <localRepository>./target/repo</localRepository>
  
  <profiles>
    <profile>
      <id>_local</id>
      <activation>
        <activeByDefault>true</activeByDefault>
      </activation>
      <repositories>
        <repository>
          <id>_local</id>
          <url>file://${user.home}/.m2/repository</url>
          <releases>
            <enabled>true</enabled>
          </releases>
          <snapshots>
            <enabled>true</enabled>
          </snapshots>
        </repository>
      </repositories>
      <pluginRepositories>
        <pluginRepository>
          <id>_local</id>
          <url>file://${user.home}/.m2/repository</url>
          <releases>
            <enabled>true</enabled>
          </releases>
          <snapshots>
            <enabled>true</enabled>
          </snapshots>
        </pluginRepository>
      </pluginRepositories>
    </profile>
  </profiles>
</settings>
```

And:

```
mvn -s settings.xml  verify
```

squakez:
I've created this issue to track the new requests: #6185 

The scope of this PR will remain according original #6100 - caching Docker layer.

squakez:
Yeah, still have to commit the changes as suggested by @nicolaferraro. I am having some issue with minishift today, I'll push as soon I can confirm locally works fine. Thanks!

PR #6172, opened 2019-07-16
fix for DB procedure name changed to text input instead of dropdown

### LINKED ISSUE #5690: DB procedure name changed to text input instead of dropdown

KurtStam:
Fixes #5690

### DIFF
--- a/app/connector/sql/src/main/resources/META-INF/syndesis/connector/sql.json
+++ b/app/connector/sql/src/main/resources/META-INF/syndesis/connector/sql.json
@@ -151,7 +151,7 @@
                 "required": false,
                 "secret": false,
                 "tags": [],
-                "type": "string"
+                "type": "select"
               },
               "template": {
                 "componentProperty": false,
@@ -207,7 +207,7 @@
                 "required": true,
                 "secret": false,
                 "tags": [],
-                "type": "string"
+                "type": "select"
               },
               "schedulerExpression": {
                 "defaultValue": 60000,
--- a/app/common/model/src/main/java/io/syndesis/common/model/action/ConnectorDescriptor.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/action/ConnectorDescriptor.java
@@ -52,41 +52,60 @@ public interface ConnectorDescriptor extends ActionDescriptor, WithConfiguredPro
             return this;
         }
 
-        public Builder replaceConfigurationProperty(
-            final String propertyName,
-            final Consumer<ConfigurationProperty.Builder> configurationPropertyConfigurator) {
+        public ConfigurationProperty findProperty(
+                final String propertyName) {
+
+            PotentialStep potentialStep = findStepForProperty(propertyName);
+            if (potentialStep!=null) {
+                return potentialStep.step.getProperties().get(propertyName);
+            } else {
+                return null;
+            }
+        }
+
+        /* But what if steps have the same property name? It may return the wrong step */
+        private PotentialStep findStepForProperty(
+                final String propertyName) {
 
             final ConnectorDescriptor definition = build();
             final List<ActionDescriptorStep> steps = definition.getPropertyDefinitionSteps();
 
-            int stepIdx;
-            ActionDescriptorStep step = null;
-            for (stepIdx = 0; stepIdx < steps.size(); stepIdx++) {
+            for (int stepIdx = 0; stepIdx < steps.size(); stepIdx++) {
                 final ActionDescriptorStep potentialStep = steps.get(stepIdx);
 
                 final Map<String, ConfigurationProperty> properties = potentialStep.getProperties();
                 if (properties.containsKey(propertyName)) {
-                    step = potentialStep;
-                    break;
+                    return new PotentialStep(stepIdx, potentialStep);
                 }
             }
 
-            if (step == null) {
+            return null;
+        }
+
+        public Builder replaceConfigurationProperty(
+            final String propertyName,
+            final Consumer<ConfigurationProperty.Builder> configurationPropertyConfigurator) {
+
+            PotentialStep potentialStep = findStepForProperty(propertyName);
+
+            if (potentialStep == null) {
                 // found no property to replace, lets just ignore it
                 return this;
             }
 
-            final ConfigurationProperty configurationProperty = step.getProperties().get(propertyName);
+            final ConfigurationProperty configurationProperty = potentialStep.step.getProperties().get(propertyName);
             final ConfigurationProperty.Builder configurationPropertyModifier = new ConfigurationProperty.Builder().createFrom(configurationProperty);
 
             configurationPropertyConfigurator.accept(configurationPropertyModifier);
 
             final ActionDescriptorStep.Builder stepModifier = new ActionDescriptorStep.Builder()
-                .createFrom(step)
+                .createFrom(potentialStep.step)
                 .putProperty(propertyName, configurationPropertyModifier.build());
 
+            final ConnectorDescriptor definition = build();
+            final List<ActionDescriptorStep> steps = definition.getPropertyDefinitionSteps();
             final List<ActionDescriptorStep> modifiedSteps = new ArrayList<>(steps);
-            modifiedSteps.set(stepIdx, stepModifier.build());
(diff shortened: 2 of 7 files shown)


app/connector/sql/src/main/resources/META-INF/syndesis/connector/sql.json
```diff
@@ -151,7 +151,7 @@
                 "required": false,
                 "secret": false,
                 "tags": [],
-                "type": "string"
+                "type": "select"
```
gashcrumb: Is this also gonna supply the `enum` property?  The form component is gonna need some values to populate this.

KurtStam: ahh - you need more then the type. What is the syntax of that?

zregvart:
I think this bit needs to change:

https://github.com/syndesisio/syndesis/blob/1af40ec5573fa15cbe6a710a1842c17c59771109/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectionActionHandler.java#L195-L204

We need to remove `if` statements that look into `suggestions.getValue().size()`.

I guess we had this case handled in the Angular UI not to show any choice when there is only a single choice. So be wary, this will impact all connectors if changed.

KurtStam:
@zregvart y was just looking at that. The good news is it's already enum syntax for > 1

				```
	"type": "select",
					"enum": [
						{
							"label": "create_lead",
							"value": "create_lead"
						},
						{
							"label": "create_lead2",
							"value": "create_lead2"
						}
					]
```

gashcrumb:
Yeah, if it's a single-value enum then the UI will just show a select element with one possible option, and default to using that option (the UI picks the first entry if no `defaultValue` is set).  We can probably improve on this in the future.

zregvart:
Also of note is #5191 that removes `enum` if there is only a single suggestion from the meta. (cc @christophd)

KurtStam:
That is exactly what's causing this no?

christophd:
I have found it very confusing to have the indication of a select drop down and when you open it you only have one single element in it.

christophd:
maybe add a new tag to the configured property that prevents auto removal of enum during meta lookup. 

So this tag would be for preserved enum value properties where the user should not have a chance to input something (e.g. procedure names). For other single enum value properties without the tag it is a text field with default value.

gashcrumb:
I think that complicates the semantics even more.  For fields that are open-ended there's a `dataList` property that auto-form supports, that then acts as a typeahead control.  When you use `enum` you're saying "these are the only possible values".  Now, in the past that wasn't the case, but here I'm putting my foot down and saying now it's the case. :smile:

christophd:
I am not tied to the fix in #5191 for now but we should find a better solution for the single element drop downs then in some next release

KurtStam:
I guess I don't understand why a single element dropdown is an issue.

gashcrumb:
It's very contentious!

No, it's not the hugest deal but it's kinda weird as it gives the impression that there's a choice, when in fact there is not.

christophd:
So I don't care too much for single value drop downs or not but there must be a way to have a single value property suggestion to be rendered as a text field. 

This is when the meta data lookup suggests a default value (maybe from some 3rd party API lookup) but the user is able to overwrite the suggestion with a custom value. This is used in GoogleSheets for instance where the meta lookup catches the column names (A,B,C,D, ...) from the data range and returns this as a single property suggestion. But the user can overwrite this suggestion with custom value in a text field, too.

I think this use case was the reason for the fix in #5191.

gashcrumb:
Yep in that case use `text` and then use `dataList` for the suggestions.  This stuff will make it into the product docs, for now I've a writeup [here](https://issues.jboss.org/browse/FUSEDOC-2916?focusedCommentId=13754716&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13754716).  The `dataList` attribute is a new thing for the react codebase, it's currently used on the basic filter page.

KurtStam:
OK plan is to rollback the fix in #5191, then commit this PR to mark it as a 'select'. Agreed?

christophd:
@KurtStam I guess we also need to apply this new `dataList` (not sure what to do here in detail) thing to the GoogleSheets property to not break this one by reverting #5191

christophd:
btw it is this property in google-sheets connector that I am talking about: https://github.com/syndesisio/syndesis/blob/077597925081f628d568dd0676f1f15641e9a8e6/app/connector/google-sheets/src/main/resources/META-INF/syndesis/connector/sheets.json#L273
https://github.com/syndesisio/syndesis/blob/077597925081f628d568dd0676f1f15641e9a8e6/app/connector/google-sheets/src/main/resources/META-INF/syndesis/connector/sheets.json#L414

and these get suggested by meeta data here: https://github.com/syndesisio/syndesis/blob/42955e96157019e2fd9164bf8e5339f0074268fa/app/connector/google-sheets/src/main/java/io/syndesis/connector/sheets/GoogleSheetsMetadataRetrieval.java#L64

KurtStam:
So try setting `"type": "string"`

https://github.com/syndesisio/syndesis/blob/077597925081f628d568dd0676f1f15641e9a8e6/app/connector/google-sheets/src/main/resources/META-INF/syndesis/connector/sheets.json#L284

to `dataType`

christophd:
removing the backport to 1.7.x label as the issue has been downgraded to p1

christophd:
as this PR is no longer affecting the 1.7.x branch we can also just revert #5191 and use the single value drop down for enum values as suggested before. 

and then see what that implies to the google-sheets property usage and try to fix this in a separate PR/issue then.

KurtStam:
@christophd yeah no worries, there is no time pressure on this. If it turns out just the change in property I will take care of it in the PR.

zregvart:
Perhaps we need to involve @syndesisio/uxd folk in this, and design our decisions around a principle. I see the following cases:

 1. multiple choices offered, user must pick one of the provided choices
 1. multiple choices offered, user can input their own value not present in the offered choices
 1. single choice offered, user must pick it
 1. single choice offered, user can input their own value different from the one offered

(perhaps others also)

I think we need to map these cases to UI implementation and agree on the values/conditions so that the UI can determine what implementation to use based on the data provided from the server.

gashcrumb:
@KurtStam @christophd I need to make a proper example story in our storybook, but [here's](https://github.com/syndesisio/syndesis/blob/master/app/ui-react/syndesis/src/modules/integrations/components/editor/ruleFilter/WithRuleFilterForm.tsx#L87) the only example at the moment using `dataList` with a `text` input.  In this case the `filterOptions.paths` value (this comes back from the server) is an array of strings.

gashcrumb:
> Perhaps we need to involve @syndesisio/uxd folk in this, and design our decisions around a principle.

@zregvart That's the dream :smile: 

Ideally I can make autoform storybook examples of these things too so there's concrete examples to look at.

KurtStam:
@christophd @gashcrumb I think this will do the trick, it now supports dataList for suggestions needed for googleSheets. So @gascrumb I guess google sheet break with this PR until the UI starts using dataList on type 'text'.

gashcrumb:
>  So @gascrumb I guess google sheet breaks with this PR until the UI starts using dataList on type 'text'.

What are you talking about?  `dataList` is *intended* to be used alongside a `text` input.  Was my example not clear enough then?

@christophd btw it should look pretty much as it does on the basic filter page, a downward pointing caret if there's values available that the user can click on.

KurtStam:
@gashcrumb so here is the logic : backend sets a collection of suggested values

- if the `type` is 'select' it populated the `enums` property with an array of enums (key, value)
- if the `type` is 'text' or 'string' it populates the `dateType` property with ban array of strings
- if the `type` is anything else it sets the first value in the collection in a property call `defaultValue`

gashcrumb:
ok perfect, was confused :smile: 

Support is already there, so it should just work.  Theoretically.  Unless it's broken.

Uh, but it shouldn't be I think.

PR #6183, opened 2019-07-16
chore: Update to AtlasMap 1.41.1

igarashitm:


gashcrumb:
Does this need to be backported to 1.7.x?

igarashitm:
Thanks! nope, this should only go master as it contains a bunch of non-p0 changes including enhancements and refactorings.

PR #6187, opened 2019-07-17
fix(6141): FTP Verifier handles empty username & password properties

phantomjinx:
* ConnectorOptions
 * Provides function for returning a default value rather than null if
   property in options map is not present or if it is empty

* FtpVerifierExtension
 * Uses the ConnectorOptions to extract options and ensure they are either
   null or empty as required.

### DIFF
--- a/app/connector/ftp/pom.xml
+++ b/app/connector/ftp/pom.xml
@@ -37,6 +37,11 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
+      <groupId>io.syndesis.connector</groupId>
+      <artifactId>connector-support-util</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
       <groupId>org.apache.camel</groupId>
       <artifactId>camel-core</artifactId>
     </dependency>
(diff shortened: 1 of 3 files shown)


zregvart:
Thanks!

I'm not too sure about having a dependency on `io.syndesis.common:common-util` in connectors, I think connectors should depend only on stuff in `io.syndesis.connector`, we probably do already. That pattern to `implements StringConstants` to provide few characters shorter access to a constant is really not my cup of tea, feels 2000s, and we have static imports now if that's what we'd like.

This is all just a matter of personal taste, nothing to prevent this from merging.

phantomjinx:
> Thanks!

No worries.

> I'm not too sure about having a dependency on `io.syndesis.common:common-util` in connectors, I think connectors should depend only on stuff in `io.syndesis.connector`, we probably do already. That pattern to `implements StringConstants` to provide few characters shorter access to a constant is really not my cup of tea, feels 2000s, and we have static imports now if that's what we'd like.

Removed. It was 1 literal so not necessary.
 
> This is all just a matter of personal taste, nothing to prevent this from merging.

(Am dev of the 2000s so such habits might be difficult to break!)

Repushed so awaiting a rebuild.

phantomjinx:
@zregvart 
Would you mind approving this for me (when you have a minute)?
Thx.

PR #6189, opened 2019-07-17
Dont use the same file that is used by template install

lgarciaaco:
Use a different file for upstream installation, dont mess around the template installation file

Add a option to target a new github user, for testing purposes mainly, default to _syndesisio_

Fixes https://github.com/syndesisio/syndesis/issues/5925


heiko-braun:
@nicolaferraro I see what you mean, but would say this is good enough for now.

lgarciaaco:
> It would be more flexible if you could set the whole URL pattern.. e.g. setting a "https://...%s/file.yaml" where you'll replace later the '%s' with the tag. Otherwise this requires that you move the template to another repo..

We have talked about restricting the flexibility a user has, in order to avoid possible issues introduced by the user when setting these parameters. I think changing the org in github is enough for testing purposes (which is why I introduce it) since everyone of us should have a fork of the `fuse-online-install` with the same exact  path just the branch/tag and org change, which are parameterized

Issue #6118, opened 2019-07-10
Datamapper fails with split in some cases


mmuzikar:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
We have had problems with datamapper causing errors when using split and aggregate steps. You can import this integration to see the error:
https://drive.google.com/open?id=1aNSEt21U-NW8tyUrGo5JbL23V_tkNMQc
Since it's an API provider, you need to invoke the flow with:
```
curl -X POST <url> -d '[{"id":2,"completed":1,"task":"task2"},{"id":3,"completed":1,"task":"task3"}]'
```
This is the resulting stacktrace (there are multiple errors so choose one you find most useful :D):
[i-todo-integration-post-collection-2-6qkd4.log](https://github.com/syndesisio/syndesis/files/3377466/i-todo-integration-post-collection-2-6qkd4.log)

mmelko:
I'm seeing the same issue in integration from db -> split -> dm -> aggregate -> google sheets:
```java
o.atlasmap.api.AtlasException: Errors: [A Target field contained in a collection can not be selected with other Target field: ['/<>/A']: docId='mapping.121775', path='null'], [A Target field contained in a collection can not be selected with other Target field: ['/<>/B']: docId='mapping.121775', path='null'], [A Target field contained in a collection can not be selected with other Target field: ['/<>/C']: docId='mapping.121775', path='null'], [A Target field contained in a collection can not be selected with other Target field: ['/<>/D']: docId='mapping.121775', path='null'], [A Target field contained in a collection can not be selected with other Target field: ['/<>/E']: docId='mapping.121775', path='null'], [A Target field contained in a collection can not be selected with other Target field: ['/<>/F']: docId='mapping.121775', path='null'], [A Target field contained in a collection can not be selected with other Target field: ['/<>/G']: docId='mapping.121775', path='null'], 
	at org.apache.camel.component.atlasmap.AtlasEndpoint.onExchange(AtlasEndpoint.java:262) ~[camel-atlasmap-1.40.0.jar!/:1.40.0]
	at org.apache.camel.impl.ProcessorEndpoint$1.process(ProcessorEndpoint.java:71) ~[camel-core-2.21.0.fuse-740028.jar!/:2.21.0.fuse-740028]
[...]

PR #6193, opened 2019-07-17
fix: explicitly set IDs for steps too; add comments

gashcrumb:
part of #6118


gashcrumb:
`stepKind` coming into that code had an `id: undefined` in it.  Since `id` is so important, maybe it's best to just ensure it's explicitly set :smile:

Issue #6111, opened 2019-07-10
[1.7.9] Komodo server doesn't start


avano:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
Because this property should be true/false, but it is not set during the template generation:

https://github.com/syndesisio/syndesis/blob/308b4cffd29cf85cb3276bec39f8ed775bce1efb/install/syndesis.yml#L2105-L2106

that results in this error in komodo-server:

```
***************************
APPLICATION FAILED TO START
***************************
Description:
Binding to target org.komodo.rest.KomodoConfigurationProperties@4466af20 failed:
    Property: controllers.exposeVia3scale
    Value: 
    Reason: Failed to convert property value of type 'java.lang.String' to required type 'boolean' for property 'exposeVia3scale'; nested exception is org.springframework.core.convert.ConverterNotFoundException: No converter found capable of converting from type [java.lang.String] to type [boolean]

```

heiko-braun:
@rkorytkowski Since we've been talking about operator changes recently, do you mind taking a look at this one too?

rkorytkowski:
I'm on it!

heiko-braun:
@rkorytkowski tnx!

rkorytkowski:
It needs to be back ported to 1.7.x.

heiko-braun:
@rkorytkowski You need to backport it to 1.7.x

heiko-braun:
Usually a label `backport 1.7.x` on the PR is sufficient, but I guess the operator related differences in both branches would ask for a manual PR

heiko-braun:
@rkorytkowski 

```
[ERROR] createAndGetIntegration(io.syndesis.server.runtime.IntegrationsITCase)  Time elapsed: 4.376 s  <<< ERROR!
java.lang.IllegalStateException: Failed to load ApplicationContext
[...]

Issue #5995, opened 2019-06-28
Remove circular dependency between API handlers for environment management


dhirajsb:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

There is a circular dependency between `PublicAPIHandler` and `IntegrationHandler` due to the need to handle environment tags removal when deleting an integration. Refactor this out into an `environment helper` class when creating a separate environment entity (with it's own ids and not just a property of integration) to simplify handlers and remove this circular dependency.

dhirajsb:
This concern was raised in #5952

PR #6208, opened 2019-07-19
Adds a GraphQL Endpoint

chirino:


chirino:
Still a wip, this is just a rebase of some previous work I had done to add a GraphQL endpoint.  Basically I need update the UI to actually do something useful with it

app/server/runtime/src/main/java/io/syndesis/server/runtime/GraphQLConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+public class GraphQLConfiguration {
+
+    @Autowired
+    IntegrationHandler integrationHandler;
+
+    @Bean
+    GraphQLSchema schema() throws IOException {
+
+        String schema;
+        try (InputStream is = Thread.currentThread().getContextClassLoader().getResourceAsStream("schema.graphqls") )  {
```
zregvart:
```suggestion
        try (InputStream is = GraphQLConfiguration.class.getResourceAsStream("schema.graphqls") )  {
```

I think we have utility method that loads from classpath, lookup `Resources` utility class.

app/server/runtime/src/main/java/io/syndesis/server/runtime/GraphQLConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+        TypeDefinitionRegistry typeDefinitionRegistry = schemaParser.parse(schema);
+
+        SchemaGenerator schemaGenerator = new SchemaGenerator();
+        return schemaGenerator.makeExecutableSchema(typeDefinitionRegistry, newRuntimeWiring()
+
+            // Entities who's json shape matches that GraphQL schema,
+            // just need a a bit of code like this to load the data for the query.
+            .type("Query", builder -> builder
+                .dataFetcher("integrations", env->{
+                    return integrationHandler.list(null);
```
zregvart: I'd avoid this coupling of handler to a handler, we have already done this and it has caused only issues.

app/pom.xml
```diff
@@ -542,22 +546,28 @@
...
-                <arg>-XepExcludedPaths:.*/generated-sources/.*</arg>
-                <arg>-implicit:none</arg>
-                <!-- temporary disable until https://github.com/google/error-prone/issues/780 is resolved -->
-                <arg>-Xep:ParameterName:OFF</arg>
-              </compilerArgs>
-            </configuration>
+            <executions>
+              <execution>
+                <id>default-compile</id>
+                <phase>compile</phase>
```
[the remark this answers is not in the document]
chirino: thats exactly the point.  This change disable the lint checks on test cases.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/operations/PaginationOptionsFromQueryParams.java
```diff
@@ -33,19 +33,24 @@
      * @param uri The request context.
      */
     public PaginationOptionsFromQueryParams(UriInfo uri) {
-        MultivaluedMap<String, String> queryParams = uri.getQueryParameters();
-        String pageQuery = queryParams.getFirst("page");
-        if (pageQuery == null || pageQuery.isEmpty()) {
+        if( uri == null ) {
```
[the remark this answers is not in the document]
chirino: This just default faults the value if URI is null which is the case when the graphql enpoint uses it.

app/server/runtime/src/main/java/io/syndesis/server/runtime/GraphQLConfiguration.java
chirino: Cool.. I'll switch to that.

app/server/runtime/src/main/java/io/syndesis/server/runtime/GraphQLConfiguration.java
chirino: can you point me at what those issues were?

chirino:
This does have a unit test at: https://github.com/syndesisio/syndesis/pull/6208/files#diff-95425df74de456611e2f7ae014ad229dR88

app/pom.xml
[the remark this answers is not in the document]
zregvart: Than -1 on that. There's no reason for tests to be worst off that production code.

app/server/runtime/src/main/java/io/syndesis/server/runtime/GraphQLConfiguration.java
zregvart: #5995 and #6111

zregvart:
> This does have a unit test at: [syndesisio/syndesis/pull/6208/files#diff-95425df74de456611e2f7ae014ad229dR88](https://github.com/syndesisio/syndesis/pull/6208/files#diff-95425df74de456611e2f7ae014ad229dR88)

That would be an integrated test not a unit test, it starts the server and interacts with it via HTTP.

PR #6213, opened 2019-07-19
Simplify DV packages/ui index files

elvisisking:
- index file work in the UI directories
- fixed a circular dependency issue
- added code to disable polling when the VirtualizationsPage is unmounted


app/ui-react/syndesis/src/modules/data/pages/VirtualizationsPage.tsx
```diff
@@ -79,12 +79,20 @@ export function getVirtualizationsHref(baseUrl: string): string {
...
+  const {
+    handleDeleteVirtualization,
+    handlePublishVirtualization,
+    handleUnpublishServiceVdb,
   } = VirtualizationHandlers();
-  const { resource: data, hasData, error } = useVirtualizations();
+  const [disableUpdates, setDisableUpdates] = React.useState(false);
+  const { resource: data, hasData, error } = useVirtualizations(disableUpdates);
+
+  // unmount: turn off polling
```
riccardo-forina: This shouldn’t be done here, but in the cleanup function of the effect that starts the polling in first place

elvisisking: I'll fix like we talked about and submit a new PR @riccardo-forina.

PR #6215, opened 2019-07-22
chore: Add automated integration tests

christophd:
This PR adds automated integration tests for the Syndesis integration runtime using Testcontainers. 

I have introduced a new pom module `test` and moved the existing `system test` module to a submodule of this. The new `integration tests` are added as a 2nd submodule. By default the test sources are part of the project but the tests are skipped by default.

The new integration tests run with syndesis build script using:

```
syndesis integration-test
```

Tests are skipped by default to not add another 10-15 minutes to the CI pipeline. We would need to find a good way to execute these tests frequently but not with every commit.


app/pom.xml
```diff
@@ -115,7 +115,8 @@
     <json-schema-validator.version>2.2.8</json-schema-validator.version>
     <junit.version>4.12</junit.version>
     <powermock.version>2.0.0</powermock.version>
-    <testcontainers.version>1.10.6</testcontainers.version>
+    <testcontainers.version>1.10.7</testcontainers.version>
+    <citrus.version>2.7.8</citrus.version>
```
oscerd: Do we have troubles using the last bits in central?

oscerd:
Looks really great. We can setup a daily integration test job somewhere.

app/pom.xml
christophd: Is your question related to citrus version? We can not use the latest 2.8.x versions here because this is using Spring Boot 2.x I think in Syndesis we are still bound to Spring 1.5.x so we need to use citrus 2.7.8

PR #6216, opened 2019-07-22
Remove wrong ocp4 permission from install resources

nicolaferraro:
That permission is no longer present on the default "edit" clusterrole on ocp4.

It belongs to the standard permissions created by early versions of operator-sdk but we're not using it, so we can remove it.

It's only for master, we've fixed install scripts on 1.7.x.
Ref:
- https://github.com/syndesisio/fuse-online-install/pull/162
- https://github.com/syndesisio/fuse-online-install/issues/161


heiko-braun:
@lgarciaaco Mind a review?

Issue #6209, opened 2019-07-19
Syndesis S2i build failing


christophd:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

The syndesis-s2i build is failing for me on master. 

I have executed
```
syndesis build -m s2i -i -f
```

This is the error message:
```
[INFO] F8: Step 9/14 : RUN cd /tmp/artifacts/m2/project  && mvn --batch-mode -s /tmp/settings.xml -Dmaven.repo.local=/tmp/artifacts/m2 package -DskipTests -Dmdep.skip=true -e -Dfabric8.skip=true  && rm -rf /tmp/artifacts/m2/project  && chgrp -R 0 /tmp/artifacts/m2  && chmod -R g=u /tmp/artifacts/m2
[INFO] F8:  ---> Running in 3e4409035f51
[INFO] F8: [INFO] Error stacktraces are turned on.
[INFO] F8: [INFO] Scanning for projects...
[INFO] F8: [ERROR] [ERROR] Some problems were encountered while processing the POMs:
[INFO] F8: [ERROR] Non-resolvable import POM: Could not find artifact io.syndesis.integration:integration-bom:pom:1.8-SNAPSHOT @ line 45, column 19
[INFO] F8: [ERROR] 'dependencies.dependency.version' for io.atlasmap:camel-atlasmap:jar is missing. @ line 57, column 17
[INFO] F8: [ERROR] 'dependencies.dependency.version' for io.syndesis.connector:connector-activemq:jar is missing. @ line 63, column 17
[INFO] F8: [ERROR] 'dependencies.dependency.version' for io.syndesis.connector:connector-amqp:jar is missing. @ line 67, column 17
[INFO] F8: [ERROR] 'dependencies.dependency.version' for io.syndesis.connector:connector-api-provider:jar is missing. @ line 71, column 17

...

[INFO] F8: [ERROR] The build could not read 1 project -> [Help 1]
[INFO] F8: org.apache.maven.project.ProjectBuildingException: Some problems were encountered while processing the POMs:
[INFO] F8: [ERROR] Non-resolvable import POM: Could not find artifact io.syndesis.integration:integration-bom:pom:1.8-SNAPSHOT @ line 45, column 19

...

[...]

PR #6217, opened 2019-07-22
#6209 - Replacing xargs -a to make it compatible with MacOS

squakez:
#6209 - Just changing the script to be cross compatible. @christophd please make some test and approve as your test are positive.


zregvart:
@squakez can we remove the `pr/wip` to get this merged?

Issue #6141, opened 2019-07-12
Cannot validate FTP connection without credentials


asmigala:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
When adding a FTP connection without credentials (i.e. no username or password), clicking on Validate causes a "Authentication failed" message to appear.

The reason seems to be that the UI now sends `"username": ""` as part of the payload to the `/api/v1/connectors/ftp/verifier` endpoint. If I completely remove that `username` key/value from the request and resend it, the verifier returns OK

asmigala:
Verified with latest master

PR #6232, opened 2019-07-23
chore(6199): Refactor connector options handling

phantomjinx:
* Refactors connectors to extract values from options map using the
  ConnectorOptions class. This handles checking for null or empty, maps
  to alternative types if required and returns as String or other type.

* Ensures uniform methodology for accessing key/values from map,
  simplifying boilerplate code.

* Provides unit tests for testing possible option extraction use-cases.

### DIFF
--- a/app/connector/activemq/pom.xml
+++ b/app/connector/activemq/pom.xml
@@ -67,7 +67,6 @@
     <dependency>
       <groupId>io.syndesis.connector</groupId>
       <artifactId>connector-support-util</artifactId>
-      <version>${project.version}</version>
     </dependency>
     <dependency>
       <groupId>io.syndesis.integration</groupId>
--- a/app/connector/activemq/src/main/java/io/syndesis/connector/activemq/ActiveMQConnectorVerifierExtension.java
+++ b/app/connector/activemq/src/main/java/io/syndesis/connector/activemq/ActiveMQConnectorVerifierExtension.java
@@ -28,6 +28,7 @@ import org.apache.camel.component.extension.verifier.ResultErrorBuilder;
 import org.apache.camel.component.extension.verifier.ResultErrorHelper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import io.syndesis.connector.support.util.ConnectorOptions;
 
 public class ActiveMQConnectorVerifierExtension extends DefaultComponentVerifierExtension {
     private static final Logger LOG = LoggerFactory.getLogger(ActiveMQConnectorVerifierExtension.class);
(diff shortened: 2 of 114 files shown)


phantomjinx:
This mitigates issues that we have been seeing such as #6141.

heiko-braun:
@phantomjinx Given the extensive nature of this PR, would it make sense to invite 2-3 people to a BJ's session to review this together?

phantomjinx:
> @phantomjinx Given the extensive nature of this PR, would it make sense to invite 2-3 people to a BJ's session to review this together?

I don't think that would be necessary if I provide some context maybe on how best to review:
- In each connector module, eg. ftp, odata, there is a set of classes containing methods with a parameter of the type `Map<String, Object>`. These object tend to be called 'options', 'parameters' or 'preferences';
- Typically, most of the time these maps have key/value pairs of strings so the most common code written has been along the lines of `String myValue = (String) options.get("myKey")`;
- In a small number of cases, objects are added to these maps then retrieved later on so something like `MyObject obj = (MyObject) options.get("myObjectKey")`;
- Frequently, the string values fetched from the map will be mapped to something else, eg. to convert the string into an Enum;
- A number of issues have popped up recently that have all been related to the change of UI. Rather than not sending empty key/values, the UI now sends all keys, inc. those that are empty. This has meant a number of connectors have developed regression bugs since they have code like `options.get("myKey") != null`. This does not work correctly if 'myKey' exists in the map but has an empty string value;
- To fix such occurences and provide some order to the extraction of these maps, the class [ConnectorOptions](https://github.com/phantomjinx/syndesis/blob/master/app/connector/support/util/src/main/java/io/syndesis/connector/support/util/ConnectorOptions.java) has been developed. This provides a set of static functions for dealing with the use-cases presented by these maps and handles the various possible scenerios including the empty-string so described.

Therefore, to review this, I would suggest:
- Reviewing ConnectionOptions to ensure the work there is sane (and see if there are any holes);
- Reviewing ConnectionOptionsTest to determine if the tests are adequate and whether there are any more use-cases that might need to be considered in the future;
- The actual replacements for each connector class are incredibly repetitive, esp. calls to `ConnectorOptions.extractOption(...)` so maybe a review of a subset of those is all that's required.
- The poms simply add `connector-support-util` to the connectors since that is the location of ConnectorOptions.

In the end, all options, parameters and preferences in the connector modules will be extracted using a single function class so extracted consistently.

@heiko-braun - if this is insufficient or requires further explanation, I will of course arrange a meeting.

lburgazzoli:
Has a quick look and most of the changes are related to simply replace invoking the map directly to using the new utility so this change seems ok fro my pov

Issue #6224, opened 2019-07-23
Conditional flow duplicate flow ids when used in API Provider


mmuzikar:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
If I try to create an API Provider integration using Conditional flow, I have to get the condition right on the first try. When I try to edit the Conditional flow again, the integration crashes with this stacktrace:
```
org.apache.camel.RuntimeCamelException: org.apache.camel.FailedToStartRouteException: Failed to start route -LkSSIqHna4CNPrkrGeP because of duplicate id detected: -LkSSIqHna4CNPrkrGeP. Please correct ids to be unique among all your routes.
	at org.apache.camel.util.ObjectHelper.wrapRuntimeCamelException(ObjectHelper.java:1830) ~[camel-core-2.21.0.fuse-740038.jar!/:2.21.0.fuse-740038]
	at org.apache.camel.spring.SpringCamelContext.start(SpringCamelContext.java:136) ~[camel-spring-2.21.0.fuse-740038.jar!/:2.21.0.fuse-740038]
	at org.apache.camel.spring.SpringCamelContext.onApplicationEvent(SpringCamelContext.java:174) ~[camel-spring-2.21.0.fuse-740038.jar!/:2.21.0.fuse-740038]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener(SimpleApplicationEventMulticaster.java:172) ~[spring-context-4.3.19.RELEASE.jar!/:4.3.19.RELEASE]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:165) ~[spring-context-4.3.19.RELEASE.jar!/:4.3.19.RELEASE]
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:139) ~[spring-context-4.3.19.RELEASE.jar!/:4.3.19.RELEASE]
	at org.springframework.context.support.AbstractApplicationContext.publishEvent(AbstractApplicationContext.java:393) ~[spring-context-4.3.19.RELEASE.jar!/:4.3.19.RELEASE]
[...]

Issue #6226, opened 2019-07-23
Conditional flow fails with collections in API Provider


mmuzikar:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
Conditional flows cannot be used with collections in Syndesis, but API Provider makes a collection around its request body, even when it is specified in the API specification to contain only one object of specified type.
Example:
[conditional-provider.zip](https://github.com/syndesisio/syndesis/files/3420607/conditional-provider.zip)
In this integration you can see that in the specification in the POST method it expects request body of Task. I'd expect that to be one Task. So if I create a condition `${body.completed}` it should work, or at least `${body["completed"]}`. I tried also splitting it, but split doesn't work well on these types of collections (see #4862). Also I tried `${body[0]["completed"]}` but that also didn't work for some reason. The log reported something along the lines of `cannot call method [0]["completed"] on ArrayList`.

christophd:
I think the correct expression would be `${body.body.completed}`

This is because the 1st body references the Camel exchange `body` and the 2nd `body` the body of the unified JSON payload that is used in API provider. I think we would have the same expression when using an advanced filter step in API provider.

I think it is a bit uncomfortable and unclear to the user but that's another part of the story.

@zregvart Am I right with this regarding unified JSON?

christophd:
@mmuzikar I think this has nothing to do with collections on API provider. The error message was misleading I guess. can you please verify using the expression `${body.body.completed}`?

mmuzikar:
@christophd Yes you are right, that works. It is certainly not the most obvious solution though. If this is the intended solution, it should be at least documented somewhere (@TovaCohen can you elaborate if it is already documented, or add a mention of that in the documentation?). But this is a thing you don't really expect that you have to read documentation for, maybe a little mention in the UI where there is the "Provide a condition that you want to evaluate (for example, `${in.header.type} == 'note'` or `${in.body.title} contains 'Important'`)." Or in a perfect world if the datatype is a JSON schema there could be a note added or something. (cc @syndesisio/uxd )

zregvart:
> @zregvart Am I right with this regarding unified JSON?

Yup, and that's consistent with the data shape.

TovaCohen:
In the user guide, in the doc for adding conditional flows, a prerequisite for adding a conditional flow step is: 

[...]

PR #6235, opened 2019-07-23
fix: revisit fix for #6118, ensure step IDs are only overridden when undefined

gashcrumb:
for #6224


christophd:
is really related to #6226? Should be related to #6224 instead?

gashcrumb:
Damn, did I fat-finger that?  Lemme fix that.

heiko-braun:
Lets add it to 1.7 despite the fact that its p1

Issue #5797, opened 2019-06-18
Continuous Integration improvement: Develop parallel connector build


balejos:
As a result of our retrospective, we agreed to develop a parallel build that would execute the connectors so that the main build can run faster.

squakez:
I'd like to give an hand on that, but it may take some time for me as I have to learn some basic about CircleCI. If we don't have any rush I can work on it

christophd:
I can also join and give that a try.

christophd:
This could be worth a try: https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3

christophd:
Let me share some idea how we could do parallel builds in CircleCi.

We could use a parameterized job for connectors (https://circleci.com/docs/2.0/reusing-config/#using-the-parameters-declaration)

So we use something like:
```yaml
  # Connector parameterized depends on connector-base, mount workspace .m2
  connector:
    <<: *job_defaults
    environment:
      <<: *common_env
    parameters:
      artifact:
        type: string
    steps:
      - checkout
      - restore_cache:
          key: syndesis-mvn-connector-{{ checksum "app/connector-<< parameters.artifact >>/pom.xml" }}
      - <<: *load_m2
      - run:
          name: Build Connector << parameters.artifact >>
          command: |
            ./tools/bin/syndesis build --batch-mode --module :<< parameters.artifact >> | tee build_log.txt
      - <<: *save_junit
      - store_test_results:
          path: /workspace/junit
      - store_artifacts:
          path: build_log.txt
      - save_cache:
          key: syndesis-mvn-connector-{{ checksum "app/connector-<< parameters.artifact >>/pom.xml" }}
          paths:
            - ~/.m2
      - <<: *save_m2
```

Notice the parameter `artifact` that is used to build only the specified Maven submodule with syndesis build script.

In the workflow definition we could explicitly list all connectors and let those run in parallel:

```yaml
workflows:
  version: 2
  syndesis:
    jobs:
      - system-test:
          filters:
            branches:
              only: system-test
      - doc:
          filters:
            branches:
              only: master
      - license-check
      - ui
      - ui-legacy
      - common
      - operator
      - extension:
          requires:
           - common
      - integration:
          requires:
            - extension
      - connector-base:
          requires:
            - integration
      - connector:
          artifact: connector-activemq
          requires:
            - connector-base
      - connector:
          artifact: connector-api-provider
          requires:
            - connector-base
      - connector:
[...]

PR #6241, opened 2019-07-24
chore: Update CircleCi config to use parallel connector jobs

christophd:
PR uses parallel jobs for connectors on CircleCi. 

- connector-support modules are built first in sequential order
- all connectors are built in parallel to each other
- connector-support-catalog is built at the very end to collect all connectors
- nightly job still uses "old" way of running all connectors with single job

Relates to #5797

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-version: 2
+version: 2.1
 
 common_env: &common_env
   MAVEN_OPTS: -Xmx1024m
@@ -113,13 +113,13 @@ jobs:
       - save_cache:
           key: syndesis-yarn-{{ checksum "app/ui-react/yarn.lock" }}
           paths:
-          - /usr/local/share/.cache/yarn/v1
+            - /usr/local/share/.cache/yarn/v1
       - store_artifacts:
           path: build_log.txt
       - save_cache:
           key: syndesis-mvn-ui-{{ checksum "app/ui-react/pom.xml" }}
           paths:
-          - ~/.m2
+            - ~/.m2
       - <<: *push_images
 
   ui-legacy:
@@ -153,17 +153,17 @@ jobs:
       - save_cache:
           key: syndesis-yarn-legacy-{{ checksum "app/ui-angular/yarn.lock" }}
           paths:
-          - /usr/local/share/.cache/yarn/v1
+            - /usr/local/share/.cache/yarn/v1
       - store_artifacts:
           path: build_log.txt
       - save_cache:
           key: syndesis-mvn-ui-legacy{{ checksum "app/ui-angular/pom.xml" }}
           paths:
-          - ~/.m2
+            - ~/.m2
       - <<: *push_images
 
-  # Connector depends on integration, mount workspace .m2
-  connector:
+  # All connectors job depends on integration, mount workspace .m2
+  connectors:
     <<: *job_defaults
     environment:
       <<: *common_env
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -513,17 +560,260 @@ workflows:
...
+            - telegram
+            - timer
+            - twitter
+            - servicenow
+            - webhook
+            - concur
+            - kudu
+            - odata
+            - knative
+            - flow
```
squakez: This is a bit ugly to see... too much duplication. Isn't there any way to group those and reference with a unique label instead?

christophd: agreed, but yet I have not found a better way to do it

christophd:
the config seems to have some issue because CircleCi complains about it - Darn! YAML!

.circleci/config.yml
squakez: Maybe someone with experience on the tool can tell (@zregvart ?)

.circleci/config.yml
```diff
@@ -162,31 +162,78 @@ jobs:
...
       <<: *common_env
     steps:
       - checkout
       - restore_cache:
-          key: syndesis-mvn-connector-{{ checksum "app/connector/pom.xml" }}
+          key: syndesis-mvn-connector-{{ checksum "app/connector-support/pom.xml" }}
       - <<: *load_m2
       - run:
-          name: Build Connectors
+          name: Build Connector Util
```
zregvart: Perhaps we can introduce two modules one containing all support modules for connectors and one containing only connectors. Would make these bits a bit simpler.

.circleci/config.yml
```diff
@@ -162,31 +162,78 @@ jobs:
...
           path: build_log.txt
       - save_cache:
-          key: syndesis-mvn-connector-{{ checksum "app/connector/pom.xml" }}
+          key: syndesis-mvn-connector-{{ checksum "app/connector-support/pom.xml" }}
           paths:
           - ~/.m2
       - <<: *save_m2
 
+  # Connector parameterized depends on connector-base, mount workspace .m2
+  connector:
```
zregvart: Here `connector` is a job, I think it needs to be a [command](https://circleci.com/docs/2.0/reusing-config/#using-the-parameters-declaration)

christophd: I wanted to have a parameterized job to call it from the workflow: https://circleci.com/docs/2.0/reusing-config/#authoring-parameterized-jobs

.circleci/config.yml
christophd: I have improved a bit to run all modules in one run command. Also it is a bit more complicated than I thought. We need to build some support modules first then all connectors and at the very end the connector-catalog.

christophd:
> the config seems to have some issue because CircleCi complains about it - Darn! YAML!

YAML has not been the problem. I should have read the documentation of parameterized jobs carefully: `Note: Invoking jobs multiple times in a single workflow and parameters in jobs are available in configuration version 2.1 and later.`

So we are using version 2 at the moment. Any objections to using version 2.1?

.circleci/config.yml
zregvart: Oh cool.

.circleci/config.yml
```diff
@@ -509,20 +560,232 @@ workflows:
...
+            - connector-webhook
+            - connector-concur
+            - connector-kudu
+            - connector-odata
+            - connector-knative
+            - connector-flow
       - meta:
           requires:
-            - connector
+            - connector-catalog
```
squakez: Tidier now!

christophd:
so the build config is still failing with version 2.1 set. I saw something on the circleci Q&A which might be the case here, too:

> I strongly suspect the issue here is you don’t have Build Processing enabled in your project settings (under “Advanced” on your project’s settings page). Your original config runs fine for me when I have that on. The bug in this case may be that we need a much better error handling when passing 2.1

I do not see the project settings nor can I adjust these @zregvart can you?

zregvart:
@christophd found [the setting](https://circleci.com/gh/syndesisio/syndesis/edit#advanced-settings) and turned 2.1 support to on.

zregvart:
We have a [build error](https://circleci.com/gh/syndesisio/syndesis/90629) now, something with YAML.

christophd:
@zregvart @squakez I have finally sorted out all problems with the parallel connector build and the workflow is running successfully.

I have removed WIP label now can you two please have another look?

We still need to see if this all brings us a lot of speed in the build jobs. 

This one here with parallel jobs took 39 minutes: 
https://circleci.com/workflow-run/923c5821-09a4-41cd-9f84-6790e7762151

This one with sequential jobs took 45 minutes:
https://circleci.com/workflow-run/2321aa81-0252-431f-b90c-cbb7bd498f93

This one with sequential jobs took 33 minutes:
https://circleci.com/workflow-run/484860c9-0d14-4fdb-b14e-29b2a05c0423

squakez:
As you've already mentioned, the parallelism is a good thing if there are enough resources to really run in parallel, otherwise it ends up in a sequential execution anyhow. Even if there is not yet any big gain it is a nice example to see how parallel builds work, so we can leverage that for the future. Nice work!

PR #6249, opened 2019-07-25
fix: CircleCi integration-test config

christophd:
This PR fixes some issues regarding the integration-test config on CircleCi so job is able to run the tests now.

Tests are still failing though and fixing needs to be continued in subsequent PRs

### DIFF
--- a/app/test/integration-test/pom.xml
+++ b/app/test/integration-test/pom.xml
@@ -199,6 +199,13 @@
           <skip>true</skip>
         </configuration>
       </plugin>
+      <!-- Required only for CircleCi build as the used JVM to run the tests has limited cryptography -->
+      <plugin>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
+      </plugin>
 
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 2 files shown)


app/test/integration-test/pom.xml
```diff
@@ -199,6 +199,13 @@
           <skip>true</skip>
         </configuration>
       </plugin>
+      <!-- Required only for CircleCi build as the used JVM to run the tests has limited cryptography -->
+      <plugin>
+        <artifactId>maven-enforcer-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
zregvart: Could we try switching to [adoptopenjdk/openjdk8](https://hub.docker.com/r/adoptopenjdk/openjdk8) image in place of `openjdk:8-jdk` instead.

christophd: as I am using Testcontainers in the tests I need to use the machine executor in CircleCi for this job. And this is using a ubuntu vm image with some JDK in it that has this cryptography restrictions. So not sure how to adjust this used JDK in this virtual machine.

zregvart: Perhaps we could install the [AdoptOpenJDK debian package](https://adoptopenjdk.net/installation.html#linux-pkg)?

christophd: Let me merge this one and tackle the cryptography afterwards so the hourly failing builds disappear for now.

Issue #6222, opened 2019-07-22
Regression in Basic Filter input field descriptions - they used to be more helpful


TovaCohen:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[X ] UI text issue or request
</code></pre>

## The problem

In the staging site, the UI for the basic filter looks like this: 
![image](https://user-images.githubusercontent.com/25067106/61665335-b6057080-aca2-11e9-877c-c20296387325.png)

I think that the descriptions, "Property Name is required" and "Keywords is required" do not provide enough information for the business user to know what to insert. In previous releases, we had: 

"The data you want to evaluate" in place of "Property Name is required"
"For this value" in place of "Keywords is required"

I see this as a regression because we changed what was there and made it more difficult to know what to do. 

## Expected behavior

Please put it back the way it was. The three field descriptions form a sentence: 

The data you want to evaluate must meet this condition for this value. 

Replace "Property Name is required" with: "The data you want to evaluate".

Replace "Keywords is required" with "For this value". 

And, can asterisks indicate the required fields?

gashcrumb:
Hmm, that's validation kicking in but I guess the element structure doesn't
match the css that would make that text red.  Still, that help text
shouldn't be getting overwritten until the field is clicked in and then
unfocused.

Also the asterisk unfortunately is part of the label normally for a
patternfly form, but this is a kinda customized thing which doesn't use
labels.

TovaCohen:
Oh, yes, when I click in a field, I get the right description. 
But then when I add another rule, the fields have the Property Name and Keywords descriptions. So something is amiss.

gashcrumb:
Hmm, so yes I can disable that validation, *however* the problem is that then, the overall form isn't validated, meaning you can click "add another rule" and not type anything in, then click "Done" because reasons, creating an invalid configuration.

I wonder if it'd be better maybe to move these text prompts from the "help text" area underneath the form controls and instead have them be labels, leaving the help text available to prompt for required fields.  So as a rough draft instead of:

![image](https://user-images.githubusercontent.com/351660/61733978-f91d1d80-ad4e-11e9-8c62-a39793dd9ea8.png)

Could be more like:

![image](https://user-images.githubusercontent.com/351660/61734247-92e4ca80-ad4f-11e9-9ef2-ce4fc49dd8a3.png)

@syndesisio/uxd

[...]

PR #6253, opened 2019-07-25
fix(basicfilter): rework condition set form UI

mcoker:
fixes https://github.com/syndesisio/syndesis/issues/6222

@gashcrumb this is as close as I could get while working with the existing markup. I couldn't figure out the auto form components to be able to modify the markup to make this layout easier/better. I'm also not sure where to inject the paragraph of description text that goes above the form fields.

In a perfect world, each of the `<fieldset>` elements would have a nested `<div>` that wraps the form groups so that I could give that a `grid` layout to place the the form groups inside of it.

A more flexible route would be to then wrap the 3 form group elements that create the condition in a `<div>`, then wrap the action/array controls in another element, that would allow me to use `flex` layout, which works in IE (since `grid` does not).

But what's there works, too, I just had to use `absolute` positioning for the delete action button, and %/relative spacing between the form groups and the delete action so it scales properly down to mobile.

Does this look ok, or can you (or point me in the right direction to) change the markup? And the same for injecting the description text before the form fields?

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/ruleFilter/WithRuleFilterForm.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/ruleFilter/WithRuleFilterForm.tsx
@@ -132,8 +132,7 @@ export const WithRuleFilterForm: React.FunctionComponent<
   const validator = (values: IRuleFilterConfig) =>
     validateRequiredProperties(
       definition,
-      (field: string) =>
-        t('integrations:editor:ruleForm:fieldRequired', { field }),
+      (field: string) => '', // return an empty string here so the help text isn't replaced
       values
     );
   return (
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/ruleFilter/WithRuleFilterForm.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/ruleFilter/WithRuleFilterForm.tsx
@@ -103,13 +103,15 @@ export const WithRuleFilterForm: React.FunctionComponent<
       },
       arrayDefinitionOptions: {
         arrayControlAttributes: {
-          className: 'col-md-3 form-group',
+          className: 'form-group with-rule-filter-form__action',
         },
+        /*
         controlLabelAttributes: {
           style: { display: 'none' },
         },
+        */
         formGroupAttributes: {
-          className: 'col-md-3',
+          className: 'with-rule-filter-form__group',
         },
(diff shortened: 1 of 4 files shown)


app/ui-react/syndesis/src/modules/integrations/components/editor/ruleFilter/WithRuleFilterForm.tsx
```diff
@@ -103,13 +104,15 @@ export const WithRuleFilterForm: React.FunctionComponent<
       },
       arrayDefinitionOptions: {
         arrayControlAttributes: {
-          className: 'col-md-3 form-group',
+          className: 'form-group with-rule-filter-form__action',
         },
+        /*
```
gashcrumb: We can delete this block.

gashcrumb:
So the markup for each element in the array is produced [here](https://github.com/syndesisio/syndesis/blob/master/app/ui-react/packages/auto-form/src/widgets/FormArrayComponent.tsx) which maps those classes you specify in the array type.  

There's also a storybook for auto-form ([example](https://syndesisio.github.io/syndesis-react/auto-form/index.html?path=/story/autoform--array)) so we could actually just take that form definition and plug it into that story or just create a new one which we could use to figure out the correct structure of the elements.  You'd just need to `cd` into `packages/auto-form` and then run `yarn storybook`.  

Now, the idea is that the form array stuff kinda needs to be really flexible so yeah, I kinda kept most of the layout logic open so that the code using auto-form kinda has to specify everything.

gashcrumb:
> And the same for injecting the description text before the form fields?

This is a tough one, in my quick example I changed the `displayName` property and commented out the code that hides the label.  But then the problem with that is that text is then used in the validation prompt.  We *could* I guess add a new parameter but I dunno, that starts to feel a bit messy.

mcoker:
@gashcrumb added an element inside the fieldset in the formarray component and converted the layout to grid. That works for me. Not sure about the description text though...

gashcrumb:
Lemme see how it's looking here.

gashcrumb:
So yeah, I think that's a huge improvement!  Looks like you've one lint error, but other than that LGTM

gashcrumb:
As for the description text, one thought I had over the weekend at some point was I could probably just use a custom validator function for the basic filter page that just ensures validation fails but doesn't set the error text, which would then not hide the initial description text that's supposed to show up.  Let me look into that, I might push a commit with that onto this PR.

mcoker:
@gashcrumb thanks!!

Issue #6168, opened 2019-07-15
Cannot login to syndesis


KurtStam:
## This is a...

<pre><code>
[ ] Feature request
[*] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
After logging in I get the following screen

<img width="573" alt="Screen Shot 2019-07-15 at 4 06 43 PM" src="https://user-images.githubusercontent.com/35576/61245550-c35ab200-a71a-11e9-9130-7e79ae032781.png">

Logging in the auth pod:

```
2019/07/15 20:03:03 provider.go:593: 200 GET https://172.30.0.1/apis/user.openshift.io/v1/users/~ {"kind":"User","apiVersion":"user.openshift.io/v1","metadata":{"name":"developer","selfLink":"/apis/user.openshift.io/v1/users/developer","uid":"d1a4b0d7-a73a-11e9-a4f7-2a297d346f44","resourceVersion":"2230","creationTimestamp":"2019-07-15T19:57:47Z"},"identities":["anypassword:developer"],"groups":["system:authenticated","system:authenticated:oauth"]}
2019/07/15 20:03:03 provider.go:593: 201 POST https://172.30.0.1/apis/authorization.openshift.io/v1/subjectaccessreviews {"kind":"SubjectAccessReviewResponse","apiVersion":"authorization.openshift.io/v1","namespace":"myproject","allowed":true,"reason":"RBAC: allowed by ClusterRoleBinding \"syndesis-extra-permissions\" of ClusterRole \"syndesis-extra-permissions\" to User \"developer\""}
2019/07/15 20:03:03 oauthproxy.go:676: 172.17.0.1:42398 authentication complete Session{developer@cluster.local token:true}
```

fails on both Safari and Chrome. This is on a fresh install using 
bash <(curl -sL https://syndes.is/start)

Also fails in incognito mode. All pods seem happy:

```
NAME                          READY     STATUS      RESTARTS   AGE
syndesis-db-1-qwhqj           2/2       Running     0          22m
syndesis-meta-1-f9nq4         1/1       Running     0          22m
syndesis-oauthproxy-1-bxgvn   1/1       Running     0          22m
syndesis-operator-1-w64jn     1/1       Running     0          23m
syndesis-prometheus-1-fxknp   1/1       Running     0          22m
syndesis-server-1-2mthp       1/1       Running     1          22m
[...]

Issue #3422, opened 2018-08-22
Update of Syndesis CR should be picked up by the operator


rhuss:
When editing an existing custom resource "Syndesis" and storing it back to OpenShift, then changes (like in `maxIntegrations`) aren't picked up by the operator.

It is expected that, depending on what has changed, the operator will perform the operations needed (like updating a configmap and restarting a pod) to take over the changes.

This might also changes the state model.

And while we are at it, it would be awesome that any change would be recorded in the status section like it happens for other Kubernetes resources.

rhuss:
@heiko-braun this is the other issue I think would be important to include for 7.2

claudio4j:
I think this issue was resolved by #9257 (any updates to configmap may trigger a pod respin). Going to close this one.

PR #6256, opened 2019-07-26
feat: #3422 Update of Syndesis CR or syndesis-global-config secret is now applied to the deployed resources.

chirino:


chirino:
This also brings restores OpenShift resources which are modified behind the back of the operator, so that they always contain the settings that are being enforced by the operator.

chirino:
This change also updates the syndesis CR spec so that it shows that values that are being applied to the deployment configuration.

chirino:
This should lets us more easily update deployment templates can get those changes applied to a running syndesis by redeploying the operator since you don't need to reset the installation to get the changes picked up.

install/operator/cmd/manager/main.go
```diff
@@ -45,7 +45,6 @@ func printVersion() {
 
 func main() {
 	configuration.TemplateConfig = pflag.StringP("template-config", "t", "/conf/template-config.yml", "Path to template config used for installation")
-	configuration.AddonsDirLocation = pflag.StringP("addons", "a", "", "Path to the addons directory used for installation")
```
zregvart: Don't we want this to be configurable? I'm even thinking that this should be a `StringSliceP` so one could specify multiple addon directories. Use cases like the ones we ship vs the one you'd like to test and mount via volume?

install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
```diff
@@ -91,16 +91,16 @@ type Resources struct {
 }
 
 type ResourcesWithVolume struct {
-	v1.ResourceRequirements `json:",inline,omitempty"`
-	VolumeCapacity          string `json:"volumeCapacity,omitempty"`
+	Resources      `json:",inline,omitempty"`
```
zregvart: Do we need to do [a upgrade procedure](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version) for this (whatever that is?)

install/operator/pkg/generator/assets/install/03-syndesis-ui.yml.mustache
```diff
@@ -74,7 +74,7 @@
           ports:
           - containerPort: 8080
           volumeMounts:
-          - mountPath: /usr/share/nginx/html/config
+          - mountPath: /opt/app-root/src/config
```
zregvart: Good catch!

install/operator/pkg/generator/assets/install/addon-legacyui.yml
```diff
@@ -0,0 +1,355 @@
+{{if eq (or .Syndesis.Spec.Addons.legacyui.enabled "false") "true"}}
```
zregvart: I think we can drop the legacy UI entirely, no need to drag that along.

install/operator/pkg/generator/assets/route/route.yml.mustache
```diff
@@ -0,0 +1,20 @@
...
+  metadata:
+    labels:
+      app: syndesis
+      syndesis.io/app: syndesis
+      syndesis.io/type: infrastructure
+    annotations:
+      console.alpha.openshift.io/overview-app-route: "true"
+    name: syndesis
+  spec:
+    host: {{Env.ROUTE_HOSTNAME}}
```
zregvart: I'm a bit confused, so this might not be an issue, should this be:
```suggestion
    host: {{.Env.ROUTE_HOSTNAME}}
```

chirino:
I've also tacked on a commit which unifies addon handling so that they are dealt with the same way that other resources are handled.  This way addons and regular resources are removed is no longer enabled.  And addons are enabled/disabled using the Sysndesis CR.

install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
chirino: This change does not affect the compatibility of json/encoding decoding, but it does allow common functions to work with with Resources or ResourcesWithVolume more easily.

install/operator/pkg/generator/assets/install/03-syndesis-ui.yml.mustache
chirino: Oh yeah this PR also fixes #6168

install/operator/pkg/generator/assets/route/route.yml.mustache
chirino: yeah.. the generator now support both go (.yml.tmpl) and mustache (.yml.mustache) templates.  mustache template variables don't start with '.' like the go template ones.

I find the mustache templates easier to read, but the go templates do give you more functionality if you need it.

install/operator/cmd/manager/main.go
lgarciaaco: well no, the problem before is that we could not enable/disable individual components, so we either we include the `addons` dir (and all its content) or we didn't. Now that we can toggle components we can always load the `addons` content

phantomjinx: I would agree with @zregvart in this instance. In my experience there are always reasons for customers to want to install addons of their own that cannot be included openly (secure/sensitive in nature being one example). So if the addons directory can be configured it really can only enhance in my view.

install/operator/pkg/generator/assets/install/04-syndesis-meta.yml.tmpl
```diff
@@ -76,14 +76,14 @@
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
-{{#Debug}}
+{{if .Debug}}
           - name: JAVA_DEBUG
             value: "true"
```
phantomjinx: I've probably missed it but can you confirm if this debug env var has been bubbled up into the install scripts? I can't see it & if not it would be most useful if a switch could be added to have JAVA_DEBUG turned on upon installation so that devs don't have to manually turn it on after.
WDYT?

install/operator/pkg/syndesis/action/install.go
```diff
@@ -107,110 +136,355 @@ func (a *installAction) Execute(ctx context.Context, syndesis *v1alpha1.Syndesis
...
+				from := value.(map[string]interface{})
+				mergeMap(mergePath+"/metadata", to, from)
+			default:
+				existing.Object[key] = value
+			}
+		}
+
+		//if d.GetKind() == "DeploymentConfig" && d.GetName() == "syndesis-meta" {
+		//	Debug("existing:", existing, "(index .Object.spec.template.spec.containers 0).resources.limits.memory")
+		//}
```
phantomjinx: Do you still want to keep this commented-out code?

install/operator/cmd/manager/main.go
chirino: I would not feel comfortable making our operator an open planform for install any user created content.  So I'd rather not allow user created resources to be managed by our operator.  Remember, the smaller the surface area of what can be modified by the user, the better it is for us in terms of support and future flexibility to change this as we want.  Once a user start using something and we want to change it later, now we need to think about how can we migrate him, or have a compatibility story.

install/operator/pkg/syndesis/action/install.go
chirino: It comes in handy when you need to debug stuff.  It's easy to get into a situation where the same resource keeps getting updated by the operator for no good reason.   I've run in to a couple of case of that: For example the the limit you put into the template, returned with a different type of unit, and so that causes a update.

install/operator/pkg/generator/assets/install/04-syndesis-meta.yml.tmpl
chirino: I've not looked into it either too much.  I was just porting what was existing.  Can you open a new issue to looking to that.

phantomjinx: Yeah, sure thing. I'm working on the bash scripts atm so can add to my list.

install/operator/pkg/generator/assets/install/addon-legacyui.yml
chirino: Don't mind dropping legacy!  https://github.com/syndesisio/syndesis/issues/6263

install/operator/pkg/generator/assets/install/04-syndesis-meta.yml.tmpl
chirino: How does https://github.com/syndesisio/syndesis/issues/6264 sound?

PR #6258, opened 2019-07-26
fix: Make sure split data shape lookup adapts previous step

### LINKED ISSUE #6257: Split step always adapts 1st output data shape
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Split step should adapt the output data shape of the previous step relative to its location in a flow. At the moment split always adapts the output data shape of the 1st step in the flow regardless of its position in the flow.

With this in place you can not use split on steps other than the 1st step in a flow.

christophd:
Fixes #6257


heiko-braun:
@christophd I don't know if there will be another 1.7.x release before 7.4.1, but feel free to merge it onto 1.7.x anyways so we have it place for the next release.

PR #6259, opened 2019-07-26
Fix #6255: fix knative metadata retrieval logic

### LINKED ISSUE #6255: Can't start an integration from a Knative channel
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description

This affects master and CR2.

Metadata lookup when starting an integration from a Knative channel fails in the second phase (data shape selection) because of a combo of slightly wrong logic in the connector and behavior change in the new react UI wrt hidden connection fields.

Error got is: `Unsupported Knative type: .`

nicolaferraro:
Fix #6255

Set it as WIP because it's not a p0 issue, hence it needs to go to 7.4.1.


zregvart:
@nicolaferraro we need this on `1.7.x`?

nicolaferraro:
I think it's sufficient to have it fixed in 1.8. Knative is TP.

PR #6275, opened 2019-07-29
chore: e2e tests for common create+delete user flows

kahboom:
Before continuing to write any additional ones, I want to make sure the team is ok with these basic E2E tests written using Cypress (no Selenium server).

## Changes:
- [x] Add Lerna script to run E2E tests from root
- [x] Add ability to set env vars for providing tests with data, and example config*
- [x] Update README for BE recorder/replayer, env config, and testing guidelines
- [x] Recorded some BE network payloads as tapes, available to be used with something like `BACKEND=https://syndesis-staging.b6ff.rh-idev.openshiftapps.com yarn replay`
- [x] Upgrade Cypress to `v3.4.0` due to bugs with previous version
- [x] Additional test IDs across the app for improved testability
- [x] Basic e2e tests (see below) for creating and deleting in common user flows

_* so that setting up CI/CD would be easier with using either machine env vars or config json ones_

## Tests:
- [x] Homepage/Dashboard, navigation + loading
- [x] Create + Delete Connection
- [x] Create +Delete Integration
- [x] Create + Delete Data Virtualization
- [x] Tag Integration for Release + CI/CD Management (needs improvement)
- [x] Additional Integration Actions (e.g. start, stop)

## Example:
Kapture 2019-07-29 at 16 53 55

Apologies for all the commits, I did try to squash some of them, this PR has been on hold for a while. Also, let me know if it's too many tapes and if I should cut down on a few.


riccardo-forina:
I’ll start with ❤️ for the moment! I’ll try and review this tomorrow :D

app/ui-react/README.md
```diff
@@ -649,42 +653,65 @@ Alternatively, you can run `$ yarn test --coverage --color --runInBand` from the
...
 
 You can find the configuration settings for Jest in `./packages/ui/package.json` under the `jest` prop, or under `./syndesis/package.json`.
 
 
-#### E2E testing
+#### E2E Testing
 
-To run all E2E tests, run the following from the [`./syndesis`](syndesis) dir: `$ yarn e2e`
+_Configuration_
+First, configure your environment. We provide a sample configuration file for your environment, located at `syndesis/cypress.env.example.json`. Duplicate (don't simply rename) this file, and name as `cypress.env.json`. By default, Cypress will load these vars for your tests. If, in turn, you'd like to override those, you always can set vars in the terminal. See [here]() for  more information. In tests, you can access these variables like this:
```
gashcrumb: Missing a link for how to override variables?

gashcrumb: I mean that you've got `[here]()` where there's no link :-)

app/ui-react/syndesis/cypress/integration/create-connection.spec.js
```diff
@@ -0,0 +1,146 @@
...
+  it('ensures the page renders properly', () => {
+    /**
+     * Sidebar should be collapsed
+     */
+    cy.get('.pf-c-page__sidebar-body').should('be.hidden');
+
+    /**
+     * Test Cancellation modal
+     */
+    cy.get('[data-testid=connection-creator-layout-cancel-button]')
```
gashcrumb: Should this test maybe be in it's own scenario specific to testing the cancel dialog?

app/ui-react/syndesis/cypress/integration/create-integration.spec.js
```diff
@@ -0,0 +1,281 @@
...
+
+    cy.location('pathname').should('eq', '/connections/');
+
+    cy.get('.form-control').type(connectionName + '{enter}');
+    cy.get('[data-testid|=connection-card-' + connectionSlug + ']').should(
+      'exist'
+    );
+  }
+
+  function deleteConnection() {
```
gashcrumb: I dunno, this is duplicated from the previous file.  If the initial state needs to be a certain way there's a "test mode" that can be enabled for the backend that adds operations to truncate the database at will, in which case these tests could pre-load whatever state is needed that's outside of doing the actual test.

gashcrumb: This flag -> https://github.com/syndesisio/syndesis/blob/master/tools/bin/commands/install#L34

I think QE uses it for example.

app/ui-react/syndesis/cypress/integration/create-integration.spec.js
```diff
@@ -0,0 +1,281 @@
...
+  };
+  const nameInt = randomInteger();
+  const connectionName = 'E2E Todo Connection ' + nameInt;
+  const integrationName = 'E2E Todo Integration ' + nameInt;
+  const connectionSlug = 'e2e-todo-connection-' + nameInt;
+  const integrationSlug = 'e2e-todo-integration-' + nameInt;
+
+  function createConnection() {
+    /**
+     * Create To Do Connection
```
gashcrumb: I kinda think we're gonna find this is too much setup for this test.  Another option might be to upload an exported integration and delete the integration, which would then leave the connections in-place to be used by the integration test.  The problem here is that this basically duplicates the create connection test, but if that test fails, this test will also fail, but not because creating an integration is broken, but instead because creating a connection is broken.

app/ui-react/syndesis/cypress/integration/create-connection.spec.js
kahboom: This is where it gets tricky, that's probably what the unit tests should do. I am trying to keep E2E tests page-specific as opposed to functionality-specific, but the lines kind of start to blur when there is overlapping functionality. I can maybe separate it out into a shared spec for those cases.

app/ui-react/syndesis/cypress/integration/create-integration.spec.js
kahboom: Cool, I'll look into that. 👍

app/ui-react/syndesis/cypress/integration/create-integration.spec.js
kahboom: I've been thinking about this a lot. I completely agree with you about the setup for the test, it's too much. The problem was that we wanted to keep each E2E test completely self-contained, but we didn't want to use mock data. We wanted to test the actual API though, so it's a bit of a catch 22.

With tapes, we can probably get rid of this function, so long as tests are _always_ run with the tape (which will guarantee that we have a certain connection/integration/whatever there), but there is no direct interaction with the API when you use the tapes, defeating part of the purpose of the test. Could you point me to some more info about how the `--test-support` flag works and if that might be a good alternative for working with data? Otherwise I'm leaning more toward swagger-generated mock data tbh. cc @riccardo-forina 

This is gonna be a problem for integration tests as well, so the sooner we come up with something, the better.

kahboom: btw I'm continuing to write tests because despite which route we take, assertions should stay the same and tests can still be run (just maybe not ready for CI/CD or git hooks).

app/ui-react/syndesis/cypress/integration/create-connection.spec.js
kahboom: Then again, it's also good to know that the the cancellation modal works from the perspective of the user in each page. Even if we did have a unit test for it, or a separate E2E test for it, it wouldn't give us the confidence that it would work in each page it's supposed to. So, in the regard, I think even though it's repetitive, it does make sense to do. I guess it depends how important this functionality is and how well we want it tested.

app/ui-react/syndesis/cypress/integration/create-integration.spec.js
gashcrumb: oh of course!  Yeah that makes sense.  We can ignore the noise I think, as some tests are better than no tests :smile:

kahboom:
Some temporary improvements:
- using Cypress custom commands to consolidate frequent tasks (e.g. `createConnection`, `deleteConnection`)
- check for necessary data prior to running tests (e.g. an existing Connection for Create Integration tests), otherwise, create it.
- added some additional tapes so you can run the tests without making network requests in the meantime (though integrations with log step not working for me atm).

These are temporary changes for the tests to be able to run properly (and to be able to keep writing assertions), because once we enable snapshots of the DB we will no longer need this for running tests. It possibly can be reused for setting up the db though.

gashcrumb:
I don't think the cypress screenshots should be checked in, should they?

kahboom:
@gashcrumb - Yep I didn't realize they were getting checked in, I'll remove them now and will disable screenshots on failure. I personally only think they are useful when you are running the tests yourself locally, or if they're browseable like you said.

At the moment, we're not storing any test results in the code base at all, just `stdout` I guess because of CI, and otherwise we just run things locally. We can if you think it's useful. Cypress can [use junit as the reporter](https://docs.cypress.io/guides/tooling/reporters.html#Reporter-Options), and then we'd be able to store it in a specified xml file (or per spec). By default it uses Mocha's reporter (`spec` but I think I should switch to `nyan` 😸).

Cypress also has a [dashboard](https://www.cypress.io/blog/2019/07/29/github-integration-for-the-cypress-dashboard/) that integrates with GH, where we can watch videos of failing tests, that they can host on their servers, so we can look into that eventually if we wanted additional monitoring.

PR #6284, opened 2019-07-30
feat: The sysdesis operator binary can now be used to install syndesis into a mini shift cluster.

chirino:


heiko-braun:
> The sysdesis operator binary can now be used to install syndesis into a mini shift cluster

Can you elaborate on that?

nicolaferraro:
Nice work, that's the right direction!

I get installation stuck at:
```
syndesis-operator-1-hsw5k syndesis-operator {"level":"info","ts":1564518979.896481,"logger":"action","msg":"Installing Syndesis resource","type":"install","name":"app"}
syndesis-operator-1-hsw5k syndesis-operator {"level":"info","ts":1564518987.141394,"logger":"action","msg":"Failed to create or replace resource","type":"install","kind":"Role","name":"syndesis-editor","namespace":"chirino"}
```

I'm running as admin.

Maybe better printing more information about the error, it should be some permission missing in the `syndesis-operator` role.

chirino:
@nicolaferraro operator needed more perms.  Should be better now.

chirino:
@heiko-braun that's what I hope will happen, not needing those scripts anymore.

chirino:
Also added a `install --dev` option which uses images which were built locally against your openshift cluster (via s2i builds).

chirino:
This also opens up the path to moving upgrade functions that are in the upgrade image into the operator too (it would sill spin an upgrade pod, but reuse the operator image).

chirino:
rebased and squashed a bit.

chirino:
Merging so I can start working on my next set of changes.

Issue #6240, opened 2019-07-24
Convert remaining card views from PF3 to PF4


seanforyou23:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## The problem
We still have [PF3 cards](https://patternfly-react.surge.sh/patternfly-3/?knob-Match%20Height=true&selectedKind=patternfly-react%2FCards&selectedStory=Base%20Card&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) being used in some places within the UI. PF4 cards are already integrated in some places, we just need to finish swapping them out.

## Expected behavior
All the card views should be [PF4 cards](https://pf4.patternfly.org/components/Card/examples/).

## Screenshot
For example, the connection card grid still uses PF3 cards.

<img width="1644" alt="Screen Shot 2019-07-24 at 10 30 56 AM" src="https://user-images.githubusercontent.com/5942899/61802284-31c80000-adfe-11e9-8783-14c784d6125d.png">

A quick search for instances of `card-pf` CSS class finds occurrences in the following components & their associated stylesheets. This is where we're using patternfly 3 core markup/css.

- [ ] ConnectorConfigurationForm
- [ ] SqlClientForm
- [ ] ViewConfigurationForm
- [ ] IntegrationActionSelectorCard
- [ ] ApiProviderSetInfo
- [ ] ChoiceCardHeader
- [ ] DescribeDataShapeForm
- [ ] IntegrationEditorForm
- [ ] IntegrationEditorNothingToConfigure
- [ ] EditorPageCard
- [ ] IntegrationSaveForm

Another quick search for instances of `<Card`, which catches Card, CardGrid, etc. turns up the following occurrences. This is where we're using patternfly 3 react card components.

- [x] StoryWrapper
- [x] ConnectionCard
- [x] ConnectionsGrid
- [x] ConnectionSetupOAuthCard
- [x] ConnectionDetailsOauthCard
- [x] ConnectionDetailsForm
- [x] OpenApiSelectMethod
- [x] OpenApiReviewActions
- [x] AggregatedMetricCard
- [x] IntegrationDetailMetrics
- [x] ImportPageUI
- [x] DvConnectionSkeleton
- [ ] DvConnectionsGridCell
- [ ] DvConnectionsGrid
- [x] DvConnectionCard
[...]

PR #6287, opened 2019-07-30
chore(cards): replace dashboard/connection pf3 cards/layout with pf4

seanforyou23:
This PR upgrades several Card views from PF3 to PF4. I also swapped out any usages of PF3 Card Grid while I was in there. There's a little bit of fine-tuning to do regarding spacing in the main dashboard layout but wanted to go ahead and raise this PR so we can start getting any feedback people may have.

It's an incremental step toward https://github.com/syndesisio/syndesis/issues/6240

Addresses card views in the following components.

- ConnectionCard
- ConnectionsGrid
- ConnectionsGridCell
- ConnectionsMetric
- Dashboard
- IntegrationBoard
- RecentUpdatesCard
- TopIntegrationsCard
- UptimeMetric
- AggregatedMetricCard

Screen Shot 2019-07-30 at 12 51 08 PM

Screen Shot 2019-07-30 at 12 51 32 PM

Screen Shot 2019-07-30 at 12 51 46 PM

### DIFF
--- a/app/ui-react/packages/ui/src/Connection/ConnectionsGrid.tsx
+++ b/app/ui-react/packages/ui/src/Connection/ConnectionsGrid.tsx
@@ -1,12 +1,8 @@
-import { CardGrid } from 'patternfly-react';
+import { Gallery } from '@patternfly/react-core';
 import * as React from 'react';
 
 export class ConnectionsGrid extends React.Component {
   public render() {
-    return (
-      <CardGrid fluid={true} matchHeight={true}>
-        <CardGrid.Row>{this.props.children}</CardGrid.Row>
-      </CardGrid>
-    );
+    return <Gallery gutter={'sm'}>{this.props.children}</Gallery>;
   }
 }
--- a/app/ui-react/packages/ui/src/Connection/ConnectionsGridCell.tsx
+++ b/app/ui-react/packages/ui/src/Connection/ConnectionsGridCell.tsx
@@ -1,12 +1,7 @@
-import { CardGrid } from 'patternfly-react';
 import * as React from 'react';
 
 export class ConnectionsGridCell extends React.Component {
   public render() {
-    return (
-      <CardGrid.Col xs={6} md={3}>
-        {this.props.children}
-      </CardGrid.Col>
-    );
+    return <>{this.props.children}</>;
   }
 }
--- a/app/ui-react/packages/ui/src/Dashboard/Dashboard.css
+++ b/app/ui-react/packages/ui/src/Dashboard/Dashboard.css
@@ -11,16 +11,6 @@
   padding: 0 10px;
 }
 
-/* .dashboard__metrics .pf-l-grid__item {
-  margin-bottom: var(--pf-global--spacer--md);
-}
-
-@media screen and (min-width: 768px) {
-  .dashboard__metrics .pf-l-grid__item {
-    margin-bottom: 0;
-  }
-} */
-
 .dashboard__metrics .pf-c-card.aggregate-status {
   padding: 10px 10px 16px 10px;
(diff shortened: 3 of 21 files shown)


app/ui-react/packages/ui/src/Connection/ConnectionsGridCell.tsx
```diff
@@ -1,12 +1,7 @@
-import { CardGrid } from 'patternfly-react';
 import * as React from 'react';
 
 export class ConnectionsGridCell extends React.Component {
```
gashcrumb: Maybe this one should just be removed then?

app/ui-react/packages/ui/src/Connection/ConnectionsGrid.tsx
```diff
@@ -1,12 +1,8 @@
...
 import * as React from 'react';
 
 export class ConnectionsGrid extends React.Component {
   public render() {
-    return (
-      <CardGrid fluid={true} matchHeight={true}>
-        <CardGrid.Row>{this.props.children}</CardGrid.Row>
-      </CardGrid>
-    );
+    return <Gallery gutter={'sm'}>{this.props.children}</Gallery>;
```
mcoker: This creates the empty state as a grid item, too. Can we only include the gallery layout if there are connections?

Screen Shot 2019-07-30 at 5 09 07 PM

mcoker: Also looks like you're missing a global variable (`--pf-l-gallery--m-gutter--GridGap`) that the gallery uses to define space responsively. That var comes from the core base CSS. Is it possible you have a newer version of the `Gallery` component than the base CSS?

Screen Shot 2019-07-30 at 5 15 43 PM

app/ui-react/packages/ui/src/Dashboard/Dashboard.css
```diff
@@ -3,19 +3,54 @@
...
+}
+
 .dashboard__connections__actions {
   padding: 0 10px;
 }
 
-.dashboard__metrics .card-pf.card-pf-aggregate-status {
-  padding: 0 10px 24px 10px;
+.dashboard__metrics .pf-l-grid__item {
+  margin-bottom: var(--pf-global--spacer--md);
```
mcoker: this is from the missing var mentioned in https://github.com/syndesisio/syndesis/pull/6287/files#r308968199. If you pull in that var, you shouldn't need this.

app/ui-react/packages/ui/src/Dashboard/Dashboard.css
```diff
@@ -3,19 +3,54 @@
...
+}
+
+@media screen and (min-width: 768px) {
+  .dashboard__metrics .pf-l-grid__item {
+    margin-bottom: 0;
```
mcoker: same. ideally we would get rid of this.

app/ui-react/packages/ui/src/Connection/ConnectionsGridCell.tsx
seanforyou23: If it's ok, I'd like to make this the very first thing I do in a follow-up PR?

seanforyou23:
Ok just pushed some updates, thx for the feedback. @mcoker I had to manually set [that gutter variable](https://github.com/syndesisio/syndesis/pull/6287/files#diff-4e966155b50ec615c88fe7bb7ec04303R24), not sure why but I think it is to do with the order of stylesheet injection. Let me know if you spot anything obvious, otherwise, I'll work on correcting this in a follow-up PR so that we properly receive the default from core. Lemme know what you think!

PR #6293, opened 2019-08-01
fix(ci): nightly CircleCi integration test execution

christophd:
PR fixes integration test execution on CircleCi

The issues were related to copying/mounting files (the Syndesis integration project) to the S2i image. The file permissions got messed up as the container is run with another user that is not the owner of the files copied/mounted.

For some reason this is not an issue on MacOS with Docker for desktop but it was on CircleCi. That made error analysis a bit more complicated.

I solved the issue with taking care of file permissions as recommended by S2i best practices.

### DIFF
--- a/app/test/integration-test/pom.xml
+++ b/app/test/integration-test/pom.xml
@@ -36,13 +36,14 @@
 
     <syndesis.integration.runtime>spring-boot</syndesis.integration.runtime>
     <syndesis.camel.k.customizers>health,logging,syndesis</syndesis.camel.k.customizers>
-    <syndesis.logging.enabled>false</syndesis.logging.enabled>
+    <syndesis.logging.enabled>true</syndesis.logging.enabled>
     <syndesis.server.port>8080</syndesis.server.port>
     <syndesis.management.port>8081</syndesis.management.port>
     <syndesis.debug.port>5005</syndesis.debug.port>
     <syndesis.debug.enabled>false</syndesis.debug.enabled>
     <syndesis.s2i.build.enabled>false</syndesis.s2i.build.enabled>
     <syndesis.output.directory>${project.build.directory}/integrations</syndesis.output.directory>
+    <syndesis.project.mount.path>/tmp/src</syndesis.project.mount.path>
     <syndesis.image.tag>${syndesis.version}</syndesis.image.tag>
   </properties>
 
@@ -199,13 +200,6 @@
           <skip>true</skip>
         </configuration>
       </plugin>
-      <!-- Required only for CircleCi build as the used JVM to run the tests has limited cryptography -->
-      <plugin>
-        <artifactId>maven-enforcer-plugin</artifactId>
-        <configuration>
-          <skip>true</skip>
-        </configuration>
-      </plugin>
 
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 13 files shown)


christophd:
here is a successful CircleCi workflow executing the tests: https://circleci.com/workflow-run/b1407a88-9f0b-42b5-8eb8-75d946375b83

app/test/integration-test/pom.xml
```diff
@@ -36,13 +36,14 @@
 
     <syndesis.integration.runtime>spring-boot</syndesis.integration.runtime>
     <syndesis.camel.k.customizers>health,logging,syndesis</syndesis.camel.k.customizers>
-    <syndesis.logging.enabled>false</syndesis.logging.enabled>
```
Delawen: This is for debugging or do we want to keep it on?

app/test/integration-test/pom.xml
```diff
@@ -36,13 +36,14 @@
...
+    <syndesis.logging.enabled>true</syndesis.logging.enabled>
     <syndesis.server.port>8080</syndesis.server.port>
     <syndesis.management.port>8081</syndesis.management.port>
     <syndesis.debug.port>5005</syndesis.debug.port>
     <syndesis.debug.enabled>false</syndesis.debug.enabled>
     <syndesis.s2i.build.enabled>false</syndesis.s2i.build.enabled>
     <syndesis.output.directory>${project.build.directory}/integrations</syndesis.output.directory>
+    <syndesis.project.mount.path>/tmp/src</syndesis.project.mount.path>
```
Delawen: Are we assuming linux/unix here? Is that fine? No Windows?

tools/bin/commands/integration-test
```diff
@@ -37,6 +42,29 @@ maven_args() {
         args="$args -Dsyndesis.version=${release_version}"
     fi
 
+    local image_tag="$(readopt --image)"
+    if [ -n "${image_tag}" ]; then
+        args="$args -Dsyndesis.image.tag=${image_tag}"
+    fi
+
+    if [ "$(hasflag --logging)" ]; then
+        args="$args -Dsyndesis.logging.enabled=true"
```
Delawen: Shouldn't this refer to the property ${syndesis.logging.enabled}?

app/test/integration-test/pom.xml
christophd: yes, it enables some additional logs for debugging/troubleshooting

app/test/integration-test/pom.xml
christophd: windows? what's that? 😄 just kidding. good point. in Maven this will be difficult to set according to os but I think we can automatically take care of proper paths in Java code. Let me fix this

tools/bin/commands/integration-test
christophd: No not really as this is a shortcut flag on the Syndesis cli tool and it overwrites the property used in Maven

app/test/integration-test/pom.xml
christophd: while trying to fix I realized that this is a mount path inside of the Docker container. And this will always be a unix style environment. so no need to support windows here.

app/test/test-support/src/main/java/io/syndesis/test/container/dockerfile/SyndesisDockerfileBuilder.java
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
```
apupier: 2016?

christophd: I think all of our files are using that date and it is protected by an enforcer rule. I once got build errors while using another date here.

apupier: ok

Issue #6306, opened 2019-08-05
Push images built in PRs to quay.io


zregvart:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

When we build PRs on CircleCI the images we build need to be pushed to quay.io and tagged with the PR number. This will help to run the smoke tests on individual PRs

zregvart:
If we push from CircleCI to quay.io then we need to provide credentials to PR jobs, meaning that we expose our passwords to anyone that files a PR.

We need to rethink this.

christophd:
Let me share some idea: 

What about triggering another build via webhook or something (e.g. on fabric8 Jenkins) when a CircleCi PR build has been successful. The 2nd job could pull the artifacts and push those to quay.io. I think in fabric8 Jenkins we can provide the credentials in a more secure way to not expose these to all users.

So the idea is to let CircleCi check the PR for any problems and the 2nd job on Jenkins is there to push the images.

@zregvart what do you think?

zregvart:
@christophd this is the approach taken in #6311, doesn't work due a docker issue that's a bit hard to troubleshoot

christophd:
Ok I see. What a bummer!

christophd:
The issue with `docker save` has been solved by #6553. The CircleCI PR build now produce Docker image tar.gz artifacts (e.g. /images/syndesis-server.tar.gz). We can now continue to push these to quay.io

heiko-braun:
@christophd Any idea how we can take them quay? 

What about this? https://circleci.com/docs/2.0/private-images/ Does it hve security implications?

christophd:
@heiko-braun I think we can now let our Jenkins take over and do the push to quay

PR #6311, opened 2019-08-05
chore(ci): store docker images as artifacts

zregvart:
The idea here is that the stored artifacts can be downloaded and then imported into quay.io

Ref #6306


christophd:
@zregvart closing this one as the issue has been fixed by another PR based on your changes. See #6553

Issue #6315, opened 2019-08-06
Docker version installed with Fedora does not have all features required


phantomjinx:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
Changes made to the syndesis [build scripts](https://github.com/syndesisio/syndesis/blob/master/install/operator/.lib.sh#L138) utilise a 'recent' docker feature whereby a Dockerfile can be redirected to the docker command via -f, ie.
```
/usr/local/bin/docker build -t "my-test-image" . -f - << EOF
FROM golang:1.12.0
EOF
```
This is a feature apparently implemented back in [2017](https://github.com/moby/moby/issues/27393), where moby is a component of the docker binary.

However, this feature does not work on docker installs available via Fedora's rpm system. Versions from both Fedora 29 & 30 fail with the following error:
```
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /home/phantomjinx/docker/-: no such file or directory
```

The versioning system of the Fedora docker is a little different. Whereas [docker community-edition (CE)](https://docs.docker.com/install/linux/docker-ce/fedora) starts at 18.09.1, the Fedora version of Docker has a version of 1.13.1.

So,
- Up until this point, it would appear that just have docker installed has been enough to do everything required in the syndesis build scripts;
- Seems to be the first occasion where a feature has been required yet missing from certain versions;

Next,
- Do we specify a specific minimum version in both syndesis script env checks and in documentation?
-- Downside: Forcing many users to uninstall distro-specified version and [install new repository](https://docs.docker.com/install/linux/docker-ce/fedora/) and new version;
- Do we try to keep the scripts as backward compatible as possible?
-- Downside: Reverting features and maintaining backward compatibility.

lburgazzoli:
There is a downside of replacing distro specific version with the one from docker: they do provide repositories for newer fedora release with huge delays. I do not know if it still holds true but docker-ce didn't had support for SElinux

Should we investigate using buildah ?

chirino:
Just `cat > tmpfile << EOF` and the. Feed the tmpfile to docker.

gashcrumb:
Theoretically in more recent fedora versions you can install `podman` and do `alias docker=podman` and most things should "just work".  I haven't personally tried that with our build though :smile:

gashcrumb:
[...]

Issue #6231, opened 2019-07-23
We should add build env check


KurtStam:
I'm running `syndesis build -f` to get a quick compile of all my code after. I ran into 2 issues:

- my go was too old for the install/operator
- Building camel-k, ERROR: $GOPATH has to be set for camel-k build

it takes a long time these days to do a quick compile, so I think we should do a quick env check and fail fast.

mmuzikar:
@KurtStam @phantomjinx is there anything that QE should verify?

KurtStam:
I guess not untill we have this env check in place, then you should see if fail fast.

phantomjinx:
@mmuzikar 
If you look at the PR [comment](https://github.com/syndesisio/syndesis/pull/6322#issue-304758847), it details a list of tests I ran against while adding in the env checks. These could be used as a basis for any QE testing. WDYT?

mmuzikar:
I did a basic verification of running the `syndesis build -f` command, and the env checks worked well for me. I am closing this as verified since if there is an error there should be a new issue opened for that anyways.

PR #6322, opened 2019-08-06
Fixes for #6231 & #6315

phantomjinx:
Done a lot of testing with different permutations, including
1. go does not exist
2. go is too low a version
3. Neither docker nor go installed
4. go is a higher version (1.13)
5. docker does not exist
6. docker not running - docker info fails
7. operator-sdk not installed
8. GO_PATH not set
9. GO_PATH is wrong
10. Project not under GO_PATH

All or some of these will basically warn/stop the operator or camel-k from building.

### DIFF
--- a/install/operator/.lib.sh
+++ b/install/operator/.lib.sh
@@ -39,71 +39,51 @@ readopt() {
     echo $default
 }
 
-
-docker_is_available() {
-    if [[ "$(which docker)" == "" ]] ; then
-        echo "warn: docker command not installed"
-        echo
-        return 0
-    fi
-
-    if docker info > /dev/null 2> /dev/null ; then
-        return 1
-    else
-        echo "warn: the docker comamand is not conected to a server"
-        echo
-        return 0
-    fi
-}
-
 build_operator()
 {
     local strategy="$1"
     shift
-    local gopackage="$1"
-    shift
+
+    local hasgo=$(go_is_available)
+    local hasdocker=$(docker_is_available)
 
     if [ "$strategy" == "auto" ] ; then
-
-        if docker_is_available; then
-            if [[ "$(which go)" == "" ]] ; then
-                echo 'error: you must have either docker or go installed to build this project.'
-                exit 1
-            else
-                strategy="go"
-            fi
+        if [ "$hasgo" == "OK" ]; then
+            # Prefer go over docker - TODO: do we prefer this?
+            strategy="go"
+        elif [ "$hasdocker" == "OK" ]; then
+            # go not available so try docker
+            printf "WARN: Building with 'docker' since 'go' command is not available ... \n\t\t$hasgo\n"
+            strategy="docker"
         else
-            if [[ "$(which go)" == "" ]] ; then
-                echo 'warn: building with docker since you do not have go installed.'
-                strategy="docker"
-            else
-                strategy="go"
-            fi
+            echo "ERROR: Building the operator requires either 'docker' or 'go' commands to be installed."
+            exit 1
         fi
     fi
 
-
     case "$strategy" in
     "go")
+        if [ "$hasgo" != "OK" ]; then
+            echo "$hasgo"
+            exit 1
+        fi
+
         echo ======================================================
         echo Building executable with go tooling
         echo ======================================================
         export GO111MODULE=on
 
         go mod vendor
-        if [[ "$(which operator-sdk)" != "" ]] ; then
-            if [[ "$(pwd)" != "$GOPATH/src/${gopackage}" ]] ; then
-                echo
-                echo "warnning: operator-sdk only works on project's in the \$GOPATH"
-                echo "          can't use it to update the generated code"
-                echo "          please move this project under the \$GOPATH so that :'$(pwd)'"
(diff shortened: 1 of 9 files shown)


phantomjinx:
Added in the easiest workaround at the moment for #6315. If we want to explore a better solution then no problem.

install/operator/.lib.sh
```diff
@@ -39,71 +39,51 @@ readopt() {
...
-                echo "          please move this project under the \$GOPATH so that :'$(pwd)'"
-                echo "          is located at '$GOPATH/src/${gopackage}'"
-                echo
-            else
-                operator-sdk generate k8s
-                # operator-sdk generate openapi
-            fi
+
+        local hassdk=$(operatorsdk_is_available)
+        if [ "$hassdk" = "OK" ]; then
```
zregvart:
```suggestion
        if [ "$hassdk" == "OK" ]; then
```

install/operator/.lib.sh
```diff
@@ -39,71 +39,51 @@ readopt() {
...
-            if [[ "$(which go)" == "" ]] ; then
-                echo 'error: you must have either docker or go installed to build this project.'
-                exit 1
-            else
-                strategy="go"
-            fi
+    local hasgo=$(go_is_available)
+    local hasdocker=$(docker_is_available)
+
+    if [ "$strategy" = "auto" ] ; then
```
zregvart:
```suggestion
    if [ "$strategy" == "auto" ] ; then
```

install/operator/.lib.sh
```diff
@@ -39,71 +39,51 @@ readopt() {
...
+        if [ "$hasgo" = "OK" ]; then
```
zregvart:
```suggestion
        if [ "$hasgo" == "OK" ]; then
```

install/operator/.lib.sh
```diff
@@ -39,71 +39,51 @@ readopt() {
...
+            # Prefer go over docker - TODO: do we prefer this?
+            strategy="go"
+        elif [ "$hasdocker" = "OK" ]; then
```
zregvart:
```suggestion
        elif [ "$hasdocker" == "OK" ]; then
```

install/operator/.lib.sh
```diff
@@ -177,15 +165,31 @@ build_image()
...
 
+    local hasdocker=$(docker_is_available)
+    local hasoc=$(is_oc_available)
+
     if [ "$strategy" == "auto" ] ; then
-        strategy=docker
-        if [ -n "$(which oc)" ] ; then
-            strategy=s2i
+
+        if [ "$hasoc" = "OK" -o "$hasoc" = "$SETUP_MINISHIFT" ]; then
```
zregvart:
```suggestion
        if [ "$hasoc" == "OK" -o "$hasoc" == "$SETUP_MINISHIFT" ]; then
```

install/operator/.lib.sh
```diff
@@ -177,15 +165,31 @@ build_image()
...
+            strategy="s2i"
+        elif [ "$hasdocker" = "OK" ]; then
```
zregvart:
```suggestion
        elif [ "$hasdocker" == "OK" ]; then
```

install/operator/.lib.sh
```diff
@@ -177,15 +165,31 @@ build_image()
...
+            strategy="docker"
+        else
+            echo "ERROR: Building an image requires either 'docker' or 'oc' commands to be installed."
+            exit 1
         fi
     fi
 
     case "$strategy" in
     "s2i")
+        if [ "$hasoc" = "$SETUP_MINISHIFT" ]; then
```
zregvart:
```suggestion
        if [ "$hasoc" == "$SETUP_MINISHIFT" ]; then
```

phantomjinx:
> Couple of typos found. One concept we have used thus far in the build scripts was to call functions via `check_error`, this departs from this a bit. Would it be possible to replace (some of?) these comparisons of `[ "$..." != "OK" ]`/`[ "$..." == "OK" ]` with `check_error $...`?

check_error calls `exit 1` if thrre is an ERROR. In these circumstances, this is not appropriate as a failure can allow another logic route to be chosen with `else` block.

install/operator/.lib.sh
phantomjinx: Didn't the scripts need to be shell agnostic and avoid specific implementations?

zregvart: Even if that was the case, which I'm fairly certain it isn't, that ship has sailed, we're very bash-specific in our shell scripts.

phantomjinx: Was just something Roland commented on in one of my first PRs late last year.

zregvart: I'm sure he would want us to be better at that, I'm not sure if that's even possible, I think we're using quite a lot of bashisms at this point; also not sure if that's required bash is pretty much universal `sh`, the only use case would be within a busybox based distribution, and I don't think we have that use case.

phantomjinx:
@zregvart 
Are you good with the explanation above or do you want to discuss further?

PR #6325, opened 2019-08-06
Fix #6292 : exclude org.json:json from openshift-client

### LINKED ISSUE #6292: server/endpoint/pom.xml issue with org.json:json dependency
## This is a...

Trying to build syndesis:master and getting the following in server/endpoint.

```
[INFO] --- maven-dependency-plugin:3.1.1:analyze-only (basepom.default) @ server-endpoint ---
[WARNING] Used undeclared dependencies found:
[WARNING]    org.json:json:jar:20140107:compile
[INFO] ------------------------------------------------------------------------
```
```
[INFO] Server :: Inspector ................................ SUCCESS [ 13.996 s]
[INFO] Server :: Credential ............................... SUCCESS [ 24.949 s]
[INFO] Server :: Endpoint ................................. FAILURE [ 16.563 s]
```

I'm not sure how to respond to this - if I exclude org.json:json from openshift-client (which is bringing it in), then I get an error 

```
[INFO] --- dependency-management-maven-plugin:0.11:analyze (basepom.default) @ server-endpoint ---
[ERROR] Exclusions must be removed for managed dependency io.fabric8:openshift-client:jar
```

If I add a dependency to server/endpoint for org.json:json, I get a complaint that android-json has classes that are duplicated from org.json:json.

cunningt:
https://github.com/syndesisio/syndesis/issues/6292

Excluding org.json:json from openshift-client.    I believe the reason that I'm seeing this in the dev build and it's not seen in syndesis ci is that we're building and aligning to a newer version of kubernetes-client (which produces io.fabric:openshift-client)

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -2462,8 +2462,18 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
+        <groupId>io.fabric8</groupId>
+        <artifactId>openshift-client</artifactId>
+        <version>${kubernetes.client.version}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
+          </exclusion>
+        </exclusions>
+      </dependency>
     </dependencies>
-
   </dependencyManagement>
 
   <!-- Metadata need to publish to central -->


zregvart:
Would it make sense to align the kubernetes client version (i.e. set the `kubernetes.client.version` to a common value)?

cunningt:
@zregvart We run into the issue of the builds of kubernetes-client disappearing after two weeks.    I think maybe the best solution is to update to the 7.4 version once 7.4 goes GA, but I'd like to get this change in in the meantime so that I can start 7.5 builds.    If that sounds okay, I'll create a task for myself to update the version once 7.4 goes GA and the k-c artifact is available.

zregvart:
Sounds good, I think we need to fix the CI build (not caused by this PR) to get our PRs merged first.

cunningt:
Logged https://github.com/syndesisio/syndesis/issues/6339 on updating k-c and any other Fuse components to 7.4 versions

PR #6345, opened 2019-08-09
[operator] Fix datavirt installation

### LINKED ISSUE #6314: Unable to deploy from master due to malformed config.json
## This is a...

<pre><code>
[ ] Feature request
[ ] Bug report  
</code></pre>

## Description
Deploying syndesis from master results in a non-functional UI, only shows the `Something is wrong` page.

As a workaround, one can edit the `syndesis-ui-config` map and editing the `config.json` key there to have `"enabled": 0` instead of just `"enabled:"`

Furthermore, any change to the `syndesis-ui-config` is reverted by the operator, so for the above workaround to actually help, one needs to scale down the operator to 0 first.

asmigala:
Seems to fix #6314 but needs another set of eyes whether it makes sense (first time writing operator code or even go).

Also needs polishing

### DIFF
--- a/install/operator/pkg/syndesis/configuration/configuration.go
+++ b/install/operator/pkg/syndesis/configuration/configuration.go
@@ -43,6 +43,7 @@ const (
 	EnvIntegrationStateCheckInterval  SyndesisEnvVar = "INTEGRATION_STATE_CHECK_INTERVAL"
 	EnvSarNamespace                   SyndesisEnvVar = "SAR_PROJECT"
 	EnvKomodoMemoryLimit              SyndesisEnvVar = "KOMODO_MEMORY_LIMIT"
+	EnvDatavirtEnabled                SyndesisEnvVar = "DATAVIRT_ENABLED"
 
 	EnvSyndesisServerTag              SyndesisEnvVar = "SYNDESIS_SERVER_TAG"
 	EnvSyndesisUITag              	  SyndesisEnvVar = "SYNDESIS_UI_TAG"
@@ -110,6 +111,7 @@ var (
 		envPrometheusTag,
 
 		envUpgradeVolumeCapacity,
+		envDatavirtEnabled,
 	}
 
 	setters = []SyndesisEnvVarSetter{
@@ -663,5 +665,16 @@ func exposeVia3Scale(config map[string]string, syndesis *v1alpha1.Syndesis) {
 		if b, err := strconv.ParseBool(v); err != nil {
 			syndesis.Spec.Components.Server.Features.ExposeVia3Scale = &b
 		}
+	}
+}
+
+func envDatavirtEnabled(syndesis *v1alpha1.Syndesis) *SyndesisEnvVarConfig {
+	v := "0"
+	if komodo := syndesis.Spec.Addons["komodo"]["enabled"]; komodo == "true" {
+		v = "1"
+	}
+	return &SyndesisEnvVarConfig{
+		Var:   EnvDatavirtEnabled,
+		Value: v,
 	}
 }
--- a/install/operator/pkg/syndesis/configuration/configuration_test.go
+++ b/install/operator/pkg/syndesis/configuration/configuration_test.go
@@ -17,10 +17,11 @@ func TestStandardConfig(t *testing.T) {
 	}
 
 	config := GetEnvVars(&syndesis)
-	assert.Len(t, config, 3)
+	assert.Len(t, config, 4)
 	assert.Contains(t, config, string(EnvOpenshiftProject))
 	assert.Contains(t, config, string(EnvSarNamespace))
 	assert.Contains(t, config, string(EnvExposeVia3Scale))
+	assert.Contains(t, config, string(EnvDatavirtEnabled))
 }
 
 func TestSpecificConfig(t *testing.T) {
--- a/install/operator/pkg/syndesis/template/install.go
+++ b/install/operator/pkg/syndesis/template/install.go
@@ -92,6 +92,13 @@ func SetupRenderContext(renderContext *generator.Context, syndesis *v1alpha1.Syn
 		syndesis.Spec.Addons["todo"]["enabled"] = "true"
 	}
 
+	if syndesis.Spec.Addons["komodo"] == nil {
+		syndesis.Spec.Addons["komodo"] = v1alpha1.Parameters{}
+	}
+	if syndesis.Spec.Addons["komodo"]["enabled"] == "" {
+		syndesis.Spec.Addons["komodo"]["enabled"] = "false"
+	}
+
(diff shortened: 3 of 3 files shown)


install/operator/pkg/syndesis/configuration/configuration_test.go
```diff
@@ -17,10 +17,11 @@ func TestStandardConfig(t *testing.T) {
 	}
 
 	config := GetEnvVars(&syndesis)
-	assert.Len(t, config, 3)
+	assert.Len(t, config, 4)
```
asmigala: The tests were failing without this (and the line below), but I'm not sure why it's these four env vars that need to be specified, when there's lots of others in configuration.go.

install/operator/vendor/modules.txt
```diff
@@ -360,9 +360,9 @@ k8s.io/apimachinery/pkg/runtime/serializer
 k8s.io/apimachinery/pkg/runtime/serializer/json
 k8s.io/apimachinery/pkg/runtime/serializer/versioning
 k8s.io/apimachinery/pkg/util/json
-k8s.io/apimachinery/pkg/api/meta
 k8s.io/apimachinery/pkg/labels
 k8s.io/apimachinery/pkg/api/resource
+k8s.io/apimachinery/pkg/api/meta
```
asmigala: No idea where the changes to this file came from

install/operator/.lib.sh
```diff
@@ -135,7 +135,7 @@ build_operator()
           OPTS="$OPTS '$i'"
         done
 
-        docker build -t "${BUILDER_IMAGE_NAME}" . -f - <<EODockerfile
+        cat > Dockerfile.tmp <<EODockerfile
```
asmigala: This is a workaround for #6315 , will remove once the fix for that is merged

lgarciaaco: I think https://github.com/syndesisio/syndesis/pull/6322/files has a more complete solution for this issue, maybe we should remove the bits here

install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -498,3 +500,16 @@ func exposeVia3Scale(config map[string]string, syndesis *v1alpha1.Syndesis) {
...
 }
+
+func envDatavirtEnabled(syndesis *v1alpha1.Syndesis) *SyndesisEnvVarConfig {
+	if komodo := syndesis.Spec.Addons["komodo"]["enabled"]; komodo == "true" {
+		return &SyndesisEnvVarConfig{
+			Var:   EnvDatavirtEnabled,
+			Value: "1",
+		}
+	}
+	return &SyndesisEnvVarConfig{
```
lgarciaaco: retunr `nil` here. The place to set defaults is: 
https://github.com/syndesisio/syndesis/blob/3e529cc10e9cf6ef83b21bbf7946ae222923d1f6/install/operator/pkg/syndesis/template/install.go#L84

install/operator/.lib.sh
asmigala: Yep, will do

install/operator/pkg/syndesis/configuration/configuration_test.go
zregvart: Stab in the dark: this might be because of the default value `EnvDatavirtEnabled` has

install/operator/pkg/syndesis/configuration/configuration.go
asmigala: ok, a question, should I provide a default for the `syndesis.Spec.Addons["komodo"]["enabled"]` value?

The whole problem is with #6314 is that https://github.com/syndesisio/syndesis/blob/3e529cc10e9cf6ef83b21bbf7946ae222923d1f6/install/operator/pkg/generator/assets/infrastructure/03-syndesis-ui.yml.mustache#L122
references `Env.DATAVIRT_ENABLED` which, when unset, results in config.json like 
```
"datavirt": {
  "dvUrl": "/vdb-builder/v1/",
  "enabled": 
}
```
so we need to set the env var to either 0 or 1.

I'm thinking I should do it similarly to https://github.com/syndesisio/syndesis/blob/3e529cc10e9cf6ef83b21bbf7946ae222923d1f6/install/operator/pkg/syndesis/configuration/configuration.go#L483-L492
where we check the value from CR and set the env var accordingly.

wdyt?

install/operator/vendor/modules.txt
asmigala: I don't think so, I haven't opened that file and I edited the others in vim, so no automatic reformating. Is it possible the go compiler reformatted it? I'll revert it back anyway, just wondering

install/operator/pkg/syndesis/configuration/configuration.go
asmigala: I think I got it now, will update shortly

asmigala: With latest change the operator correctly enables or disables datavirt when `komodo.enabled` is `true`/`false` or missing entirely.

install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -498,3 +500,16 @@ func exposeVia3Scale(config map[string]string, syndesis *v1alpha1.Syndesis) {
 		}
 	}
 }
+
```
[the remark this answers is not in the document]
asmigala: Added the setter, though not sure it's useful at all, since I also added the default value in `install.go`, so form my understanding the setter will never actually set anything

install/operator/pkg/syndesis/configuration/configuration_test.go
asmigala: yep, that was it, I tried not setting the default value but then I would run into #6314 when komodo was set to disabled.

install/operator/pkg/syndesis/configuration/configuration.go
lgarciaaco: Yes you should set that 0 default value the same way as https://github.com/syndesisio/syndesis/blob/3e529cc10e9cf6ef83b21bbf7946ae222923d1f6/install/operator/pkg/syndesis/template/install.go#L107

only with your variable

zregvart:
Needs to be rebased since we merged a change from another PR to the operator...

asmigala:
@zregvart we had a call with @lgarciaaco and we decided to change this a bit, will rebase and push a new version

asmigala:
After further discussion with @lgarciaaco we decided that this is good for now, doing it properly (i.e. not using `Env` at all and instead getting the values directly from the CR) would be too much work and should be done separately.

heiko-braun:
Good job everyone!

PR #6349, opened 2019-08-09
chore: update form widgets to pf4; api provider return form definition and related enhancements

gashcrumb:
@seanforyou23 converts auto-form to use pf4 components, lookie:

image

Also @mcoker has an impact on the basic filter work you're doing.  I added a basic filter storybook, it doesn't contain any of the custom css you had added though for getting it working in the app:

image

Also @KurtStam started coming up with a [definition](https://github.com/syndesisio/syndesis/pull/6349/files#diff-3281d7774d21f7a8c35b65e250e10a0cR42-R83) for the error mapping to come up with:

image

Still have to add the filter and the reset button, that may or may not be in a subsequent PR.


gashcrumb:
btw @seanforyou23 I noticed the pf4 TextArea component throws this:

image

Also am not sure why the margin next to the checkboxes isn't right, I stared at the example and what auto-form produces a bunch too :-)

gashcrumb:
Oh, also the popovers in the labels don't appear to work @seanforyou23 but that seemed like the place to put that info popover.  Also am looking at the impact on the app at the moment :-)

gashcrumb:
Updated screenshot for error code form and link to form definition.

gashcrumb:
> Question: Do we want to do anything about aligning versions of PF between UI and auto-form pkg? Even if it's just a manual effort of keeping them in sync and no programmatic solution that enforces it, I think it may benefit us down the road.

Yeah, we should align those for sure.

gashcrumb:
> For some reason, the styles are broken for me;

Oh right, so that particular storybook uses the 'array' type and doesn't specifically set any classes or have any CSS to control formatting.  I'd like to get auto-form to produce a form for this case that "just works" and looks alright.  

So far in the app where 'array' is actually used we cheat a bunch and style up the array type to look like the relevant design.

Issue #6353, opened 2019-08-12
Create e2e and integration test for operator


lgarciaaco:
There is currently no way to do end 2 end integration tests for the operator. We introduced changes last sprint that broke installation of the operator and we didn't realize until some QE manually tested the operator; which is unacceptable.

From all I've seen out there, the [operator sdk test suite](github.com/operator-framework/operator-sdk/pkg/test) seems to be the best fit.

PR #6354, opened 2019-08-12
Add integration tests

lgarciaaco:
This PR creates the infrastructure for e2e tests for the operator and adds two basic integration tests. `installWithDefaultsTest` test the installation of sydensis when an empty CR is created. `installTest` uses a populated CR.

Did some cleanup.

Closes https://github.com/syndesisio/syndesis/issues/6353


install/operator/tests/e2e/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Running the tests
+### Prerequisites
+- a running k8s cluster and kube config.
+- [opearator sdk](https://github.com/operator-framework/operator-sdk) installed
+
+### Running
+
+You can run your local code with
+```$xslt
```
zregvart: weird did you mean
```suggestion
```shell
```

install/operator/deploy/operator.yaml
```diff
@@ -16,12 +16,10 @@ spec:
       containers:
         - name: syndesis-operator
           # Replace this with the built image name
-          image: REPLACE_IMAGE
+          image: lgarciaac/syndesis-operator:latest
```
KurtStam: Why are you pulling from your account?

lgarciaaco: Forgot to set that back!!!

install/operator/deploy/role.yaml
```diff
@@ -1,58 +1,206 @@
-apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
zregvart: this changed from `rbac.authorization.k8s.io/v1` to `rbac.authorization.k8s.io/v1beta1`, should it have?

install/operator/deploy/role_binding.yaml
```diff
@@ -1,11 +1,16 @@
 kind: RoleBinding
-apiVersion: rbac.authorization.k8s.io/v1
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
zregvart: Should this be?

```suggestion
apiVersion: rbac.authorization.k8s.io/v1
```

install/operator/deploy/service_account.yaml
```diff
@@ -2,3 +2,8 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: syndesis-operator
+  labels:
+    app: syndesis
+    syndesis.io/app: syndesis
+    syndesis.io/type: operator
```
zregvart: Perhaps like this, to distinguish from the operator? Not sure what labels we set on other service accounts...

```suggestion
    syndesis.io/type: service-account
```

install/operator/tests/e2e/install_test.go
```diff
@@ -0,0 +1,171 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package e2e
+
+import (
+    goctx "context"
```
zregvart: does `go-fmt` like this? I thought we were using tabs in Golang code

lgarciaaco: Not sure what yo meant here?

install/operator/deploy/role_binding.yaml
lgarciaaco: you probably right, I blindly copy from https://github.com/syndesisio/syndesis/blob/e14633aa7e591f635a9466ba4e28e6b5fa66557a/install/operator/pkg/generator/assets/install/operator.yml.tmpl#L13, maybe @chirino can shed some lights

install/operator/deploy/service_account.yaml
lgarciaaco: It seems to refers to the components the system account is managing, for instance:
https://github.com/syndesisio/syndesis/blob/e14633aa7e591f635a9466ba4e28e6b5fa66557a/install/operator/pkg/generator/assets/infrastructure/04-syndesis-server.yml.mustache#L1-L9

or
https://github.com/syndesisio/syndesis/blob/e14633aa7e591f635a9466ba4e28e6b5fa66557a/install/operator/pkg/generator/assets/infrastructure/02-syndesis-service-accounts.yml.mustache#L1-L9

so I guess operator is acceptable, but not sure whether we have some guidelines for it, this was again a c&p from:

https://github.com/syndesisio/syndesis/blob/e14633aa7e591f635a9466ba4e28e6b5fa66557a/install/operator/pkg/generator/assets/install/operator.yml.tmpl#L9

lgarciaaco:
@zregvart 
> I'm wondering what happens to the `syndesis-operator` role creation when cluster doesn't have some of the resources that the role references installed (Camel K, Integreatly CRs?)
> 
The test suite uses the resources in https://github.com/syndesisio/syndesis/tree/master/install/operator/deploy, if the CRD is missing it will attempt to create it. This tests only need syndesis CRDs atm,  I guess as we incorporate more tests that use other CRDs (like kamel), we would have to move those tho the crd folder
 
> The tests look at the status of the Syndesis CR to assert it was installed, should we put more asserts towards asserting that deployment configs or other resources are created?

Definitely!! This is just the bare structure to start adding more on top of that the fun is just starting ;)

install/operator/deploy/role.yaml
lgarciaaco: same as the next comment

PR #6355, opened 2019-08-12
Fixes view header display issue when navigating back and forth from view output tab to SQL tab

elvisisking:
- removed the not needed `useVirtualization` call as the virtualization is available from the route state
- added a loader for when the SQL tab is loading
- created SqlClientContentSkeleton and story


elvisisking:
Is there an issue for this @mdrillin? I couldn't seem to find it.

Issue #6344, opened 2019-08-09
[MongoDB Connector] Operations refinment


squakez:
Provide the available operations for the connector through an enumeration and provide related unit test:

* findById
* findAll
* insert
* save
* update
* remove
* count

AC:
* Producer operation must show a select-able list of operation
* All operation must have a passed unit test

Issue #6323, opened 2019-08-06
[MongoDB Connector] Meta exception thrown


squakez:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

```
2019-08-06 13:54:21.056  INFO 1 --- [localhost:27017] org.mongodb.driver.cluster               : Exception in monitor thread while connecting to server localhost:27017
com.mongodb.MongoSocketOpenException: Exception opening socket
	at com.mongodb.connection.SocketStream.open(SocketStream.java:63) ~[mongo-java-driver-3.4.3.jar!/:na]
	at com.mongodb.connection.InternalStreamConnection.open(InternalStreamConnection.java:115) ~[mongo-java-driver-3.4.3.jar!/:na]
	at com.mongodb.connection.DefaultServerMonitor$ServerMonitorRunnable.run(DefaultServerMonitor.java:113) ~[mongo-java-driver-3.4.3.jar!/:na]
```

Likely due to missing autoconfiguration spring boot loader.

AC: exception is no longer shown in meta startup load.

PR #6358, opened 2019-08-12
#6344 - operation refinement and #6323 - meta exception

squakez:
#6344 - Refined supported operation for the connector, and provided unit test for each.
#6323 - minor fix, removed startup log exception


app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/MongoClientCustomizer.java
```diff
@@ -15,36 +15,46 @@
...
 
     @Override
     public void customize(ComponentProxyComponent component, Map<String, Object> options) {
+        // Validate operation parameter, set only once
+        if (operation == null) {
+            Object operation = options.get("operation");
+            if (operation instanceof String) {
+                this.operation = (String) operation;
+            }
+            component.setBeforeProducer(this::checkValidOperation);
```
christophd: Why do we have to check the operation using a before producer? Can't we just validate the operation in the customize and throw exception right away. This way we would not need to save the operation to a local member variable in this class.

app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/MongoClientCustomizer.java
```diff
@@ -58,18 +68,28 @@ public void customize(ComponentProxyComponent component, Map<String, Object> opt
...
             } else {
                 LOGGER.warn(
-                        "Not enough information provided to set-up the MongoDB client. Required host, user and password.");
+                    "Not enough information provided to set-up the MongoDB client. Required at least host, user and password.");
             }
         }
     }
+
+    // Validate the operation among a list of possible values
+    @SuppressWarnings("PMD")
```
christophd: 😢 can we find code that is valid according to PMD? At least check for null value to avoid the NPE? And the MongoProducerOperation could provide a isValid helper providing the formatted IllegalArgumentException

app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/MongoClientCustomizer.java
squakez: I tried to find a way to know if the `customize` method is called by a producer or a consumer, since the check must be performed only by a producer endpoint. Did not find a way to identify that, any suggestion? an alternative would be to create a further customizer and configure for producer only, but I'd like to listen any alternative approach.

app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/MongoClientCustomizer.java
squakez: I don't like that either, I fell in the temptation to simply put a suppress warning as passing a non valid enumeration just throw the same exception... gonna refactor that!

app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/MongoClientCustomizer.java
squakez: I finally created a dedicated customizer just to check the operation.

PR #6359, opened 2019-08-12
fix #5493: SQL connector doesn't return generated IDs when added as start step

### LINKED ISSUE #5493: SQL connector doesn't return generated IDs when added as start step
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
follow-up to [#4466](https://github.com/syndesisio/syndesis/issues/4466), when performing the INSERT statement as a start step it doesn't return the generated IDs. that means:
 - Timer -> SQL (insert) -> Log works just fine, the log shows the generated ID
 - periodic SQL (insert) -> Log doesn't return anything

KurtStam:
fix #5493: SQL connector doesn't return generated IDs when added as start step


KurtStam:
why are these modules blocked?

- connector-catalog - Success
- meta - Blocked
- server - Blocked
- s2i - Blocked

PR #6366, opened 2019-08-13
chore(auto-form): update auto-form array layout

seanforyou23:
This PR upgrades patternfly core in auto-form and also applies a few html/css adjustments to the auto-form array layout.

update pf react-core
form array layout adjustments
update auto-form types to align with latest pf

### DIFF
--- a/app/ui-react/packages/auto-form/src/widgets/FormArrayComponent.css
+++ b/app/ui-react/packages/auto-form/src/widgets/FormArrayComponent.css
@@ -1,6 +1,6 @@
 .form-array-control__row-title {
-  text-align: right;
-  margin-top: 4px;
+  /* text-align: right;
+  margin-top: 4px; */
 }
 .form-array-control__array-controls {
   white-space: nowrap ;
@@ -8,4 +8,9 @@
 
 .form-array-control__array-controls .fa {
   font-size: 20px;
+}
+
+.form-array-layout {
+  display: grid;
+  grid-gap: var(--pf-c-form--GridGap);
 }
--- a/app/ui-react/packages/auto-form/src/widgets/FormArrayComponent.css
+++ b/app/ui-react/packages/auto-form/src/widgets/FormArrayComponent.css
@@ -1,7 +1,3 @@
-.form-array-control__row-title {
-  /* text-align: right;
-  margin-top: 4px; */
-}
 .form-array-control__array-controls {
   white-space: nowrap ;
 }
--- a/app/ui-react/packages/auto-form/src/widgets/FormSelectComponent.tsx
+++ b/app/ui-react/packages/auto-form/src/widgets/FormSelectComponent.tsx
@@ -25,8 +25,8 @@ export const FormSelectComponent: React.FunctionComponent<
   const updatedValue =
     isMultiple && typeof value === 'string' ? JSON.parse(value) : value;
   const handleChange = (
-    _: string,
-    event: React.ChangeEvent<HTMLSelectElement>
+    value: string,
+    event: React.FormEvent<HTMLSelectElement>
   ) => {
     if (isMultiple) {
       const newValue = getSelectedValues(event.currentTarget);
@@ -35,7 +35,7 @@ export const FormSelectComponent: React.FunctionComponent<
       onChange(event);
     }
   };
-  const handleBlur = (event: React.ChangeEvent<HTMLSelectElement>) =>
+  const handleBlur = (event: React.FormEvent<HTMLSelectElement>) =>
     handleChange('', event);
   return (
     <FormGroup
--- a/app/ui-react/packages/auto-form/src/widgets/FormSelectComponent.tsx
+++ b/app/ui-react/packages/auto-form/src/widgets/FormSelectComponent.tsx
@@ -25,6 +25,7 @@ export const FormSelectComponent: React.FunctionComponent<
   const updatedValue =
     isMultiple && typeof value === 'string' ? JSON.parse(value) : value;
   const handleChange = (
+    // tslint:disable-next-line:no-shadowed-variable
     value: string,
     event: React.FormEvent<HTMLSelectElement>
   ) => {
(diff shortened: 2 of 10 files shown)


mcoker:
Not sure if this is within the scope of this PR, but I would also like to see classes on some of these elements. Particularly ones that only have the `.form-array-layout` class, so that we can target them specifically in CSS.

The first child of the form (`#arrayofthings`) could have like `.form-array-container`. Then the element that follows it (`.pf-c-form__group.pf-m-action`) could have `.form-array-action`.

The section could have `.form-array-section`.

The `<div>` that wraps the `.pf-c-title` could have `.form-array-section__header`. And the title within it could change from `.form-array-control__row-title` to `.form-array-section__title`? Unless there's a reason (based on the component name or something?) to keep that as `.form-array-control__row-title`.

Should/could we change `.form-array-fields to `.form-array-section__fields`?

The div that wraps the actions in each section could be `.form-array-section__action`.

I think that's it. Basically just add a unique class to some of these elements that don't have them in case we need to build these bespoke layouts, like the basic filter UI we've been working on.

app/ui-react/packages/auto-form/src/widgets/FormArrayComponent.css
```diff
@@ -1,6 +1,6 @@
 .form-array-control__row-title {
-  text-align: right;
-  margin-top: 4px;
+  /* text-align: right;
```
gashcrumb: Should just delete these then?

app/ui-react/packages/auto-form/stories/Array.stories.tsx
```diff
@@ -30,10 +30,10 @@ stories.add('General Example', () => {
       },
       arrayDefinitionOptions: {
         arrayControlAttributes: {
-          className: 'col-md-1',
+          className: '',
```
gashcrumb: it's safe to delete this or even the whole `arrayControlAttributes` if needed.

app/ui-react/packages/auto-form/stories/Array.stories.tsx
```diff
@@ -30,10 +30,10 @@ stories.add('General Example', () => {
...
         },
         arrayRowTitleAttributes: {
-          className: 'col-md-2',
+          className: '',
```
gashcrumb: Feel free to delete this whole `arrayRowTitleAttributes`, if these aren't needed anymore.  Ideally it'd be cool if the array just kinda worked well in like 80% of cases :-)

gashcrumb:
@mcoker I'd be cool with that, WDYT @seanforyou23?  Awesome stuff BTW and the UI tests passed too :-)  I only had a couple things.

gashcrumb:
Overall looking nice BTW, just fired it up locally.

I don't know if maybe I'm passing an empty string to the label or something but I think there's some blank `label` element that's occupying all this space that I want the select control to be in:

image

app/ui-react/packages/auto-form/stories/Array.stories.tsx
seanforyou23: I think they provide nice hooks for styling in case we run into a scenario where we need a little more fine-grain control over a specific form, so I combined it with @mcoker's feedback and applied BEM convention to the class names. Now that we have classes dedicated for layout/structure, we can apply styling against these new classes for non-structual styling adjustments.

app/ui-react/packages/auto-form/stories/Array.stories.tsx
seanforyou23: I removed `col-md-2` class and replaced it with one that doesn't change the layout. This should achieve the 80% rule, as the class is not actually required to make anything work. Just a hook in case you need it on the CSS side.

seanforyou23:
Thanks for the feedback! Just pushed some updates. @gashcrumb I remember seeing an empty label yesterday while I was working on this and planned to circle back to it, but haven't re-discovered this yet today so still looking. That said, I did notice we have nested forms, which is likely adding to the problem. Do you have any idea of how we might undo that so we only have a single form element?

Screen Shot 2019-08-14 at 10 20 38 AM

Screen Shot 2019-08-14 at 10 22 09 AM

gashcrumb:
@seanforyou23 Yup, [I cheated](https://github.com/syndesisio/syndesis/blob/master/app/ui-react/packages/auto-form/src/widgets/FormMapsetComponent.tsx#L103) I think trying to work around it initially :smile:

app/ui-react/packages/auto-form/src/widgets/FormTextAreaComponent.tsx
```diff
@@ -10,8 +10,9 @@ export const FormTextAreaComponent: React.FunctionComponent<
   const { value, onChange, ...field } = props.field;
   const id = toValidHtmlId(field.name);
   const handleChange = (
-    _: string,
-    event: React.FormEvent<HTMLInputElement>
+    // tslint:disable-next-line:no-shadowed-variable
+    value: string,
```
gashcrumb: maybe rename to `eventValue` so `tslint` is happy?

app/ui-react/packages/auto-form/src/widgets/FormSelectComponent.tsx
```diff
@@ -25,8 +25,9 @@ export const FormSelectComponent: React.FunctionComponent<
   const updatedValue =
     isMultiple && typeof value === 'string' ? JSON.parse(value) : value;
   const handleChange = (
-    _: string,
-    event: React.ChangeEvent<HTMLSelectElement>
+    // tslint:disable-next-line:no-shadowed-variable
+    value: string,
```
gashcrumb: I'd rename this to `eventValue` instead.

mcoker:
LGTM! Thanks @seanforyou23

seanforyou23:
@gashcrumb made those updates - also I furthered the cheat and replaced the nested `<form>` element with a div and used the `pf-c-form` class. Still has the layout issue, and to be honest, this is kinda abusing the patternfly class. That said, at least it fixes the error about nested forms. I figure we can tackle that in a follow issue if it's ok. Lemme know if you see anything else!

Issue #6227, opened 2019-07-23
Comma in the integration title crashes the integration


mmuzikar:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
I created an integration which had a comma in the name, it caused the integration to crash. Without any feedback why it happened or what did I do wrong. Since there's a possibility the user couldn't have access to Openshift and the logs, they wouldn't know what's wrong and most likely wouldn't suspect the integration name to be cause of their troubles.
This is the stacktrace produced:
```
Caused by: javax.management.MalformedObjectNameException: Could not create ObjectName from: org.apache.camel:context=test,,type=context,name="test,". Reason: javax.management.MalformedObjectNameException: Invalid character ',' in key part of property
	at org.apache.camel.management.DefaultManagementNamingStrategy.createObjectName(DefaultManagementNamingStrategy.java:418) ~[camel-core-2.21.0.fuse-740038.jar!/:2.21.0.fuse-740038]
	at org.apache.camel.management.DefaultManagementNamingStrategy.getObjectNameForCamelContext(DefaultManagementNamingStrategy.java:106) ~[camel-core-2.21.0.fuse-740038.jar!/:2.21.0.fuse-740038]
	at org.apache.camel.management.DefaultManagementLifecycleStrategy.onContextStart(DefaultManagementLifecycleStrategy.java:167) ~[camel-core-2.21.0.fuse-740038.jar!/:2.21.0.fuse-740038]
	... 40 common frames omitted
```
I believe there are many other characters that cause this behavior... I've found ':' to be problematic as well with quick search on the interwebz. 
## What I'd expect to happen
Either we shouldn't allow these "problematic" characters in the integration name
![image](https://user-images.githubusercontent.com/46345469/61691164-3e2f5880-ad2b-11e9-9693-507e2f76bb25.png)
Here the Name input should shine red, tell the user what's wrong, disable the save and publish buttons.
OR 
Use commas and other characters just for display, I can imagine someone using : or , in the title.

squakez:
The failure is caused by the [ObjectName](https://docs.oracle.com/javase/8/docs/api/javax/management/ObjectName.html) class which is used for jmx managment. This set of characters is going to provoke failure: `,:*?=`

squakez:
[...]

Issue #3951, opened 2018-10-24
Don't use user input in names, labels or annotations of k8s objects


zregvart:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

We should not use any user provided input in key-like properties of k8s objects. We currently rely on name sanitization which can yield non-unique names in several cases:

 1. two different names equal in the first 63 characters
 1. two different names equal in alphanumeric characters and placements, non equal in non-alphanumeric characters and placements (e.g. `hello!` and `hello?`)

We should instead either hash the user provided input or generate random values.

gashcrumb:
+1!  It also imposes arbitrary restrictions on what users can use for names that don't really make sense these days.

squakez:
I can give a look on this. @zregvart do you have some example where I can look at? right now I'm working on  #6371 but I don't have enough context for this. Thanks!

squakez:
Development done to partially accomplish the feature is available here: https://github.com/squakez/syndesis/tree/feature/3951_sanitized_collisions. Dropped in favor of a different redesign of the feature.

PR #6371, opened 2019-08-14
fix(core): overriding integration name

squakez:
Setting a pattern to allow certain chars only and avoid integration crash as described in #6227 - we may include a different set of allowed chars.


gashcrumb:
I'd rather we suppress the invalid characters where we're using a user-supplied name for a resource vs limiting what characters the user can enter personally.

squakez:
> I'd rather we suppress the invalid characters where we're using a user-supplied name for a resource vs limiting what characters the user can enter personally.

This can be still done by replacing chars when the user is creating the integration. In any case I'd keep the server side check to make sure that no strange char is going to mess up neither the db insertion, neither the future integration. With this PR I'd like to "fail fast" instead of waiting the integration to run, as reported in the issue.

gashcrumb:
I think we should avoid constraining what characters the user can use when naming their integration or connection.  It's bad UX.

Make the conversion to safe characters (like `_`) when a resource such as an object name needs to be created vs mutating the user input.

zregvart:
I agree with @gashcrumb rather than forcing the user to adapt we should. Also of interest is #3951.

squakez:
Thanks for the feedback. I've gone through the implementation and I tend to think that can be a bug in Camel: https://github.com/apache/camel/compare/master...squakez:fix/object_name_context?expand=1#diff-6630627c35eabc53e1eec02a91192421L200

I've commented on dev mailing list to see how to progress further, not sure of the deeper implication of such change.

squakez:
It seems that quoting context is not a solution as it would affect the overall design. Then, I will find a way to replace the unwanted characters that finally ends up in the project generator [springboot.name](https://github.com/syndesisio/syndesis/blob/05431de3015812ad39ac8bb1a5836fb46622e713/app/integration/project-generator/src/main/resources/io/syndesis/integration/project/generator/templates/application.properties.mustache#L12) application property. Please, notice that also the [jaeger service](https://github.com/syndesisio/syndesis/blob/05431de3015812ad39ac8bb1a5836fb46622e713/app/integration/project-generator/src/main/resources/io/syndesis/integration/project/generator/templates/application.properties.mustache#L36) may be affected by the same. I guess I can use the same integration name replacement, there. Please advice.

squakez:
I've introduced a default sanitize method that will be executed during the project integration assembly phase. It can be overridden by any specific further implementation if it needs to be extended for any platform.

@chirino, @gashcrumb, @zregvart  can you have a look now please?

gashcrumb:
LGTM, thanks!

zregvart:
We also have this:
https://github.com/syndesisio/syndesis/blob/16498025712e59d3cf4e9ac473f3575779234b1e/app/common/util/src/main/java/io/syndesis/common/util/Names.java#L52

Might be good to have all that logic in one place.

app/integration/api/src/main/java/io/syndesis/integration/api/IntegrationResourceManager.java
```diff
@@ -202,10 +208,21 @@ default Integration sanitize(Integration integration) {
...
     }
 
+    /**
+     * Sanitize an integration name according a specific set of characters
+     *
+     * @param name the name to sanitize
+     * @return a sanitized name replacing illegal characters
+     */
+    default String sanitize(String name) {
+        return name == null ? null : Names.sanitize(name);
```
zregvart: I'd either move this to `Names` or make it throw NullPointerException, I think integration name is required (not sure how enforced though)

squakez: I partially agree on that. Actually the `sanitize()` method is meant to be generic, not necessarily specific to the name. We already have in place a validation of the `null` value in the model https://github.com/syndesisio/syndesis/blob/39513aa3af7ab70633f31875ec2d155651abeb07/app/common/model/src/main/java/io/syndesis/common/model/WithName.java#L28
Since it's something generic I prefer to have the check since I don't know before hand what kind of values will be provided there.

PR #6374, opened 2019-08-14
TEIIDTOOLS-731 changing image name to syndesis-dv

shawkins:


shawkins:
Requires https://github.com/teiid/teiid-syndesis/pull/204

Issue #6373, opened 2019-08-14
Operator is not able to install the syndesis-default service account


lgarciaaco:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
It throws:
```
{"level":"error","ts":1565692702.8354821,"logger":"kubebuilder.controller","msg":"Reconciler error","controller":"syndesis-controller","request":"syndesis/example","error":"yaml: line 9: did not find expected key:\n- apiVersion: v1\n  kind: ServiceAccount\n  metadata:\n    name: syndesis-default\n    labels:\n      app: syndesis\n      syndesis.io/app: syndesis\n      syndesis.io/type: infrastructure\n      syndesis.io/component: syndesis-infrastructure\n  - name: \"syndesis-pull-secret\"\n\n","stacktrace":"github.com/go-logr/zapr.(*zapLogger).Error\n\t/home/circleci/project/install/operator/vendor/github.com/go-logr/zapr/zapr.go:128\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/home/circleci/project/install/operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:217\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1\n\t/home/circleci/project/install/operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:158\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/home/circleci/project/install/operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/home/circleci/project/install/operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:134\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/home/circleci/project/install/operator/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:88"}
```

Apparently there is a problem when creating the resources in https://github.com/syndesisio/syndesis/blob/master/install/operator/pkg/generator/assets/infrastructure/02-syndesis-service-accounts.yml.mustache

In the meantime if you hit this bug, remove the following lines that are needed only for productized images:
[...]

PR #6376, opened 2019-08-14
Move service accounts to service-account file

lgarciaaco:
No idea what causes the bug, but changing to go template seems to fix it, and this is the way to go anyway

Fixes https://github.com/syndesisio/syndesis/issues/6373


chirino:
LGTM

PR #6389, opened 2019-08-15
fix(6264): Links up setting --dev in syndesis script with Debug in op…

phantomjinx:
…erator

* When --dev is applied to install of syndesis script, the DevImages flag
  is set to true. This implies that development images should be used
  rather than docker hub released images.

* In addition, in this situation, the Debug flag is set to true which
  in turn sets JAVA_DEBUG to true on all pods, allowing them to be
  debugged.


install/operator/pkg/syndesis/template/install.go
```diff
@@ -160,6 +159,16 @@ func SetupRenderContext(renderContext *generator.Context, syndesis *v1alpha1.Syn
 		return fmt.Errorf("required config var not set: %s", configuration.EnvSarNamespace)
 	}
 
+	//
+	// Apply DevImages flag value to Debug
+	//
+	// If using dev images then doing development so
+	// useful to have JAVA_DEBUG set to allow for image debugging
+	//
+	if syndesis.Spec.DevImages != nil {
```
zregvart: Perhaps this would be a good opportunity to rename `DevImages` to `DevSupport` (like we already have `TestSupport`?

PR #6390, opened 2019-08-15
feat(typescript): update loaders, tweak webpack config

seanforyou23:
This PR replaces `awesome-typescript-loader` with `ts-loader`, a move we've seen necessary before. awesome-typescript-loader doesn't seem to be maintained in about a year. We're having a problem where some basic code from [patternfly's table component](https://www.patternfly.org/v4/documentation/react/components/table/) is running fine two other apps, but storybook doesn't seem to like it and we get compile errors. So, this PR is mainly to see if it by chance happens to be the use of this out of date loader that's causing problems.

I gave it a quick run and the UI seems to work ok and auto-form/ui package storybooks work also. Test suite runs. I'm having trouble connecting to the backend at the moment so I haven't been able to test very thoroughly - please test this well before merging!

@elvisisking ping

Those new options experimentalWatchApi and transpileOnly are nice to have, can remove them if they cause any problems, they should help speed up the time it takes to start dev server and compile stuff.


seanforyou23:
I think live reloading may be broken with these changes. Working on a fix but need more time. If the storybook stuff works for @elvisisking then I'll resume the rest of this work. Let me know how it goes! 🤞

elvisisking:
@seanforyou23: I pulled your PR branch into my branch that uses `react-table`. Then, in `packages/ui`, I upgraded `react-table` to `2.17.1` and upgraded `react-core` to `3.86.3`.  A full build worked fine. Storybook now starts fine and startup is **WAY** faster than usual. My table story works  and hot code replace of story and table component works fine.

Issue #6336, opened 2019-08-07
Basic filter UI enhancements


mcoker:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

As noted in https://github.com/syndesisio/syndesis/issues/6222, the form fields in the basic filter conditions are as wide as the viewport allows, which is too wide. The form elements on the page should be limited to a max-width around 600px.

You lose the "continue only if incoming data match" label/select after entering a few conditions and the page starts to scroll. That condition block should be sticky to the top of that scrollable area so it's always in view.

## Tasks involved / Steps to Reproduce

1. Create integration.
2. Add basic filter step
3. Create enough condition sets that you can scroll the page.
4. Note that the top "continue only if..." field scrolls out of view and, on desktop/wide viewports, the condition set form fields are as wide as the viewport.

PR #6396, opened 2019-08-16
fix(basicfilter): rework UI for bordered sections and sticky top condition

mcoker:
fixes https://github.com/syndesisio/syndesis/issues/6336 and https://github.com/syndesisio/syndesis/issues/6222

@seanforyou23  there is an empty label for the trash icon. Should that have text, or can we remove it if it's empty?

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/ruleFilter/WithRuleFilterForm.css
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/ruleFilter/WithRuleFilterForm.css
@@ -1,37 +1,39 @@
-.with-rule-filter-form,
-.with-rule-filter-form .pf-c-page__main-section {
+.with-rule-filter-form {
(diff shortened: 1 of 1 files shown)


gashcrumb:
I'm liking it!  I wonder if we should go a similar route with the conditional flow page as well, that also uses the auto-form array functionality including the up/down arrows.  At the moment I've naturally kinda hosed it up with my pf4 changes, so it's lookin' like:

image

mcoker:
@dongniwang what do you think about the conditional flow page UI in @gashcrumb's comment?

dongniwang:
@mcoker @gashcrumb - conditional flow design update in progress, I can definitely try the border treatment. I would go ahead with what we have for the basic filter right now. The conditional flow page needs some review.

mcoker:
@gashcrumb is there anything else you'd like to see in this PR?

Issue #6399, opened 2019-08-16
DV Wrong image is being referenced in Operator


rareddy:
## This is a...

<pre><code>
[ ] Feature request
[x ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
With the name change from `komodo-server` to "syndesis-dv` the docker reference for the image is still wrong in the Syndesis Operator

PR #6400, opened 2019-08-16
#6399: correcting the docker name reference for the sysdesis-dv

rareddy:


heiko-braun:
please also change the references in fuse-online-install

PR #6403, opened 2019-08-19
fix: [6068] Configuration required warnings for conditional flows steps

christophd:
When an integration uses conditional flows step the warning "Configuration Required" always showed up after save. This was because data shapes in connection descriptors were automatically adjusted for each subflow.

For some reason (I still do not understand the javascript magic) the data shape was adjusted not only in `flow->step->action->descriptor->datashape` but also in `flow->step->action->connection->descriptor->datashape`. The latter connection descriptor data shape should never be changed by a data shape auto adapt. This difference of the connection configuration caused the "Configuration Required" warning to always show up.

The data shape adjustment on the connector descriptor took place in https://github.com/syndesisio/syndesis/blob/master/app/ui-react/packages/api/src/helpers/integrationFunctions.ts#L1478

I tried to fix and then found out that the data shape adjustment at this point is not necessary as the shapes get set later on during step reconcile mechanism. So removing the initial data shape adjustment completely fixed the issue.

### DIFF
--- a/app/connector/flow/src/main/resources/META-INF/syndesis/connector/flow.json
+++ b/app/connector/flow/src/main/resources/META-INF/syndesis/connector/flow.json
@@ -7,9 +7,11 @@
         "componentScheme": "direct",
         "connectorCustomizers": [],
         "inputDataShape": {
+          "name": "No shape",
           "kind": "none"
         },
         "outputDataShape": {
+          "name": "Any shape",
           "kind": "any"
         },
         "propertyDefinitionSteps": [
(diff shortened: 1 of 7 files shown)


gashcrumb:
> For some reason (I still do not understand the javascript magic) the data shape was adjusted not only in flow->step->action->descriptor->datashape but also in flow->step->action->connection->descriptor->datashape

Looking at the code you're removing, I'd say I probably should have used the spread operator to ensure a new object was being set on the descriptor, i.e. something like:

```
    step.action!.descriptor!.outputDataShape =
      { ...thisStep.action.descriptor.inputDataShape };
```

Otherwise `outputDataShape` refers to *the same object* as `inputDataShape`.  The joy of javascript :smile:

PR #6412, opened 2019-08-20
fix: Reintroduce Firefox login fix

### LINKED ISSUE #6282: Can't login with Firefox < 61.0b13
# This is a Bug report/probable regression

<pre><code>
[ ] Feature request
[ X ] Bug report  
</code></pre>

## Description

I'm trying to login to Syndesis using the default OAuth in **Firefox 60.8.0esr** (64-bit) (default version on Debian stable/buster). But a CORS error makes the config.json request to fail:

imagen

This is not reproducible since Firefox v61.0.b13 due to [implementation of the updated spec](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch):

> By default, fetch won't send or receive any cookies from the server, resulting in unauthenticated requests if the site relies on maintaining a user session (to send cookies, the credentials init option must be set).
> Since Aug 25, 2017. The spec changed the default credentials policy to same-origin. Firefox changed since 61.0b13.

I suggest modifying the fetch call to explicitly add the cookies:

imagen

imagen

asmigala:
fixes #6282


zregvart:
Thanks!

PR #6416, opened 2019-08-21
Fixes #3708 Allow the citizen integrator to choose how the integratio…

### LINKED ISSUE #3708: Allow the citizen integrator to choose how the integration is exposed
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description

With #3504 we expose the integration via 3scale if the backend is configured to do so. So every integration that ends up being exposed (API provider, webhook) will be exposed either via 3scale API gateway or via OpenShift Route.

We should allow the user to choose how to expose integrations on a per-integration basis (from this evil comment https://github.com/syndesisio/syndesis/issues/3504#issuecomment-424816704).

rkorytkowski:
…n is exposed

Initial PR to check if build passes and verify approach.

Screenshots:

image

image


app/common/model/src/main/java/io/syndesis/common/model/integration/IntegrationOverview.java
```diff
@@ -60,6 +61,10 @@ default IntegrationBulletinBoard getBoard() {
 
     Optional<String> getUrl();
 
+    Optional<String> getUrlTo3scale();
```
rkorytkowski: Could it actually be exposed via 'config.json'? Does operator modify configmap, if we change an environment variable after installation? Or will one have to modify both the OPENSHIFT_ULR_TO3SCALE environment variable and config map?

zregvart: It could be added to `config.json`, not sure about operator changing the ConfigMap perhaps @lgarciaaco can answer that. I think it would make sense to make this part of the Syndesis custom resource instead of environment variable...

app/common/model/src/main/java/io/syndesis/common/model/integration/IntegrationBase.java
```diff
@@ -92,6 +96,36 @@ default boolean isDeleted() {
...
+    @JsonIgnore
+    default boolean isExposable() {
+        return getFlows().stream().flatMap(f -> f.getSteps().stream())
+            .flatMap(step -> Optionals.asStream(step.getAction()))
+            .flatMap(action -> action.getTags().stream())
+            .anyMatch("expose"::equals);
+    }
+
+    @JsonIgnore
+    default boolean isExposeVia3scale(String urlTo3Scale) {
```
zregvart: Perhaps 

```suggestion
    default boolean isExposableVia3Scale(String urlTo3Scale) {
```

app/common/model/src/main/java/io/syndesis/common/model/integration/IntegrationOverview.java
zregvart: Perhaps

```suggestion
    Optional<String> getManagementUrl();
```

This way it's not tied to 3scale and we can point to the OpenShift console route if we'd like to.

app/common/model/src/main/java/io/syndesis/common/model/integration/IntegrationOverview.java
```diff
@@ -60,6 +61,10 @@ default IntegrationBulletinBoard getBoard() {
...
+
+    Optional<List<String>> getExposureOptions();
```
zregvart: No need for `Optional` with collections

```suggestion
    Set<String> getExposureOptions();
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/IntegrationHandler.java
```diff
@@ -74,11 +75,11 @@
     Creator<Integration>, Deleter<Integration>, Updater<Integration>, Validating<Integration> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(IntegrationHandler.class);
-
     final APIGenerator apiGenerator;
     private final EncryptionComponent encryptionSupport;
     private final Inspectors inspectors;
     private final OpenShiftService openShiftService;
+    private final OpenShiftConfigurationProperties openShiftProperties;
```
zregvart: Perhaps we could store only the 3scale management URL here instead of `OpenShiftConfigurationProperties`

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/IntegrationOverviewHelper.java
```diff
@@ -45,7 +48,8 @@ private IntegrationOverviewHelper() {
 
     @SuppressWarnings("PMD.NPathComplexity")
     static IntegrationOverview toCurrentIntegrationOverview(final Integration integration,
-        final DataManager dataManager) {
+                                                            final DataManager dataManager,
+                                                            final OpenShiftConfigurationProperties openShiftProperties) {
```
zregvart: Passing the whole `OpenShiftConfigurationProperties` seems a bit much, perhaps we could refactor this class to be initialized with `OpenShiftConfigurationProperties` and make methods non-static.

app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftConfigurationProperties.java
```diff
@@ -57,6 +57,8 @@
 
     private int integrationLivenessProbeInitialDelaySeconds;
 
+    private String urlTo3scale;
```
zregvart: Perhaps

```suggestion
    private String _3scaleManagementUrl;
```

app/common/model/src/main/java/io/syndesis/common/model/integration/IntegrationOverview.java
zregvart: Oh I see that it is part of the custom resource, then the operator can definitely update the `config.json`, not sure if it does currently.

app/common/model/src/main/java/io/syndesis/common/model/integration/IntegrationBase.java
```diff
@@ -92,6 +96,36 @@ default boolean isDeleted() {
...
+            String exposure = getExposure().orElse("3scale");
+            return exposure.equals("3scale");
+        } else {
+            return false;
+        }
+    }
+
+    @JsonIgnore
+    default List<String> getExposureOptions(String urlTo3scale) {
+        if (!getUsedConnectorIds().contains("webhook") && StringUtils.isNotBlank(urlTo3scale)) {
```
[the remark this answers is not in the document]
rkorytkowski: The webhook exemption logic was there for the exposeVia3scale setting so I just kept it. I'm not sure why webhook was excluded from 3scale in the first place.

app/common/model/src/main/java/io/syndesis/common/model/integration/IntegrationBase.java
```diff
@@ -92,6 +96,36 @@ default boolean isDeleted() {
...
+        if (getExposureOptions(urlTo3Scale).contains("3scale")) {
+            String exposure = getExposure().orElse("3scale");
+            return exposure.equals("3scale");
+        } else {
+            return false;
+        }
+    }
+
+    @JsonIgnore
+    default List<String> getExposureOptions(String urlTo3scale) {
```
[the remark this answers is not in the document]
rkorytkowski: I was hesitant to put that in the domain model as well. Let's move it out.

app/common/model/src/main/java/io/syndesis/common/model/integration/IntegrationBase.java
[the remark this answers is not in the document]
zregvart: Do you remember @nicolaferraro why we don't want webhooks to be exposed via 3scale?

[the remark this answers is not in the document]
nicolaferraro: I've not followed the 3scale integration discussion, but I think it can be done. Webhook integrations currently don't expose any openapi spec.

app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftConfigurationProperties.java
rkorytkowski: I renamed it to managementUrlFor3scale to avoid the use of '_', which is not liked by go (can only export variables starting with [A-Z]), and keep same names for go and java code.

PR #6422, opened 2019-08-22
add uninstall command

### LINKED ISSUE #6381: [operator] Implement operator uninstall command which deletes the app
This is about deleting the app when a user runs `operator uninstall`. 

https://github.com/syndesisio/fuse-online-install/blob/ba221a3854cbbaccd46e2a7b65e5cd7e83e9ee90/install_ocp.sh#L538-L543

lgarciaaco:
Closes #6381


install/operator/pkg/cmd/internal/uninstall/uninstall.go
```diff
@@ -0,0 +1,87 @@
...
+    for _, res := range sl.Items {
+        err = c.Delete(o.Context, &res)
+        if err != nil {
+            if !errors.IsNotFound(err) {
+                fmt.Println(err, "could not deleted", "custom resource", res.Name, "namespace", res.GetNamespace())
+            }
+        } else {
+            fmt.Println("resource deleted", "custom resource", res.Name, "namespace", res.GetNamespace())
+        }
+    }
```
chirino: This might be a good start.. but I think we also need to render the install resources (to get a list of things we create to get the operator installed) and then delete those too.  So we put back the cluster to how it looked before we did an install.

chirino:
I wonder if we should just set the ownerRef of those additional resources we install to be the syndesis CRD.  So that way if we delete the syndesis CRD, it should auto delete those too..

install/operator/pkg/cmd/internal/uninstall/uninstall.go
lgarciaaco: well that is a new feature ... this only closes https://github.com/syndesisio/syndesis/issues/6381

lgarciaaco:
> I wonder if we should just set the ownerRef of those additional resources we install to be the syndesis CRD. So that way if we delete the syndesis CRD, it should auto delete those too..

Good idea, I rather have the CR as owner

zregvart:
Test fails on CI:
```
--- FAIL: TestUninstall (0.00s)
    uninstall_test.go:49: 	Test: When run without any CRs, `operator uninstall` should not fail
    uninstall_test.go:51: 	✗	 got an error when uninstalling the app: [could not locate a kubeconfig]
FAIL
FAIL	github.com/syndesisio/syndesis/install/operator/pkg/cmd/internal/uninstall	0.026s
```

:cry: :trumpet:

lgarciaaco:
Really a pity, we need a way to run tests against a real cluster

PR #6455, opened 2019-08-23
chore: enable incremental builds again using SNAPSHOT deployments to a git based mvn GitHub mvn repo.  fixes #5632

### LINKED ISSUE #5632: Incremental builds on CircleCI
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description

The builds on CircleCI are getting longer and this prevents the UI folk from switching over to the upstream (syndesis) git repository.

We recently introduced the `--incremental` switch to `syndesis build` and enabled it on CI. This caused [some](https://circleci.com/gh/syndesisio/syndesis/75107) [issues](https://circleci.com/gh/syndesisio/syndesis/75097) that need to be refined before it's enabled again.

This is what I think needs to be done:
 - For builds on `master` the Maven artifacts in `io.syndesis` group installed in the local Maven repository need to be saved in a separate cache not tied to the hash of the `pom.xml`. This will act as a makeshift Maven repository, so when a build of a module that another module depends on is skipped the previously built version of the module (from `master`) will be used.
 - CircleCI configuration needs to be a bit more robust taking into account that some jobs can be skipped and not produce any artifacts which need to be cached/archived.

As an option, it would be good if there were a way to trigger a full build i.e. disable incremental build via some convention or parameter so if an incremental build causes issues a build can be retried as a full build. Not sure how that can be done, need to investigate it.

chirino:


zregvart:
@chirino can you finish this or revert this as it's not working at all?

PR #6479, opened 2019-08-30
feat(connector): convert input/output to json

### LINKED ISSUE #6290: [MongoDB Connector] Datashape conversion
Both producer and consumer are using [Document](https://api.mongodb.com/java/3.2/org/bson/Document.html) as an object. For consumer the conversion from either a `java object` or a `json text` is automatic, so, nothing to care about. Producer though should convert such `java object` to a `json text` that may be feeding any further step in the integration.

squakez:
Decorating the proxy by adding a method to convert any input/output to a json text, usable by other steps

closes #6290

### DIFF
--- a/app/connector/mongodb/src/main/resources/META-INF/syndesis/connector/mongodb3.json
+++ b/app/connector/mongodb/src/main/resources/META-INF/syndesis/connector/mongodb3.json
@@ -10,7 +10,7 @@
       "descriptor": {
         "componentScheme": "mongodb3",
         "inputDataShape": {
-          "kind": "json-schema"
+          "kind": "any"
         },
         "outputDataShape": {
           "kind": "json-schema"
@@ -102,7 +102,7 @@
       "descriptor": {
         "componentScheme": "mongodb3",
         "inputDataShape": {
-          "kind": "json-schema"
+          "kind": "none"
         },
         "outputDataShape": {
           "kind": "json-schema"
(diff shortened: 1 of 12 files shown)


app/connector/mongodb/src/main/resources/META-INF/syndesis/connector/mongodb3.json
```diff
@@ -10,7 +10,7 @@
       "descriptor": {
         "componentScheme": "mongodb3",
         "inputDataShape": {
-          "kind": "json-schema"
+          "kind": "any"
```
zregvart: We're no longer generating the input data shape?

app/connector/mongodb/src/test/java/io/syndesis/connector/mongo/MongoDBConnectorCountTest.java
```diff
@@ -48,9 +48,9 @@ public void mongoCountSingleTest() {
         collection.insertOne(Document.parse(json2));
         // Given
         String countArguments = "{\"test\":\"single\"}";
-        long count = (long) template().sendBody("direct:start", ExchangePattern.InOut, countArguments);
+        Document result = Document.parse(template.requestBody("direct:start", countArguments, String.class));
         // Then
-        assertEquals(1, count);
+        assertEquals(Integer.valueOf(1), result.getInteger("count"));
```
zregvart: Could we use AssertJ assertions? Something like:
```suggestion
        assertThat(result.getInteger("count")).isEqualTo(1);
```

app/connector/mongodb/src/test/java/io/syndesis/connector/mongo/MongoDBConnectorInsertTest.java
```diff
@@ -47,11 +49,16 @@ public void mongoInsertTest() {
         // When
         String uniqueId = UUID.randomUUID().toString();
         String message = String.format("{\"test\":\"unit\",\"uniqueId\":\"%s\"}", uniqueId);
+        Document doc = Document.parse(message);
         // Given
-        template().sendBody("direct:start", message);
+        Document result = Document.parse(template.requestBody("direct:start", message, String.class));
```
zregvart: Might be easier to debug if the result of `template.requestBody(...)` was put in a variable.

PR #6491, opened 2019-09-03
[operator] take operator tag as default tag for syndesis comp

### LINKED ISSUE #6468: [operator] Change source for default tags for syndesis components
Currently, if no tags are specified in the CR, the value taken comes from https://github.com/syndesisio/syndesis/blob/7ed87cab23c5183b0644e1176c0d611bf62c8d4a/install/operator/build/conf/config.yaml#L39

Changes the source of the default value to be the tag of the operator image

lgarciaaco:
Accept a parameter tag, the user can now specify the image and the tag of the operator docker image in two separate tags: 
```
syndesis install --image syndesis/syndesis-operator --tag 1.8.1-20190903
```
If no tags are supplied for meta, s2i, ui and server components,  they then take the operator tag version for it.

As we move away from statics bits in the operator, we have to sets Syndesis.Status.Version to the version of the operator which before was coming from config. Also, fix when the upgrade triggers by comparing Syndesis.Status.Version to the version of the operator.

Fixes #6468


heiko-braun:
Does this fix https://issues.jboss.org/browse/ENTESB-11274 ?

lgarciaaco:
> Does this fix https://issues.jboss.org/browse/ENTESB-11274 ?

yes

avano:
just a note that we are discussing this over chat with Luis

Issue #6093, opened 2019-07-08
Conditional Flow Enhancements


christophd:
See also **https://issues.jboss.org/browse/ENTESB-11494**

This epic collects a set of findings and missing features to bring conditional flows feature from tech preview to GA.

Agreed list of features not in scope:
- Nesting conditional flows
- Building conditions from multiple rules (with and/or)
- Reusing/sharing flows between integrations

heiko-braun:
Migrated to **https://issues.jboss.org/browse/ENTESB-11494**

PR #6495, opened 2019-09-03
fix: hide the configure buttons for the flow start/end connections

gashcrumb:
for #6093

Was on my list for the last PR and totally forgot to do it.  This PR makes this happen for these two connections in the editor:

image

and

image

@christophd FYI


gashcrumb:
Fixed my fat-fingered commit message.

Issue #6094, opened 2019-07-08
[Conditional Flow] add basic language expression builder


christophd:
At the moment the user has to provide the complete simple language expression in order to define the condition used for evaluation, e.g. `${body.company} contains 'Red Hat'`(like in advanced filter).

This is the most powerful way but requires the user to know the expression syntax. 

We should add a convenient way to build simple language expressions from available message body properties (just like basic filter is doing).

PR #6496, opened 2019-09-03
feat [6094]: Add basic condition expression builder to conditional flows step

christophd:
Relates to #6094 

Adds rule based form controls to condition expressions in conditional flows step. At the moment just adds the controls next to the advanced expression input field. The user can either provide an advanced expression or use the rule based form components to build the condition expression.

We will need to add some advanced form validation so that user needs to fill in either advanced or rule based form controls.

Backend has already been prepared to handle both advanced and rule based configuration so conditions are evaluated correctly at runtime.

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/choice/ChoiceStepPage.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/choice/ChoiceStepPage.tsx
@@ -1,7 +1,8 @@
 import {
   createConditionalFlow,
   FlowKind,
-  getFlow, getPreviousIntegrationStepWithDataShape,
+  getFlow,
+  getOutputDataShapeFromPreviousStep,
   getStep,
   getSteps,
   reconcileConditionalFlows,
@@ -10,7 +11,7 @@ import {
   WithIntegrationHelpers,
 } from '@syndesis/api';
 import * as H from '@syndesis/history';
-import {DataShape, Integration, StringMap} from '@syndesis/models';
+import { Integration, StringMap } from '@syndesis/models';
 import {
   ChoiceCardHeader,
   ChoicePageCard,
@@ -60,19 +61,7 @@ export class ChoiceStepPage extends React.Component<IChoiceStepPageProps> {
               <WithRouteData<IChoiceStepRouteParams, IChoiceStepRouteState>>
                 {(params, state, { history }) => {
                   const positionAsNumber = parseInt(params.position, 10);
-                  let dataShape = {} as DataShape;
-                  try {
-                    const prevStep = getPreviousIntegrationStepWithDataShape(
-                      state.integration,
-                      params.flowId,
-                      positionAsNumber
-                    );
-                    dataShape =
-                      prevStep!.action!.descriptor!.outputDataShape ||
-                      ({} as DataShape);
-                  } catch (err) {
-                    // ignore
-                  }
+                  const dataShape = getOutputDataShapeFromPreviousStep(state.integration, params.flowId, positionAsNumber);
                   const step = state.step;
                   // parse the configured properties
                   const configuration = createChoiceConfiguration(
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/choice/ChoiceStepPage.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/choice/ChoiceStepPage.tsx
@@ -1,15 +1,16 @@
 import {
   createConditionalFlow,
   FlowKind,
-  getFlow,
+  getFlow, getPreviousIntegrationStepWithDataShape,
   getStep,
   getSteps,
   reconcileConditionalFlows,
   WithConnection,
+  WithFilterOptions,
   WithIntegrationHelpers,
 } from '@syndesis/api';
 import * as H from '@syndesis/history';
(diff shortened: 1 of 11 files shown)


app/ui-react/syndesis/src/modules/integrations/components/editor/choice/ChoiceStepPage.tsx
```diff
@@ -1,15 +1,16 @@
 import {
   createConditionalFlow,
   FlowKind,
-  getFlow,
+  getFlow, getPreviousIntegrationStepWithDataShape,
```
gashcrumb: `getPreviousIntegrationStepWithDataShape` should go on the next line.  Ideally you'd want to install whatever extension adds `prettier` support to your IDE, then it should pick up the rules and automatically apply them.  I'd say run `yarn format` and commit the result but testing locally I see that there's a few other unrelated files that haven't been formatted using `prettier` at the moment.

app/ui-react/syndesis/src/modules/integrations/components/editor/choice/ChoiceStepPage.tsx
```diff
@@ -59,6 +60,19 @@ export class ChoiceStepPage extends React.Component<IChoiceStepPageProps> {
...
+                      state.integration,
+                      params.flowId,
+                      positionAsNumber
+                    );
+                    dataShape =
+                      prevStep!.action!.descriptor!.outputDataShape ||
+                      ({} as DataShape);
+                  } catch (err) {
+                    // ignore
+                  }
```
gashcrumb: Would it make more sense to maybe extract this block of logic out into a separate function in `utils.ts`?

app/ui-react/syndesis/src/modules/integrations/locales/integrations-translations.en.json
```diff
@@ -91,15 +91,21 @@
...
-      "conditionDescription": "Provide a condition that you want to evaluate (for example, ${in.header.type} == 'note' or ${in.body.title} contains 'Important').",
+      "conditionDescription": "Provide a condition that you want to evaluate (for example, ${header.type} == 'note' or ${body.title} contains 'Important').",
       "conditionName": "Condition",
-      "conditionPlaceholder": "Simple language expression",
+      "conditionPlaceholder": "Condition expression",
       "addCondition": "Add another condition",
       "addConditionTitle": "When",
       "forAllIncomingData": "For all incoming data",
-      "otherwise": "If incoming data does not match all of the above conditions",
+      "otherwise": "When none of the above conditions matches to the incoming data",
```
gashcrumb: Typo, I think you want `When none of the above conditions match the incoming data`

app/ui-react/syndesis/src/modules/integrations/components/editor/choice/utils.ts
```diff
@@ -20,3 +20,35 @@ export function createChoiceConfiguration(configuredProperties: {
...
+ * Builds a sane flow form option description from either condition expression or path, op, value properties
+ * @param option
+ */
+export function getFlowDescription(option: IFlowFormOption) {
+  if (option.condition) {
+    return option.condition;
+  }
+
+  if (option.path) {
+    return '${body.' + option.path + '} ' + option.op + " '" + option.value + "'";
```
gashcrumb: you could do this via an interpolated string with backticks which should make this more readable, I think it'd be like:

```
return `\${body.${option.path}} ${option.op} '${option.value}'`
```

Just need to escape the first `$` so it's printed in the resulting string.

app/ui-react/syndesis/src/modules/integrations/components/editor/choice/utils.ts
```diff
@@ -20,3 +20,35 @@ export function createChoiceConfiguration(configuredProperties: {
...
+ * Builds a sane flow option condition expression from either given condition expression or path, op, value properties
+ * @param option
+ */
+export function getConditionExpression(option: IFlowOption) {
+  if (option.condition) {
+    return option.condition;
+  }
+
+  if (option.path) {
+    return '${body.' + option.path + '} ' + option.op + " '" + option.value + "'";
```
gashcrumb: Same comment as above, I think an interpolated string might be more clear and less quote-y.  Also if this function and the last function do the same thing, why have two of them?

gashcrumb:
Looking good!  I added some things, also it'd be cool to kinda keep [the storybook I created](https://github.com/syndesisio/syndesis/blob/master/app/ui-react/packages/auto-form/stories/SyndesisUIs.stories.tsx#L280) for this up to date a bit, makes it easier to kinda design the form.  You just need to cd into `packages/auto-form` and run `yarn storybook` to see it in action.

app/ui-react/syndesis/src/modules/integrations/components/editor/choice/ChoiceStepPage.tsx
christophd: Thanks, I have now installed some IntelliJ plugin - lets see how that goes

app/ui-react/syndesis/src/modules/integrations/locales/integrations-translations.en.json
christophd: 👍 fixed

PR #6512, opened 2019-09-04
ENTESB-11263: Add scheduled to CR for meta, server, ui and s2i

lgarciaaco:
For ImageStreams, tweak `importPolicy.scheduled` by setting a parameter in the CR

For example:
```
kind: Syndesis
metadata:
  name: app
spec:
  components:
    meta:
      scheduled: true
```

Will end up in IS syndesis-meta having
```
      importPolicy:
        scheduled: true
```
or false.


heiko-braun:
What if we make it global settings that applies to all ImageStreams? I think you might want to enable for all or none.

lgarciaaco:
One downside of this approach is that the default value of a boolean is false, therefore if `Syndesis.Spec.Components.Scheduled` is omitted in the CR, `importPolicy.scheduled` will go to false which is not what we want by default.

heiko-braun:
> importPolicy.scheduled will go to false which is not what we want by default.

I don't know, for product builds that seems to be right, for development probably not

PR #6521, opened 2019-09-05
Get default value for syndesis components from conf file

lgarciaaco:
This PR partially reverts https://github.com/syndesisio/syndesis/pull/6491

- It is still possible to define the operator image tag by using `--tag tagValue`, but this will only set the tag for the operator, as well as the tag name.
- The default value for images comes from file. What this value should be, is yet to be defined.
- e2e test are working again


lgarciaaco:
Related with https://github.com/syndesisio/syndesis/pull/6491

PR #6524, opened 2019-09-05
chore: fix JDBC URL typo

zregvart:


zregvart:
@TovaCohen I think this will get pulled into product documentation at 

https://access.redhat.com/documentation/en-us/red_hat_fuse/7.4/html-single/connecting_fuse_online_to_applications_and_services/index#create-database-connection_db

https://access.redhat.com/documentation/en-us/red_hat_fuse/7.3/html-single/connecting_fuse_online_to_applications_and_services/index#create-database-connection_db

https://access.redhat.com/documentation/en-us/red_hat_fuse/7.2/html-single/connecting_fuse_online_to_applications_and_services/index#create-database-connection_db

Do we need to backport it to more branches?

TovaCohen:
@zregvart  Thanks very much for alerting me about this, and for fixing it.
Looks like I was very consistent, and the doc for every release has the same error.
I'll make sure this gets updated in the published downstream doc on the customer portal. At the end of every sprint, we republish any documents that have updates. So this should happen on Monday. 
You decide if you want to backport it to all the earlier upstream releases.

Issue #5456, opened 2019-05-23
[Conditional-Flow] User defined output data shape


christophd:
Conditional flows step should be handled as "shapeless connection" so the user is prompted to provide a shape when the step is added to the flow.

When the user provides an output shape for the conditional flow step all flows added to this step must produce this very same output.

In case a subsequent step already exists, the UI should offers also an option to adopt the input shape of that subsequent step.

When a shape is given all conditional flows have this as fixed finish step shape. This ensures that all conditional flows produce the same output and we can continue with the integration flow after the conditional flows step accessing this output.

gashcrumb:
Sorry I had the above comment open but not posted for like an hour :-)

PR #6528, opened 2019-09-05
feat [5456]: Add user defined output data shape on conditional flows

christophd:
Relates tp #5456 

This PR adds the describe data shape page to the conditional flows configuration. This way the user can define an output data shape for the step. All subflows need to produce this data shape so the primary flow can continue to work with this shape as outcome of conditional flows step.

Up to now we only had the describe data shape page for connection actions and not for steps. I decided to add this functionality to the conditional flows step only. Other steps might want to use in future but for now I felt more safe to only add it to the CF step.

Also I needed to reconcile the integration flows setting the data shape to the flow end steps which is kind of unique to the conditional flows data shape describe page.


app/ui-react/packages/api/src/helpers/integrationFunctions.ts
```diff
@@ -686,25 +686,35 @@ export async function setFlow(
...
+    const startFlowStep = setDataShapeOnStep(
       { ...flow.steps![0] },
-      updatedDataShape,
+      flowStartDataShape,
       false
     );
-    return applyUpdatedStep(flow, newStep, 0);
+    const updatedFlow = applyUpdatedStep(flow, startFlowStep, 0);
+
+    const endStepPosition = flow.steps!.length-1;
```
gashcrumb: Smallest nit ever, use [one of these](https://github.com/syndesisio/syndesis/blob/master/app/ui-react/packages/api/src/helpers/integrationFunctions.ts#L946-L966).  I've always thought that at any point we could decide to stop using an array, so I've thought it'd be good to kinda hide that implementation detail.

app/ui-react/syndesis/src/modules/integrations/components/editor/choice/DescribeChoiceDataShapePage.tsx
```diff
@@ -0,0 +1,193 @@
...
+                        ...state,
+                        step: stepWithUpdatedDescriptor,
+                        updatedIntegration: reconciledIntegration,
+                      } as IChoiceStepRouteState
+                    )
+                  );
+                };
+                return (
+                  <>
+                    <PageTitle title={'Specify Conditional Flows Data Type'} />
```
gashcrumb: i18n in here would be good if you've time.

gashcrumb:
Great work!  I only spotted kinda nitpicky things :smile:

christophd:
thanks! I tried to add i18n but hit some problems so I will come back to this in a separate PR when there is more time

PR #6538, opened 2019-09-06
feat: add a backup and restore sub command which can eventually be used in the upgrade process.

chirino:


chirino:
@lgarciaaco we could also do that eventually.  BTW I was envisioning on using this by running `syndesis backup` as a job eventually, kicking it off on a schedule  would be good too.

PR #6543, opened 2019-09-06
Fix for syndesis upgrade

lgarciaaco:
- Remove steps related with the template
- Fix some lint issues
- Fixes ENTESB-11576
- Fixes ENTESB-11321
- Fixes ENTESB-11320


lgarciaaco:
Putting this up in case someone has a better idea on how to patch the s2i image stream tag from configmap.

zregvart:
The gist is to provide the S2I image tag via env variable? Looks like a good approach to me.

lgarciaaco:
> The gist is to provide the S2I image tag via env variable? Looks like a good approach to me.

yes that's the idea

lgarciaaco:
The code here is not needed, I don't think this 
 was actually needed anyways. After this PR is merged the syndesis version is not updated from the pod. The assumption is that if the upgrade pod finishes successfully, then the upgrade itself was successfully and we can directly proceed to change the syndesis version in the CR 

https://github.com/syndesisio/syndesis/pull/6543/files#diff-269c3070ecfab4ab96e7fba6794ebf14L109-L126

lgarciaaco:
I am completely removing tools/upgrade/migration/resource/01_server_config_map_upgrade.sh because the new operator will update config maps with the new content.

lgarciaaco:
I dont fully understand what this snippet is doing: https://github.com/syndesisio/syndesis/pull/6543/files#diff-269c3070ecfab4ab96e7fba6794ebf14L152-L156

But it won't work because it attempts to load the old template which is not there ... I am dropping it.

zregvart:
> I dont fully understand what this snippet is doing: https://github.com/syndesisio/syndesis/pull/6543/files#diff-269c3070ecfab4ab96e7fba6794ebf14L152-L156

I guess it needs to set the owner back reference to the Syndesis CR so that garbage collection works correctly, might be already a part of the object creation when operator is creating objects.

lgarciaaco:
> > I dont fully understand what this snippet is doing: https://github.com/syndesisio/syndesis/pull/6543/files#diff-269c3070ecfab4ab96e7fba6794ebf14L152-L156
> 
> I guess it needs to set the owner back reference to the Syndesis CR so that garbage collection works correctly, might be already a part of the object creation when operator is creating objects.

Is it needed? The only two objects that don't have the reference are the upgrade pod and the volume for the pod, and both are deleted when the first reconciliation kicks in

lgarciaaco:
@avano merged so @zregvart can do his branching with this included. We continue over pm testing this further

PR #6550, opened 2019-09-09
TTOOLS-748 Added "Loading query results" spinner to preview results panel

blafond:


app/ui-react/syndesis/src/modules/data/pages/viewEditor/ViewEditorOutputPage.tsx
```diff
@@ -213,9 +213,11 @@ export const ViewEditorOutputPage: React.FunctionComponent = () => {
                 'data:virtualization.preview.resultsTableEmptyStateInfo'
               )}
               i18nHidePreview={t('data:virtualization.preview.hidePreview')}
+              i18nLoadingQueryResults={t('data.virtualization.preview.loadingQueryResults')}
```
mdrillin: This i18n needs to have a colon - 'data:' - same as the ViewEditorSqlPage

app/ui-react/syndesis/src/modules/data/pages/viewEditor/ViewEditorSqlPage.tsx
```diff
@@ -201,6 +203,7 @@ export const ViewEditorSqlPage: React.FunctionComponent = () => {
           0
         );
       }
+      setIsLoadingPreview(false);
```
mdrillin: Add a "setIsLoadingPreview(false)" in the catch block, so the spinner will dismiss when there is an error condition

blafond: Done

PR #6562, opened 2019-09-10
ENTESB-10826, ENTESB-11252 - DV Wizard layout fixes

mdrillin:
Modifications to the DV wizard layouts to make consistent with other syndesis wizards
- ViewCreateLayout and ViewsImportLayout - using layout css consistent with other syndesis wizards.  Most of the previous css could be eliminated.


app/ui-react/packages/ui/src/Data/Virtualizations/Views/ViewCreateLayout.tsx
```diff
@@ -69,59 +69,58 @@ export const ViewCreateLayout: React.FunctionComponent<
...
+                      data-testid={'view-create-layout-next-button'}
+                      onClick={onNext}
+                      href={nextHref}
+                      as={'primary'}
+                      className={'wizard-pf-next'}
+                      disabled={isNextLoading || isNextDisabled}
+                    >
+                      {isNextLoading ? <Loader size={'xs'} inline={true} /> : null}
+                      {isLastStep ? (
+                        'Done'
```
elvisisking: This file and `ViewsImportLayout` should be i18n'd. Maybe open up another issue for that.

Issue #4156, opened 2018-11-23
Gracefuly degrading smarter and super awesome syndesis operator


zregvart:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
With #3576 we introduced `OAuthClient` resource creation in the operator, subsequently we converted [install/operator/deploy/syndesis-operator.yml](https://github.com/syndesisio/syndesis/blob/53b53c9a83e6e26193c95df475e5bf07fa53a096/install/operator/deploy/syndesis-operator.yml) to deal with the creation of `ClusterRole` and `ClusterRoleBinding` needed to grant the permissions to create and delete `OAuthClient` resource.

We also introduced the [`syndesis-privileged.yml`](https://github.com/syndesisio/syndesis/blob/53b53c9a83e6e26193c95df475e5bf07fa53a096/install/syndesis-privileged.yml) template that utilizes the `OAuthClient` instead of the service account for OAuth configuration of `oauth-proxy`.

Now when we install we have a `ClusterRole` and a `ClusterRoleBinding` created per syndesis installation, which is superfluous and leads to issues like #4133, #4131, #4093 and #4126.

This makes us have a hard requirement on the installation of `syndesis-operator` to have `ClusterRole` and `ClusterRoleBinding` created for the operator to function.

We should handle this situation a bit more gracefully by checking in the operator if it can create the `OAuthClient` resource and select a template (`syndesis.yml` or `syndesis-privileged.yml`) based on the privileges given to the operator.

These should be our goals:
 1. installation without any privileges via:
    - oauth-proxy service account ([install/support/serviceaccount-as-oauthclient-restricted.yml](https://github.com/syndesisio/syndesis/blob/master/install/support/serviceaccount-as-oauthclient-restricted.yml))
    - [install/sydesis.yml](https://github.com/syndesisio/syndesis/blob/master/install/syndesis.yml) template
 2. installation via the operator with permissions to create custom resource
    - via OLM
    - via descriptors [install/operator/deploy/syndesis-operator.yml](https://github.com/syndesisio/syndesis/blob/master/install/operator/deploy/syndesis-operator.yml), [install/operator/deploy/syndesis-crd.yml](https://github.com/syndesisio/syndesis/blob/master/install/operator/deploy/syndesis-crd.yml)
[...]

PR #6587, opened 2019-09-13
remove generator folder, not needed anymore and can lead to confusion

lgarciaaco:
With https://github.com/syndesisio/syndesis/pull/6543 merged, this is officially not needed anymore


zregvart:
We're still using other bits? Like [install/syndesis.yml](https://github.com/syndesisio/syndesis/blob/master/install/syndesis.yml) that is the result of the generator scripts?

Would be also cool to remove the pre-commit bits that check that the template was re-created:

https://github.com/syndesisio/syndesis/blob/7d1d8dc7f2851f8b8481c0fcd57e22dcccc0d288/.pre-commit-config.yaml#L14-L18

(and [install/support/validate-generated-templates.sh](https://github.com/syndesisio/syndesis/blob/master/install/support/validate-generated-templates.sh))

lgarciaaco:
> We're still using other bits? Like [install/syndesis.yml](https://github.com/syndesisio/syndesis/blob/master/install/syndesis.yml) that is the result of the generator scripts?
> 
> Would be also cool to remove the pre-commit bits that check that the template was re-created:
> 
> https://github.com/syndesisio/syndesis/blob/7d1d8dc7f2851f8b8481c0fcd57e22dcccc0d288/.pre-commit-config.yaml#L14-L18
> 
> (and [install/support/validate-generated-templates.sh](https://github.com/syndesisio/syndesis/blob/master/install/support/validate-generated-templates.sh))

Where are we using install.yml? I think last references to this file was for the upgrade pod. Now that I think of it, maybe there are some build script that will try to copy the file when building the upgrade docker, that should be removed also

KurtStam:
@lgarciaaco  So if we delete these, how can we install to OpenShift if you don't have cluster admin rights? See the instructions here: https://github.com/syndesisio/syndesis/wiki/Deploying-to-a-Remote-OpenShift-from-the-Syndesis-Template

zregvart:
@lgarciaaco to address @KurtStam's question: I think you added the `eject` option to the operator when used as a CLI, so you can now do something like `syndesis eject | oc apply` or thereabouts, right?

lgarciaaco:
> @lgarciaaco So if we delete these, how can we install to OpenShift if you don't have cluster admin rights? See the instructions here: https://github.com/syndesisio/syndesis/wiki/Deploying-to-a-Remote-OpenShift-from-the-Syndesis-Template. This is used pretty widely at the moment.

Although there is a replacement for template generation added by https://github.com/syndesisio/syndesis/pull/6569, my understanding is that we won't support template installation any longer. We might need template support for 7.5 but only for the eval clusters, and only if we can't find an alternate solution.

The content I am deleting is obsolete and not used by the operator in its current form. The templates generated by the generator most likely ... wont work

@heiko-braun can correct me if I am wrong

lgarciaaco:
> @lgarciaaco to address @KurtStam's question: I think you added the `eject` option to the operator when used as a CLI, so you can now do something like `syndesis eject | oc apply` or thereabouts, right?

Right ... but this is not what we want to support as far as I can say. Operator install and in the future, installations via OLM, are the preferable ways.

IIRC in the f2f one of the conclusions was, that the case were the user doesn't have admin rights, is a corner case that we should not support ... but don't quote me on that

heiko-braun:
> IIRC in the f2f one of the conclusions was, that the case were the user doesn't have admin rights, is a corner case that we should not support ... but don't quote me on that

Right, that was my understanding as well. And, even if we would support templates, you do need admin rights anyway, no? I think certaion settings have to applied by an admin user (i.e. oauth proxy setup)

zregvart:
> Right, that was my understanding as well. And, even if we would support templates, you do need admin rights anyway, no? I think certain settings have to applied by an admin user (i.e. oauth proxy setup)

That was not a requirement thus far. If installed from templates the only bit that would currently fail are additional permissions needed to run Camel-K Integrations.

zregvart:
What's important to note is that if we don't have a way to install without admin rights that makes it difficult for someone to evaluate Syndesis on their own if they're given (only) access to a OpenShfit project, so without any cluster admin rights.

It also makes it difficult for us to maintain the staging environment where we don't have (currently) the ability to use custom resources.

Thus far there was a considered effort to avail Syndesis installation to environments with constrained permissions. In fact it was used as an argument by @heiko-braun to revert #3576.

We also have some ideas in #4156 on how operator could be installed without OLM/CR support and when dealing with lack of permissions.

I think that `syndesis eject` approach could lead us to a solution that satisfies both use cases.

lgarciaaco:
@zregvart this is not about installing with admin right vs not admin rights. This is about deprecation of the installation via templates. There are ways to install syndesis without admin rights, using the operator (an admin needs to create the CRD and grant the user the permissions, and off you go to install the operator). The operator binary supports granting permissions to a user. Basically and admin would need to run:
```
syndesis install cluster
syndesis grant --user lgarciaac
```
And the rest the user is left with one command to run:
```
syndesis install operator
```

If @KurtStam worry is about how to install syndesis without admin rights, because the current documentation is pointing out community to use templates, then the solution is to point the community to use operator base installation, and give them instructions about what to do in case they aren't admin.

More over, with OLM, all the steps where an admin account would be required, the framework just does it for you, no admin needed anylonger. 

This is about transitioning to an operator install mode, operator only.

zregvart:
@lgarciaaco the counterpoint to that is that we're still unable to use this workflow on our own staging environment. Meaning, if the user can't do everything by themselves then the getting started experience is broken.

I think that requiring the admin to grant permissions to the user is in the end the same as requiring admin rights, at least in the sense of what's needed to install.

I don't have any objections to this getting merged, I'd just like to voice the concern that we shouldn't generate obstructions for anyone trying to install/evaluate Syndesis.

lgarciaaco:
> @lgarciaaco the counterpoint to that is that we're still unable to use this workflow on our own staging environment. Meaning, if the user can't do everything by themselves then the getting started experience is broken.
> 
> I think that requiring the admin to grant permissions to the user is in the end the same as requiring admin rights, at least in the sense of what's needed to install.
> 
> I don't have any objections to this getting merged, I'd just like to voice the concern that we shouldn't generate obstructions for anyone trying to install/evaluate Syndesis.

Ok. But shouldn't the preferable way of installing syndesis be, to install it in latest stable version of ocp or openshift, therefore just using the embedded  operatorhub catalog and installing using the workflows provided by OLM? Because we are talking about a 3 clicks process if installed using OLM, and no need of any special permissions.

I think yo guys don't want to let die what is already dead ...

KurtStam:
@lgarciaaco please update the instructions found here: https://github.com/syndesisio/syndesis/wiki/Deploying-to-a-Remote-OpenShift-from-the-Syndesis-Template with your proposed solution so that we can install to https://api.rh-idev.openshift.com 

That's all I'm asking for. Please...

heiko-braun:
@lgarciaaco Looks like this broke the daily builds: https://ci.fabric8.io/job/syndesis-release-daily/586/

KurtStam:
@lgarciaaco  You realize with this merge and no workaround you are preventing folks from working on the codebase right?

zregvart:
It's OK that the stuff breaks, it needs to so we can know what breaks. Let's work with @lgarciaaco to get stuff that we need working.

zregvart:
I think daily releases should be fixed with #6618.

PR #6589, opened 2019-09-13
fix wrong library in eject command

lgarciaaco:
Wrong yaml library was messing up the way unstructured objects were printed


asmigala:
When I pulled the changes and built the operator locally, I still get the same issues:

* `host: todo-<no value>`
* ` allowedOrigins: https://<no value>`
* no `syndesis-oauth` service account

lgarciaaco:
> When I pulled the changes and built the operator locally, I still get the same issues:
> 
> * `host: todo-<no value>`
> * ` allowedOrigins: https://<no value>`
> * no `syndesis-oauth` service account

@asmigala I need more input to reproduce

asmigala:
> @asmigala I need more input to reproduce

@lgarciaaco 
```
./build.sh --image-build=skip
dist/linux-amd64/operator eject -n syndesis --operator-config build/conf/config.yaml | grep 'no value'
```

lgarciaaco:
> > @asmigala I need more input to reproduce
> 
> @lgarciaaco
> 
> ```
> ./build.sh --image-build=skip
> dist/linux-amd64/operator eject -n syndesis --operator-config build/conf/config.yaml | grep 'no value'
> ```

@asmigala noted. https://github.com/syndesisio/syndesis/issues/6591

heiko-braun:
@lgarciaaco Does eject create an openshift template?

lgarciaaco:
> @lgarciaaco Does eject create an openshift template?

No. Only the openshift resources

heiko-braun:
@lgarciaaco I see, does `--operator-config build/conf/config.yaml` provide the parameters to the resources?

heiko-braun:
Thanks for your review @avano !

lgarciaaco:
> @lgarciaaco I see, does `--operator-config build/conf/config.yaml` provide the parameters to the resources?

To some of them, it provides 'sane' defaults, for other, they have to be specified via the CR. This is the case for what @asmigala pointed out:

https://github.com/syndesisio/syndesis/blob/53a661087fec1108523297c04130a7b03f7f8444/install/operator/pkg/syndesis/action/install.go#L93-L96

Now, I am not sure if this is right. Previously it was provided as a parameter, and it was the only mandatory parameter for the openshift template:
https://github.com/syndesisio/syndesis/blob/master/install/README.md#required-input-parameters

heiko-braun:
Does the CR support everything that previously existed as `Parameters with default values` in https://github.com/syndesisio/syndesis/blob/master/install/README.md#required-input-parameters?

lgarciaaco:
> Does the CR support everything that previously existed as `Parameters with default values` in https://github.com/syndesisio/syndesis/blob/master/install/README.md#required-input-parameters?

I would say no, I didnt do it with that in mind

heiko-braun:
>  I didn't do it with that in mind

That's fine, we can add those when needed.

PR #6590, opened 2019-09-13
chore(cci): Tag PR images with PR-SHA combination

avano:


heiko-braun:
@avano What's the purpose of adding the sha? When I think of the consumer end it seems more complicated to update (or refer to images) this way:

`syndesis/${image}:${CIRCLE_PR_NUMBER}-${CIRCLE_SHA1}`

.circleci/config.yml
```diff
@@ -130,8 +130,8 @@ push_images: &push_images
         done
       elif [ -n "${CIRCLE_PR_NUMBER}" ]; then
         for image in ${IMAGES_TO_PUSH} ; do
-          docker tag syndesis/${image}:latest syndesis/${image}:${CIRCLE_PR_NUMBER}
-          docker save syndesis/${image}:${CIRCLE_PR_NUMBER} | gzip > ${CIRCLE_WORKING_DIRECTORY}/${image}.tar.gz
+          docker tag syndesis/${image}:latest syndesis/${image}:${CIRCLE_PR_NUMBER}-${CIRCLE_SHA1}
+          docker save syndesis/${image}:${CIRCLE_PR_NUMBER}-${CIRCLE_SHA1} | gzip > ${CIRCLE_WORKING_DIRECTORY}/${image}.tar.gz
```
heiko-braun: I don't know, this requires clients that want to pull the image to know the Circle CI SHA in the first place. That's seems a lot to ask for.

avano: @heiko-braun is there any other "client" than QE smoke tests? We need to differentiate between the rebuilds of one PR (with code changes)

heiko-braun: I think there might be other use too, yes. Like staging a PR on a preview environment. Is there a different to carry to this information, that doesn't rely on the tag?

heiko-braun: What if we introduce a moving tag in addition to the one you suggest here? So that each build creates both `syndesis/${image}:${CIRCLE_PR_NUMBER}` and `syndesis/${image}:${CIRCLE_PR_NUMBER}-${CIRCLE_SHA1}`. The former would be the moving tag for the PR.

avano: yes, that will definitely work as well

PR #6595, opened 2019-09-13
feat(connector): mongoDB cluster support

squakez:
Enabling the configuration to provide a list of nodes and a replica set name to connect to

closes ENTESB-11590


squakez:
> What does replicaset do? I am not familiar with that property.

It's a way to identify the cluster: https://docs.mongodb.com/manual/replication/

PR #6616, opened 2019-09-17
Add ability to specify custom maven repositories

nicolaferraro:
Fixes ENTESB-11264

Custom repositories can be now specified as in:
```
kind: Syndesis
metadata:
  name: app
spec:
  mavenRepositories:
    another: https://yyy.xxx/v2/
    myrepo: http://xxx.yyy/path
  # ...
```

They are picked up by both classic and camel k runtimes.


install/operator/pkg/generator/assets/infrastructure/03-syndesis-server-config.yml.tmpl
```diff
@@ -64,17 +64,25 @@
...
+      maven:
+        repositories:
+    {{- range $id, $url := .Syndesis.Spec.MavenRepositories}}
+          {{ $id }}: {{ $url }}
+    {{- end }}
+{{- else}}
+    {{- if .Productized}}
       maven:
         repositories:
           central: https://repo.maven.apache.org/maven2/
```
lgarciaaco: Why don't we rather set a default value for `.Syndesis.Spec.MavenRepositories` so we dont have to ask every time whether is has a value or not?

nicolaferraro: That's a better solution. Iirc default values handling is still work in progress, isn't it?

lgarciaaco: We have this section starting here, it seems the right place to do it: https://github.com/syndesisio/syndesis/blob/9fab4e22937b38a7e18bbadbc684aa1c04488689/install/operator/pkg/syndesis/template/install.go#L146-L151

zregvart:
@lgarciaaco @nicolaferraro do we need to do something else here other than a rebase?

nicolaferraro:
@zregvart I'm working on the fix suggested by @lgarciaaco.

nicolaferraro:
@lgarciaaco can you have another look now?

nicolaferraro:
Any hint on how to unlock the "syndesis" check?

zregvart:
> Any hint on how to unlock the "syndesis" check?

Seems that CircleCI has made some changes to the checks, we're no longer getting the cumulative checks as `Syndesis` but each job as separate check, I've disabled mandatory `Syndesis` check for `master`...

PR #6617, opened 2019-09-17
fix(atlasmap-assembly): remove routing related console warning

riccardo-forina:
Since Atlasmap doesn't use the routing, we can avoid enabling it. This
gets rid of the warning we were having about the router not being happy
about being initialized in an iframe without an src attribute.

Also, fixes the minishift setup/restore script to work with the
operator. Since the operator reverts back any local change to operator's
managed resources, we have to scale it down to 0 when running the setup
script. The restore script brings the operator back to 1 replica.

Related to ENTESB-11628

### DIFF
--- a/app/ui-react/syndesis/scripts/minishift-restore.sh
+++ b/app/ui-react/syndesis/scripts/minishift-restore.sh
@@ -83,3 +83,6 @@ oc replace --force -f - <<EOF
   ]
 }
 EOF
+
+# also scale back up the operator
+oc scale --replicas=1 dc syndesis-operator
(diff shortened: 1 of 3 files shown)


app/ui-react/syndesis/scripts/minishift-restore.sh
```diff
@@ -83,3 +83,6 @@ oc replace --force -f - <<EOF
   ]
 }
 EOF
+
+# also scale back up the operator
+oc scale --replicas=1 dc syndesis-operator
```
riccardo-forina: @zregvart is that `dc` bit there needed? @igarashitm made me notice it, maybe it's not needed?

PR #6624, opened 2019-09-17
feat: add a reifiedSpec field so that we don’t update the user’s original spec declaration.

chirino:


zregvart:
Feels bit too complex. Wasn't the original idea to bake the tag into the operator and then we can either have the tag `latest` in the CR to denote that the user wants to always upgrade to the latest or a specific tag that would denote that the user would like to stick to a particular version.

The discussion is [still ongoing](https://issues.jboss.org/browse/ENTESB-11366) from what I understand.

Also, no tests :cry:

lgarciaaco:
Yea I really don't think this should be the solution. I rather stop writing default values to the CR at all. This adds another layer of complexity which I don't understand what is there for, what are the values in `ReifiedSpec` used for? Is it to increase visibility to the user? so the users know what is set by default? If this is the case then `syndesis.Status` might be a better placeholder for it.

Now I am going to take this PR to continue with the test topic, which is something everyone is nt doing but ... we should stop accepting PRs in the operator section that don't include at least unit tests, and some sort of integration test (which we might not be able to run in CI but at least we can run locally with `-short`) ... ideally also some e2e tests. By now we have examples of how to do this. Thoughts? I don't want to be the only one -1ing because of no test :-)

install/operator/pkg/syndesis/action/install.go
```diff
@@ -170,22 +151,37 @@ func (a *installAction) Execute(ctx context.Context, syndesis *v1alpha1.Syndesis
 			return err
 		}
 	}
+	addRouteAnnotation(original, syndesisRoute)
+
+	// The above section of code update the syndesis.Spec, lets not modify the CR
+	// lets copy it to the ReifiedSpec field so users can see how it was reified but not change the original
+	// spec on the CR
+	syndesis.ReifiedSpec = syndesis.Spec.DeepCopy()
+	original.ReifiedSpec = syndesis.Spec.DeepCopy()
```
phantomjinx: Once we have copied it, do we actually do anything with it? If it is purely for user analysis, could it not be dumped to a log file rather than sitting in the CR?

lgarciaaco:
@chirino what about we create an `upgrade` command which kicks off the upgrade process. It would look like:

- scale operator dc to 0.
- modify CR (and here you could set the tags to empty to solve your issue), update CRD, update whatever is needed
- change operator imagestream to target version
- scale operator to 1

install/operator/pkg/syndesis/action/install.go
chirino: User might not have access to operator logs.

chirino:
Yeah we could go the simple way and just stop writing default values to the CR.  I'm fine with that too.  Was just hoping to make it a little easier for our users to modify the CR config.   I'll close this out and just stop updating the CR with the default.

PR #6627, opened 2019-09-18
Fixes ENTESB-11653: Fixed incomplete unit test, fixed export endpoint by looking up environment ids instead of names

dhirajsb:


dhirajsb:
Do we need a backport label to add this to next 7.5 CR?

PR #6636, opened 2019-09-18
[ENTESB-11669] Remove Syndesis registry from Prometheus image prefix

asmigala:
This would be a problem with productized builds:

* QE needs to pull the productized images from either brew or quay
* the prometheus image is not built as part of the the productization, an already released image is used

This solves the problem:
* in upstream builds, nothing changes, since `docker.io` is the default when the registry is not specified
* for productized builds, `PrometheusImagePrefix` in `config.yaml` will be changed to `registry.redhat.io/openshift3` and the released image will be used

cc @orpiske


heiko-braun:
Thanks for the contribution @asmigala !

PR #6638, opened 2019-09-18
fix: update ci/cd page validation and state handling

gashcrumb:
fixes [ENTESB-11386](https://issues.jboss.org/browse/ENTESB-11386)

Refactors a few methods to work with the kinda standard Syndesis error object


dhirajsb:
Verified, works as expected.

PR #6643, opened 2019-09-18
ENTESB-11643 Rework of virtualization create form

mdrillin:
Rework the virtualization create page based on feedback from UX
- The VirtualizationCreateForm component was created in packages/ui - now utilizes pf4 components
- The button positions follow pf4 guidelines (cancel button to the right of create button)
- created story for the component
- VirtualizationCreateForm was incorporated into the VirtualizationCreatePage.  Added handler for cancel button - redirects to virtualizations summary page.  Reworked error handling so that the error is more appropriately shown above the form - instead of a toast notification.

### DIFF
--- a/app/ui-react/packages/ui/src/Data/Virtualizations/VirtualizationCreateForm.tsx
+++ b/app/ui-react/packages/ui/src/Data/Virtualizations/VirtualizationCreateForm.tsx
@@ -1,0 +1,86 @@
+import { ActionGroup, 
+         Alert, 
+         Button, 
+         Card, 
+         CardBody, 
+         Form } from '@patternfly/react-core';
+import * as React from 'react';
+
+export interface IVirtualizationCreateValidationResult {
+  message: string;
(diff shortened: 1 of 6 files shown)


app/ui-react/packages/ui/src/Data/Virtualizations/VirtualizationCreateForm.tsx
```diff
@@ -0,0 +1,88 @@
...
+   */
+  handleSubmit: (e?: any) => void;
+
+  /**
+   * The callback for cancel.
+   */
+  onCancel: () => void;
+}
+
+export class VirtualizationCreateForm extends React.Component<
```
elvisisking: Can you make this a `FunctionComponent`?

mdrillin: Good catch - I changed and pushed an update

PR #6646, opened 2019-09-19
feat(s2i): add a PNC maven profile

squakez:
Using the profile won't trigger the S2I template project generation

closes ENTESB-11682

### DIFF
--- a/app/s2i/pom.xml
+++ b/app/s2i/pom.xml
@@ -35,6 +35,8 @@
     <!-- Must be set empty here so that it can be overridden from the outside-->
     <!-- Don't delete this line -->
     <maven.repo.local></maven.repo.local>
+    <s2i.sample-generation.skip>false</s2i.sample-generation.skip>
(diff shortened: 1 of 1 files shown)


app/s2i/pom.xml
```diff
@@ -257,5 +261,18 @@
         <mavenMirrorOf>external:*</mavenMirrorOf>
       </properties>
     </profile>
+    <!-- Build by PNC -->
+    <profile>
+      <id>pnc</id>
```
zregvart:
```suggestion
      <id>with-proxy</id>
```

Perhaps?

squakez: Yeah, I see that the same exists on `ui-react` module: https://github.com/syndesisio/syndesis/blob/master/app/ui-react/pom.xml#L149 - I think such a profile could be ported to the parent pom. Any inconvenience if I go on that direction?

squakez: The previous idea is not feasible as it would imply porting also variables in the parent pom, sticking to the `with-proxy` name instead.

PR #6648, opened 2019-09-19
Add a profile to set a custom-cafile

cunningt:
Adding a profile to set a custom-cafile.


cunningt:
@zregvart I'll do some testing to determine whether that's still needed.

zregvart:
@cunningt let's merge this and then we can see about another PR for removing the `insecure` profile.

PR #6652, opened 2019-09-19
chore: remove dependency on oc for the backup and restore sub commands. Handle shrinking arrays on resource updates.  Support using the NAMESPACE env var to set the default namespace commands execute against.

chirino:


heiko-braun:
@chirino Is this a 7.6 thing?

heiko-braun:
@lgarciaaco Mind a review?

PR #6653, opened 2019-09-19
fix(ui): show the error message on failed saves

riccardo-forina:
Fixes ENTESB-11685

save-error


heiko-braun:
Thanks @riccardo-forina !

PR #6657, opened 2019-09-19
fix(ENTESB-11247) - Detail page of Virtualization doesn't contain state of publishing

elvisisking:
- See [ENTESB-11247](https://issues.jboss.org/browse/ENTESB-11247)
- added `useVirtualization` hook in order to get updates to the virtualization
- no longer need `state` from the `useRouteData` hook
- now refreshing the published state every 5 seconds to see if it has changed

### DIFF
--- a/app/ui-react/syndesis/src/modules/data/pages/VirtualizationViewsPage.tsx
+++ b/app/ui-react/syndesis/src/modules/data/pages/VirtualizationViewsPage.tsx
@@ -1,8 +1,14 @@
 import {
+  usePolling,
   useViewDefinitionDescriptors,
+  useVirtualization,
   useVirtualizationHelpers,
 } from '@syndesis/api';
-import { QueryResults, ViewDefinitionDescriptor } from '@syndesis/models';
+import {
+  QueryResults,
+  ViewDefinitionDescriptor,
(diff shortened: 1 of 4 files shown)


elvisisking:
I agree @mdrillin. I'll change to display a spinner when first loading the state.

app/ui-react/syndesis/src/modules/data/pages/VirtualizationViewsPage.tsx
```diff
@@ -146,8 +166,8 @@ export const VirtualizationViewsPage: React.FunctionComponent = () => {
     await handlePublishVirtualization(pVirtualizationId, hasViews);
   };
 
-  const doUnpublish = async (virtualizationName: string) => {
-    await handleUnpublishVirtualization(virtualizationName);
+  const doUnpublish = async (serviceVdbName: string) => {
```
mdrillin: this should be virtualizationName, not serviceVdbName

elvisisking: Fixed

mdrillin:
Looks good - thanks Dan!

Issue #6356, opened 2019-08-12
Consolidate fields allowed in syndesis CR


lgarciaaco:
As we move toward an operator full supporting CR, we have to update documentation in some aspects:

- Description of all suported CR fields
- Documentation on how to install syndesis for customers
- Documentation on how to install syndesis for development
- Documentation on how to debug installation steps

KurtStam:
Can we add comments in the CR? And can we add all possible config options in it? I'm thinking like a linux config file, where values are commented out if not in use.

TovaCohen:
Will this have an impact on the Fuse Online installation doc that we provide here: 
https://access.redhat.com/documentation/en-us/red_hat_fuse/7.4/html/installing_and_operating_fuse_online_on_openshift_container_platform/installation-of-fuse-online-on-ocp_ocp

lgarciaaco:
> Will this have an impact on the Fuse Online installation doc that we provide here:
> https://access.redhat.com/documentation/en-us/red_hat_fuse/7.4/html/installing_and_operating_fuse_online_on_openshift_container_platform/installation-of-fuse-online-on-ocp_ocp

@TovaCohen yes!!

claudiocampisi:
Is there a way to have some documentation for syndesis CR before it will be published officially. Actually i need to know how to provide http proxy information when building integrations.

heiko-braun:
@lgarciaaco Can you help @claudiocampisi ?

lgarciaaco:
@claudiocampisi is this of any help: https://github.com/syndesisio/syndesis/blob/master/install/operator/README.md

claudiocampisi:
Thanks @lgarciaaco, i already read that readme file. I was wondering how to configure S2IConfiguration component in order to specify http proxy environment information

PR #6669, opened 2019-09-20
[operator] replace eject for install standalone 

lgarciaaco:
This PR replace `operator eject` with `operator install standalone`

```
perform a fire an forget installation

Usage:
  operator install standalone [flags]

Flags:
      --addons string            a coma separated list of addons that should be enabled
  -h, --help                     help for standalone
      --operator-config string   Path to the operator configuration file. (default "/conf/config.yaml")
```

`install standalone` aims to provide a way of installation where no cluster admin rights are required. It will parse the syndesis resources and directly install them in openshift.

There is no eject mode for `install standalone`.

`install standalone` supports addons installation, but some addons require 3rd party CRDs ... which brings us back to the problem we are trying to solve.

```
operator install standalone --addons camelk,opds --operator-config build/conf/config.yaml
```

`install standalone` requires the conf/config.yaml


lgarciaaco:
related with https://github.com/syndesisio/syndesis/issues/6356

lgarciaaco:
Updated [documentation](https://github.com/syndesisio/syndesis/wiki/Deploying-to-a-Remote-OpenShift-from-the-Syndesis-Template) accordingly

lgarciaaco:
@heiko-braun are we going to use this for eval clusters? If so, we rather get QE to test it

zregvart:
Perhaps `operator install unmanaged`?

heiko-braun:
@lgarciaaco The backport needs some handholding.

PR #6667, opened 2019-09-20
reflect images tag in deploymentconfigs

lgarciaaco:
Fixes ENTESB-11683

PR #6670, opened 2019-09-20
1.8.x

lgarciaaco:
Same as https://github.com/syndesisio/syndesis/pull/6667


heiko-braun:
@lgarciaaco this would be avoided `backport master`. Please always do it the other way around: master first, maintanance branch second

PR #6676, opened 2019-09-20
fix: documentation build in the daily release

zregvart:
This adds a `index.adoc` to pull in all documents from modules within the `doc` directory. Currently the daily build is failing because of missing `master.adoc`.

@fbolton, @TovaCohen I don't think this makes a difference in the product doc build, but just to make sure...

### DIFF
--- a/doc/index.adoc
+++ b/doc/index.adoc
@@ -1,0 +1,14 @@
+:experimental:
+include::shared/attributes.adoc[]
+
+= Syndesis User Documentation
+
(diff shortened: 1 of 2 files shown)


TovaCohen:
Right. This toplevel index.adoc file will not get in the way of the product doc build, which we do downstream. Previously, there was a master.adoc file, which I removed to cause this build error. Sorry. Thanks for resolving this.

PR #6688, opened 2019-09-23
add some documentation for syndesis CR

lgarciaaco:
Documentation for CR fields. 

This is not the final documentation, but a temporally one so we can have some reference in the meanwhile. @phantomjinx is working in a dynamic way of generating the documentation from the Syndesis object itself

- Fixes ENTESB-11339


lgarciaaco:
Related with https://github.com/syndesisio/syndesis/issues/6356
Fixes FUSEDOC-3450

heiko-braun:
Thanks, lgtm.

PR #6690, opened 2019-09-23
fix(ENTESB-11697): Render data testid property in text buttons

christophd:
Adds data testid to the "add another item" button on form arrays.

Fixes ENTESB-11697

### DIFF
--- a/app/ui-react/packages/auto-form/src/widgets/TextButton.tsx
+++ b/app/ui-react/packages/auto-form/src/widgets/TextButton.tsx
@@ -2,40 +2,20 @@ import * as React from 'react';
 
 import './TextButton.css';
 
-export interface ITextButtonProps {
-  className?: string;
+export interface ITextButtonProps extends React.HTMLAttributes<HTMLElement> {
   linkText?: string;
-  onClick?: (event: React.MouseEvent<HTMLButtonElement>) => void;
-  children?: React.ReactNode;
   enable?: boolean;
 }
 
-export class TextButton extends React.Component<ITextButtonProps> {
-  public render() {
-    const enable =
-      typeof this.props.enable !== 'undefined' ? this.props.enable : true;
-    const className = this.props.className || '';
-    return (
-      <>
-        {this.props.linkText ? (
-          <button
-            type="button"
-            onClick={this.props.onClick}
-            className={`btn btn-link ${className}`}
-            dangerouslySetInnerHTML={{ __html: this.props.linkText! }}
-            disabled={!enable}
-          />
-        ) : (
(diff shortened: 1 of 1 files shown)


app/ui-react/packages/auto-form/src/widgets/FormArrayComponent.tsx
```diff
@@ -111,7 +111,7 @@ export const FormArrayComponent: React.FunctionComponent<
       })}
       <div className={'form-array-control__array-add'}>
         <TextButton
-          data-testid="form-array-control-add-another-item-button"
+          id="form-array-control-add-another-item-button"
```
riccardo-forina: No, I think this should stay `data-testid` for consistency sake.

app/ui-react/packages/auto-form/src/widgets/TextButton.tsx
```diff
@@ -1,8 +1,10 @@
 import * as React from 'react';
 
+import { toValidHtmlId } from "./helpers";
 import './TextButton.css';
 
 export interface ITextButtonProps {
```
riccardo-forina: We can extend the `TextButton` to inherit props that are valid for the underlying component, like this `export interface ITextButtonProps extends React.HTMLAttributes<HTMLElement>`. That should allow all `data` attributes to be passed to the component.

app/ui-react/packages/auto-form/src/widgets/TextButton.tsx
```diff
@@ -23,13 +26,15 @@ export class TextButton extends React.Component<ITextButtonProps> {
             onClick={this.props.onClick}
             className={`btn btn-link ${className}`}
             dangerouslySetInnerHTML={{ __html: this.props.linkText! }}
+            data-testid={id}
```
riccardo-forina: With the previous suggestions, this can become `{...props}`, and all `this.props` can be dumped.

app/ui-react/packages/auto-form/src/widgets/TextButton.tsx
```diff
@@ -15,6 +17,7 @@ export class TextButton extends React.Component<ITextButtonProps> {
     const enable =
       typeof this.props.enable !== 'undefined' ? this.props.enable : true;
     const className = this.props.className || '';
+    const id = toValidHtmlId(this.props.id!);
     return (
```
riccardo-forina: With the `extend` suggested above, here all props that are custom to the `TextButton` should be extracted from the props. Something along the lines of:
`const { enable, linkText, className, ...props } = this.props;`

PR #6709, opened 2019-09-24
feat(common): enhanced name service sanitization

### LINKED ISSUE #3951: Don't use user input in names, labels or annotations of k8s objects

squakez:
* Adding 4 digits calculated from the hash to sanitization service in order to avoid clashing of names with same base
* Changed a few test cases in order to match the new name structure
* Beware, the change does not guarantee uniqueness but it limits collisions in known cases

closes #3951

With this change we ensure to avoid that sanitized names with same base (ie `test?` and `test!`) are not returning the same result. There is still some possibility for collision though it would be very rare to find any, but a real hashing would require a very long code to append (20 chars). Through this change we have the test cases prepared to accept any additional hash if we change it in the future (ie using any other algorithm).

Since there is a new name for each pod created, I need a peer review to confirm that this would not affect any possible external dependency I am not aware of.


squakez:
@syndesisio/backend any volunteer to have a look at this? or should I drop the PR?

zregvart:
I'd rather rethink the naming entirely and come up with solution that doesn't depend on the integration name at all, and there's no need to sanitize any user input. For that we would need to discuss if the current (weak) link between the integration name and the pod/service/secret is something that brings us value.

My opinion is that there's no value in linking the integration name with the Kubernetes objects and drop the name sanitization.

squakez:
Thanks @zregvart, I understand the point and agree with the need of a different design. However, since the development is ready and it is not adding any further complexity to the actual solution, do you think we can merge this and still keep the feature request as open for an eventual redesign? or should we better drop it at all?

zregvart:
I'd drop anything that prolongs this issue and concentrate the effort into doing the proper solution.

(2c)

squakez:
Thanks for the comments. I'll close the PR but I'll leave the branch development open in case it can provide hints to future developments.

PR #6711, opened 2019-09-24
fix(ENTESB-11578): Improve conditional flow usability

christophd:
Adds a select config mode (basic, advanced) page to step configuration so user can choose how to provide condition expressions.

Fixes ENTESB-11578

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/EditorApp.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/EditorApp.tsx
@@ -8,6 +8,7 @@ import { ReviewActionsPage } from './apiProvider/ReviewActionsPage';
 import { SelectMethodPage } from './apiProvider/SelectMethodPage';
 import { ChoiceStepPage } from './choice/ChoiceStepPage';
 import { DescribeChoiceDataShapePage } from './choice/DescribeChoiceDataShapePage';
+import { SelectChoiceConfigModePage } from "./choice/SelectChoiceConfigModePage";
 import { DataMapperPage } from './dataMapper/DataMapperPage';
 import { EditorRoutes } from './EditorRoutes';
 import { EditorSidebar } from './EditorSidebar';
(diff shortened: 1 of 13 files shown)


app/ui-react/syndesis/src/modules/integrations/components/editor/EditorApp.tsx
```diff
@@ -207,15 +206,14 @@ export const EditorApp: React.FunctionComponent<IEditorApp> = ({
     <DescribeChoiceDataShapePage
       cancelHref={cancelHref}
       sidebar={props => (
-        <EditorSidebar {...props} isAdding={false} />
+        <EditorSidebar {...props} isAdding={mode === 'adding'} />
```
riccardo-forina: Good catch

app/ui-react/syndesis/src/modules/integrations/components/editor/choice/SelectChoiceConfigModePage.tsx
```diff
@@ -0,0 +1,142 @@
...
+ * [state]{@link ISelectConfigModeRouteState} to be properly set in the route
+ * object.
+ *
+ * **Warning:** this component will throw an exception if the route state is
+ * undefined.
+ */
+export class SelectChoiceConfigModePage extends React.Component<ISelectChoiceConfigModePageProps> {
+  public render() {
+    return (
+      <WithRouteData<ISelectConfigModeRouteParams, ISelectConfigModeRouteState>>
```
riccardo-forina: For the future, there is an equivalent hook for this called `useRouteData`. No need to change this now, eventually we'll update the whole routing thing when `react-router` will hit v6.

dongniwang:
Hey @christophd - thanks for the PR. Would it be possible to include some key screenshots? Staging is down at the moment for me...and I'd like to take a look at it as soon as possible. Thank you!

christophd:
@dongniwang I have added a screencast to the ENTESB issue

PR #6717, opened 2019-09-24
e2e test data setup commands

kahboom:
Some changes to the E2E tests we have set up for this repo to prepare for working strictly with tapes, predictable data, and no network requests (unless we run a specific setup command).

- set up the db with `yarn e2e:setup`, which offers separate tests to reset the db, create data (e.g. connection, integration), and create a snapshot of the db. these are normally excluded when running `yarn e2e`.
- non-setup tests can be run with the tapes, because they should not be making any backend requests, just testing the UI. (up for discussion)
- ability to run e2e tests headless with `yarn e2e:terminal`.
- ability to take a snapshot of the db programmatically, which is stored alongside the codebase in `/syndesis/cypress/snapshots` as a JSON file.
- extract CUD commands to cypress custom commands, custom plugins for writing setup code.

Screenshot 2019-09-24 09 55 04

- restore db is acting up, though i'm able to restore the db using the stored file with the proper curl command, think it has to do with working behind a proxy + cypress. can look into this, but really, restoring a snapshot is not a priority since we are re-creating the data manually (as its own set of tests) and then resetting the db anyway. it would be a convenience feature. for now I'm disabling these tests, as they might still be useful, but will fail atm.

next steps are to write some more tests that simply double check that everything in the UI loads as expected, now that we have predictable data. these should be a very small number of specific tests, the remaining should be integration and unit tests (some of which we already have).


kahboom:
Some videos to help see how things work, particularly for the setup command:

1. [Starting up Cypress (`yarn e2e`)](https://drive.google.com/open?id=1bywtqdWH9vEw0DigICuZGWE9DwfK470N)
2. [Running tests without expected data causes failure](https://drive.google.com/open?id=18bfJ5suR3vKG_v_4QiC4IaChInAJ5TFt)
3. [Setup: Reset the database (`yarn e2e:setup`)](https://drive.google.com/open?id=1RREOUelufpKs5W-dVxMSyrOGwMBEuvKJ)
4. [Setup: Create data expected by tests](https://drive.google.com/open?id=1AI1WdkFxzF_88SC_m8Ixe9vPJ7L8Vr-4)
5. [Setup: Take a snapshot of existing DB for future use](https://drive.google.com/open?id=1dVRM8loKg-ffHHXOLBm3BIIqnslWFDZa)
6. [Tests now pass with expected data (`yarn e2e`)](https://drive.google.com/open?id=1Q0_g7hMNA98_FTdZ7dn7AF8u151JcFfV)

Once you have the data you can run tests like normal with `yarn e2e`, and even create/use tapes as you normally would.

If these videos are useful, then we can add them to the README. Similarly, if you'd like to see how these would work in conjunction with the tapes in a few videos, let me know and I'll record some.

cc @gashcrumb @riccardo-forina

app/ui-react/syndesis/cypress/integration/setup/01-create.setup.spec.js
```diff
@@ -0,0 +1,25 @@
...
+ * 01: CREATE DATA
+ *
+ * The purpose of this "test" is to setup data as the user would in the UI.
+ * First, the database is reset, and then custom Cypress commands are used
+ * to create the data manually.
+ */
+
+const constants = require('../../fixtures/constants');
+
+describe('Create Data', () => {
```
gashcrumb: Is this the right way to do initialization for tests?  Is this a one-time thing or an every test thing?

I get kinda suspicious of tests that need to run in a particular order.

kahboom: The reason I separated these setup tasks into separate tests is for our own convenience. For instance, I could throw all of these into a single setup test, but then that would mean that each and every time we run that test we are resetting the DB, creating the data, and taking a snapshot. I figured that there may be a time that someone would want only a snapshot of the DB, or maybe only to reset the DB, and so on.

The reason they are named in a specific order with numbers in the name is because Cypress runs tests alphabetically, and it wouldn't make sense to create a snapshot and _then_ create the data, for instance. If it were all in one test, this probably wouldn't be an issue because we set up the test manually, but of course means everything would be bundled into one.

I can put these all into a single test if you'd like. Likewise, I can create an npm command for each of these tasks/tests and then the naming and order really wouldn't matter. However, running `yarn e2e:setup` and then clicking 'Run All Tests', would run them in alphabetical order, which wouldn't work.

Last thing I thought I'd mention is that the reason these are written as tests is because the environment Cypress provides is the best way to use the Cypress command, execute node commands, and perform API requests.

kahboom:
Closing for now.

Issue #6713, opened 2019-09-24
[operator] don't log resource diffs by default since they could contain secrets.


chirino:
## This is a...

<pre><code>
[X] Bug report  
</code></pre>

Logging resource diffs could be considered a security risk since they can contain secrets.  We should only do this if an option is enabled in the operator for debugging purposes.

PR #6731, opened 2019-09-25
fix: #6713 don't log resource diffs by default since they could conta…

chirino:
…in secrets.

Also reran go mod tidy and go mod vendor


chirino:
rebased

PR #6752, opened 2019-09-26
Fix upgrade

lgarciaaco:
Fixes ENTESB-11320

- Dot write back syndesis CR in install or upgrade
- Fix copy and paste error
- Bring back `POSTGRESQL_USER` in global secret, it is needed for upgrade
- Don't deepcopy the cr in some scenarios, or install kicks in (it doesnt know that upgrade is running)


lgarciaaco:
@avano This seems to fix the issues. Mind a look?

avano:
@lgarciaaco hi Luis, looks ok!

I built the operator image from your branch with the config from 1.8.1-20190927 (because there were no changes to other images)

The upgrade pod can be tweaked a bit - I think the https://github.com/syndesisio/syndesis/blob/master/tools/upgrade/steps/upgrade_20_stop_all and https://github.com/syndesisio/syndesis/blob/master/tools/upgrade/steps/upgrade_40_update_resources can be left out completely, as currently the upgrade pod stops all deployments, then does nothing (upgrade_40_update_resources is not doing any changes) and then scales everything back to 1.

Then the upgrade pod finishes and operator replaces all resources with newer ones, which trigger another round of re-deployments.

If you don't want to address that due to time issues, I'm ok with that :-)

lgarciaaco:
@avano thanks! I add those changes straight away!

avano:
also there is a problem with rebuilding the old integration after the upgrade: https://issues.jboss.org/browse/ENTESB-11766 however I think that may be a problem in the syndesis server code

PR #6757, opened 2019-09-26
fix(ENTESB-11758): Fix form field rendering on metadata lookup

christophd:
Restores old behavior (before data list feature) where we use select dropdown by default. Only use dataList for those fields marked accordingly in the connector json (just used by Google Sheets at the moment).

When only one single value is suggested pick this as default value so it is preselected in dropdowns and dataList components.

Fixes ENTESB-11758


christophd:
This is related to former changes in metadata lookup logic

https://github.com/syndesisio/syndesis/issues/5690
https://github.com/syndesisio/syndesis/pull/5191

PR #6763, opened 2019-09-27
Don’t punish the default settings on the CR since this will cause pro…

chirino:
…blems with future updates since it locks in the default settings (like image tag names).

(cherry picked from commit feed49af338ab9900a3fb27ba93237a1b0bd7c7d)


lgarciaaco:
@chirino those changes are included in https://github.com/syndesisio/syndesis/pull/6752/files

PR #6768, opened 2019-09-27
[ENTESB-11965] Fix DV UI enabled by default

asmigala:
This is just a quick fix, the proper way to fix it would probably be to have the UI read `true`/`false` values rather than `1`/`0`


heiko-braun:
Thanks @asmigala !

PR #6775, opened 2019-09-27
chore: upgrade apicurio to 0.2.40

gashcrumb:
fixes [ENTESB-11759](https://issues.jboss.org/browse/ENTESB-11759)


riccardo-forina:
This should not be backported I think, to avoid adding load on QA. Do you confirm @paoloantinori?

paoloantinori:
yes, not backported. this was an activity planned for 7.6

gashcrumb:
Ah, didn't see that, thanks!

heiko-braun:
+1 on not backporting. This might carry unintended side effects.

PR #6784, opened 2019-09-27
fix(ci): ignore `cherry-pick-backport` branches

zregvart:
During the creation of backport PR the backport bot creates a temporary ref[1] that gets picked up by CircleCI to build. That build always fails so we get check failures on the PR. This adds ignores to all jobs so that there is no build of this temporary ref.

[1] https://github.com/tibdex/github-cherry-pick/blob/15767bab6cf1c60d2032f9d4ba89fe47e3a3c582/src/index.ts#L318

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,6 +90,11 @@ yaml-templates:
       paths:
         - repository/io/syndesis
 
+  ignore_cherry_pick_branches: &ignore_cherry_pick_branches
+    filters:
+      branches:
+        ignore: /cherry-pick-backport.*/
+
 common_env: &common_env
   MAVEN_OPTS: -Xmx1024m
   DOCKER_VERSION: 18.06.1-ce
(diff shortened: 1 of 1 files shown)


zregvart:
Example of this issue is on #6782 (and many others):

https://circleci.com/workflow-run/56b45bbb-3dd4-4661-a151-51e3d71ba186

vs

https://circleci.com/workflow-run/fee1ef7d-2271-4aea-91ca-211af9d4a6d9

zregvart:
Not sure [what's going on here](https://circleci.com/gh/syndesisio/syndesis/139028#config/containers/0) @christophd any ideas?

.circleci/config.yml
```diff
@@ -1129,17 +1180,21 @@ workflows:
...
           requires:
             - integration
             - connector-catalog
             - common
       - s2i:
+          <<: *ignore_cherry_pick_branches
           requires:
             - server
       - upgrade:
+          <<: *ignore_cherry_pick_branches
```
christophd: I can only imagine that this upgrade job is already using a filter setting and bringing in the template to ignore cherry pick branches screws up the config

zregvart: Good catch, should be fixed, thanks!

PR #6789, opened 2019-09-28
ENTESB-11573 add external database support in operator

aidenkeating:
currently, in the syndesis operator, the creation of the in-cluster
postgres instance is hardcoded. there is no way to specify an
external database through the custom resource.

this change adds a new field to the syndesis custom resource named
externalDbSecretRef which should be set to the name of a secret in
the same namespace. this secret contains information about an
external database.

this new field is optional, if it is not specified, an in-cluster
postgres will be deployed as normal, to keep the operator backwards
compatible.


zregvart:
My knee jerk first thought is that we should perhaps expose the JDBC URL for the database from the `db` section and have the username/password managed via Secret in all cases. So rather than making this case special, unify the approach so it fits all cases.

aidenkeating:
@zregvart Thanks for the feedback. Just to confirm, you mean something like the following flows:

## Created database
- CR doesn't have `.components.db.url` set
- [Optional] CR has `.components.db.secretRef` set, if not set a default is used
- Postgres credentials are generated and stored in Secret
- Postgres instance is created using the generated credentials

## External database
- [Prerequisite] Secret is created manually with the username and password specified
- CR has `.components.db.url` set, to `postgres://example:5432`
- [Optional] CR has `.components.db.secretRef` set, if not set a default is used
- Postgres doesn't generate credentials and reads credentials from Secret
- Postgres instance creation is skipped

If so, that makes sense to me. Would we also need to allow the username to be specified via the CR to maintain backwards compatibility?

zregvart:
@aidenkeating exactly like that, thank you for putting it to words so nicely. I think username can be a part of the Secret also.

Let's have some time for other reviewers to voice their opinions.

lgarciaaco:
Do you still need a review for this or there is another PR being prepared?

aidenkeating:
@lgarciaaco I think we're still in need of reviews here, if you have the time?

And perhaps an opinion on the comments above?

install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
```diff
@@ -119,6 +119,8 @@ type DbConfiguration struct {
 	User                 string              `json:"user,omitempty"`
 	Database             string              `json:"database,omitempty"`
 	ImageStreamNamespace string              `json:"imageStreamNamespace,omitempty"`
+	// Options for configuring an external database instead of deploying one
+	ExternalDBSecret string `json:"externalDBSecretRef,omitempty"`
```
chirino: You should not put secrets in a CR since they are not encoded.

install/operator/pkg/generator/assets/database/syndesis-db.yml.tmpl
```diff
@@ -212,11 +212,11 @@
         containers:
         - env:
           - name: POSTGRESQL_USER
-            value: {{.Syndesis.Spec.Components.Db.User}}
+            value: {{.Env.POSTGRESQL_USER}}
```
chirino: Your going backwards here from what we want to do.

install/operator/pkg/syndesis/action/install.go
```diff
@@ -101,6 +101,22 @@ func (a *installAction) Execute(ctx context.Context, originalSyndesis *v1alpha1.
 	}
 	config[string(configuration.EnvOpenShiftOauthClientSecret)] = token
 
+	// Handle an external database being defined
+	externalDbSecName := syndesis.Spec.Components.Db.ExternalDBSecret
```
chirino: I think you should just use an ExternalDbService or ExternalDbURL field to know if the mode is enabled.

install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
aidenkeating: I'm not sure I understand? This is a reference to a secret name, not the secret itself

install/operator/pkg/generator/assets/database/syndesis-db.yml.tmpl
aidenkeating: I'm not familiar with what the team wants to do.

This was moved to an env to handle that an external database has the user defined in a secret, not in the CR spec. I did this to keep all of the information together in a secret and to avoid providing a password in the CR.

I agree that it's not ideal to have the divide between the two ways of deploying. So I'm happy to make changes to bring the "created" and "external" database logic branches closer together, if you have suggestions?

install/operator/pkg/syndesis/action/install.go
aidenkeating: Do you mean include the entire connection string in the CR. I've not seen passwords be passed in via a CR before. The usual pattern I'd seen is that the CR references out to a secret if a password is needed.

If you're happy with the connection string being in the CR, I'd be happy to make that change

install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
chirino: ah.. if it's a ref, then that's ok.  Note that your json name has to match  the go field name or else you'll get an error when you generate the openapi docs.

But then again, we already have a secret where we store the DB password.  Why don't we keep using it?

install/operator/pkg/generator/assets/database/syndesis-db.yml.tmpl
chirino: The CR is still controlling the configuration and deployment of Syndesis.  
We just need the CR to say don't deploy the DB, just connect to an external one with the smallest amount of change possible.

install/operator/pkg/syndesis/action/install.go
chirino: I mainly meant the hostname:port mainly.  I think we have fields in the CR and our secrets to configure everything else needed for an external connection.

install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
aidenkeating: I think that secret is `syndesis-global-config`, right?

The reason I went with not reusing that is that it contains other information, almost all the content of that secret makes sense to be created by the operator in all cases. So, even though there ends up being some duplication, this way `syndesis-global-config` stays "owned" and maintained by the operator, and the user-created secret only has connection information and stays "owned" by the user.

However, this made more sense when keeping all connection info about the Postgres instance was going into a secret. So I'm happy to change it if needed.

install/operator/pkg/syndesis/action/install.go
aidenkeating: Ok cool. I'll make changes to re-use as much of the existing CR/Secrets as possible instead of adding an extra Secret. Thanks for the review.

aidenkeating:
@chirino @zregvart Thanks for the reviews. I have made changes based on the feedback.

Basically the flow for an external database is now.

- In the CR `spec`, set:
     - `.components.db.user` to an existing user
     - `.components.db.database` to an existing database name
     - `externalDbURL` to the host of the cluster, like `postgresql://myhost.svc:5432`. If a database is not included here, it will be added from `.components.db.database`
- Create a secret named `syndesis-global-config`, that at least contains, in `data`:
    - `POSTGRESQL_PASSWORD` as the password of the existing user defined in the CR

This avoids putting any passwords in the CR and uses existing resources, the only new part is `externalDbURL` in the CR now.

aidenkeating:
@zregvart @lgarciaaco @chirino Any update on this?

From what I can see from the CircleCI workflow, every log says success, so I'm not fully sure why the test is failing. But I'll look into it more.

zregvart:
@aidenkeating there are [issues in the tests](https://circleci.com/gh/syndesisio/syndesis/149101):
```
# github.com/syndesisio/syndesis/install/operator/pkg/syndesis/configuration
pkg/syndesis/configuration/configuration.go:88:37: syntax error: unexpected ConfigSpec, expecting comma or }
pkg/syndesis/configuration/configuration.go:89:2: syntax error: non-declaration statement outside function body
ok  	github.com/syndesisio/syndesis/install/operator/pkg/cmd/internal/grant	0.053s
FAIL	github.com/syndesisio/syndesis/install/operator/pkg/cmd/internal/install [build failed]
ok  	github.com/syndesisio/syndesis/install/operator/pkg/cmd/internal/uninstall	0.037s
FAIL	github.com/syndesisio/syndesis/install/operator/pkg/generator [build failed]
FAIL	github.com/syndesisio/syndesis/install/operator/pkg/syndesis/action [build failed]
ok  	github.com/syndesisio/syndesis/install/operator/pkg/util	0.019s
ERROR: Last command exited with 2
```

aidenkeating:
@zregvart Apologies, I hadn't pushed my latest changes, pushed them now. My previous comment with steps for external databases are still accurate.

aidenkeating:
@zregvart @lgarciaaco @chirino updated based on suggestions and rebased. could you take a look when you get the chance, thanks!

PR #6795, opened 2019-09-30
[ENTESB-11786] Fix hardcoded dc config

asmigala:


heiko-braun:
Thanks @asmigala, but shouldn't this go on master first?

asmigala:
@heiko-braun looks like master is not using the `ProductName` in the templates at all (e.g. https://github.com/syndesisio/syndesis/blob/master/install/operator/pkg/generator/assets/infrastructure/03-syndesis-ui.yml.tmpl#L100 ), so adding it there would be pretty invasive, especially since we are moving away from imagestreams for 7.6.

asmigala:
@heiko-braun in fact the `ProductName` was specifically removed as part of the move away from imagestreams:

https://github.com/syndesisio/syndesis/commit/9945174053a45acdbc2e7091fc10c0453859f410#diff-af3e83fd83e7f4724c8033b2d609fd0fL89

heiko-braun:
> move away from imagestreams

Right, I forgot about that

heiko-braun:
@lgarciaaco Can you please review this?

PR #6796, opened 2019-09-30
[ENTESB-11649] Adding a table name validation

Delawen:


Delawen:
@oscerd the next thing I am going to review is what else may be missing on apache camel and pull :)

PR #6801, opened 2019-09-30
chore: Update version number so doc URLs will be correct for 7.5

gashcrumb:
Fixes the documentation links for the next release

@TovaCohen this look okay?


TovaCohen:
LGTM!

PR #6810, opened 2019-10-01
Change binary name from operator to syndesis

lgarciaaco:
`operator` is very generic, change the binary to `syndesis`

@orpiske you tweak to some nods on your end.

### DIFF
--- a/install/operator/.lib.sh
+++ b/install/operator/.lib.sh
@@ -93,14 +93,14 @@ build_operator()
         for GOARCH in amd64 ; do
           for GOOS in linux darwin windows ; do
             export GOARCH GOOS
-            echo building ./dist/${GOOS}-${GOARCH}/operator executable
-            go build  "$@" -o ./dist/${GOOS}-${GOARCH}/operator \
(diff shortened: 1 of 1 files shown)


orpiske:
@lgarciaaco thanks, noted!

asmigala:
@lgarciaaco won't it be kind of confusing, since there will now be two executables called `syndesis` ? The other being https://github.com/syndesisio/syndesis/blob/master/tools/bin/syndesis

lgarciaaco:
> @lgarciaaco won't it be kind of confusing, since there will now be two executables called `syndesis` ? The other being https://github.com/syndesisio/syndesis/blob/master/tools/bin/syndesis

I dont know ... maybe. @heiko-braun ^^

zregvart:
Needs a change here:

https://github.com/syndesisio/syndesis/blob/7a18a509fbdb70ad9c71c67f27af52d20a49fa73/install/operator/build/Dockerfile#L8

Also would be better to do these changes on `master` and then backport it to `1.8.x`

lgarciaaco:
> Also would be better to do these changes on `master` and then backport it to `1.8.x`

ok, closing this doing it that way

PR #6836, opened 2019-10-04
fix for ENTESB-11787 - Reconfiguration of SQL query step changes text to dropdown

KurtStam:
https://issues.jboss.org/browse/ENTESB-11787

### DIFF
--- a/app/connector/sql/src/main/resources/META-INF/syndesis/connector/sql.json
+++ b/app/connector/sql/src/main/resources/META-INF/syndesis/connector/sql.json
@@ -44,7 +44,7 @@
                 "order": 1,
                 "required": true,
                 "secret": false,
-                "type": "string"
+                "type": "text"
               },
               "batch": {
                 "defaultValue": false,
@@ -112,7 +112,7 @@
                 "order": "1",
                 "required": true,
                 "secret": false,
-                "type": "string"
+                "type": "text"
               },
               "schedulerExpression": {
                 "defaultValue": 60000,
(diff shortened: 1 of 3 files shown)


KurtStam:
@zregvart There are 2 failing Salesforce Tests, should the type be a 'select' rather than a 'string'?

zregvart:
I think you want @christophd's and @gashcrumb's help on this one, I don't know what the current consensus is.

oscerd:
Yes, it should be select, I changed that with another PR. The objectName is now a select and not a text.

KurtStam:
I will squash the commits if the tests pass

christophd:
@KurtStam @zregvart @gashcrumb this fix breaks all other connectors that are using "text"/"string" typed properties that should be rendered as dropdown (see https://issues.jboss.org/browse/ENTESB-11719 and https://issues.jboss.org/browse/ENTESB-11843). 

There was a decision made to revert the logic on the backend server to what was been used before (see https://issues.jboss.org/browse/ENTESB-11758).

The fix made in this PR does not align with this decision and breaks some other connectors again.

I agree that the best fix would be to force all connectors to use "select" as property type where appropriate. IMO we decided to not do this for 7.5 as it implies too much analysis (@heiko-braun @paoloantinori). 

Instead we should keep the backend server logic as it was before and the SQL connector "query" property can just use the new "dataList" property type which renders as editable input text field with suggestions. This way we can keep the initial dropdown rendering logic as it was before and all other connectors that require dropdown rendering for "text"/"string" typed properties are fine.

In a long term perspective we should revisit all connectors and fix the property types from "text"/"string" to "select" where appropriate.

PR #6841, opened 2019-10-04
Release syndesis binaries for real

lgarciaaco:
The [release section](https://github.com/syndesisio/syndesis/releases) has a zipped content of the github repo, which doesn't add much value. I want to start releasing the syndesis binaries instead of what it is released right now.

@zregvart suggested that currently this is set up in the github project. I didn't find how to disable this, but my access is restricted (I don't see the settings tab).

My long term goal is to use those binaries to replace the installation methods in `tools/bin/commands/minishift`, in [quickstart](http://syndesis.io/quickstart/) point users to the binaries.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -849,7 +849,7 @@ jobs:
               fi
             done
             cd ${BASEDIR}
-            ghr -u ${GITHUB_USER} -t ${GITHUB_TOKEN} -r ${CIRCLE_PROJECT_REPONAME} -c ${CIRCLE_SHA1} -replace -prerelease -draft ${VERSION} "${BASEDIR}/archives"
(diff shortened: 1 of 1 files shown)


zregvart:
So I don't really understand how we ended up doing releases via CircleCI? We're already doing them on Jenkins. Perhaps a discussion on how releases are done should be made first.

lgarciaaco:
> So I don't really understand how we ended up doing releases via CircleCI? We're already doing them on Jenkins. Perhaps a discussion on how releases are done should be made first.

That's a very good point, I can setup a meeting.

zregvart:
> That's a very good point, I can setup a meeting.

I'd prefer a discussion on the Syndesis Google group, that would be most inclusive.

lgarciaaco:
https://groups.google.com/forum/#!topic/syndesis/C0kU-H7byQs

PR #6847, opened 2019-10-05
Fix for ENTESB 11633 - setting limit to prevent timeout.

KurtStam:
https://issues.jboss.org/browse/ENTESB-11633

### DIFF
--- a/app/connector/sql/src/main/java/io/syndesis/connector/sql/common/DbMetaDataHelper.java
+++ b/app/connector/sql/src/main/java/io/syndesis/connector/sql/common/DbMetaDataHelper.java
@@ -18,6 +18,7 @@ package io.syndesis.connector.sql.common;
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.JDBCType;
+import java.sql.PreparedStatement;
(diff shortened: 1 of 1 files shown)


app/connector/sql/src/main/java/io/syndesis/connector/sql/common/DbMetaDataHelper.java
```diff
@@ -163,6 +163,13 @@ public ResultSet fetchProcedures(final String catalog,
         }
     }
 
+    @SuppressFBWarnings("OBL_UNSATISFIED_OBLIGATION")
```
zregvart:
```suggestion
    @SuppressFBWarnings("OBL_UNSATISFIED_OBLIGATION") // needed for https://github.com/spotbugs/spotbugs/issues/432
```

zregvart:
Would be cool to point to a spotbugs issue so when it's fixed and we review the `@SuppresWarnings` we can remove it. Other than that LGTM.

(We can merge without this, let me know and I can approve)

KurtStam:
Does this need backporting to 1.8?

heiko-braun:
@kurt not at this point anymore

PR #6853, opened 2019-10-07
change syndesis installation to use the operator binary for crc

lgarciaaco:
Use the operator binary instead of old templates. Download the operator binary if it is not present

- `--route` is not needed anymore, the operator takes care of figuring it out
- change defaults so it works properly with latest crc version
- the whole login to a different user logic is removed, why: Is it anyone really using it? With CRC it is needed to pass a password when login as kubeadmin, so the current logic just doesnt work. We can keep it and ask the user for a password, but I would like to challenge an argument whether this is really needed or not.

I am picking the binary from a manual release, which is far from optimal, but with https://github.com/syndesisio/syndesis/pull/6841 I will improve it later

### DIFF
--- a/tools/bin/commands/crc
+++ b/tools/bin/commands/crc
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 DEFAULT_OPENSHIFT_VERSION="v3.11.0"
-DEFAULT_CPUS="2"
-DEFAULT_RAM="8GB"
+DEFAULT_CPUS="4"
+DEFAULT_RAM="8192"
 
 crc::description() {
     echo "Initialize and manage a developer environment using OCP4 CodeReady Containers"
@@ -10,22 +10,20 @@ crc::description() {
 
 crc::usage() {
     cat <<EOT
-    --install                 Install templates to a running OCP4 CodeReady Containers cluster.
+    --install                 Install syndesis to a running OCP4 CodeReady Containers cluster.
 -p  --project                 Install into this project. Delete this project if it already exists.
                               By default, install into the current project (without deleting)
     --reset                   Reset and initialize the crc installation by
                               'crc delete && crc start'.
     --full-reset              Full reset and initialie by
                               'crc stop && rm -rf ~/.crc/* && crc start'
-    --operator-only           Only install the operator but no resource
-                              connected cluster.
+    --operator-only           Only install the operator but no custom resource
     --memory <mem>            How much memory to use when doing a reset. Default: $DEFAULT_RAM
     --cpus <nr cpus>          How many CPUs to use when doing a reset. Default: $DEFAULT_CPUS
     --pull-secret-file <file> File holding the OCP4 pull secret
     --vm-driver <driver>      Which virtual machine driver to use (depends on OS)
     --bundle <path-to-bundle> Path to system bundle (required when using virtualbox vm-driver)
     --local                   Use the local resource files instead of fetching them from GitHub
--r  --route <route>           Use the given route.
                               By default "syndesis.<project_name>.apps-crc.testing" is used.
 -o  --open                    Open Syndesis in the browser
 -y  --yes                     Assume 'yes' automatically when asking for deleting
@@ -47,6 +45,11 @@ EOT
 crc::run() {
     source "$(basedir)/commands/util/openshift_funcs"
     source "$(basedir)/commands/util/camel_k_funcs"
+    source "$(basedir)/commands/util/operator_funcs"
+
+    OPERATOR_BINARY="$(basedir)/commands/binaries/syndesis"
+
+    download_operator_binary || print_error_and_exit "unable to download the operator binary, exit"
 
     # Check that crc is installed
     which crc &>/dev/null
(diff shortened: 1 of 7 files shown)


lgarciaaco:
If this approach is accepted, I will create another PR changing `minishift` installation so that it also downloads the operator binary.

lgarciaaco:
https://github.com/syndesisio/syndesis/pull/6841 is related

tools/bin/commands/crc
```diff
@@ -1,31 +1,29 @@
 #!/bin/bash
 
 DEFAULT_OPENSHIFT_VERSION="v3.11.0"
-DEFAULT_CPUS="2"
-DEFAULT_RAM="8GB"
+DEFAULT_CPUS="4"
+DEFAULT_RAM="8192"
```
zregvart: Is 8GB needed or can it be lower?

tools/bin/commands/crc
```diff
@@ -47,6 +45,11 @@ EOT
 crc::run() {
     source "$(basedir)/commands/util/openshift_funcs"
     source "$(basedir)/commands/util/camel_k_funcs"
+    source "$(basedir)/commands/util/operator_funcs"
+
+    OPERATOR_BINARY="$(basedir)/commands/binaries/syndesis"
```
zregvart:
```suggestion
    OPERATOR_BINARY="$HOME/.syndesis/bin/syndesis"
```

Perhaps download outside of the source tree?

tools/bin/commands/util/operator_func.bats
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+load operator_funcs
+
+setup() {
+    touch /tmp/foo
```
zregvart:
```suggestion
    touch /tmp/foo
```

Can we use `mktemp (1)`?

tools/bin/commands/util/operator_func.bats
```diff
@@ -0,0 +1,36 @@
...
+
+@test "invoke check_operator_binary with a nonexistent file returns returns 1" {
+    run check_operator_binary path file
+    [ "$status" -eq 1 ]
+}
+
+@test "check default binary path" {
+    fail=1
+    run default_binary_path
+    if [[ $output == *"syndesis/tools/bin/commands/binaries"* ]]; then
```
zregvart:
```suggestion
    if [[ $output == *"$HOME/.syndesis/bin/syndesis"* ]]; then
```

tools/bin/commands/util/operator_func.bats
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
```
zregvart: What runs this test? Is it a part of the CI?

tools/bin/commands/util/operator_funcs
```diff
@@ -36,3 +36,75 @@ check_operator_requirements() {
 
     echo "OK"
 }
+
+default_binary_path() {
+    local abs_path="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../binaries" >/dev/null 2>&1 && pwd )"
```
zregvart:
```suggestion
    local abs_path="$HOME/.syndesis/bin"
```

tools/bin/commands/util/operator_funcs
```diff
@@ -36,3 +36,75 @@ check_operator_requirements() {
...
+
+    if [[ -z "$abs_path" ]] ; then
```
zregvart:
```suggestion
    if [[ -d "$abs_path" ]] ; then
```

tools/bin/commands/util/operator_funcs
```diff
@@ -36,3 +36,75 @@ check_operator_requirements() {
...
+        fi
+    fi
+    return 1
+}
+
+# download the operator binary
+download_operator_binary() {
+    local location=${1:-"$(default_binary_path)"}
+    local binary=${2:-"syndesis"}
+    local base_url="https://github.com/syndesisio/syndesis/releases/download/1.8.1-20191004/syndesis-1.8.1-20191004-"
```
zregvart: You might want to `curl` it: `curl -s -L -H 'Accept: application/json' https://github.com/syndesisio/syndesis/releases/latest |jq .tag_name`

christophd:
Argh! pure-bot merges fast

tools/bin/commands/crc
christophd: 8GB is the minimum

tools/bin/commands/crc
lgarciaaco: why? it is included in .gitignore

tools/bin/commands/util/operator_func.bats
lgarciaaco: I run it locally to test my stuff. CI could run it if we include bats in it.

tools/bin/commands/util/operator_funcs
lgarciaaco: I cant do that because this is a one time release, the binaries arent part of the release

tools/bin/commands/crc
zregvart: This might be recursive: so it doesn't have to be in .gitignore :) Also might be better not to tie it to a single git checkout, this is different for shell scripts that are cross platform.

tools/bin/commands/util/operator_func.bats
zregvart: Please do, otherwise there's little value in it, no one will remember to run it and then we'll need to fix it after it was broken...

lgarciaaco: I will do a separate PR for this

tools/bin/commands/crc
lgarciaaco: I still don't buy that idea, I don't see how this improves anything. We have the syndesis binary under `tools/bin`, why having this in a different place.

BUT! I will do it

tools/bin/commands/util/operator_funcs
lgarciaaco: Now with releases containing all files, I can improve this bit. I will include it in a different PR

tools/bin/commands/util/operator_funcs
lgarciaaco: it is a matter of taste and clarity ... `-z` catches issues as well as `-d`. Including it in a different PR

PR #6879, opened 2019-10-09
chore: Update Minishift installation scripts to use the new operator binary

christophd:
Update bash scripts to use new syndesis-operator binary


tools/bin/commands/minishift
```diff
@@ -19,8 +19,7 @@ minishift::usage() {
                               'minishift delete && minishift start'.
     --full-reset              Full reset and initialie by
                               'minishift stop && rm -rf ~/.minishift/* && minishift start'
-    --operator-only           Only install the operator but no resource
-                              connected cluster.
+    --operator-only           Only install the operator but no custom resource
```
zregvart:
```suggestion
    --operator-only           Only install the operator but no Syndesis
                              or Camel-K custom resources
```

PR #6881, opened 2019-10-09
fix(connector): default authentication database

squakez:
* Using default connection database instead of "admin"
* Removed useless faulty integration creation in verifier test

fixes ENTESB-11853


squakez:
@syndesisio/backend does anyone mind having a look? it's quite straightforward ;)

PR #6888, opened 2019-10-10
fix(ENTESB-11787): Reconfiguration of SQL query step changes text to dropdown

christophd:
Use dataList component for SQL query so we can revert meta data lookup on server to use the original logic. This way other connectors that rely on "text"/"string" getting automatically rendered as dropdown are not affected.

We will straighten the meta logic and the connector json property types on master as part of https://issues.jboss.org/browse/ENTESB-11969


christophd:
@zregvart thanks! `displayValue` should be used in select/dataList dropdown components and defines what is shown to the user as value there

PR #6893, opened 2019-10-10
fix(ENTESB-11945): roll back to use Json schema draft 4

christophd:
Json schema draft 6 was causing problems in jackson library when parsing unified Json data shapes. So roll back to using jackson compatible draft version.

Started to incorporate a library that supports Json schema draft 6 but this is a really big change as we use the jackson schema quite a lot in server, meta and connectors. so we would need a bigger refactoring. so being a little less hip might do the trick for us, too.


heiko-braun:
Does it impact 1.8.x?

christophd:
yes

PR #6900, opened 2019-10-11
Bump camel version to 2.21.0.fuse-750031

claudio4j:


oscerd:
I promoted the artifacts to EA for you.

zregvart:
I guess we need the `org.apache.kafka:kafka-clients:jar:2.3.0.redhat-00002` artefact published somewhere where it can be fetched.

claudio4j:
There are many transitive dependencies (meta, integrations, templates, support/catalog/) on org.apache.kafka:kafka-clients:jar 2.3.0.redhat-00002. It is better to wait for them to become available on a public repo.

PR #6908, opened 2019-10-14
[ENTESB-11767] Fixing PutItem returning an empty json 

Delawen:
Due to caching issues on the DDB side. 

Plus, some implementation improvements and test extensions I did while fixing this.

### DIFF
--- a/app/connector/aws-ddb/src/main/java/io/syndesis/connector/aws/ddb/customizer/DDBConnectorCustomizerQuery.java
+++ b/app/connector/aws-ddb/src/main/java/io/syndesis/connector/aws/ddb/customizer/DDBConnectorCustomizerQuery.java
@@ -15,6 +15,9 @@
  */
 package io.syndesis.connector.aws.ddb.customizer;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
 import com.amazonaws.services.dynamodbv2.model.AttributeValue;
 import com.google.common.base.Splitter;
 import io.syndesis.connector.aws.ddb.util.Util;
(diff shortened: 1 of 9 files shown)


app/connector/aws-ddb/src/main/java/io/syndesis/connector/aws/ddb/customizer/DDBConnectorCustomizerQuery.java
```diff
@@ -15,30 +15,29 @@
...
 import io.syndesis.connector.aws.ddb.util.Util;
 import io.syndesis.connector.support.util.ConnectorOptions;
 import org.apache.camel.Exchange;
 import org.apache.camel.component.aws.ddb.DdbConstants;
 import org.apache.camel.component.aws.ddb.DdbOperations;
-
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import org.apache.commons.lang3.StringUtils;
```
heiko-braun: I think for simple things like this we shouldn't bring in a new dependency like commons-lang. Keep in mind hat everything you add, needs to be supported and maintained, inclduing commons-lang.

app/connector/aws-ddb/pom.xml
```diff
@@ -92,6 +92,10 @@
       <groupId>io.syndesis.connector</groupId>
       <artifactId>connector-support-util</artifactId>
     </dependency>
+    <dependency>
```
heiko-braun: Please see my comment below.  Adding it to a `pom.xml` seems cheap, but we have to maintain these libraries for a long time. I would recommend removing this, unless there is a really good reason for intriducing commons-lang.

Delawen: Ok!

app/connector/aws-ddb/src/main/java/io/syndesis/connector/aws/ddb/customizer/DDBConnectorCustomizerQuery.java
Delawen: Removed the dependency. It was a transitioned one so it was included anyway, but you are right, adding here explicitly adds maintenance.

zregvart:
@Delawen some conflicts need to be resolved here. @heiko-braun the review was accepted anything else before we merge this?

zregvart:
@Delawen the [ENTESB-11767](https://issues.jboss.org/browse/ENTESB-11767) was marked as a blocker, that means the either it needs to be reclassified as non blocking or we need to backport this (manually) to 1.8.x branch.

Delawen:
I think on Jira we blocked it? @heiko-braun

heiko-braun:
@zregvart We need to look at `Target Release` and the `Priority`. https://issues.jboss.org/browse/ENTESB-11767 is targetting 7.6 and hence doesn't have to be backported.

PR #6914, opened 2019-10-14
fix(ENTESB-11945): Support Json schema draft-4 and draft-6

christophd:
Make Json schema parsing in Syndesis compatible with version draft-3, draft-4 and draft-6. API provider generates Json schema draft-6 version again.

@zregvart we are hip again 😄

Fixes ENESB-11945


zregvart:
This looks a bit too broad and patchy.

I think a better solution would be to use [a mixin](https://github.com/FasterXML/jackson-docs/wiki/JacksonMixInAnnotations) to add a `@JsonAlias` for `$id` to the `id` property of `ObjectSchema`.

christophd:
that's the classic hipness-to-patchy trade-off 😄 I will have a look how that mixin works ...

zregvart:
@christophd you know that JSON Schema draft 06 is from 2017 right? Draft 04 is from 2013.

christophd:
@zregvart must be that 2017 feels like yesterday here in Bavaria!  you guys in Berlin are moving forward way too fast 😄

I have used mixin for `$id` now and it worked fine. Thanks for the proposal

app/common/util/src/main/java/io/syndesis/common/util/Json.java
```diff
@@ -85,6 +97,39 @@ public static ObjectMapper copyObjectMapperConfiguration() {
...
+     * for supporting draft-03, draft-04 and draft-06 level at the same time.
+     *
+     * Auto converts "$id" to "id" property for draft-04 compatibility.
+     *
+     * In case the provided schema specification to read uses draft-04 and draft-06 specific features such as "examples" or a list of "required"
+     * properties as array these information is more or less lost and auto converted to draft-03 compatible defaults. This way we can
+     * read the specification to draft-03 compatible objects and use those.
+     * @return
+     */
+    public static ObjectReader defaultJsonSchemaReader() {
```
zregvart: Naming this method like this it's very likely for someone to use it without realizing it deals with specific issue in parsing JSON schemas (think someone just auto completing `defaultJson...`). In general this functionality is not _common_ so I'd move it somewhere else. Preferably out of the common module.

app/common/util/src/main/java/io/syndesis/common/util/Json.java
```diff
@@ -17,32 +17,44 @@
...
+
     private static final ObjectMapper OBJECT_MAPPER;
     private static final ObjectWriter OBJECT_WRITER;
     private static final ObjectReader OBJECT_READER;
 
     static {
         OBJECT_MAPPER = new ObjectMapper()
             .registerModules(new Jdk8Module())
-            .setPropertyInclusion(JsonInclude.Value.construct(JsonInclude.Include.NON_EMPTY, JsonInclude.Include.NON_EMPTY))
+            .setDefaultPropertyInclusion(JsonInclude.Value.construct(JsonInclude.Include.NON_EMPTY, JsonInclude.Include.NON_EMPTY))
```
zregvart: I don't suppose this is related to the JSON schema version support change?

zregvart:
Cool. Do you think it would be better if we make those changes local to the ObjectMapper used in `server/inspector/` here:

https://github.com/syndesisio/syndesis/blob/3daffed5b8c9b8a781a55dca5fd634b7cc3fe7b1/app/server/inspector/src/main/java/io/syndesis/server/inspector/JsonSchemaInspector.java#L40

app/common/util/src/main/java/io/syndesis/common/util/Json.java
christophd: No, not related to Json schema changes. It's just that `setPropertyInclusion` is deprecated

app/common/util/src/main/java/io/syndesis/common/util/Json.java
christophd: I can change the name to not use `default`. Something like `jsonSchemaReader()`

Though we need to keep it in some common module as it is being used not only by schema inspector but also within split/aggregate step handlers and split/aggregate meta data handlers.

christophd: it is quite important to also use the other default object mapper settings as basis of this Json schema reader instance. We need the enable/disable settings done here: https://github.com/syndesisio/syndesis/blob/ef2d1189b3ac8e183d32f781a8f58a01b419becd/app/common/util/src/main/java/io/syndesis/common/util/Json.java#L55

zregvart: Perhaps a separate class for JSON Schema, like we have one for OpenAPI stuff. I think it should concern us to put more stuff into common, it's starting to smell...

christophd: So where would I put this new class? into a separate module like https://github.com/syndesisio/syndesis/tree/master/app/server/api-generator?

zregvart: If it needs to be in common i.e. a lot of modules depend on it, so that makes it easier, then perhaps a separate class in a separate package like we have for OpenAPI: https://github.com/syndesisio/syndesis/tree/master/app/common/util/src/main/java/io/syndesis/common/util/openapi

christophd: so if this all stays in common module I think we should use https://github.com/syndesisio/syndesis/tree/master/app/common/util/src/main/java/io/syndesis/common/util/json and move everything related to Json utils to this package then. I would like to do this in a separate PR as this is not related to the initial bug that we try to fix here.

zregvart: Well, we're discussing the placement of code added in this PR, so not sure about a separate PR to follow up on that. I'd prefer a separate class as anything regarding JSON schema should be optional when depending on `common-util` and putting it in the Json class makes it mandatory. I'd prefer a separate package as any class (the mixin for example) can be potentially made package private as we would not want to expose it as an API that might get depended on.

app/common/util/src/main/java/io/syndesis/common/util/json/schema/JsonSchemaUtils.java
```diff
@@ -0,0 +1,94 @@
...
+public final class JsonSchemaUtils {
+
+    /** Logger */
+    private static final Logger LOG = LoggerFactory.getLogger(JsonSchemaUtils.class);
+
+    /**
+     * Prevent instantiation for utility class.
+     */
+    private JsonSchemaUtils() {
+        super();
```
zregvart:
```suggestion
        // utility class
```

christophd: I do not know what is better here. I remember just having a comment as method body gives me some other style warnings.

zregvart: I think the style issues is with empty blocks, so adding a comment fixes that, and should be a best practice. Calling Object::super doesn't amount to much. Not hugely important though.

Issue #6918, opened 2019-10-15
Changes required for getting syndesis working on CRC


phantomjinx:
Placeholder to log issues with installation and execution of syndesis on [CRC](https://github.com/code-ready/crc).

Log issues as comments and include `crc version` information.

phantomjinx:
```
$ crc version
version: 1.0.0-rc.0+34371d3
OpenShift version: 4.2.0-0.nightly-2019-09-26-192831 (embedded in binary)
```
- Encountered error with oc version checks in syndesis scripts (corrected in [PR-6919](https://github.com/syndesisio/syndesis/pull/6919))
- Postgresql 9.5 is not available from crc by default so upgrade to 9.6 ([PR-6919](https://github.com/syndesisio/syndesis/pull/6919))
- Found 1st install failed with Postgresql 9.6 reporting "insufficient memory"
  - Limit / requirement set at 512MB
  - Lowered limit to 100MB and installed correctly
  - Logged [issue](https://github.com/code-ready/crc/issues/706) & followed suggestion to increase memory of CRC from 12GB to 16GB which was successful in running all syndesis pods
  - With Postgresql being able to run on only 8GB, it remains to be seen whether this was a one-off or whether such steep memory limits are required for syndesis (shouldn't be!)
- Cannot login with developer account & unable therefore to route to syndesis application

phantomjinx:
Duplicate of #6905

PR #6919, opened 2019-10-15
Chore(GH-6918): Updates required for crc

phantomjinx:
* CRC does not appear to contain postgresql 9.5 and usages results in the
  syndesis-db not installing due to "invalid image"

* Adds extra regexp processing to ensure the CRC version of oc is
  correctly parsed


tools/bin/commands/util/openshift_funcs
```diff
@@ -11,7 +11,7 @@ check_oc_version()
     local minimum=${OC_MIN_VERSION}
     local test=$(oc version | grep "^oc" | tr -d oc\ v | cut -f1 -d "+")
     if [ "$test" = "" ]; then
-        local test=$(oc version | grep 'Client Version' | sed "s/^.*GitVersion:\"v\(.*\)\", GitCommit.*$/\1/")
+        local test=$(oc version | grep 'Client Version' | sed "s/^.*GitVersion:\"v\(.*\)\", GitCommit.*$/\1/" | sed "s/Client Version: v//")
```
zregvart: For oc downloaded from https://mirror.openshift.com/pub/openshift-v4/clients/ocp-dev-preview/latest/:

```
$ ./oc version
Client Version: openshift-clients-4.2.0-201910041700
Kubernetes Version: v1.11.0+d4cacc0
$ ./oc version | grep 'Client Version' | sed "s/^.*GitVersion:\"v\(.*\)\", GitCommit.*$/\1/" | sed "s/Client Version: v//"
Client Version: openshift-clients-4.2.0-201910041700
```

In that case this might work better:

```suggestion
        local test=$(oc version | grep -i 'client version'|sed -e 's/.*: \(openshift-clients-\)\{0,1\}//')
```

Might be a good idea to add tests for this.

Also of note is that there's a yaml/json output in the 4.2 of `oc version`, doesn't seem to be much useful at the moment:

```
$ ./oc version --client -o json
{
  "clientVersion": {
    "major": "",
    "minor": "",
    "gitVersion": "openshift-clients-4.2.0-201910041700",
    "gitCommit": "d5465d715197862b15c8cf05e864d7c3cfea6917",
    "gitTreeState": "clean",
    "buildDate": "2019-10-07T15:16:39Z",
    "goVersion": "go1.12.8",
    "compiler": "gc",
    "platform": "linux/amd64"
  }
}
```

phantomjinx: The suggested change broke when dealing with crc due to the 'v' so taken it to bits and managed to do a single regexp with grep. Added some bats tests to prove it works.
WDYT?
@zregvart

PR #6921, opened 2019-10-15
Release fixes

zregvart:
@asmigala can you take a look at 1254975, not sure if that was intentional or not, we need to change `install/operator/build/conf/config.yaml` for all builds I think, not only snapshot builds.


asmigala:
@zregvart so my thinking was that this shouldn't be done for regular "latest" builds from circleci. I also assumed the non-daily numbered builds would be semi-manual and someone would go and change the values in config.yaml (as is the case for prod builds)

zregvart:
Let me see if I can convince the builds from `master` on CircleCI to set the image tags in `/conf/config.yaml` to `latest`. Would that work?

asmigala:
@zregvart I think so. I personally wouldn't mind if the circleci build also had hardcoded specific image version, but I remember someone having reservations about that, perhaps @lgarciaaco or @chirino

lgarciaaco:
> Let me see if I can convince the builds from `master` on CircleCI to set the image tags in `/conf/config.yaml` to `latest`. Would that work?

But in master the `/conf/config.yaml` file looks totally different, there is not suck a thing like a version for all components anymore.

zregvart:
From what I see the tag should be `latest` by default unless otherwise specified:

https://github.com/syndesisio/syndesis/blob/3d331d202adfdb683c21b0bcb25a4f86f3c8d7b8/tools/bin/commands/build#L179

zregvart:
@lgarciaaco can you elaborate? This is what I see on some of the latest PR built images:

```
...
  components:
    psql:
      image: docker.io/wrouesnel/postgres_exporter:v0.4.7
    s2i:
      image: docker.io/syndesis/syndesis-s2i:latest
    operator:
      image: docker.io/syndesis/syndesis-operator:latest
    ui:
      image: docker.io/syndesis/syndesis-ui:latest
    server:
      image: docker.io/syndesis/syndesis-server:latest
    meta:
      image: docker.io/syndesis/syndesis-meta:latest
    db:
      imageStreamNamespace: openshift
      image: postgresql:9.6
    komodo:
      image: docker.io/teiid/syndesis-dv:latest
    oauth:
      image: quay.io/openshift/origin-oauth-proxy:v4.0.0
    prometheus:
      image: docker.io/prom/prometheus:v2.1.0
    upgrade:
      image: docker.io/syndesis/syndesis-upgrade:latest
...
```

lgarciaaco:
@zregvart my observations is the following:

In 1.8.x, the code https://github.com/syndesisio/syndesis/blob/032c17f4771cc99984585d671cedda4a0c964d88/tools/bin/commands/release#L224-L227

Relies in the following structure of the config file:
https://github.com/syndesisio/syndesis/blob/9685ba1f50b4acd39f53ccca8b4c0f576edd29b7/install/operator/build/conf/config.yaml#L40-L41

Which is not what master has after Hiram changes(as you pointed out in https://github.com/syndesisio/syndesis/pull/6921#issuecomment-542609931). So in master, we don't have a mechanisms to set the tags of the syndesis components to something different than master. I dont think this is important at the moment, but just saying

.circleci/config.yml
```diff
@@ -759,132 +618,8 @@ jobs:
...
-            - operator-build-linux-386
-            - operator-build-linux-amd64
-            - operator-build-darwin-386
-            - operator-build-darwin-amd64
-            - operator-build-windows-386
-            - operator-build-windows-amd64
-            - operator-build-linux-arm
-            - operator-build-linux-arm64
-            - operator-test
-
```
lgarciaaco: But this here doesn't hurt ...

zregvart: It does as it's not using the `syndesis` shell script, the `syndesis` shell script is **the** way of running any build. If there are special cases for the operator those need to be added to the `syndesis` shell script, otherwise we won't be portable accross CI environments and contributors won't be able to run any check performed by CI locally.

lgarciaaco: you are right ... but this very snippet is not the cause the our issues. By removing it now, you leave my without a place where the binaries are uploaded, `syndesis crc --install` and `syndesis minishift --install` won't work, very likely.

Can we find a compromise here? I can work migrating them into the release script asap

zregvart: @lgarciaaco was the author of that in search of a compromise, or a solution that would be in line with the workflows and tools already in place. Could I also be allowed this. As [I've mentioned](https://groups.google.com/forum/#!topic/syndesis/O7Ih9Oi6wZc) in the Syndesis Google group, you're free to add this support to the `syndesis` shell script.

zregvart:
I need this merged to backport and try a new release of 1.8.x, can I get a +1 from someone?

heiko-braun:
@avano @lgarciaaco I guess the last https://github.com/syndesisio/syndesis/pull/6921#issuecomment-542643665 is intended for you

.circleci/config.yml
christophd: @zregvart when these bits break the release remove it. If this is not the case I suggest to keep 'em and to not break the `syndesis crc --install` procedure until we have a better way (aligned with the release scripts).

AFAIU @lgarciaaco already volunteered to consolidate this with the release scripts. After that we can safely remove this.

PR #6928, opened 2019-10-16
Adding a PR template

Delawen:
Which was missing here: https://github.com/syndesisio/syndesis/community


zregvart:
Rebase [doesn't work across forks apparently](https://github.com/syndesisio/syndesis/commit/ca05e03e5a0678ec4e8380aec7f10ae61cceed85/checks?check_suite_id=269200072).

zregvart:
The CI fails because of dependency that was not pinned so it's no longer available, if rebased it should work, so to save some CI cycles I'll merge this now.

PR #6929, opened 2019-10-16
Adding a Code of Conduct

Delawen:
This is a simple one and very straight forward.


zregvart:
The CI fails because of dependency that was not pinned so it's no longer available, if rebased it should work, so to save some CI cycles I'll merge this now.

PR #6938, opened 2019-10-16
fix: fetch updates on the detail page when change events are received

gashcrumb:
for [ENTESB-11405](https://issues.jboss.org/browse/ENTESB-11405)

@claudio4j this is what I was thinking in regards to the server events.

Looking deeper I think I led you astray a bit, `useConnection` returns a `read()` function that can be used to trigger a fetch, so another option would have been to call `read()` after calling the save function.  But this approach should work as well, after going through the oauth dance once there's a change made to the connection the UI should then fetch an updated connection.

I can't seem to fully see this working in development mode though, it seems like the dev proxy isn't receiving the change events for some reason.  Can you test this in your environment?  You'd want to build the console and deploy it to verify I think, open the network tab and you should see a `GET` occur shortly after the ui does that `PUT`.

### DIFF
--- a/app/ui-react/packages/api/src/useConnection.tsx
+++ b/app/ui-react/packages/api/src/useConnection.tsx
@@ -7,6 +7,8 @@ import {
 } from './helpers';
 import { useApiResource } from './useApiResource';
 import { transformConnectorResponse } from './useConnector';
+import { useServerEvents } from './useServerEvents';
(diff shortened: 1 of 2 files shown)


claudio4j:
This change fixes the https://issues.jboss.org/browse/ENTESB-11405 bug.

PR #6939, opened 2019-10-16
ENTESB-11772: Fixes for outstanding errors

phantomjinx:
* Fixes the ENDPOINTS_TEST_SUPPORT_ENABLED var as it has been renamed
  to TEST_SUPPORT_ENABLED

* Sorts the parameters list

### DIFF
--- a/install/operator/pkg/cmd/internal/install/install_forge.go
+++ b/install/operator/pkg/cmd/internal/install/install_forge.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 	"unicode"
 
@@ -175,6 +176,13 @@ func (o *Install) installForge() error {
 }
 
 func convertToParam(name string) string {
+	//
+	// Exceptional situation where the
+	// ENDPOINTS_TEST_SUPPORT_ENABLED var
+	// is renamed to TEST_SUPPORT_ENABLED
+	//
+	name = strings.TrimPrefix(name, "ENDPOINTS_")
+
 	evar := "${" + name + "}"
 	if _, ok := params[evar]; !ok {
 		displayName := strings.ToLower(name)
@@ -430,6 +438,13 @@ func exportTo(resources []unstructured.Unstructured, format string) error {
 
 		paramList = append(paramList, m)
 	}
+
+	sort.Slice(paramList,
+		func(i, j int) bool {
+			f := paramList[i].(map[string]interface{})
+			g := paramList[j].(map[string]interface{})
+			return f["name"].(string) < g["name"].(string)
+		})
 	unstructured.SetNestedSlice(exportTemplate.UnstructuredContent(), paramList, "parameters")
 
 	//


zregvart:
Perhaps a better approach would be not to workaround the issue but fix the issue at the cause. And by that I mean either change this:

https://github.com/syndesisio/syndesis/blob/2fb60c6807ee526aab7d07bda10b48f39c4f7182/install/operator/pkg/syndesis/configuration/configuration.go#L28

To use `ENDPOINTS_TEST_SUPPORT_ENABLED`, or change this:

https://github.com/syndesisio/syndesis/blob/2fb60c6807ee526aab7d07bda10b48f39c4f7182/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/tests/TestSupportHandler.java#L51

To be `test_support.enabled`.

With this the workaround would be applied to any `ENDPOINTS_` variable, which could cause issues in the future.

zregvart:
> @zregvart
> I would agree with you but am not sure I can do that. The aim of the exercise is to cause the generate to produce a template that looks like [this](https://github.com/syndesisio/fuse-online-install/blob/1.7.x/resources/fuse-online-template-oh.yml#L1197) [](/syndesisio/fuse-online-install/blob/HEAD@{2019-10-17T15:18:24Z}/resources/fuse-online-template-oh.yml#L1197). Note. the template contains both the TEST_SUPPORT_ENABLED flag and the ENDPOINTS_TEST_SUPPORT_ENABLED variable.

I'm pretty sure, but I'll allow that I'm wrong, that template parameters are one time thing only, and with that yesterdays template doesn't have to bare any resemblance to the today's template.

If I'm wrong on this, perhaps the replacement could be made explicit by changing the string `ENDPOINTS_TEST_SUPPORT_ENABLED` to `TEST_SUPPORT_ENABLED`, to prevent future issues.

phantomjinx:
> > @zregvart
> > I would agree with you but am not sure I can do that. The aim of the exercise is to cause the generate to produce a template that looks like [this](https://github.com/syndesisio/fuse-online-install/blob/1.7.x/resources/fuse-online-template-oh.yml#L1197) . Note. the template contains both the TEST_SUPPORT_ENABLED flag and the ENDPOINTS_TEST_SUPPORT_ENABLED variable.
> 
> I'm pretty sure, but I'll allow that I'm wrong, that template parameters are one time thing only, and with that yesterdays template doesn't have to bare any resemblance to the today's template.
> 
> If I'm wrong on this, perhaps the replacement could be made explicit by changing the string `ENDPOINTS_TEST_SUPPORT_ENABLED` to `TEST_SUPPORT_ENABLED`, to prevent future issues.

@chirino @lburgazzoli 
Are you able to give any steerage on this, please?

lburgazzoli:
I have a very little knowledge about what the operator does nowadays but in term of naming, I think we should keep "endpoints." for property and "ENDPOINTS_" for env var as this is a quite common convention in spring boot.

phantomjinx:
> I have a very little knowledge about what the operator does nowadays but in term of naming, I think we should keep "endpoints." for property and "ENDPOINTS_" for env var as this is a quite common convention in spring boot.

Sorry, @lburgazzoli, I actually meant to put in @lgarciaaco rather than yourself. However, my thanks for the contribution - much appreciated.

PR #6947, opened 2019-10-17
ENTESB-12038 credentials of a connector attached to a connection are plain text

claudio4j:
https://issues.jboss.org/browse/ENTESB-12038
ENTESB-12005 NullPointerException when saving an oauth connection description
https://issues.jboss.org/browse/ENTESB-12005

- [x] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code

### DIFF
--- a/app/server/credential/src/main/java/io/syndesis/server/credential/CredentialFlowState.java
+++ b/app/server/credential/src/main/java/io/syndesis/server/credential/CredentialFlowState.java
@@ -90,4 +90,8 @@ public interface CredentialFlowState {
     Type type();
 
     CredentialFlowState updateFrom(HttpServletRequest request);
+
+    @JsonIgnore
+    @Value.Derived
+    boolean isApplicable();
 }
--- a/app/server/credential/src/main/java/io/syndesis/server/credential/OAuth1CredentialFlowState.java
+++ b/app/server/credential/src/main/java/io/syndesis/server/credential/OAuth1CredentialFlowState.java
@@ -94,4 +94,9 @@ public interface OAuth1CredentialFlowState extends CredentialFlowState {
 
         return new OAuth1CredentialFlowState.Builder().createFrom(this).verifier(verifier).build();
     }
+
+    @Override
+    default boolean isApplicable() {
+        return getAccessToken() != null;
+    }
 }
--- a/app/server/credential/src/main/java/io/syndesis/server/credential/OAuth2CredentialFlowState.java
+++ b/app/server/credential/src/main/java/io/syndesis/server/credential/OAuth2CredentialFlowState.java
@@ -99,4 +99,8 @@ public interface OAuth2CredentialFlowState extends CredentialFlowState {
         return new OAuth2CredentialFlowState.Builder().createFrom(this).code(code).build();
     }
 
+    @Override
+    default boolean isApplicable() {
+        return getAccessGrant() != null;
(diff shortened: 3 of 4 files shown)


app/server/credential/src/main/java/io/syndesis/server/credential/OAuth1Applicator.java
```diff
@@ -46,8 +46,10 @@ public Connection applyTo(final Connection connection, final OAuthToken token) {
         final Connection.Builder mutableConnection = new Connection.Builder().createFrom(connection)
             .lastUpdated(new Date());
 
-        Applicator.applyProperty(mutableConnection, accessTokenValueProperty, token.getValue());
-        Applicator.applyProperty(mutableConnection, accessTokenSecretProperty, token.getSecret());
+        if (token != null) {
```
zregvart: Can `token` be null at this point? If so perhaps it would be better to fail rather than to ignore the `null` possibility.

zregvart:
Hmm, looking at the description of ENTESB-12005 it seems that in that case an attempt to apply unfinished (without the callback being invoked) credential flow state to connection was attempted.

Perhaps a better approach would be to add `isApplicable` method to `CredentialFlowState` and in the `Credentials::apply` first check if the flow state is appliable, and only if it is perform any further logic.

That `isApplicable` would need to be `@Value.Derived` and `@JsonIgnore`-ed. For OAuth1 it would check if access token was set and for OAuth2 if access grant was set.

claudio4j:
> That `isApplicable` would need to be `@Value.Derived` and `@JsonIgnore`-ed. For OAuth1 it 
> would check if access token was set and for OAuth2 if access grant was set.

It is fixed in the ConnectionHandler.applyCredentialFlowStateTo.

> Should we encrypt properties on store rather than on fetch?

There is one fetch for the Connection and other for the Connector, the fetch(Connector) reads the unencrypted values and set it in the Connection. The ConnectionHandler.create stores only the Connection, not the Connector.

Thanks @zregvart

zregvart:
> There is one fetch for the Connection and other for the Connector, the fetch(Connector) reads the unencrypted values and set it in the Connection. The ConnectionHandler.create stores only the Connection, not the Connector.

Are we also missing encryption of Connector properties? Looks like it's missing here:

https://github.com/syndesisio/syndesis/blob/ad80c2990853cf6fc4a88f7f64557c38fcdd4c24/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/setup/OAuthAppHandler.java#L128

If that's the case, we probably are missing property decryption when used in the OAuth flow.

zregvart:
@claudio4j are we done here, can this be merged? Looks good from my end.

claudio4j:
> @claudio4j are we done here, can this be merged? Looks good from my end.

About the connector encryption, if the credentials are encrypted in OAuthAppHandler, it fails to reconnect, I need more time to investigate this. In this meantime, I think this can be merged and a later fix to the connector can be done.

zregvart:
> About the connector encryption, if the credentials are encrypted in OAuthAppHandler, it fails to reconnect, I need more time to investigate this. In this meantime, I think this can be merged and a later fix to the connector can be done.

We probably need to also decrypt the secrets where we use them.

Issue #6856, opened 2019-10-08
Syndesis-server imagestream not found


KurtStam:
## This is a...

<pre><code>
[ ] Feature request
[*] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

Patching imagestream syndesis-server
Error from server (NotFound): imagestreams.image.openshift.io "syndesis-server" not found
ERROR: Last command exited with 1
~/git/rh/syndesis/syndesis/app

PR #6950, opened 2019-10-18
some fixes to install scripts

lgarciaaco:
Remove `patch_imagestreams_for_initial_image`, we dont have image streams in mastaers and in 1.8.x the operator set the changes back.

Fix race condition when --open is given and the route doesnt exist before `oc get routes syndesis` is called

Adapt `syndesis install` command to use the operator binary

Fixes https://github.com/syndesisio/syndesis/issues/6856


lgarciaaco:
@chirino I don't think this should be failing ...

install/operator/.lib.sh
```diff
@@ -93,15 +93,17 @@ build_operator()
...
-                if [ "$(git diff)" != "" ] ; then
-                    echo ===========================================
-                    echo   Looks like some generated source code
-                    echo   not previously checked in.  See diff:
-                    echo ===========================================
-                    echo
-                    git diff
-                    exit 1
-                fi
+        	    for file in pkg/apis/syndesis/v1alpha1/zz_generated.deepcopy.go pkg/generator/assets_vfsdata.go; do
```
lgarciaaco: @chirino can you have a look at this, I don't understand why we force this now if we generate the resources when we build ...

PR #6965, opened 2019-10-21
BACKPORT - ENTESB-11639: Adds the deploymentconfigs/finalizers permissions to op…

phantomjinx:
…erator

* deploy/syndesis-operator
* role.yml.tmpl
 * Gives operator permission to assign this permission to the
   syndesis-editor role

* 05-syndesis-security-yml.tmpl
 * Gives syndesis-editor role the permission to patch finalizers. Once
   allowed the syndesis-editor is allowed to remove replication-controllers


heiko-braun:
@lgarciaaco Mind a review?

install/syndesis.yml
```diff
@@ -2409,6 +2409,7 @@ objects:
     resources:
     - deploymentconfigs
     - deploymentconfigs/scale
+    - deploymentconfigs/finalizers
```
lgarciaaco: This file can be deleted

tools/bin/commands/util/openshift_funcs
```diff
@@ -475,6 +475,7 @@ rules:
   resources:
   - deploymentconfigs
   - deploymentconfigs/scale
+  - deploymentconfigs/finalizers
```
lgarciaaco: Not needed to add it here. All permissions are applied through the operator. We rather remove this method and fix references to it

install/operator/pkg/generator/assets/install/role.yml.tmpl
```diff
@@ -223,7 +224,7 @@
     - apps.openshift.io
     resources:
     - deploymentconfigs
-    - deploymentconfigs/scale
+    - deploymentconfigs/finalizers
```
lgarciaaco: Don't we need this? If an addon is enabled / disabled, I think the operator has to scale up/down the deploymentconfig. Once the operator takes full control of the upgrade process, this will needed.

tools/bin/commands/util/openshift_funcs
phantomjinx: Would rather avoid doing this big a change in a backport. So will simply revert this file for now.

```
commands/minishift:        add_user_permissions_for_operator "$user" "true"
commands/install:        local result=$(add_user_permissions_for_operator "$user_to_prepare" $(hasflag --cluster))
commands/util/openshift_funcs:add_user_permissions_for_operator() {
install-syndesis:add_user_permissions_for_operator() {
install-syndesis:    result=$(add_user_permissions_for_operator "$user_to_prepare" $(hasflag --cluster))
```

install/syndesis.yml
phantomjinx: Think we need a proper chore / cleanup commit which addresses these extra files. Will simply revert change for the moment.

heiko-braun:
@lgarciaaco Is it good to merge now?

PR #7000, opened 2019-10-23
feat(connector): Adding "strict mode" connector validation

squakez:
As highlighted in a [recent document](https://syndesis.io/docs/connector-schema/) we have in place a connector schema validation running at connector build time. However this is not strict, in the sense that any connector can add arbitrarily any additional configuration, allowing a non homogeneous growth.

With this PR we are limiting the possibility to add any further configuration without changing this file (and possibly without adding any documentation). We should see this file as a contract between UI and Backend, and we should keep documentation aligned.

Thanks @phantomjinx for the initial development of the schema.

### DIFF
--- a/app/connector/support/maven-plugin/src/main/resources/connector-schema.json
+++ b/app/connector/support/maven-plugin/src/main/resources/connector-schema.json
@@ -17,8 +17,14 @@
             "descriptor": {
               "type": "object",
               "properties": {
+                "componentScheme": {
+                  "type": "string"
+                },
                 "configuredProperties": {
                   "type": "object"
+                },
+                "connectorFactory": {
+                  "type": "string"
                 },
                 "connectorCustomizers": {
                   "type": "array",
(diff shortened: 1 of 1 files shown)


zregvart:
Next up

heiko-braun:
Not sure I understand your message @zregvart

zregvart:
> Not sure I understand your message @zregvart

Look at #7000

zregvart:
> Did I win something!? laughing

Delawen:
In muy head this should be two different PR: one for Mongo and another one for the general schema.

squakez:
> If this schema fits with all current connectors I am all for restricting it and making sure we have some strict validation on one of the maven steps.
> 
> Besides that, please, move the mongo changes to a separated PR, in case we have to rollback this at some point.

Thanks @Delawen - actually it's funny but that change was the first error discovered by the new json-schema connector! As it's a property that was set by mistake in a place where it should not be. If I remove that, this build will be failing!

Issue #6976, opened 2019-10-22
Syndesis install using build.sh is not working on master branch


dhirajsb:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[X] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

Using build.sh in app directory, installation is failing on master branch using command:
```
./build.sh install --local -p syndesis -y
```
The operator installs all deployment configs except the DB for some reason. 
```
rh-wfh-dbokde:app dbokde$ oc get dc
NAME                  REVISION   DESIRED   CURRENT   TRIGGERED BY
syndesis-meta         1          1         1         config
syndesis-oauthproxy   1          1         1         config
syndesis-operator     1          1         1         config,image(syndesis-operator:latest)
syndesis-server       1          1         1         config
syndesis-ui           1          1         1         config
```
Operator log shows the following repeated error
```
[...]

PR #7012, opened 2019-10-24
Publish operator binaries from release script

lgarciaaco:
With this PR, when doing a release by running `syndesis release` (not a snapshot release), a [release in github](https://github.com/syndesisio/syndesis/releases) is created and the operator binaries are included.

---
How to use:
```
 syndesis release --release-version tag_name 
```
Replace tag_name with with the real tag, 1.9 ... for instance

---
What produces:
The result is similar to [this release](https://github.com/syndesisio/syndesis/releases/tag/1.9.99999). To the description, `git log` with commits added since the last release will be added, it should look like:
```
f36f417 - Merge branch 'master' into refactor/json-utils
e13475a - fix: warnings in Maven POM (#6994)
56e2884 - refactor: remove unused imports (#6992)
5a91bc7 - refactor: remove needless else (#6991)

```

The release will be created when `syndesis release` is ran without `--snapshot-release`

PR #7026, opened 2019-10-25
Publish operator binaries from release script

lgarciaaco:
With this PR, when doing a release by running `syndesis release` (not a snapshot release), a [release in github](https://github.com/syndesisio/syndesis/releases) is created and the operator binaries are included.

---
How to use:
```
 syndesis release --release-version tag_name 
```
Replace tag_name with with the real tag, 1.9 ... for instance

---
What produces:
The final releases will be similar to like [this](https://github.com/lgarciaaco/syndesis/releases). The body of the release is generated with [gren](https://github.com/github-tools/github-release-notes), further tuning of parameters can be done in the future. 

The release will be created when `syndesis release` is ran without `--snapshot-release`


lgarciaaco:
Original PR: https://github.com/syndesisio/syndesis/pull/7012

tools/bin/commands/release
```diff
@@ -235,6 +240,88 @@ update_image_versions() {
...
+    check_github_access_token
+    check_github_username
+    # check_gren_access_token
+
+    local data="{\
+        \"tag_name\": \"${tag}\", \
+        \"name\": \"${release_name}\" \
+    }"
+
+	local upload_url=$(curl -q -s --fail \
```
zregvart: tab

tools/bin/commands/release
```diff
@@ -235,6 +240,88 @@ update_image_versions() {
...
+    local release_dir=$1
+    local tag=$2
+
+    check_github_access_token
+    check_github_username
+    # check_gren_access_token
+
+    local data="{\
+        \"tag_name\": \"${tag}\", \
+        \"name\": \"${release_name}\" \
```
zregvart: Not sure $release_name is defined anywhere, perhaps

```suggestion
        \"name\": \"${tag}\" \
```

tools/bin/commands/release
```diff
@@ -235,6 +240,88 @@ update_image_versions() {
...
+
+publish_artifacts() {
+    local release_dir=$1
+    local tag=$2
+
+    check_github_access_token
+    check_github_username
+    # check_gren_access_token
+
+    local data="{\
```
zregvart: Would be good to set the `prerelease` to true for snapshot releases.

lgarciaaco:
Fixes https://github.com/syndesisio/syndesis/issues/6976

Issue #6997, opened 2019-10-23
Clarify ways to install Syndesis


christophd:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[x] Documentation issue or request
</code></pre>

## Description
We should clarify the different ways how to install Syndesis. 

Let's collect all known install scripts and scenarios in this issue and see how we can consolidate or remove obsolete things in order to have less maintenance effort.

christophd:
So I am aware of the following scripts dealing with Syndesis installation:

- [tools/bin/syndesis](tools/bin/syndesis)
  - subcommand `minishift --install`
  - subcommand `crc --install`
  - subcommand `install`
- [tools/bin/install-syndesis](tools/bin/install-syndesis)
- `curl -sL https://syndes.is/start`
- [app/build.sh](app/build.sh)
  - subcommand `install --local -p syndesis -y`

Anything else I am missing?

zregvart:
I think all of those scripts end up using the `tools/bin/syndesis` shell script.

christophd:
@zregvart true except `tools/bin/install-syndesis` which is something different.

It says:
```
# ====================================================
# Standalone script for deploying syndesis
#
# This is an alternative to "syndesis install". If you
# have the choice, try to use "syndesis install" as
# it has more features
```

Do we really need all those entry points into the very same binary then? And why do we have something like `syndesis install --local`?

zregvart:
> Do we really need all those entry points into the very same binary then? And why do we have something like `syndesis install --local`?

Nope, we need something to support local development and we need the install for end users via operatorhub.io. Nothing else.

slim-benhammouda:
Hello,
Is there a documentation describing how to install Syndesis in production? All the documentation is about minishift!!
Many thanks,

zregvart:
@slim-benhammouda we're working on a way to install and run on Kubernetes follow #6200 for updates.

SvenC56:
I also wanted to ask for a production installation Guide on OKD/ OpenShift. I'm not 100% sure how to setup a custom domain and enable data-virt. 

Where do I have to specify these settings?
```
# Switch to admin
$ oc login -u system:admin

# Register CRD and grant permissions to "developer"
$ syndesis install --setup --grant developer --cluster

# Switch to account developer
$ oc login -u developer

# Install Syndesis
$ syndesis install
```

[...]

PR #7027, opened 2019-10-25
Do some documentation cleanup for install

lgarciaaco:
We shouldn't use `tools/bin/install-syndesis` for installation, remove it and update documentation.

- Remove `install-syndesis` and old installation methods that aren't used anymore. 
- Remove `--route` section in documentation, it is not a parameter from `syndesis install` anymore.
- Remove `--console` section in documentation, it is not a parameter from `syndesis install` anymore.
- Remove **Quick installation**, it references `install-syndesis`


christophd:
relates to #6997

PR #7033, opened 2019-10-28
feat(connector): mongodb operation validity check

squakez:
Adding proper customizer to check validity of operations

closes ENTESB-12091

### DIFF
--- a/app/connector/mongodb/src/test/java/io/syndesis/connector/mongo/MongoDBConnectorNotSupportedSyndesisOperationTest.java
+++ b/app/connector/mongodb/src/test/java/io/syndesis/connector/mongo/MongoDBConnectorNotSupportedSyndesisOperationTest.java
@@ -18,7 +18,9 @@ package io.syndesis.connector.mongo;
 import io.syndesis.common.model.integration.Step;
 
 import org.apache.camel.CamelExecutionException;
+import org.apache.camel.FailedToCreateRouteException;
 import org.apache.camel.component.mongodb3.CamelMongoDbException;
+import org.junit.Before;
 import org.junit.Test;
 
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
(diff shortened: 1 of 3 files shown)


app/connector/mongodb/src/test/java/io/syndesis/connector/mongo/MongoDBConnectorNotSupportedSyndesisOperationTest.java
```diff
@@ -39,9 +49,7 @@
 
     @Test
     public void mongoTest() {
-        assertThatExceptionOfType(CamelExecutionException.class).isThrownBy(() -> template().sendBody("direct:start","Anything...."))
-            .withMessageContaining("Exception occurred during execution on the exchange")
-            .withCause(new CamelMongoDbException("Invalid payload for aggregate"));
+        assertTrue(true);
```
zregvart: Is this needed?

squakez: It's a workaround to avoid the test failure as there is no `@Test` method annotated:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M3:test (default-test) on project connector-mongodb: No tests were executed!  (Set -DfailIfNoTests=false to ignore this error.) -> [Help 1]
```
To solve I guess I could overwrite the variable above on the plugin alternatively.

zregvart: It's fairly confusing to see a test not asserting anything. When I look at the test I don't know what it tests. I'd rather remove such a test than have it be as unclear as it is.

squakez: What I am testing is the setup of the integration in this case. As this is performed by the abstract class in a `Before` method, I could only test by overriding it and capturing the exception during the initialization. I wish I could do that from a `Test` method instead, not sure if there is any easy way to do that - any hint? In any case it would be for the next iteration as I can see this PR was already merged by the bot.

zregvart: What's preventing you from executing that code from a `@Test` method? Do you need to subclass `MongoDBConnectorTestSupport`? Could you test the `MongoProducerCustomizer::customize` method directly?

PR #7039, opened 2019-10-29
More refactorings and stricter errorprone

zregvart:
Contains:

refactor: remove unused imports (e116d41)

refactor: warning suppression tuning (97d9c9c)

This removes unnecessary @SuppressWarnings and adds additional ones that
are needed to pass the stricter checks.

refactor: test cleanup (0ca4cb7)

refactor: field can be final (4a71242)

refactor: group overridden methods (15bd8be)

refactor: field could be final (4130261)

In this case since Plexus will initialize the field with the default
value when the Mojo is configured there's no need to initialize the
field at construction, that would be cleaner, but it would mark the
field for could be made final check.

chore: stricter errorprone checks (ac71470)

This adds additional errorprone checks by changing their severity from
suggestion to warning. And makes sure that the autofix profile can
perform fixes and that the autofix inherits the configuration from the
default profile.

cc @syndesisio/backend


squakez:
@zregvart could we merge this? I need to rebase with some change there before opening another PR. Thanks!

PR #7042, opened 2019-10-29
fix: added vertical max height and scroll TEIIDTOOLS-784

indraraj:
- Jira Issue https://issues.jboss.org/browse/TEIIDTOOLS-784

- With this pr add the vertical scroll property to ConnectionSchemaList with max-height of 45em.

- UI Look & feel 
Screenshot from 2019-10-29 20-42-55

### DIFF
--- a/app/ui-react/packages/ui/src/Data/Virtualizations/Views/ConnectionSchemaList.css
+++ b/app/ui-react/packages/ui/src/Data/Virtualizations/Views/ConnectionSchemaList.css
@@ -4,3 +4,7 @@
 .connection-schema-list-empty-create-connection-button {
   margin-top: 10px;
 }
+.connection-schema-list {
+    max-height: 45em;
+    overflow-y: auto;
+  }
--- a/app/ui-react/packages/ui/src/Data/Virtualizations/Views/ConnectionSchemaList.css
+++ b/app/ui-react/packages/ui/src/Data/Virtualizations/Views/ConnectionSchemaList.css
@@ -5,6 +5,6 @@
   margin-top: 10px;
 }
 .connection-schema-list {
-    max-height: 45em;
+    max-height: 40em;
     overflow-y: auto;
   }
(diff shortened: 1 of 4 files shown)


mdrillin:
I put a comment on the jira.  We might need to adjust the vertical height - I'm getting double scrollbars

indraraj:
Hey @mcoker can you please suggest some better way to provide height here so that it fits on the screen without creating the outer page section scrollable.

mcoker:
Hey @indraraj my apologies for missing your initial comment. I took a look at this and I think using the existing page stucture, you can do this with the following CSS.

```
.integration-editor-layout__contentInner,
.integration-editor-layout__contentInner > .pf-c-page__main-section,
.integration-editor-layout__contentInner > .pf-c-page__main-section > .pf-c-card,
.views-import-layout-card-body {
  max-height: 100%;
}

.views-import-layout-card-body {
  min-height: 0;
  display: flex;
  flex-direction: column;
}

.views-import-layout-card-body > .list-view-toolbar {
  margin-left: 0;
  margin-right: 0;
}

.views-import-layout-card-body > .view-info-list > .list-group {
  margin-bottom: 0;
}
```

Note, I was only able to test this in the chrome browser. When I try to visit syndesis staging in a browser other than chrome, I get this error message. Can you test this in the other supported browser, or let me know why I'm getting this error in the other browsers?

Screen Shot 2019-11-20 at 1 38 47 PM

cc @dlabaj

gashcrumb:
> When I try to visit syndesis staging in a browser other than chrome, I get this error message.

Is that firefox?  I wonder if you still have the service worker installed, looks like you can go to about:debugging#workers and double-check that there isn't one installed for staging, if there is it can be disabled.  I'd also check the 'Storage' tab and clear the cache and the oauth cookie if it still persists.  Firefox seems to work alright here for me against staging.

mcoker:
@gashcrumb that did it, thanks! The changes look good in Firefox, too.

PR #7051, opened 2019-10-30
feat(connector): mongodb operation split refactory

squakez:
* Added an action for each producer operation
* Adding an abstract customizer to provide generic conversion methods
* Updated unit test with improvements and changes according the refactored code

closes ENTESB-11869


app/connector/mongodb/pom.xml
```diff
@@ -151,6 +154,11 @@
       <artifactId>hamcrest-library</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-core</artifactId>
```
zregvart: I think we're using AssertJ more than hamcrest in the codebase, I was actually thinking of refactoring hamcrest and replace it with AssertJ

app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/MongoAbstractCustomizer.java
```diff
@@ -0,0 +1,73 @@
...
+
+public abstract class MongoAbstractCustomizer implements ComponentProxyCustomizer {
+    private static final Logger LOGGER = LoggerFactory.getLogger(MongoAbstractCustomizer.class);
+
+    /**
+     * Used to convert any result MongoOperation (either {@link DeleteResult} or {@link UpdateResult}
+     * to a {@link Long}
+     * @param exchange
+     */
+    protected void convertMongoResultToLong(Exchange exchange) {
```
zregvart: Methods like these can be made static, that also makes them a bit easier to test and it clearly shows that they're side effect free. Also ErrorProne's new severity fails the build if a method can be made static. Try the `autofix` Maven profile I think it can refactor these for you, you might need to run it a couple of times as it usually fails on the first file.

app/connector/mongodb/src/test/java/io/syndesis/connector/mongo/MongoDBConnectorCappedCollectionConsumerAllOptionsTest.java
```diff
@@ -86,7 +82,7 @@ public void mongoTest() throws Exception {
                 JsonNode jsonNode = MAPPER.readTree(doc.get(0));
                 Assertions.assertThat(jsonNode).isNotNull();
             } catch (IOException ex) {
-                log.error("Test failed because: ",ex);
+                log.error("Test failed because: ", ex);
```
zregvart: Perhaps it would be better if the test fails and the error is seen in the unit test report, then we don't need to read the logs and correlate test to exception.

app/connector/mongodb/src/test/java/io/syndesis/connector/mongo/MongoDBConnectorCappedCollectionConsumerTest.java
```diff
@@ -71,7 +71,7 @@ public void mongoTest() throws Exception {
                 String value = jsonNode.get("someKey").asText();
                 return id <= globalId && "someValue".equals(value);
             } catch (IOException ex) {
-                log.error("Test failed because: ",ex);
+                log.error("Test failed because: ", ex);
```
zregvart: Same as above.

squakez:
> Quite a big PR and I don't have enough knowledge on MongoDB, so just a couple of things that I noticed.

Yep, this is after the feedback you provided about splitting operation ENTESB-11869. It needed a whole refactory, so most of the code appear as new because class name changes.

app/connector/mongodb/pom.xml
squakez: Okey, I can give a look and replace with the assertj instead.

PR #7057, opened 2019-10-31
Grant permissions to syndesis-editor

lgarciaaco:
Apparently syndesis server needs access CRD integrations.camel.apache.org. Grant syndesis-editor Role access to this CRD


nicolaferraro:
I've seen your mail @lgarciaaco . In Camel K mode, the syndesis server creates integrations.camel.apache.org and monitors them, delegating the role of materializing them to Camel K but displaying their status in the console. That's why it needs access to those resources.

PR #7058, opened 2019-11-01
Push tags at the end of the release

lgarciaaco:
Force push the tags after they were created by the github release.

### DIFF
--- a/tools/bin/commands/release
+++ b/tools/bin/commands/release
@@ -56,7 +56,7 @@ get_release_version() {
 
 release::run() {
     source "$(basedir)/commands/util/maven_funcs"
-    source "$(basedir)/commands/util/operator_funcs"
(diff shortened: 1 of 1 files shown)


tools/bin/commands/release
```diff
@@ -56,7 +56,7 @@ get_release_version() {
 
 release::run() {
     source "$(basedir)/commands/util/maven_funcs"
-    source "$(basedir)/commands/util/operator_funcs"
+    # source "$(basedir)/commands/util/operator_funcs"
```
zregvart: Perhaps we remove it rather than comment it out?

```suggestion
```

lgarciaaco: This belongs to the `gren` bits, which I commented out and will add back once `release` works again

PR #7064, opened 2019-11-01
refactor: cleanup exception handling

zregvart:


zregvart:
I think #7074 will fix that PMD issue, it's a false positive in lambda support for that check.

PR #7073, opened 2019-11-04
chore(deps): camel fuse bump

squakez:
* Adding first 7.6 camel fuse development release dependency

### DIFF
--- a/app/extension/bom/pom.xml
+++ b/app/extension/bom/pom.xml
@@ -28,7 +28,7 @@
 
   <properties>
     <spring-boot.version>1.5.13.RELEASE</spring-boot.version>
-    <camel.version>2.21.0.fuse-750033</camel.version>
+    <camel.version>2.21.0.fuse-760006</camel.version>
   </properties>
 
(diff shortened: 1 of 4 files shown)


claudio4j:
@squakez You should update camel version in 
install/operator/pkg/generator/assets_vfsdata.go
install/operator/pkg/generator/assets/addons/camelk/camel-catalog-2.21.0.fuse-750033.yaml.tmpl
and rename this last file to match camel version

squakez:
> @squakez You should update camel version in
> install/operator/build/conf/config.yaml
> install/operator/pkg/generator/assets_vfsdata.go
> install/operator/pkg/generator/assets/addons/camelk/camel-catalog-2.21.0.fuse-750033.yaml.tmpl
> and rename this last file to match camel version

Nice catch @claudio4j. I was not aware of those, I guess they were recently introduced in the operator part. Then I think we must find a unique manifest with the camel version or some procedure to upgrade it, we cannot rely on copying in tens of different files as we're doing now.

PR #7076, opened 2019-11-04
bash script to change camel version

claudio4j:
Those who uses mac, review please.

### DIFF
--- a/tools/misc/syndesisCamelVersion.sh
+++ b/tools/misc/syndesisCamelVersion.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+#
+# Change camel version in syndesis
+#
+# Ussage:
+#   - ./syndesisCamelVersion.sh <camel version>
+#
+# Example:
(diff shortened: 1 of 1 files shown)


squakez:
Great stuff. I suggest to include it in `syndesis` CLI instead, so we won't need to mantain separate scripts and it would be autodocumented there.

tools/misc/syndesisCamelVersion.sh
```diff
@@ -0,0 +1,43 @@
...
+    app/pom.xml \
+    app/extension/bom/pom.xml \
+    app/integration/bom-camel-k/pom.xml \
+    app/integration/bom/pom.xml \
+    install/operator/build/conf/config.yaml \
+    install/operator/pkg/generator/assets_vfsdata.go \
+    ${current_camel_catalog}
+
+git mv ${current_camel_catalog} ${new_camel_catalog}
+
```
squakez: I'd add also a line statement to the user to warn that the version changed locally and must be added, committed and pushed before issuing a PR.

PR #7077, opened 2019-11-04
Fix camel version bump to 2.21.0.fuse-760006

claudio4j:
It fixes the bump merged in https://github.com/syndesisio/syndesis/pull/7073


zregvart:
I've pushed the regenerated assets_vfsdata.go to make the CI pass.

PR #7082, opened 2019-11-05
fix: thrift dependency issues (again)

zregvart:
I'm not sure if my upgraded go version (1.13.3) is to blame or something
else changed around the thrift dependency or in opencensus that pulls it
in. I was getting:

```
go: github.com/apache/thrift@v0.12.0 used for two different module paths (git.apache.org/thrift.git and github.com/apache/thrift)
```

Doing a replace by an older version (v0.0.0-20180902110319-2566ecd5d999)
instead of the version that's supposed to be used removed the issue.

I've run `go mod tidy` and that did the rest of the changes.

My concern is that this will pull in an older version of thrift.


dhirajsb:
considering that everyone and their mother is using this fix in go 1.13 supporting projects out there (like tektoncd), it should be a fairly safe bet.

dhirajsb:
See https://github.com/tektoncd/cli/commit/d3dbc043ec24b12d32cd52c9dfcc1b6aca56ef00

PR #7087, opened 2019-11-05
[TEIIDTOOLS-752] Enabling teiid-syndesis to be built using build.sh -m teiid-syndesis

vhalbert:


vhalbert:
when cloning use --recurse-submodules  or if the teiid-syndesis firectory is empty, run:
-  git submodule init
-  git submodule update

tools/bin/commands/build
```diff
@@ -269,6 +269,7 @@ MODULES=(
   "server:common extension integration connector"
   "s2i:common extension integration connector server"
   "test:common extension integration connector server s2i"
+  "teiid-syndesis"
```
rareddy: I would just call these "dv"

tools/bin/commands/build
```diff
@@ -281,6 +282,7 @@ MODULE_PATHS=(
   "meta:app/meta/"
   "server:app/server/"
   "s2i:app/s2i/"
+  "teiid-syndesis:app/teiid-syndesis/"
```
rareddy: call as "dv", teiid-syndesis is only repo name, when "dv" should be used else where to represent anything else.

app/pom.xml
```diff
@@ -615,6 +615,18 @@
       </properties>
     </profile>
 
+    <profile>
+      <id>teiid-syndesis</id>
```
rareddy: call it as  "dv"

tools/bin/commands/build
vhalbert: So your going to rename teiid-syndesis to dv when u port it into syndesis?

rareddy: It is just repo name we do not need to port anything. if you call as "dv" it will match code references during the build to the runtime names, so that it will be consistent across. IDK, but you could simply called it "dv" sub-module too, but reference `teiid-syndesis` repo.

vhalbert: the goal, i thought, was to move teiid-syndesis into syndesis.  After talking with zoran, he indicated that it should land under the app directory.    Unless you plan on changing teiid-syndesis to something else, the scripts will specifically reference this folder name.   I was staying with teiid-syndesis so that a different name doesn't have to remembered or changed later.

vhalbert: i'll rename to dv.

PR #7092, opened 2019-11-06
fix: maintain list item selections on back/next TEIIDTOOLS-679

indraraj:
- Jira Issue https://issues.jboss.org/browse/TEIIDTOOLS-679

- With this PR here is the behavior of Import Datasource Wizard
When the user selects some source on page 1 and moves to page 2 and selects few connections their
  --If the user moves back to page 1, the selected source will be maintained selected there.
  --If the user does not change selection on page 1 and move back to page 2 then the connections selected there will be maintained.
  --If the user changes the selection on page 1 then connection selected on page 2 will be reset

- Lifted the selected source state and selected connection state form SelectConnectionPage and SelecViewPage respectively to ViewsImportApp.

### DIFF
--- a/app/ui-react/syndesis/src/modules/data/ViewsImportApp.tsx
+++ b/app/ui-react/syndesis/src/modules/data/ViewsImportApp.tsx
@@ -30,29 +30,25 @@ export const ViewsImportApp: React.FunctionComponent = () => {
   };
 
   const [selectedViews, setSelectedViews] = React.useState<ViewInfo[]>([]);
-  const [hasSelectedViews, setHasSelectedViews] = React.useState(false);
 
   const handleAddView = async (view: ViewInfo) => {
-    const currentViews = selectedViews;
+    const currentViews = selectedViews.slice();
     currentViews.push(view);
     setSelectedViews(currentViews);
-    setHasSelectedViews(currentViews.length > 0);
   };
 
   const handleRemoveView = async (viewName: string) => {
-    const currentViews = selectedViews;
+    const currentViews = selectedViews.slice();
     const index = currentViews.findIndex(view => view.viewName === viewName);
 
     if (index !== -1) {
       currentViews.splice(index, 1);
     }
     setSelectedViews(currentViews);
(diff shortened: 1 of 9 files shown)


app/ui-react/syndesis/src/modules/data/ViewsImportApp.tsx
```diff
@@ -18,6 +18,43 @@ export const ViewsImportApp: React.FunctionComponent = () => {
...
+  };
+
+  const [selectedViews, setSelectedViews] = React.useState<ViewInfo[]>([]);
+  const [hasSelectedViews, setHasSelectedViews] = React.useState(false);
+
+  const handleAddView = async (view: ViewInfo) => {
+    const currentViews = selectedViews;
+    currentViews.push(view);
+    setSelectedViews(currentViews);
+    setHasSelectedViews(currentViews.length > 0);
```
elvisisking: Couldn't you just do this: `setHasSelectedViews(true);` Also was wondering what you thought about this: Since `setHasSelectedViews` is *always* called after `setSelectedViews` is called, we could move the `setHasSelectedViews` call into a `useEffect` that runs whenever `selectedViews` changes.

indraraj: Yes, Dan we could have used `setHasSelectedViews(true);` here. I may have missed this because I just lifted the state management(code) from `SelectViewsPage` to `ViewsImportApp`.
And while checking your second suggestion about using `useEffect` I found that we are incorrectly updating the React State in the case state being an array. If handled correctly we may not even require `hasSelectedViews` state. I have done some changes locally to verify this. Maybe we can discuss it before I push these changes.

indraraj:
- Updated the `selectedViews` state handling and removed the `hasSelectedViews` state. Changes in `ViewsImportApp` and `SelectViewspage`.

- Updated the `selectedSchemaNodes` state handling and removed the `selectedNodesCount` state. Changes in `ViewCreateApp` and `SelectSourcespage`.

PR #7102, opened 2019-11-07
chore: Improve docs about creating a JDBC driver library extension

christophd:
- Refer to using the Maven wrapper in docs to make sure that the build is successful. It is required to use the exact same Maven version otherwise build of extension may fail.
- Add some information how to add the correct system path when custom driver file has been copied.

### DIFF
--- a/doc/modules/customizing/p_creating-jdbc-driver-library-extensions.adoc
+++ b/doc/modules/customizing/p_creating-jdbc-driver-library-extensions.adoc
@@ -5,61 +5,63 @@
 = Creating JDBC driver library extensions
 
 To connect to a SQL database other than Apache Derby, MySQL, and
-PostgreSQL, you can create a library extension that wraps a JDBC driver for 
-the database you want to connect to. 
-After uploading this extension to {prodname}, the {prodname}-provided 
-*Database* connector can access the driver to validate and create connections 
-to the proprietary database. You do not create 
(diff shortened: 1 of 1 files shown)


christophd:
@TovaCohen FYI

doc/modules/customizing/p_creating-jdbc-driver-library-extensions.adoc
```diff
@@ -5,61 +5,63 @@
...
+... Update the value of the `Name` element to be a name that you choose
+for this extension.
 ... Update the value of the `Description` element to provide helpful
 information about this extension.
+... If you have copied the driver into `syndesis-library-jdbc-driver/lib`
+ensure that the `systemPath` in `pom.xml` points to that driver file. Optionally change
+the `groupId`, `artifactId` and `version` to reflect proper values according to the driver.
 ... If the driver is in a Maven repository, ensure that a reference to
-that Maven repository is in the `pom.xml` file. 
+that Maven dependency is in the `pom.xml` file.
```
KurtStam: Actually 'repository' is correct, this is referring to i.e. Oracle which has its drivers in their Oracle repo. So you would need to add that repository for the dependency to work.

christophd: Oracle drivers also have been officially moved to Maven central lately. There could be other drivers or versions that might require another repository setting though.

PR #7109, opened 2019-11-08
fix: selection allowed for source of active status even for loading

indraraj:
- Jira Issue https://issues.jboss.org/browse/TEIIDTOOLS-820

- with this PR here is the behaviour in Import Datasource wizard
  -- Allowed selection of any ACTIVE connection, even if loading.
  --Removed spinner for ACTIVE connections that are 'loading' (refreshing)

### DIFF
--- a/app/ui-react/packages/ui/src/Data/DvConnection/DvConnectionCard.tsx
+++ b/app/ui-react/packages/ui/src/Data/DvConnection/DvConnectionCard.tsx
@@ -47,12 +47,8 @@ export const DvConnectionCard: React.FunctionComponent<
       onClick={doToggleSelected(props.name)}
     >
       <CardHeader>
-        {props.loading ? (
-          props.dvStatus !== ConnectionStatus.ACTIVE ? (
-            <Spinner loading={true} inline={true} />
-          ) : (
-            <></>
-          )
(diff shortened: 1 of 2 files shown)


app/ui-react/packages/ui/src/Data/DvConnection/DvConnectionCard.tsx
```diff
@@ -47,7 +47,15 @@ export const DvConnectionCard: React.FunctionComponent<
       onClick={doToggleSelected(props.name)}
     >
       <CardHeader>
-        {props.loading ? <Spinner loading={true} inline={true} /> : <></>}
```
mdrillin: your change would work fine, however I think something like this would be more readable
`{ (props.loading && props.dvStatus !== ConnectionStatus.ACTIVE) ? <Spinner loading={true} inline={true} /> : <></>}`
wdyt?

indraraj: yes, it can be easy to grasp at first glance.

indraraj: updated same in PR

PR #7110, opened 2019-11-08
ENTESB-12070: add knative as addon

nicolaferraro:
fix ENTESB-12070


app/connector/knative/src/main/java/io/syndesis/connector/knative/KnativeComponentProxyFactory.java
```diff
@@ -39,6 +39,13 @@
     @Override
     public ComponentProxyComponent newInstance(String componentId, String componentScheme) {
         return new ComponentProxyComponent(componentId, componentScheme, createCatalog()) {
+
+            {
+                setBeforeConsumer(exchange -> {
+                    exchange.getMessage().removeHeader("CamelHttpUri");
```
zregvart: Would [SyndesisHeaderStrategy](https://github.com/syndesisio/syndesis/blob/master/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/util/SyndesisHeaderStrategy.java) help with this?

PR #7113, opened 2019-11-08
fix(connector/rest-swagger): support multiple values for query parame…

zregvart:
…ters

This adds support for setting multiple values for query parameters in
the custom API client connectors.

For example query parameters that are of array type and for which
multiple values are given now appear multiple times as query parameter
each time with one of the values.

For example:

    /v2/pet/findByStatus?status=available&status=pending

For `status` having a values `available` and `pending`.

Ref https://issues.jboss.org/browse/ENTESB-12139


zregvart:
@squakez yeah, we need to upgrade to a newer version of Fuse Camel fork, that's why it's a draft PR.

zregvart:
We could upgrade to `2.21.0.fuse-760011`, we also need to make sure that the version is pinned.

PR #7114, opened 2019-11-10
Operator refactor

lgarciaaco:
A bit of a refactor around the custom resource and how configuration is defined/built, also some cleanup.

### Remove legacy code:
 - Remove `install standalone`. It was meant to replace template installation but it never did.
 - Remove `install forge`. It is a temporally solution for eval cluster. 
 - Commented out some bits from upgrade, a preliminary work for the upgrade rework.
 - Remove syndesis-extra-permissions role which is not used but declared.
 - Remove some dead code.

### Consolidate properties, ENV, context, template, configs and hardheaded defaults into configuration:
We had a bunch of "configurations" serving the same purpose but defined in different places and in different ways. Consolidate them into configuration pkg. All defaults should be defined in the config file. Delete template pkg. Move all logic on how configurations are generated to configuration pkg.

### CR refactor
The custom resource is our public API to customers. Remove fields from custom resource that aren't meant to be changed. Provide a way to specify some fields via environment variables, ie docker images so QE can replace them and test.


heiko-braun:
@lgarciaaco I would keep install forge until we have migrated the eval cluster

install/operator/deploy/crds/syndesis_v1alpha1_syndesis_crd.yaml
```diff
@@ -30,255 +30,115 @@ spec:
...
               type: object
             components:
               description: Components is used to configure all the core components
                 of Syndesis
               properties:
-                db:
+                database:
                   properties:
                     database:
                       type: string
```
phantomjinx: Doesn't changing db make this attribute `components.properties.database.properties.database.type` which is a little repeatative?

install/operator/pkg/generator/assets/infrastructure/02-syndesis-secrets.yml.tmpl
```diff
@@ -19,26 +19,24 @@
...
-      {{.Env.OPENSHIFT_OAUTH_CLIENT_SECRET}}
-    OPENSHIFT_OAUTH_CLIENT_SECRET: |-
-      {{.Env.OPENSHIFT_OAUTH_CLIENT_SECRET}}
+      {{.OpenShiftOauthClientSecret}}
     POSTGRESQL_PASSWORD: |-
-      {{.Env.POSTGRESQL_PASSWORD}}
+      {{.Syndesis.Components.Database.Password}}
     POSTGRESQL_SAMPLEDB_PASSWORD: |-
-      {{.Env.POSTGRESQL_SAMPLEDB_PASSWORD}}
+      {{.Syndesis.Components.Database.SampledbPassword}}
```
phantomjinx: Any special treatment of these attributes due to them being 'sensitive'? I originally left them in ENV due to that reason.

install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -1,313 +1,430 @@
...
-	EnvSyndesisRegistry:               ConfigSpec{Value: "docker.io", Description: "Registry from where to fetch Syndesis images"},
-	EnvDemoDataEnabled:                ConfigSpec{Value: "false", Required: true, Description: "Enables starting up with demo data"},
-	EnvMaxIntegrationsPerUser:         ConfigSpec{Value: "1", Required: true, Description: "Maximum number of integrations single user can create"},
-	EnvIntegrationStateCheckInterval:  ConfigSpec{Value: "60", Required: true, Description: "Interval for checking the state of the integrations"},
-	EnvSarNamespace:                   ConfigSpec{Required: true, Description: "The user needs to have permissions to at least get a list of pods in the given project in order to be granted access to the Syndesis installation"},
-	EnvDvMemoryLimit:              ConfigSpec{Value: "1024Mi", Required: true, Description: "Maximum amount of memory the data virtualization service might use"},
-	EnvDatavirtEnabled:                ConfigSpec{Value: "0", Required: true, Description: "Set to 0 to disable data virtualization, set to 1 to enable data virtualization"},
```
phantomjinx: Seems a bit of a shame to lose all of these defaults & descriptions. Can they not find a new home in the refactored config?

install/operator/deploy/crds/syndesis_v1alpha1_syndesis_crd.yaml
lgarciaaco: We can change `database.database` to `database.schema` to refer to the name of the database. I didn't like the `db` abbreviation but if other ppl find it better I can set it back to what it was .

install/operator/pkg/generator/assets/infrastructure/02-syndesis-secrets.yml.tmpl
lgarciaaco: They receive the same treatment. Sensitive fields are randomly generated and saved in  `syndesis-server-secret` secret. 

https://github.com/syndesisio/syndesis/pull/7114/files#diff-8f63b1158cd0c8188207de68b0013e69R221-R224

This is only changed the place where those fields are now stored after being read from the secret ... or being generated.

install/operator/pkg/syndesis/configuration/configuration.go
lgarciaaco: All defaults are now in `conf/config.yaml`, same values I didn't change them. I am trying to bring some sanity and standardization to how we define configurations. Configurations defaults are now set in `config/config.yaml` only, and overwritten by: environment variables or custom resource.

Some configuration must be generated or figured out: passwords, secrets, route, console route, tokens. Those aren't overwritten.

Back to the OP, we lost comments but I had to revisit each of those configurations and the comment didn't help me much. If you want them back I can add the comments to the `Config` struct: https://github.com/syndesisio/syndesis/pull/7114/files#diff-8f63b1158cd0c8188207de68b0013e69R49

install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
```diff
@@ -11,41 +26,14 @@ import (
 // SyndesisSpec defines the desired state of Syndesis
 // +k8s:openapi-gen=true
 type SyndesisSpec struct {
-	// Set RouteHostname to the hostname of the exposed syndesis service.  Typically the operator can automatically
-	// determine this by looking at the result of the Route object it creates.
-	RouteHostname      string `json:"routeHostname,omitempty"`
```
zregvart: Looks like you didn't run `go fmt`

install/operator/deploy/crds/syndesis_v1alpha1_syndesis_crd.yaml
phantomjinx: database.schema works for me - just avoids repetition. Better than 'db'.

install/operator/pkg/syndesis/configuration/configuration.go
phantomjinx: Sure. I appreciate the standardisation. Having things in 1 place rather than 3 is SO much less confusing. If we can just retain the comments somewhere that'd be fine but if they really are not going to ever be used again then am not standing in the way of dropping them.

install/operator/deploy/crds/syndesis_v1alpha1_syndesis_crd.yaml
zregvart: Database schema is not the same as the database name. The property should reflect what's being configured. Is it schema or database name?

lgarciaaco: I will change it to `Name`. Final result will be: `Components.Database.Name`, `Components.Database.Password` and so on.

install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
lgarciaaco: Why? I do have it  enabled as File Watcher ... :thinking:

install/operator/pkg/syndesis/configuration/configuration.go
lgarciaaco: https://github.com/syndesisio/syndesis/pull/7114/commits/a972c087ea3e6e929d38e84700ac23a5bf1829ab is taking care of comments

install/operator/deploy/crds/syndesis_v1alpha1_syndesis_crd.yaml
lgarciaaco: https://github.com/syndesisio/syndesis/pull/7114/commits/a972c087ea3e6e929d38e84700ac23a5bf1829ab is taking care of naming

lgarciaaco:
@asmigala can you think of any configurations we can move from custom resource to environment variables, other than the images?

asmigala:
@lgarciaaco maybe test support, not sure whether that would work for us, cc @avano

lgarciaaco:
> We have everything in the URL already the `/syndesis` is the name of the database. Did we loose the ability to configure the schema (do we really need that?) or did we duplicate the information? Sorry to harp on this, not much I can say about the rest of the PR...

No biggies, it is a valid question. `Database.Name` is used to setup `Components.DV` which doesn't use the URL

PR #7117, opened 2019-11-11
chore(install): remove `--nodev` from `syndesis install`

zregvart:
This removes the `--nodev` option from `syndesis install`, development workflows are supported via `syndesis minishift` or `syndesis crc`. For installing to arbitrary cluster one needs to perform `syndesis install` followed by editing of the Syndesis custom resource to enable the dev mode or use the `syndesis-operator` binary with `--dev` option.

Updates the documentation.


lgarciaaco:
Why is this?

zregvart:
> Why is this?

Currently `syndesis install` doesn't install Syndesis. It does the same thing `syndesis minishift` or `syndesis crc` do.

PR #7123, opened 2019-11-12
ENTESB-12000: Allows tagged images from quay.io pr testing registry

phantomjinx:
* Loosens the checkTags function slightly to allow a x-chksum format of
  image tag as well as the original version tag.

### DIFF
--- a/install/operator/pkg/syndesis/action/install.go
+++ b/install/operator/pkg/syndesis/action/install.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


phantomjinx:
This has to be reworked entirely due to refactoring of the operator.

heiko-braun:
@phantomjinx before you do that, can you prepare a version for 1.8, i.e in another PR? It might be that we treat 1.8.x as the baseline for 7.6 and it would be good to be prepared

lgarciaaco:
@phantomjinx the checkTag function is not required anymore. Actually we didn't have tags at the moment I refactored so I deleted all of it.

phantomjinx:
No longer applicable.

PR #7138, opened 2019-11-14
[ENTESB-12157] Adding output parameters on DELETE statements too

Delawen:
(for the "RETURNING" case)


Delawen:
Adding test to ignore by default because it needs docker.

KurtStam:
@Delawen is there an GH issue for this? I'd like to understand the use cases.

asmigala:
@KurtStam 
https://issues.jboss.org/browse/ENTESB-12157

PR #7142, opened 2019-11-14
ENTESB-12268 Integration uptime doesn't contain any data

claudio4j:
https://issues.jboss.org/browse/ENTESB-12268

### DIFF
--- a/app/server/metrics/prometheus/src/main/java/io/syndesis/server/metrics/prometheus/PrometheusMetricsProviderImpl.java
+++ b/app/server/metrics/prometheus/src/main/java/io/syndesis/server/metrics/prometheus/PrometheusMetricsProviderImpl.java
@@ -15,8 +15,6 @@
  */
 package io.syndesis.server.metrics.prometheus;
 
-import javax.annotation.PostConstruct;
-import javax.annotation.PreDestroy;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.Comparator;
@@ -26,6 +24,8 @@ import java.util.Optional;
 import java.util.function.BinaryOperator;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
 
 import io.fabric8.kubernetes.api.model.LabelSelector;
 import io.fabric8.kubernetes.api.model.Pod;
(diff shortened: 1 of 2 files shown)


zregvart:
Perhaps also related to #7143, not sure if we would like to provide the time in seconds and the UI does the multiply by 1000 or the backend provides the value in miliseconds and the UI takes the value verbatim.

Any thoughts @gashcrumb?

gashcrumb:
Consistency would be better, if the UI can just use a value without having
to modify it first, that'd be better IMO

app/ui-react/packages/ui/src/Integration/Metrics/IntegrationDetailMetrics.tsx
```diff
@@ -35,7 +35,7 @@ export class IntegrationDetailMetrics extends React.Component<
   > {
   public render() {
     const okMessagesCount = this.props.messages! - this.props.errors!;
-    const startAsDate = new Date(this.props.start!);
+    const startAsDate = new Date(this.props.start! * 1000);
```
zregvart: Perhaps we can do this on the server (checkout the comments on PR)

app/ui-react/syndesis/src/modules/integrations/pages/detail/MetricsPage.tsx
```diff
@@ -105,7 +105,7 @@ export class MetricsPage extends React.Component {
                                                     parseInt(
                                                       metricsData.lastProcessed,
                                                       10
-                                                    )
+                                                    ) * 1000
```
zregvart: Let's do this on the server, so UI doesn't have to do `* 1000`

gashcrumb:
The other option btw is to maybe as a convention include the time unit as part of the property name, that way anybody coming along and looking can understand what's going on without having to stare at the value or see it convert to some date in the 70s :-)

claudio4j:
I found Zoran suggestion better, do it on server. I am testing this option and will push the commit today. Thanks for the review.

app/ui-react/packages/ui/src/Integration/Metrics/IntegrationDetailMetrics.tsx
claudio4j: Done, thanks.

app/ui-react/syndesis/src/modules/integrations/pages/detail/MetricsPage.tsx
claudio4j: Done.

claudio4j:
> Perhaps also related to #7143

This fix the above issue.

app/ui-react/packages/ui/src/Integration/Metrics/IntegrationDetailMetrics.tsx
claudio4j: Done.

Issue #6948, opened 2019-10-18
Jackson version incompatibility


zregvart:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

Seems that the change in #6914 triggered a [nightly build failure](https://circleci.com/gh/syndesisio/syndesis/157090). We most likely are using two different versions of Jackson between Syndesis and integrations.

Perhaps it would be a good idea to remove dependency on syndesis/common-util from the integration runtime/connectors.

```
01:16:58.649 [tc-okhttp-stream-788540094] INFO  INTEGRATION_RUNTIME_CONTAINER - STDOUT: Caused by: java.lang.NoSuchMethodError: com.fasterxml.jackson.databind.ObjectMapper.setDefaultPropertyInclusion(Lcom/fasterxml/jackson/annotation/JsonInclude$Value;)Lcom/fasterxml/jackson/databind/ObjectMapper;
01:16:58.649 [tc-okhttp-stream-788540094] INFO  INTEGRATION_RUNTIME_CONTAINER - STDOUT:     at io.syndesis.common.util.Json.<clinit> (Json.java:45)
01:16:58.650 [tc-okhttp-stream-788540094] INFO  INTEGRATION_RUNTIME_CONTAINER - STDOUT:     at io.syndesis.integration.runtime.IntegrationRouteBuilder.loadIntegration (IntegrationRouteBuilder.java:115)
01:16:58.650 [tc-okhttp-stream-788540094] INFO  INTEGRATION_RUNTIME_CONTAINER - STDOUT:     at io.syndesis.integration.runtime.IntegrationRouteBuilder.configure (IntegrationRouteBuilder.java:140)
01:16:58.651 [tc-okhttp-stream-788540094] INFO  INTEGRATION_RUNTIME_CONTAINER - STDOUT:     at org.apache.camel.builder.RouteBuilder.checkInitialized (RouteBuilder.java:462)
01:16:58.651 [tc-okhttp-stream-788540094] INFO  INTEGRATION_RUNTIME_CONTAINER - STDOUT:     at org.apache.camel.builder.RouteBuilder.configureRoutes (RouteBuilder.java:402)
[...]

PR #7147, opened 2019-11-15
feat(connector/mongodb): operations datashapes

squakez:
* Providing each operation with its own datashape
* Parsing dynamic parameters in filters and replacing during runtime with values coming from input source
* Adapting unit test cases properly

Close ENTESB-11589


k6-uk:
It's cool

zregvart:
License check is failing on CI:

```
[WARNING] Missing header in: /workspace/app/connector/mongodb/src/test/java/io/syndesis/connector/mongo/meta/DataShapeConnectorTest.java
[WARNING] Missing header in: /workspace/app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/meta/FilterUtil.java
```

@squakez can you run `./mvnw -N license:format` from `app/`

app/connector/mongodb/pom.xml
```diff
@@ -83,6 +83,10 @@
       <groupId>io.syndesis.connector</groupId>
       <artifactId>connector-support-util</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.syndesis.common</groupId>
+      <artifactId>common-util</artifactId>
```
zregvart: We should not depend on Syndesis modules in the connectors, that is everything but the integration-runtime. I know we have some connectors that transgress on this but we should try not to introduce new dependencies like this.

This prevents us from using different dependencies at integration runtime and in the Syndesis server.

app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/meta/FilterUtil.java
```diff
@@ -0,0 +1,91 @@
...
+     *
+     * @param filter an expression containing :#variable parameters
+     * @param jsonMessage a message formatted as JSON
+     * @return
+     */
+    public static String merge(String filter, String jsonMessage) {
+        LOGGER.debug("Filter configured by user {}", filter);
+        LOGGER.debug("Input body {}", jsonMessage);
+        String mergedFilter = filter;
+        if (mergedFilter != null) {
```
zregvart: I'd prefer early exit from method like:
```suggestion
        if (mergedFilter == null) {
        	LOGGER.debug("Skipping merge as filter was null");
        	return null;
        }
```

app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/meta/MongoDBMetadataRetrieval.java
```diff
@@ -63,33 +69,70 @@ protected SyndesisMetadata adapt(CamelContext context, String componentId, Strin
...
         } else if (actionId.contains("consumer")) {
-            return SyndesisMetadata.outOnly(schema);
+            return SyndesisMetadata.of(none("N/A"), schema);
         } else {
             throw new IllegalArgumentException(String.format("Could not find any dynamic metadata adaptation for action %s", actionId));
         }
     }
 
-    private static DataShape any(String name) {
+    @SuppressWarnings("MethodCanBeStatic")
```
zregvart: Why not make the method static?

app/connector/mongodb/src/main/resources/META-INF/syndesis/connector/mongodb3.json
```diff
@@ -13,10 +13,9 @@
           "io.syndesis.connector.mongo.MongoProducerFindCustomizer"
         ],
         "inputDataShape": {
-          "name": "Criteria",
-          "description": "Text criteria",
-          "kind": "java",
-          "type": "java.lang.String"
+          "name": "Filter parameters",
+          "description": "Flat list of parameters contained in the filter provided by user",
```
zregvart: This is a bit confusing for me but I'm not a Mongo expert, would a Mongo user understand what's requested of them?

app/connector/mongodb/src/main/resources/META-INF/syndesis/connector/mongodb3.json
```diff
@@ -156,10 +184,9 @@
           "io.syndesis.connector.mongo.MongoProducerUpdateCustomizer"
         ],
         "inputDataShape": {
-          "name": "Criteria",
-          "description": "Text criteria",
-          "kind": "java",
-          "type": "java.lang.String"
+          "name": "Filter parameters",
+          "description": "Flat list of parameters contained in the filter provided by user",
```
zregvart: This will be presented in the user interface; I don't think we need to tell the user that this is _provided by user_. Perhaps discuss with @TovaCohen for a better way to phrase these.

app/connector/mongodb/src/main/resources/META-INF/syndesis/connector/mongodb3.json
```diff
@@ -183,14 +210,30 @@
...
+              "filter": {
+                "deprecated": false,
+                "displayName": "Update expression",
+                "group": "common",
+                "javaType": "java.lang.String",
+                "kind": "path",
+                "labelHint": "Expression used as a criteria to update a set of documents. Use :#variable to set variables eventually mapped to any incoming input.",
+                "placeholder": "{\"color\": :#color}",
+                "order": "2",
+                "required": true,
```
zregvart: If we're adding a required property, we probably need a migration to set `configuredProperty` to integration's existing action steps.

app/connector/mongodb/pom.xml
squakez: Actually 20 over 33 connectors are using this utility dependency. It's handy for [metadata handling](https://github.com/syndesisio/syndesis/blob/master/app/common/util/src/main/java/io/syndesis/common/util/json/JsonUtils.java). I will rework that part, but we should consider the effort of refactoring the rest.

app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/meta/MongoDBMetadataRetrieval.java
squakez: I feel that is semantically an extension of `buildDatashape` method, and it's there to have a more concise coding.

app/connector/mongodb/pom.xml
zregvart: Yeah, we need to refactor that, let's not make it more difficult on ourselves and add more things we need to refactor. See if you can omit it and then we can apply the same pattern to the rest of the connectors.

We have had issues (#6948) upgrading the version of Jackson and than having a different version of at runtime, dictated by Camel BOM/Spring Boot BOM or the integration-runtime BOM.

This will be an ongoing effort I think trying to separate the dependency trees of integrations at runtime and the dependencies we use in the server.

app/connector/mongodb/src/main/resources/META-INF/syndesis/connector/mongodb3.json
squakez: Yes, this is something that will be reviewed and adjusted with the feedback received by doc and product team.

app/connector/mongodb/src/main/resources/META-INF/syndesis/connector/mongodb3.json
squakez: Do we have some document to explain how to provide that migration?

app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/meta/MongoDBMetadataRetrieval.java
zregvart: I don't understand what that means can you elaborate? I'm proposing that instead of `isFilterAction` being a `private` method if we can make it `private static` method instead of adding a `@SuppressWarnings("MethodCanBeStatic")`.

app/connector/mongodb/src/main/resources/META-INF/syndesis/connector/mongodb3.json
zregvart: There's a bit here: https://github.com/syndesisio/syndesis/blob/master/project/proposals/971-upgrade.adoc#step-upgrade-db have a look at the scripts [here](https://github.com/syndesisio/syndesis/tree/master/app/server/runtime/src/main/resources/migrations) for ideas on how this can be done, and make sure to upgrade the schema version [here](https://github.com/syndesisio/syndesis/blob/master/app/common/model/src/main/java/io/syndesis/common/model/Schema.java).

squakez: I don't think that is really needed. All the actions provided in 7.6 will be using a new connector ID. Migrating 7.5 to 7.6 is not compatible as there would be missing mandatory criteria and data mapping, so, if any migration is needed it must be performed through a kick off of a new integration with the new parameters required.

app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/meta/MongoDBMetadataRetrieval.java
squakez: I meant to say that it's a method used only [here](https://github.com/syndesisio/syndesis/blob/e4485cb2b206002ad9da24a7e6541d1395a2961a/app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/meta/MongoDBMetadataRetrieval.java#L72-L73). The way I see this is that the `isFilterAction` is semantically part of the `MongoDBMetadataRetrieval` instance as I could have inlined the code into the `buildDatashape`. I do expect in fact that compiler would inline the source code at compilation time as that was not intended to be static on purpose.

zregvart: I'm sorry I've never come across the use of the word _semantically_ in context of Java methods, so I really don't understand. What do you mean that the method that is not static is semantically a part of a class vs a method that is static not being semantically a part of a class?

This method is a private method and it doesn't access any instance state, I don't see a valid reason for not making it static vs increasing the amount of technical debt by adding the SuppressWarnings annotation.

squakez: The PR was merged automatically by the bot already.

But, just to clarify, I was meaning that the [final goal would be the this](https://github.com/squakez/syndesis/compare/feature/ENTESB-11589_filters...squakez:experiment/inline_method) - and I'd say that the compiler would take care on its own to do that. I can see too many static methods around just because PMD is forcing to go in that direction making our development too procedural. The design I chose was different, and the only way to achieve that was adding a `SuppressWarnings` annotation.

If you still feel that it must considered as technical debt, please log an issue and we will review that as the PR is already merged unfortunately.

KurtStam:
not-sure-if-semantics

Issue #7146, opened 2019-11-15
MongoDB daily build error


squakez:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
The embedded mongo intermittently fails to start due to the following exception:

```
06:52:38 [INFO] Running io.syndesis.connector.mongo.MongoDBConnectorCappedCollectionConsumerTest
06:52:38 05:52:38.103 [cluster-ClusterId{value='5dce3d256efa1a6e7fd5eb8b', description='null'}-localhost:27017] DEBUG org.mongodb.driver.connection - Closing connection connectionId{localValue:62, serverValue:1}
06:52:38 05:52:38.158 [main] ERROR de.flapdoodle.embed.process.runtime.AbstractProcess - failed to call start()
06:52:38 java.io.IOException: Cannot run program "/tmp/extract-44370b4c-36de-4539-a8ef-0c5ea3eff42aextractmongod": error=26, Text file busy
06:52:38 	at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
06:52:38 	at de.flapdoodle.embed.process.runtime.ProcessControl.start(ProcessControl.java:213)
06:52:38 	at de.flapdoodle.embed.process.runtime.AbstractProcess.<init>(AbstractProcess.java:99)
06:52:38 	at de.flapdoodle.embed.mongo.AbstractMongoProcess.<init>(AbstractMongoProcess.java:53)
06:52:38 	at de.flapdoodle.embed.mongo.MongodProcess.<init>(MongodProcess.java:50)
06:52:38 	at de.flapdoodle.embed.mongo.MongodExecutable.start(MongodExecutable.java:44)
06:52:38 	at de.flapdoodle.embed.mongo.MongodExecutable.start(MongodExecutable.java:34)
06:52:38 	at de.flapdoodle.embed.process.runtime.Executable.start(Executable.java:108)
06:52:38 	at io.syndesis.connector.mongo.MongoDBConnectorTestSupport.startUpMongo(MongoDBConnectorTestSupport.java:67)
06:52:38 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
06:52:38 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #7154, opened 2019-11-18
fix(connector/mongodb): embedded mongo refactory

squakez:
* Starting a unique static server instead of one for each testcase
* Reduced time to execute all testcases by half due to the above

`EmbedMongoConfiguration` will startup a unique server (with replicaset). Each unit test class will use its own collection to reduce resource consumption and limit possibility of test collisions.

Fixes #7164


squakez:
Fixes #7146

PR #7159, opened 2019-11-18
Fixes ENTESB-12178, initial version of Public API addon

dhirajsb:


dhirajsb:
@lgarciaaco can you review this PR?

install/operator/build/conf/config.yaml
```diff
@@ -26,6 +26,9 @@ Syndesis:
             Image: "fabric8/s2i-java:3.0-java8"
             CamelVersion: "2.21.0.fuse-760006"
```
claudio4j: Rebase it as camel is 2.21.0.fuse-760011

install/operator/pkg/generator/assets/addons/publicApi/addon-public-oauthproxy.yml.tmpl
```diff
@@ -158,10 +79,11 @@ objects:
       spec:
         containers:
         - name: syndesis-public-oauthproxy
+          image: '{{ .Syndesis.Components.Oauth.Image }}'
           args:
             - --provider=openshift
-            - --client-id=system:serviceaccount:${OPENSHIFT_PROJECT}:syndesis-oauth-client
-            - --client-secret=${OPENSHIFT_OAUTH_CLIENT_SECRET}
+            - --client-id=system:serviceaccount:{{.OpenShiftProject}}:syndesis-oauth-client
+            - --client-secret=$(OPENSHIFT_OAUTH_CLIENT_SECRET)
```
lgarciaaco:
```suggestion
            - --client-secret=$(.OpenShiftOauthClientSecret)
```

install/operator/pkg/cmd/internal/grant/grant.go
```diff
@@ -49,6 +51,7 @@ func New(parent *internal.Options) *cobra.Command {
 
 	cmd.PersistentFlags().BoolVarP(&o.cluster, "cluster", "", false, "add the permission for all projects in the cluster(requires cluster admin privileges)")
 	cmd.PersistentFlags().StringVarP(&o.User, "user", "u", pkg.DefaultOperatorImage, "add permissions for the given User")
+	cmd.PersistentFlags().BoolVarP(&o.publicApi, "publicApi", "", false, "add permissions required for publicApi addon in the cluster(requires cluster admin privileges)")
```
lgarciaaco: No need for this, look at following comment.

install/operator/pkg/cmd/internal/grant/grant.go
```diff
@@ -75,8 +78,16 @@ func (o *Grant) grant() error {
...
 	resources = append(resources, gr...)
+
+	if o.publicApi {
+		pubRole, err := generator.Render("./install/grant_public_api_cluster_role.yml.tmpl", o)
+		if err != nil {
+			return err
+		}
+		resources = append(resources, pubRole...)
+	}
+
```
lgarciaaco: I would rather have this snippet in `pkg/syndesis/action/install`, something like.

```
	if configuration.Syndesis.Addons.PublicApi.Enabled  {
		pubRole, err := generator.Render("./install/grant_public_api_cluster_role.yml.tmpl", o)
		if err != nil {
			return err
		}
		resources = append(resources, pubRole...)
	}
```

install/operator/pkg/generator/assets/install/grant_public_api_cluster_role.yml.tmpl
```diff
@@ -0,0 +1,17 @@
...
+    syndesis.io/type: infrastructure
+    syndesis.io/component: syndesis-public-oauthproxy
+subjects:
+- kind: ServiceAccount
+  name: syndesis-public-oauthproxy
+  namespace: {{.Namespace}}
+roleRef:
+  kind: ClusterRole
+  name: system:auth-delegator
+  apiGroup: rbac.authorization.k8s.io
```
lgarciaaco: I have to point out that this is generally a bad practice towards OLM integration because, they are picky with the verbs they allow and it can lead to the operator not being allowed to bind the `ClusterRole` in OLM if the `ClusterRole` has any verb other than `assign get list watch create update patch delete deletecollection initialize use`. I rather have a new `Role` with all the specific permissions you need, and a `RoleBinding` for it, which in my opinion also adds clarity of what is being granted.

**BUT**, since this ClusterRole is so slim, I think it is fine to leave it this way.

Nevertheless you have to add those permissions to https://github.com/syndesisio/syndesis/blob/684df965f085499789b8033385abb63027e4b255/install/operator/pkg/generator/assets/install/role.yml.tmpl#L10

install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
```diff
@@ -73,6 +73,16 @@ type OauthConfiguration struct {
...
 
+type PublicApiConfiguration struct {
+	Enabled bool `json:"enabled,omitempty"`
+	// Set RouteHostname to the hostname of the exposed syndesis API.  Typically the operator can automatically
+	// determine this by looking at the result of the Route object it creates.
+	RouteHostname string `json:"routeHostname,omitempty"`
+	// if set to true, then any authenticated user can access the API. otherwise the user
+	// needs access to get pods against the SarNamespace
+	DisableSarCheck bool `json:"disable-sar-check,omitempty"`
+}
```
lgarciaaco: I would like to see the default values for these fields in 
https://github.com/syndesisio/syndesis/blob/684df965f085499789b8033385abb63027e4b255/install/operator/build/conf/config.yaml#L1-L78

install/operator/pkg/generator/assets/addons/publicApi/addon-public-oauthproxy.yml.tmpl
dhirajsb: It's better to replace these with [environment variables support](https://github.com/openshift/oauth-proxy#environment-variables) in the oauth proxy, which avoids injecting secret values in deployment configs.

install/operator/pkg/cmd/internal/grant/grant.go
dhirajsb: If the ClusterRoleBinding is moved to the addon, the operator's service-account has to hold the same permissions in it's own list of roles. I tried that, and it's actually worse, since it turns into a laundry list of other permissions the operator must hold. 
Not to mention that if we do it this way, only a cluster admin could install the Syndesis operator preventing developers from being able to setup private development environments in their own namespaces. 
I think it's better to separate concerns and keep this cluster admin task out of the operator's runtime and to instead make it something it can do externally on behalf of a cluster admin through the grant command.

install/operator/pkg/generator/assets/install/grant_public_api_cluster_role.yml.tmpl
dhirajsb: I have added a ClusterRole to avoid referencing the OpenShift system:auth-delgator role if it's going to be an issue with OLM.

install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
dhirajsb: Added default values in config.yaml.

dhirajsb:
It's failing on a weird error in operator build
```
go: vbom.ml/util@v0.0.0-20180919145318-efcd4e0f9787: unrecognized import path "vbom.ml/util" (https fetch: Get https://vbom.ml/util?go-get=1: dial tcp: lookup vbom.ml on 127.0.0.11:53: server misbehaving)
```

Any ideas? The build works fine on my laptop.

dhirajsb:
Merged PR manually.

PR #7171, opened 2019-11-20
Set operand images to right tag when building

lgarciaaco:
When we do a release, set operand images to the version being released instead of latest

Fixes https://issues.jboss.org/browse/ENTESB-12176

### DIFF
--- a/tools/bin/commands/release
+++ b/tools/bin/commands/release
@@ -234,13 +234,11 @@ update_image_versions() {
     local topdir="$1"
     local version="$2"
     local moving_tag="$3"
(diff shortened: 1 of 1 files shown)


tools/bin/commands/release
```diff
@@ -234,13 +234,11 @@ update_image_versions() {
     local topdir="$1"
     local version="$2"
     local moving_tag="$3"
```
zregvart: Looks like this is not used any more
```suggestion
```

KurtStam: @lgarciaaco will you update?

Issue #5102, opened 2019-04-02
Failure to download spring-social-salesforce dependency


zregvart:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
Seems that the build is failing with:

```
[INFO] --- dependency-scope-maven-plugin:0.8:check (default) @ server-endpoint ---
[INFO] Downloading from central: https://repo.maven.apache.org/maven2/com/github/mikegirard/spring-social-salesforce/1.3.0/spring-social-salesforce-1.3.0.pom
[INFO] Downloading from jboss-ea: https://repository.jboss.org/nexus/content/groups/ea/com/github/mikegirard/spring-social-salesforce/1.3.0/spring-social-salesforce-1.3.0.pom
[INFO] Downloading from redhat-ga: https://maven.repository.redhat.com/ga/com/github/mikegirard/spring-social-salesforce/1.3.0/spring-social-salesforce-1.3.0.pom

...

[ERROR] Failed to execute goal com.hubspot.maven.plugins:dependency-scope-maven-plugin:0.8:check (default) on project server-endpoint: Error resolving descriptor for artifact com.github.mikegirard:spring-social-salesforce:1.3.0: Failed to read artifact descriptor for com.github.mikegirard:spring-social-salesforce:jar:1.3.0: Could not transfer artifact com.github.mikegirard:spring-social-salesforce:pom:1.3.0 from/to redhat-ga (https://maven.repository.redhat.com/ga/): Failed to transfer file: https://maven.repository.redhat.com/ga/com/github/mikegirard/spring-social-salesforce/1.3.0/spring-social-salesforce-1.3.0.pom. Return code is: 500, ReasonPhrase: Internal Server Error. -> [Help 1]
```

Perhaps the best option would be to try to eliminate that dependency and not depend on it being downloaded from jitpack.

Or would be to try to determine why it's not downloaded in this use case.

zregvart:
Occurred again on [ci](https://circleci.com/gh/syndesisio/syndesis/85465)

PR #7172, opened 2019-11-20
refactor(ENTESB-12323): Syndesis using camel-fuse-2.23.x

christophd:
Syndesis using camel-fuse-2.23.x along with Spring Boot 2.x.


app/meta/src/main/resources/application.yml
```diff
@@ -21,23 +21,26 @@ server:
 # See https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-monitoring.html
 # For details
 management:
-  port: 8080
-  security:
-    enabled: true
+  server:
+    port: 8181
```
squakez: This port differs from the `managment.port: 8080` above, is that expected?

app/connector/aws-sqs/src/main/java/io/syndesis/connector/aws/sqs/AWSSQSBatchMessagesCustomizer.java
```diff
@@ -49,6 +46,5 @@ public void beforeProducer(final Exchange exchange) throws IOException {
             }
         }
         in.setBody(messageList);
-        in.setHeader(SqsConstants.SQS_OPERATION, SqsOperations.sendBatchMessage);
```
squakez: Had a live conversation already about this, it seems it may be needed.

app/server/credential/src/main/java/io/syndesis/server/credential/SocialProperties.java
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.syndesis.server.credential;
+
+/**
+ * Basic configuration properties for spring social.
+ */
+public class SocialProperties {
```
squakez: I understand you're cloning this [java bean](https://docs.spring.io/spring-boot/docs/2.0.0.M3/api/org/springframework/boot/autoconfigure/social/SocialProperties.html) to avoid bringing in another spring library, isn't it?

app/connector/jira/pom.xml
```diff
@@ -32,11 +32,18 @@
         <!-- The 5.1.0 version uses an old guava 20 version incompatible with the default guava set on syndesis -->
         <atlassian-jira-rest-java-client-version>5.1.6</atlassian-jira-rest-java-client-version>
         <atlassian-util-concurrent-version>4.0.1</atlassian-util-concurrent-version>
+
+        <!-- TODO: Remove when using Camel 3 bits in Syndesis -->
```
squakez: Discussed in live conversation, the Jira component is here while waiting the backport to 2.23 camel branch, we can live with this in the while.

app/server/credential/src/main/java/io/syndesis/server/credential/SocialProperties.java
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.syndesis.server.credential;
+
+/**
+ * Basic configuration properties for spring social.
```
squakez: You may add an explanation and a link to the original library in order to leave a trace of where and why this is cloned here.

app/meta/src/main/resources/application.yml
christophd: I have noticed that the liveness checks have ever been using port 8181 (https://github.com/syndesisio/syndesis/blob/master/install/operator/pkg/generator/assets/infrastructure/04-syndesis-meta.yml.tmpl#L104) I do not know why the local setting was using 8080 as management port but I think it makes sense to align with what we use on OCP in the config maps.

app/connector/aws-sqs/src/main/java/io/syndesis/connector/aws/sqs/AWSSQSBatchMessagesCustomizer.java
christophd: Yes that is a problem indeed. Thanks for pointing to it! We are missing a backport to camel-fuse 2.23.x here. I have opened an issue to do the backport: https://issues.jboss.org/browse/ENTESB-12348

This is the commit that needs to be backported: https://github.com/jboss-fuse/camel/commit/53670be2a404b9c258897d02019892a36ba46664#diff-5b2805df49a4af06496342bb13db80a3

Once this is done we can add the header again in our code. I will add a comment in our code and reference the backporting issue.

app/server/credential/src/main/java/io/syndesis/server/credential/SocialProperties.java
christophd: Yes Spring Boot 2.x has removed that bean and we kinda use it heavily so this was the best way for me to align our code with newer Spring versions. Maybe requires some further refactoring to make our code also not depend on that particular class in future.

app/connector/jira/pom.xml
christophd: yes @claudio4j is already working on that. Once he is finished with backporting the camel-jira component to 2.23.x we can remove the component clone again from Syndesis.

app/server/credential/src/main/java/io/syndesis/server/credential/SocialProperties.java
christophd: will do yes

app/connector/jira/pom.xml
christophd: This is the PR we need to align 2.21 and 2.23: https://github.com/jboss-fuse/camel/pull/1185

app/server/credential/src/main/java/io/syndesis/server/credential/SocialProperties.java
zregvart: It's a long standing problem that we depend on a forked version of `spring-social` (e.g. #5102), we could (at a later point in time) refactor this code so we don't depend on it at all.

zregvart:
[gifbot:celebration]

PR #7203, opened 2019-11-22
chore(build): Maven settings support

zregvart:
Adds support for specifying `-s` or `--settings` to the `syndesis` shell
script to configure the Maven with specified Maven settings XML file.

### DIFF
--- a/tools/bin/commands/util/maven_funcs
+++ b/tools/bin/commands/util/maven_funcs
@@ -6,41 +6,43 @@ POD_MODULES="meta server ui operator"
 call_maven() {
     local args=$1
     local maven_modules=$2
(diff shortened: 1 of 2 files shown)


tools/bin/commands/util/maven_funcs
```diff
@@ -22,7 +22,11 @@ call_maven() {
       echo "Modules: $maven_modules"
       echo "=============================================================================="
       echo "./mvnw -N --batch-mode install -Pflash ### Installing parent pom.xml"
```
squakez: I'd update this `echo` with the current setting flag as well

zregvart:
I've cleaned this up a bit, @squakez care to have another look?

PR #7215, opened 2019-11-25
Align (remove) dv properties so that it uses parent pom properties

vhalbert:


app/dv/pom.xml
```diff
@@ -239,8 +239,8 @@
 
     <!-- Common dependencies -->
     <version.springfox>2.9.2</version.springfox>
-    <assertj-core.version>3.11.1</assertj-core.version>
-    <shrinkwrap.version>2.2.6</shrinkwrap.version>
+    <!--assertj-core.version>3.11.1</assertj-core.version-->
+    <!--shrinkwrap.version>2.2.6</shrinkwrap.version-->
```
zregvart: We could even remove these commented out lines

PR #7225, opened 2019-11-25
TEIIDTOOLS-830, TEIIDTOOLS-831 Add virtualization details page content

mdrillin:
Adds virtualization details tab/page, which brings virtualization versioning into data virtualization.
- addition of useVirtualizationEditions - fetch the virtualization versions list
- add revertVirtualization and startVirtualization to useVirtualizationHelpers - functions to support virtualization versions.
- VirtualizationEdition was added to models.  Virtualization model was updated to include new attributes.
- VirtualizationDetailHistoryTable was added - component for display of the version table
- VirtualizationDetailPage was updated to include the VirtualizationDetailHistoryTable component
- VirtualizationUtils was updated to add the canStart and canRevert functions, and update other logic where needed.
- updates to other components to include virtualization 'modified' and 'version' info.
- created / updated component stories


app/ui-react/packages/ui/src/Data/Virtualizations/PublishStatusWithProgress.tsx
```diff
@@ -34,17 +36,25 @@ export const PublishStatusWithProgress: React.FunctionComponent<
           }
           i18nLogUrlText={props.i18nPublishLogUrlText}
         />
+        {props.publishVersion
```
elvisisking: This should be i18n'd but can be worked in a follow-on issue. Also in a future issue, like we talked about, using an 'M' as modified indicator needs to be changed per UX guidance/mockups. There are multiple occurrences that need to be changed.

PR #7231, opened 2019-11-26
refactor(ENTESB-12218): 2nd part of OpenAPI 3.x support refactoring

christophd:
This is the 2nd portion of the Open API 3.x refactoring towards supporting both 2.x and 3.x specifications in server-api-generator module.

The PR separates 2.x specifics from common OpenAPI logic and prepares 3.x implementations by adding empty classes. 

The new package structure reflects the version separation and is as follows:
- **openapi**: common classes and interfaces shared by both 2.x and 3.x
- **openapi/util**: shared helpers and utility classes
- **openapi/v2**: 2.x specific implementations
- **openapi/v3**: 3.x specific implementations (empty at the moment)

Each commit in this PR resolves the former `swagger` package step by step to either shared `openapi` or specific `openapi/v2` packages.

Based on this PR I will now add OpenAPI 3.x logic to the `openapi/v3` package and implement the empty classes located there.


app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiValidationRules.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * This class contains Syndesis custom validation rules for Open API specifications.
+ */
+public final class OpenApiValidationRules implements Function<OpenApiModelInfo, OpenApiModelInfo> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(OpenApiValidationRules.class);
+
+    private final List<Function<OpenApiModelInfo, OpenApiModelInfo>> v2Rules = new ArrayList<>();
+    private final List<Function<OpenApiModelInfo, OpenApiModelInfo>> v3Rules = new ArrayList<>();
```
zregvart: I was hoping that using the Apicurio OpenAPI would not necessitate two different sets of validation rules, one for each version. I guess that's not so?

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/SchemeType.java
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.syndesis.server.api.generator.openapi;
+
+/**
+ * Supported security schemes.
+ */
+public enum SchemeType {
```
zregvart: Perhaps more telling name?

```suggestion
public enum SecurityScheme {
```

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiModelInfo.java
```diff
@@ -43,7 +46,30 @@
         return Collections.emptyList();
     }
 
-    Oas20Document getModel();
+    OasDocument getModel();
+
+    @JsonIgnore
+    default OpenApiVersion getApiVersion() {
```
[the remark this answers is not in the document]
christophd: Agreed, that might be a valuable info for the UI.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiSchemaValidator.java
```diff
@@ -0,0 +1,60 @@
...
+     * Validates given specification Json and add validation errors and warnings to given open api info model builder.
+     * @param specRoot the specification as Json root.
+     * @param modelBuilder the model builder receiving all validation errors and warnings.
+     */
+    void validateJSonSchema(JsonNode specRoot, OpenApiModelInfo.Builder modelBuilder);
+
+    default boolean append(final List<Violation> violations, final ProcessingMessage message, final Optional<String> requiredLevel) {
+        if (requiredLevel.isPresent()) {
+            final Optional<String> level = ofNullable(message.asJson()).flatMap(node -> ofNullable(node.get("level")))
+                .flatMap(node -> ofNullable(node.textValue()));
```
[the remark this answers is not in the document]
christophd: will have a look. To be honest just copied it over without thinking about it

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiValidationRules.java
christophd: I prepared for having different sets because I think some rules may be obsolete according to the changed spec architecture in 3.x. On the other hand we might want to add some 3.x specific rules. When I get to implementing the rules for 3.x I probably can say more about it

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/SchemeType.java
christophd: 👍

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiSchemaValidator.java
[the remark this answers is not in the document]
christophd: after having a 2nd look I can only imagine that we want to avoid having a compile time dependency to the enum com.github.fge.jsonschema.core.report.LogLevel for some reason. It must be a good reason

Issue #7086, opened 2019-11-05
`syndesis minishift --install --tag` installs `latest`


zregvart:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

Seems that when we install like:

```shell
$ syndesis minishift --install --nodev --tag 1.8.12
```

The `latest` version is installed not `1.8.12`.

cc @lgarciaaco

ryckowski:
It looks like there's no priority to fix installation the stable, it's a pity as installation from master is broken so often in the last weeks

PR #7235, opened 2019-11-26
Fix install not taking the tag

lgarciaaco:
Fixes https://github.com/syndesisio/syndesis/issues/7086

If that `--tag` is set, forde download of a new binary and download the binary from the specified release

### DIFF
--- a/tools/bin/commands/util/operator_funcs
+++ b/tools/bin/commands/util/operator_funcs
@@ -88,10 +88,16 @@ download_operator_binary() {
     esac
 
     if [[ "${pattern}" != "unknown" ]]; then
-        url=$(curl -s https://api.github.com/repos/syndesisio/syndesis/releases?per_page=1 \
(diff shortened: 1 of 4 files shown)


tools/bin/commands/util/operator_funcs
```diff
@@ -88,10 +88,16 @@ download_operator_binary() {
     esac
 
     if [[ "${pattern}" != "unknown" ]]; then
-        url=$(curl -s https://api.github.com/repos/syndesisio/syndesis/releases?per_page=1 \
-                | grep browser_download_url \
-                | grep ${pattern} \
-                | cut -d '"' -f 4)
+        release_uri="https://api.github.com/repos/syndesisio/syndesis/releases?per_page=1"
+        release_tag="$(readopt --tag)"
```
zregvart: Perhaps this should be a parameter of the function?

PR #7237, opened 2019-11-26
fix(TEIIDTOOLS-809) - ViewListItem should show icon indicating if the view is invalid

elvisisking:
- see [TEIIDTOOLS-809](https://issues.jboss.org/browse/TEIIDTOOLS-809)
- added `valid` to the `ViewDefinitionDescriptor` model object
- added a red `Label` to `ViewListItem` with the text "Invalid" when a view is invalid
- updated stories

### DIFF
--- a/app/ui-react/packages/models/src/dv.d.ts
+++ b/app/ui-react/packages/models/src/dv.d.ts
@@ -71,6 +71,7 @@ export interface ViewDefinitionDescriptor {
   id: string;
   name: string;
   description: string;
+  valid: boolean;
 }
 
 export interface ViewDefinition {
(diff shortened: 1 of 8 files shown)


amysueg:
@elvisisking Would you mind adding a screenshot to this PR so we can have a look?

elvisisking:
Wdyt @amysueg?
view-list-item-invalid

amysueg:
I think the label looks ok - but it is pretty alarming and I'd be somewhat concerned if the system were ever in a state where a number of Views are invalid at the same time. If that is possible, then perhaps using plain red text (rather than the white text on the red rectangle) would be somewhat less jarring. I left a question in the corresponding Jira about how the user gets out of the "Invalid" state. Thanks.

elvisisking:
How's this...
Screen Shot 2019-11-26 at 3 59 26 PM

amysueg:
I like, @elvisisking - thanks.

PR #7239, opened 2019-11-26
ENTESB-12262: Expose pvc name, access mode, storage-class & labels

phantomjinx:
* Exposes persistent volume metadata for connecting the syndesis-db with a
  specific kind of persistent volume
 * Openshift specifies that a volume name will be a direct bind request
   whereas the other metadata feeds into the internal algorithm for
   determining which persistent volume should be allocated

* generator_test
 * Provides tests for the different volume metadata

* install.go
 * Log the combination of volume metadata to inform how the db volume is
   selected (or not as the case maybe)

* configuration.go
 * Expose env vars for access-mode, storage-class & volume-name for testing
   and development purposes

* Automatic tidying of operator dependencies


install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
```diff
@@ -112,6 +112,23 @@ type Resources struct {
 	Memory string `json:",inline,omitempty"`
 }
 
+type VolumeAccessMode string
```
lgarciaaco: Creating a new type here only makes sense if you validate the parameter further down. https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#openapi-validation

install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -152,6 +152,15 @@ type ResourcesWithVolume struct {
 	VolumeCapacity string
 }
 
+type ResourcesWithPersistentVolume struct {
+	Memory             string
+	VolumeCapacity     string
+	VolumeName         string
+	VolumeAccessMode   string
+	VolumeStorageClass string
```
lgarciaaco: Can we get default values for all of them?

install/operator/pkg/generator/generator_test.go
```diff
@@ -293,3 +290,201 @@ func assertPropStr(t *testing.T, resource map[string]interface{}, expected strin
 		assert.Equal(t, expected, value, "rendering should be applied correctly")
 	}
 }
+
+func loadDBResource(t *testing.T, syndesis *v1alpha1.Syndesis) []unstructured.Unstructured {
+	configuration, err := configuration.GetProperties("../../build/conf/config.yaml", context.TODO(), nil, syndesis)
```
lgarciaaco: use rather config-test.yaml

install/operator/pkg/syndesis/configuration/configuration.go
phantomjinx: No, as some are supposed to be blank by default.

VolumeCapacity & VolumeAccessMode are provided with defaults in config.yaml. These are the minimum to provide the default behaviour of specifying a PVC for claiming a default file PV. The others are optional and will enhance the specification for a particular kind of PV. In fact, it is pointless to specify the others if specifying VolumeName as this bypasses the openshift algorithm for determining a PV & simply allocated the named PV if it exists.

install/operator/pkg/generator/generator_test.go
phantomjinx: Done.

install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
phantomjinx: Added the enum validation to the type. Not made any difference to any other file after re-running sdk-operator k8s.

PR #7250, opened 2019-11-27
Revert Google Calendar API upgrade

zregvart:
Ref https://issues.jboss.org/browse/ENTESB-12191


app/test/integration-test/src/test/java/io/syndesis/test/itest/googlecalendar/WebhookToGoogleCalendar_IT.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.http.HttpStatus;
+
+import com.consol.citrus.annotations.CitrusResource;
+import com.consol.citrus.annotations.CitrusTest;
+import com.consol.citrus.dsl.runner.TestRunner;
+
+public class WebhookToGoogleCalendar_IT extends SyndesisIntegrationTestSupport {
+
+    /**
+     * Webhook receives POST request and creates a Google Calendar event.
```
christophd: Cool new integration test! We could also receive the created Calendar event in the test and verify its content

zregvart: We can't really, look at the commit messages.

christophd: What is the output of the running integration then? Where is the calendar event stored? Isn't that sent to the mocked `www.googleapis.com`?

zregvart: I can't stub the `www.googleapis.com`, the resolution I can influence but the host port forwarded ends up being in the `sshd` container spun up by testcontainers and I don't know the IP of that container. Also the Jetty setup using TLS didn't seem to work.

PR #7254, opened 2019-11-27
chore: fix the alignment of the dashboard cards

gashcrumb:
This was bugging me yesterday.  Basically this PR makes the dashboard cards align better, so instead of:

image

it looks like:

image


zregvart:
My OCD was killing me, thanks!

PR #7259, opened 2019-11-27
ENTESB-12090: Fix addons parameter when executing app

phantomjinx:
* addons value should be a bool rather than a string


KurtStam:
@phantomjinx Can you add a bit of context?I was expecting to see a simple type change and this PR seems much more comprehensive.

heiko-braun:
@phantomjinx Should the `vendor` directory be included or was it by accident?

phantomjinx:
@heiko-braun @KurtStam 
The vast majority of this PR is simply auto regenerated bits n pieces updated when the operator is recompiled. I'd prefer to leave it all out but to do so would simply mean everyone else seeing it all when they compile.

The only germaine file change is https://github.com/syndesisio/syndesis/pull/7259/files#diff-9f105b2a895920f82ad1c20a09c87eafR9

heiko-braun:
> The vast majority of this PR is simply auto regenerated bits n pieces updated when the operator is recompiled. I'd prefer to leave it all out but to do so would simply mean everyone else seeing it all when they compile.


I don't know. What do others recommend? @lgarciaaco @nicolaferraro @chirino Should we commit `vendor` and other pieces as @phantomjinx suggests, or leave it out?

zregvart:
@phantomjinx this needs a rebase

PR #7272, opened 2019-11-28
refactor(ENTESB-12218): Implement basic OpenAPI 3 support

christophd:
This PR adds basic OpenAPI 3.0 support for server-api-generator. The support should be feature equal to that what we have with Swagger 2.x for API provider integrations.

In general try to reuse as much logic as possible by sharing classes on OpenAPI 2.x and 3.x implementations.

The OpenAPI 3 support includes:
- Parse OpenAPI 3.0 specification
- Apply validation rules and display warnings/errors
- Generate integration flows
- Generate unified request/response data shapes (Json and XML) from specification
- Generate api-provider connector with configured properties

Missing:
- Open API 3.0 schema validation


app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiConnectorGenerator.java
```diff
@@ -136,11 +138,11 @@ public final APISummary info(final ConnectorTemplate connectorTemplate, final Co
     protected ConnectorDescriptor createDescriptor(String connectorId, OpenApiModelInfo info, OasOperation operation) {
         switch (info.getApiVersion()) {
```
zregvart: Would it make sense to have a field with `DataShapeGenerator<T,O>` and then two instances of `OpenApiConnectorGenerator` one with the field initialized to v2 implementation and one to v3 implementation. Would reduce the amount of conditionals we use here.
Then it would perhaps make sense to have the `OpenApiConnectorGenerator` parametrized with the same `<T extends OasDocument, O extends OasOperation>`.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiConnectorGenerator.java
```diff
@@ -171,10 +173,10 @@ private Connector basicConnector(final ConnectorTemplate connectorTemplate, fina
             Optional<ConfigurationProperty> maybeProperty;
             switch (info.getApiVersion()) {
```
zregvart: Kinda the same thing here as above. Would reduce the number of object instantiations as well. I don't believe that the `DataShapeGenerator`s or `OpenApiPropertyGenerator` hold any state so they're prime for caching in a field.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiConnectorGenerator.java
```diff
@@ -301,10 +303,10 @@ protected final String determineConnectorName(final ConnectorTemplate connectorT
     private static void addGlobalParameters(final Connector.Builder builder, final OpenApiModelInfo info) {
         switch (info.getApiVersion()) {
             case V2:
-                Oas20ConnectorGeneratorSupport.addGlobalParameters(builder, info.getV2Model());
```
zregvart: And these also, we could cache them in a field instead of instantiating them.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiPropertyGenerator.java
```diff
@@ -0,0 +1,338 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import static java.util.Optional.empty;
+import static java.util.Optional.ofNullable;
+
+/**
+ * Set of property generators that may apply to the give OpenAPI specification. When applicable property is added as configuration
+ * property. Subclasses provide 2.x and 3.x OpenAPI specific implementations.
+ */
+@SuppressWarnings("PMD.GodClass")
```
zregvart: Perhaps we could refactor this (at some point, not necessarily now) and break it into several classes to make it non god class. Something like one class per property.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiSecurityScheme.java
```diff
@@ -19,13 +19,13 @@
 /**
  * Supported security schemes.
  */
-public enum SecurityScheme {
+public enum OpenApiSecurityScheme {
```
zregvart: :+1:

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/v2/UnifiedJsonDataShapeGenerator.java
```diff
@@ -38,18 +33,17 @@
...
-import io.syndesis.common.model.DataShapeMetaData;
+import io.syndesis.server.api.generator.openapi.DataShapeGenerator;
+import io.syndesis.server.api.generator.openapi.UnifiedJsonDataShapeSupport;
 import io.syndesis.server.api.generator.openapi.util.JsonSchemaHelper;
 import io.syndesis.server.api.generator.openapi.util.OasModelHelper;
 import org.apache.commons.lang3.StringUtils;
 
 import static org.apache.commons.lang3.StringUtils.trimToNull;
 
 @SuppressWarnings("PMD.GodClass")
```
zregvart: Perhaps this can be removed now, not as many methods in here any more

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/v2/UnifiedXmlDataShapeGenerator.java
```diff
@@ -38,245 +32,54 @@
...
 import io.syndesis.server.api.generator.openapi.util.XmlSchemaHelper;
-import org.dom4j.Document;
-import org.dom4j.DocumentHelper;
 import org.dom4j.Element;
 
-import static java.util.Objects.requireNonNull;
 import static java.util.Optional.ofNullable;
 import static org.apache.commons.lang3.StringUtils.trimToNull;
 
 @SuppressWarnings("PMD.GodClass")
```
zregvart: Perhaps can be also removed

app/server/api-generator/src/test/java/io/syndesis/server/api/generator/openapi/OpenApiGeneratorTest.java
```diff
@@ -44,7 +44,7 @@ public void infoShouldHandleNullModels() {
 
         assertThat(summary).isNotNull();
         assertThat(summary.getErrors()).hasSize(1)
-            .allSatisfy(v -> assertThat(v.message()).startsWith("This document cannot be uploaded. Provide an OpenAPI 2.0 document"));
+            .allSatisfy(v -> assertThat(v.message()).startsWith("This document cannot be uploaded. Please provide an OpenAPI document"));
```
zregvart: @TovaCohen I think we said not to say "Please", did I miss remember that?

app/server/api-generator/src/test/java/io/syndesis/server/api/generator/openapi/util/OpenApiModelParserTest.java
```diff
@@ -85,16 +85,40 @@ public void shouldNotReportIssuesWithSupportedVersions() {
     }
 
     @Test
-    public void shouldReportIssuesWithUnsupportedVersions() {
+    public void shouldReportIssuesWithUnsupportedV2Versions() {
         final OpenApiModelInfo validated = OpenApiModelParser.parse(
-            "{\"openapi\": \"3.0.0\", \"info\":{ \"title\": \"test\", \"version\": \"1\"}, \"paths\": { \"/api\": { \"get\": {\"responses\": { \"200\": { \"description\": \"OK\" }}}}}}",
+            "{\"swagger\": \"3.0\", \"info\":{ \"title\": \"test\", \"version\": \"1\"}, \"paths\": { \"/api\": { \"get\": {\"responses\": { \"200\": { \"description\": \"OK\" }}}}}}",
```
zregvart: Perhaps we test with `openapi` 4.0 then?

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/v2/UnifiedJsonDataShapeGenerator.java
christophd: Good call! Done

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/v2/UnifiedXmlDataShapeGenerator.java
christophd: Good call! Done

app/server/api-generator/src/test/java/io/syndesis/server/api/generator/openapi/OpenApiGeneratorTest.java
christophd: Oh yes I will revert the `Please`

app/server/api-generator/src/test/java/io/syndesis/server/api/generator/openapi/util/OpenApiModelParserTest.java
christophd: I will be using two tests one for `swagger:1.0` and one for `openapi:4.0`

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiConnectorGenerator.java
christophd: No, the `OpenApiConnectorGenerator` is a single entry point for other modules like `server-runtime`. Depends on if we want to have two separated server endpoints for the Syndesis UI to call (one for API v2 and one for API v3). Then we can separate the OpenAPI versions from the very beginning as the UI logic picks the right version. At the moment we have one API generating endpoint on the server.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiConnectorGenerator.java
christophd: Agreed, cached those in a field

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiConnectorGenerator.java
zregvart: Perhaps I didn't make it that clear, this is what I was thinking:

```java

// from client, perhaps something like
OpenApiConnectorGenerator<?, ? generator = OpenApiConnectorGenerator.forSpecification(String specification);

// in the implementation
class OpenApiConnectorGenerator<T extends OasDocument, O extends OasOperation> {
  DataShapeGenerator<T,O> dataShapeGenerator;
  OpenApiDescriptorGenerator<T,O> descriptorGenerator;
  OpenApiPropertyGenerator<T, O> propertyGenerator;

  private static final OpenApiConnectorGenerator<Oas30Document, Oas30Operation> V3IMPL = new OpenApiConnectorGenerator(new Oas30DataShapeGenerator(), new Oas30DescriptorGenerator(), new Oas30PropertyGenerators());

  private OpenApiConnectorGenerator(DataShapeGenerator<T,O> generator, dataShapeGenerator, OpenApiDescriptorGenerator<T,O> descriptorGenerator,  OpenApiPropertyGenerator<T, O> propertyGenerator) {
    this.dataShapeGenerator = dataShapeGenerator;
    this.descriptorGenerator = descriptorGenerator;
    this.propertyGenerator = propertyGenerator;
  }

  public static <T extends OasDocument, O extends OasOperation> OpenApiConnectorGenerator<T, O> forSpecification(String specification) {
    JsonNode nodes = JsonUtils.reader().readTree(specification);
    OpenApiVersion openApiVersion = OpenApiModelParser.getOpenApiVersion(tree);

	if (OpenApiVersion == OpenApiVersion.V2) {
       // could be even cached in a static field
       return new OpenApiConnectorGenerator(new Oas20DataShapeGenerator(), new Oas20DescriptorGenerator(), new Oas20PropertyGenerators())
    } else if (OpenApiVersion == OpenApiVersion.V3) {
       // like here we cache it
       return V3IMPL;
    }
  }
}
```

Does that make sense, am I off base?

zregvart:
@christophd we can merge this and refactor if needed later, looks really good!

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/OpenApiConnectorGenerator.java
christophd: Ah thanks, more clear now. And I like it! Will add it along with some upcoming work on that OpenAPI 3 support.

christophd:
@zregvart cool, many thanks!

PR #7278, opened 2019-11-29
fix: improve enablement of virtualization publish action

indraraj:
- Jira Issue https://issues.jboss.org/browse/TEIIDTOOLS-854

- With this PR the behavior of the `Publish` button is modified.
  --Apart from looking at the empty virtualization now it's also checking if the virtualization contains any valid view.

- Screenshot
invalid_View

### DIFF
--- a/app/ui-react/syndesis/src/modules/data/pages/VirtualizationMetricsPage.tsx
+++ b/app/ui-react/syndesis/src/modules/data/pages/VirtualizationMetricsPage.tsx
@@ -1,7 +1,9 @@
-import { useVirtualization } from '@syndesis/api';
-import { DvMetricsContainer, PageSection } from '@syndesis/ui';
-import { useRouteData } from '@syndesis/utils';
+import { useViewDefinitionDescriptors, useVirtualization } from '@syndesis/api';
+import { DvMetricsContainer, PageSection, ViewListSkeleton } from '@syndesis/ui';
+import { useRouteData, WithLoader } from '@syndesis/utils';
 import * as React from 'react';
+import { ApiError } from '../../../shared/ApiError';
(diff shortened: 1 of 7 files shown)


app/ui-react/syndesis/src/modules/data/pages/VirtualizationMetricsPage.tsx
```diff
@@ -25,54 +27,74 @@ export const VirtualizationMetricsPage: React.FunctionComponent = () => {
...
   const { model: virtualization } = useVirtualization(params.virtualizationId);
 
+  /**
+   * Hook to obtain view descriptors. Also does polling to get any view descriptor updates.
+   */
+  const {
+    error,
+    hasData: hasViewDefinitionDescriptors,
+    model: viewDefinitionDescriptors,
+  } = useViewDefinitionDescriptors(params.virtualizationId);
```
elvisisking: Just doesn't *feel* right to me to have the metrics page fetch view information since it doesn't use that information. It is just being passed through. I'm wondering if you just move the `useViewDefinitionDescriptors` to the `VirtualizationEditorPage`. Then you wouldn't need the `haveValidView` property. And wouldn't need the `WithLoader` below either. Wdyt @indraraj @mdrillin ?

mdrillin: After thinking about this more, I requested a change to the backend - https://issues.jboss.org/browse/TEIIDTOOLS-866 .  I think this should be a pretty straightforward change, and would eliminate some cases of us needing to make the extra call to get viewDefinitionDescriptors.

mdrillin:
Closing the PR - decided not to handle enablement in this way.  We can re-open later if needed.

PR #7283, opened 2019-11-29
fix(ENTESB-12218): Fix request body extraction for OpenAPI 3.x

christophd:
Parameter defined as "in body" has been replaced by requestBody element in OpenAPI 3.x so we need to add specific logic to extract body schemas when working on OpenAPI 3.x

### DIFF
--- a/app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/DataShapeGenerator.java
+++ b/app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/DataShapeGenerator.java
@@ -22,8 +22,8 @@ import java.util.function.Predicate;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.apicurio.datamodels.openapi.models.OasDocument;
 import io.apicurio.datamodels.openapi.models.OasOperation;
-import io.apicurio.datamodels.openapi.models.OasParameter;
 import io.apicurio.datamodels.openapi.models.OasResponse;
+import io.apicurio.datamodels.openapi.models.OasSchema;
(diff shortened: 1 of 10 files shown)


app/server/api-generator/src/main/java/io/syndesis/server/api/generator/openapi/DataShapeGenerator.java
```diff
@@ -45,20 +45,25 @@
...
-        }
+    default Optional<NameAndSchema> findBodySchema(final O operation) {
+        return Optional.empty();
+    }
 
-        final List<OasParameter> operationParameters = operation.parameters;
+    /**
+     * Combination of schema and name for request body.
+     */
+    class NameAndSchema {
```
zregvart: perhaps?

```suggestion
    class NamedSchema {
```

nitpicking as always...

christophd: haha, I had exactly this `NamedSchema` first and renamed it into `NameAndSchema` right before the PR.

christophd: My thoughts were that it is a name and schema because in 2.x name refers to the parameter name and in 3.x to the media type name. Never it is the name of the schema itself. Ah I don't know what's better here

zregvart: Leave it as is, if it ends up bothering us we can rename it... (Naming is hard, remember)

PR #7287, opened 2019-12-02
fix(build): s2i build within a container

zregvart:
`maven.repo.local` is a legacy property there for compatibility with
Maven 1, and the replacement is `settings.localRepository. When running
the build in a container the `user.home` Java property was set correctly
but the property `maven.repo.local`, was resulting into `null` (checked
via `mvn help:evaluate`), while `settings.localRepository` evaulated to
correct path of the Maven local repository.

Not sure what circumstances necessitated the workaround in the POM of
the S2I module, this seems to be solving the problem more _correctly_.


zregvart:
> I wonder if it would be more appropriate instead of putting that information directly in the `settings.xml` where it really belongs and updating it accordingly.

Not sure about this, the `settings.xml` can't be controlled by us in all situations, like when someone builds and needs to use corporate Maven repository to access all artefacts.

squakez:
> > I wonder if it would be more appropriate instead of putting that information directly in the `settings.xml` where it really belongs and updating it accordingly.
> 
> Not sure about this, the `settings.xml` can't be controlled by us in all situations, like when someone builds and needs to use corporate Maven repository to access all artefacts.

You're right. We had to deal with PNC builds last time where they use their own `settings.xml`. It seems that the new variable is not used by `maven install` phase though. Finally the idea is that the S2I `install` will download all dependencies into `/target/generated-sources/docker/m2/repo` and that the `docker build` will copy those into its own local repository in order to install the `/target/generated-sources/docker/m2/project` generated.

I haven't found any other alternative than `maven.repo.local` or, as said before, playing with `settings.xml`.

zregvart:
@squakez I'll try to have it working with just the `settings.localRepository` I'll ping as I make progress. Thanks :)

zregvart:
@squakez can you have another look I've added a substantial refactor in c2cd75f535368159f9ebf321eec5e2df19edb575

squakez:
I like the idea to bring the script into maven phases. However I have a couple of concerns:

How should I test it locally? when executing the s2i build I got the following:
```
[INFO] F8: Using BuildServiceConfig syndesis-s2i-s2i for Docker strategy
[INFO] F8: Adding to ImageStream syndesis-s2i
[INFO] F8: Starting Build syndesis-s2i-s2i
[INFO] F8: Waiting for build syndesis-s2i-s2i-68 to complete...
[INFO] F8: Step 1/11 : FROM fabric8/s2i-java:3.0-java8
[INFO] F8:  ---> f36e68604d42
[INFO] F8: Step 2/11 : ENV AB_JOLOKIA_HTTPS "true"
[INFO] F8:  ---> Using cache
[INFO] F8:  ---> 4cdf2fe90692
[INFO] F8: Step 3/11 : USER 0
[INFO] F8:  ---> Using cache
[INFO] F8:  ---> 3f3680137791
[INFO] F8: Step 4/11 : COPY settings.xml /tmp/settings.xml
[INFO] F8:  ---> Using cache
[INFO] F8:  ---> 423a5980de80
[INFO] F8: Step 5/11 : COPY m2/repo /tmp/artifacts/m2
[INFO] F8: error: build error: lstat m2/repo: no such file or directory
[ERROR] F8: Failed to execute the build [Unable to build the image using the OpenShift build service]
```
Then, I see that the PNC profiles we introduced to skip the maven repo generation and project generation are not taken in consideration. This is the ENTESB thread fyi: https://issues.redhat.com/browse/ENTESB-11682

zregvart:
> How should I test it locally?

I've run `syndesis build -m s2i --docker` to get the image built using the local docker, or `syndesis build -m s2i -i` to trigger binary S2I image build on OpenShift. Either way the fabric8-maven plugin's [default is `auto`](https://maven.fabric8.io/#build-mode), so that might cause it to detect `$HOME/.kube` (I think that's what it uses) profile and auto-configure for OpenShift build.

> Then, I see that the PNC profiles we introduced to skip the maven repo generation and project generation are not taken in consideration

The image generation is only performed when the `image` profile is active. From the description of the issue this is exactly what's requested in ENTEB-11682. @orpiske @cunningt would this interfere with the way the build is performed in PNC, or to put it plainly how do you build the S2I image in PNC without building it using the `image` profile?

squakez:
Okey, it worked without using `-f` and `-c` flag. You may check to see why `syndesis build -f -c -i -m s2i` is failing.

More important, I noticed that the integrations built upon the new S2I images are hitting the maven public repo. I used an integration and published with old and new S2I:

OLD maven lookup: [old_s2i_build.log](https://github.com/syndesisio/syndesis/files/3927831/old_s2i_build.log)
NEW maven lookup: [new_s2i_build.log](https://github.com/syndesisio/syndesis/files/3927832/new_s2i_build.log)

This may be problematic on low speed networks. If you recall, we had a chat some month ago with @asmigala thanking the new speed performance improvement for the integrations build, I think we should investigate and keep it to the improved standard.

I think that's because we're now building the base project outside the docker image, so there may be dependencies and pom descriptor that need to be refreshed.

orpiske:
> > How should I test it locally?
> 
> I've run `syndesis build -m s2i --docker` to get the image built using the local docker, or `syndesis build -m s2i -i` to trigger binary S2I image build on OpenShift. Either way the fabric8-maven plugin's [default is `auto`](https://maven.fabric8.io/#build-mode), so that might cause it to detect `$HOME/.kube` (I think that's what it uses) profile and auto-configure for OpenShift build.
> 
> > Then, I see that the PNC profiles we introduced to skip the maven repo generation and project generation are not taken in consideration
> 
> The image generation is only performed when the `image` profile is active. From the description of the issue this is exactly what's requested in ENTEB-11682. @orpiske @cunningt would this interfere with the way the build is performed in PNC, or to put it plainly how do you build the S2I image in PNC without building it using the `image` profile?

If I understand it correctly, what we do is, we generate the s2i sample on PNC:

```mvn -e -V -B - ... -Ds2i.sample-generation.skip=false clean deploy -f app/pom.xml ```

During the image build on OSBS, we download the s2i maven repo [1], include it into the image [2] and then use this environment [3] to create the s2i image contents.


1. https://github.com/jboss-fuse/fuse-ignite-s2i/blob/private-devel-fuse-7.6-openshift-rhel-7/build-image.sh#L163
2. https://github.com/jboss-fuse/fuse-ignite-s2i/blob/private-devel-fuse-7.6-openshift-rhel-7/Dockerfile#L15
3. https://github.com/jboss-fuse/fuse-ignite-s2i/blob/private-devel-fuse-7.6-openshift-rhel-7/Dockerfile#L28

**Comment by zregvart** [comment]
Okay, so I think I know what needs to change to make this work, I'll outline this below. Do we agree, though, that it needs to change? This started when I found issues in building behind a corporate Maven repository, similar to a fully offline case. The idea is to honour the existing `settings.xml` of the Maven build that builds the S2I Maven module in building the repository. I've missed a couple of things in this.

 1. I'm not 100% sure that the proper settings.xml is used, need to double check that
 1. `flash` profile doesn't build the image (will fix)
 1. the repository assembly needs to work
 1. I need to double check the repository content and bundled settings.xml in the S2I

Anything else I missed?

**Comment by zregvart** [comment]
@squakez I think I've addressed the issues in f0cfa877c803558fa62435ad4de79d609a58cd66 can you one more look?

**Comment by squakez** [comment]
> @squakez I think I've addressed the issues in [f0cfa87](https://github.com/syndesisio/syndesis/commit/f0cfa877c803558fa62435ad4de79d609a58cd66) can you one more look?

Nice, it seems it retrieves all dependencies locally now. Just a minor thing you may further look into it is that I got the following warnings when building the S2I:
```
[WARNING] No license information available for: io.syndesis.connector:connector-support-catalog:1.9-SNAPSHOT
[WARNING] No URL for license at index 0 in dependency net.iharder:base64:2.3.8
[WARNING] Unable to retrieve license from URL 'https://glassfish.dev.java.net/nonav/public/CDDL+GPL.html' for dependency 'org.glassfish:javax.el:3.0.0': Certificate for <glassfish.dev.java.net> doesn't match any of the subject alternative names: [*.java.net, java.net]
[WARNING] There were 4 download errors - check the warnings above
```

PR #7289, opened 2019-12-02
Aligning DV with syndesis parent properties and cleaning up no longer used properties

vhalbert:


vhalbert:
This aligns DV with 2.1.6.RELEASE spring-boot

rareddy:
@vhalbert note to change the Teiid Spring Boot too.

PR #7290, opened 2019-12-02
More updates regarding  DV alignment with syndesis

vhalbert:


app/dv/pom.xml
```diff
@@ -236,8 +233,8 @@
     <!-- Common dependencies -->
     <version.springfox>2.9.2</version.springfox>
 
-    <version.org.mockito>2.21.0</version.org.mockito>
-    <version.com.fasterxml.jackson>2.9.9</version.com.fasterxml.jackson>
+    <version.org.mockito>${mockito.version}</version.org.mockito>
+    <version.com.fasterxml.jackson>${jackson.version}</version.com.fasterxml.jackson>
```
zregvart: I kinda like the `version.groupId` style of property naming better than what we use currently in the Syndesis parent POM.

vhalbert: I agree.   As I make any further changes, I'll try to transition to this format.

Issue #7253, opened 2019-11-27
Unable to install via `syndesis minishift --install`


zregvart:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

If I run `syndesis minishift --install` the script will be stuck in:

```
Waiting for syndesis-operator to be scaled to 1
Sleeping 10s ...
Sleeping 10s ...
```

This is because the DeploymentConfig for syndesis-operator is pointing to the internal OpenShift image registry which at that point cannot have the tag set in the `syndesis-operator` binary.

As a workaround one can build the operator using `./build.sh --image-build s2i`. And that will create the `latest` tag in the internal OpenShift image registry. So patching the syndesis-operator DeploymentConfig is also needed:

```
$ oc patch dc syndesis-operator --type json -p '[{"op": "replace", "path": "/spec/triggers/0/imageChangeParams/from/name", "value": "syndesis-operator:latest"}]' # replace the tag 1.9.100-2019... with latest

```

PR #7304, opened 2019-12-03
Set right tag when installing operator in dev mode

lgarciaaco:
If devMode is set, the ImageStreamTag should point to latest independently from the tag that the operator has.

Fixes https://github.com/syndesisio/syndesis/issues/7253


heiko-braun:
A newbie question again: Why are contents of the `vendor` directory commited ?

zregvart:
> A newbie question again: Why are contents of the `vendor` directory commited ?

I guess the [pros outweigh the cons](https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory).

heiko-braun:
Thanks @zregvart

PR #7310, opened 2019-12-03
Backport fix for thrift dependency issue & fix for ENTESB-12350

phantomjinx:


install/operator/pkg/apis/syndesis/v1alpha1/syndesis_types.go
```diff
@@ -73,7 +73,8 @@ type ComponentsSpec struct {
 }
 
 type OauthConfiguration struct {
-	Tag string `json:"tag,omitempty"`
+	DisableSarCheck *bool  `json:"disable-sar-check,omitempty"`
```
lgarciaaco: Can we use camel here?

heiko-braun: You mean `DisableSarCheck`, rather than `disable-sar-check`?

heiko-braun: I think we should use the same property as in 1.7.x

lgarciaaco: Ok I meant dromedary case: `disableSarCheck`. All properties for CR use that case

lgarciaaco: Now I see in master it is also called `disable-sar-check` and I can't find that property in 1.7.x :thinking:

phantomjinx: Consistancy with master I think should be kept.

lgarciaaco: I will change master ...

phantomjinx: Done.

phantomjinx:
No need to port to 1.9.x branch. Associated code already present.

PR #7312, opened 2019-12-03
Adding dv module to build

vhalbert:


zregvart:
Looks good, just to doublecheck DataVirt doesn't depend on the server, I thought I saw it depend on the server in the CircleCI build. If it does then we need to codify that dependency here:

https://github.com/syndesisio/syndesis/pull/7312/files#diff-c4dddee4e8b7a9a24e8d861b3dd7ba68L278

vhalbert:
> Looks good, just to doublecheck DataVirt doesn't depend on the server, I thought I saw it depend on the server in the CircleCI build. If it does then we need to codify that dependency here:
> 
> https://github.com/syndesisio/syndesis/pull/7312/files#diff-c4dddee4e8b7a9a24e8d861b3dd7ba68L278

No DV doesn't depend on server.

PR #7327, opened 2019-12-04
fix(TEIIDTOOLS-867) - Preview query fails when view name has a dash

elvisisking:
- see [TEIIDTOOLS-867](https://issues.jboss.org/browse/TEIIDTOOLS-867)
- if view name has non-alphanumeric characters the view name is wrapped in double quotes before requesting the preview results


elvisisking:
I'll fix as you recommend @shawkins

PR #7350, opened 2019-12-06
TEIIDTOOLS-878 fixing swagger regression

shawkins:


shawkins:
All the refactoring ended up breaking the swagger integration.  Changed to reference the package name from the class rather than using a literal.

Issue #7355, opened 2019-12-09
Use GnuGrep on OS X


KurtStam:
The native BSD grep is slightly different from GNU grep that the rest of the world uses. This causes some script commands to fail.

 (build.sh on OS X #6976)

PR #7356, opened 2019-12-09
fix Use GnuGrep on OS X #7355

KurtStam:


tools/bin/commands/util/openshift_funcs
```diff
@@ -13,8 +13,10 @@ check_oc_version()
     # Removes any lines containing kubernetes or server
     # Extracts any version number of the format dd.dd.dd, eg. 3.10.0 or 4.1.0
     #
-    local test=$(oc version | grep -Eiv 'kubernetes|server' | grep -o '[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}\?')
+    set_gnugrep
 
+    local test=$(oc version | $gnugrep -Eiv 'kubernetes|server' | $gnugrep -o '[0-9]\{1,\}\.[0-9]\{1,\}\.[0-9]\{1,\}\?')
```
zregvart: I'm thinking rather than introducing the dependency on gnu grep, and make OSX folk install it if they don't have we could perhaps eliminate the need upon it. Would that be better? We could probably parse this out differently. For example for OpenShift 4.2.x there's a much nicer `oc version` implementation (https://github.com/openshift/oc/pull/5).

zregvart: I think `oc version|awk '/^oc v/ {sub(/^oc v/, ""); sub(/^oc v/, ""); sub(/\+.*/, ""); print}/Server Version/{print $3}'` should work. I'll create a PR...

PR #7369, opened 2019-12-10
ENTESB-12222: Implement jaeger backend in syndesis

phantomjinx:
* install_cluster.go
 * Look for 'cluster' sub-directory in addons and install those artifacts
   since they require a cluster-admin account

* configuration.go
 * Moves the gathering of addons into a single place rather than having
   their extraction distributed in multiple places


phantomjinx:
> I see this is installing the Jaeger operator, wouldn't we get this through the OLM? I thought there was a mechanism for one operator to depend on another one.

That would be relying on an external mechanism to facilitate the out-of-the-box experience. If a more sophisticated approach is required for jaeger then the user is free to install their own version (from OLM). That's said probably need to split the addon into jaeger-agent and jaeger-collector?

@lgarciaaco WDYT?

lgarciaaco:
> I see this is installing the Jaeger operator, wouldn't we get this through the OLM? I thought there was a mechanism for one operator to depend on another one.

OLM doesn't have such a mechanisms, sadly

Issue #7241, opened 2019-11-27
Unable to clone syndesis git repo due to question marks in file names ( Invalid argument )


gelonsoft:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
Could you please rename all files and directories with question marks ("?") to simplify syndesis development/contribution on Windows OS. 
By default you can't use "?" in file names or directory names ( also the following reserved chars: <>:"/\|?*  https://docs.microsoft.com/ru-ru/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN ).
<pre><code>
$ git clone git@github.com:syndesisio/syndesis.git
Cloning into 'syndesis'...
remote: Enumerating objects: 35, done.
remote: Counting objects: 100% (35/35), done.
remote: Compressing objects: 100% (30/30), done.
remote: Total 353378 (delta 10), reused 13 (delta 4), pack-reused 353343
Receiving objects: 100% (353378/353378), 137.96 MiB | 3.93 MiB/s, done.
Resolving deltas: 100% (186462/186462), done.
fatal: cannot create directory at 'app/ui-react/syndesis/tapes/create-connection.spec.js/api/v1/connections?per_page=50': Invalid argument
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'
</code></pre>

gashcrumb:
@kahboom did you say there was a way we could override the filenames for those integration tests?

kahboom:
@gelonsoft - Hey! Please retest and let me know if you still have any issues. Going to see if I can find a way to test on Windows as well.

gelonsoft:
@kahboom , still have issue cause ":" is also reserved char.
------------------
fatal: cannot create directory at 'app/ui-react/syndesis/tapes/create-integration.spec.js/api/v1/connections/i-LkiT24SvERwF0KppRNIz/actions/i-LkAI0ut-OrgXelAq8JAz:4babf99c-6287-47bb-862c-da7d053338af': Invalid argument
------------------

PR #7370, opened 2019-12-10
fix(UI tests): Prevent problematic characters in naming of test tape files

kahboom:
fixes https://github.com/syndesisio/syndesis/issues/7241

- Adds a regex pattern to replace troublesome characters across platforms.
- Removes offending tapes that currently exist, adds a few with the new naming format.


matskiv:
@kahboom Thanks a lot for fixing this. I literally just bumped into the issue with `?` character in the file name. My use case involved go mod to import 1.8.x branch of the operator ([logs](https://gist.github.com/matskiv/05d1fe01fbd9153186488f5128a95a0a)).

Would it be possible to get this change into 1.8.x branch? If it can be done with just a cherry-pick - let me know, I can send a PR with that.

kahboom:
@matskiv No problem!

Am I understanding correctly that you'd just want it in the 1.8.x branch for your use case, so nothing to do with it being included in the next release of Fuse? If you're intending for the Fuse 7.5.1 release, I think it might be a bit too late, but I'm not sure--maybe @heiko-braun knows?

On a side note--Heiko, if you can also clarify if it's still ok to backport into the 1.8.x branch, and what exactly it correlates to in terms of product (for future reference), that'd be great.

matskiv:
@kahboom Correct, my use case is not related to next/current Fuse patch release.

Just to give you a full picture - we are importing syndesis operator as a dependency for integreatly operator. We do that just to use structs defined in `install/operator/pkg/apis/syndesis/v1alpha1`. And I want to import branch of syndesis that matches minor version of Fuse we are currently running, and for Fuse 7.5 that would be syndesis 1.8.x, right? :)

kahboom:
@matskiv - Ah, alright I understand. Looks like yes, you'd have to cherry pick and create a PR against 1.8.x. Since I still don't know who determines what can go into `1.8.x`, I'd probably ping Gary and/or Heiko when you open it just to make sure.

matskiv:
Will do. Thanks a lot 👍

PR #7383, opened 2019-12-11
feat(connector): enable meta dynamic properties

squakez:
* Adding a new method in the MetadataRetrieval interface to allow any Connector to expose a list of dynamic properties
* Fixing a few unit test that were not using getter accessor method

Closes ENTESB-12482


squakez:
Discarded

PR #7395, opened 2019-12-12
fix: Remove embedded camel-jira component

christophd:
... and use the one provided with 2.23.2.fuse-760009

Should fix duplications that cause some issues in prod builds.


claudio4j:
Other files should be camel version updated, take a look at `tools/misc/syndesisCamelVersion.sh`
install/operator/pkg/generator/assets_vfsdata.go

christophd:
@claudio4j thanks! Hopefully I have changed all camel versions in repo now. 

@lgarciaaco can you double check for operator sources please

PR #7397, opened 2019-12-12
TEIIDTOOLS-883, TEIIDTOOLS-865: Fixing the resource leak for http con…

rareddy:
…nection leak and upgrading the openshift client libraries to 4.x


rareddy:
Tab issues got fixed.

> may be happening to use the same instance when it is closed?
I could convert to create a syndesis client each time, I figured when in exception it is more or less due to the http client, so even other one will fail anyways.

> The pom changes other than any of the api version changes should probably be separated
It is hard to see with to too with all clutter, IMO if there is another JIRA for it, I can tag that into the commit comment.

rareddy:
@shawkins 
> may be happening to use the same instance when it is closed?

Handled the threading issues, by creating the connection per use and close right after the response is read. Also renamed the file SyndesisHttpUtil --> SyndesisHttpClient to be more representative.

> The pom changes other than any of the api version changes should probably be separated

Added the TEIIDTOOLS-752 in the comment to refer to the original issue around the move to cover the references.

rareddy:
I believe indentation may be due mix of tab/spaces, we should enforce space only IMO

shawkins:
We are not picking up all of the syndesis checkstyle rules.  They have tabs set to the warning level.

christophd:
@rareddy backport bot is not working yet for 1.9 afaik. Can you please check and backport manually if applicable. Thanks!

rareddy:
@christophd thanks, I will do manual push/merge then.

zregvart:
@rareddy the configuration of the GitHub action is taken at the point the PR was created (from the HEAD of the PR). So the new backport GitHub action is not considered for this PR. I don't see this commit present on the `1.9.x`, so you'll need to perform a manual backport.

rareddy:
@zregvart I added it here https://github.com/syndesisio/syndesis/commit/50f7374df97818fd0ce6f85cf4ea7caec807b549

Is this not the right spot?

PR #7409, opened 2019-12-13
[ENTESB-12331] Adding auto-discovery for strimzi kafka/AMQ brokers

Delawen:
This takes the kubernetes configuration from the environment and tries to discover brokers. All plain discoverable brokers will be passed to the Kafka connector to auto-populate some suggestion dropdown box.

This needs some manual tweaking on privileges as the default environment user does not have privileges to discover any broker. Will just return an empty list.

### DIFF
--- a/app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaMetaDataRetrieval.java
+++ b/app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaMetaDataRetrieval.java
@@ -15,45 +15,124 @@
  */
 package io.syndesis.connector.kafka;
 
-import io.syndesis.connector.support.verifier.api.ComponentMetadataRetrieval;
-import io.syndesis.connector.support.verifier.api.PropertyPair;
-import io.syndesis.connector.support.verifier.api.SyndesisMetadata;
-import org.apache.camel.CamelContext;
-import org.apache.camel.component.extension.MetaDataExtension;
-
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Predicate;
+import io.fabric8.kubernetes.client.DefaultKubernetesClient;
+import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.dsl.base.CustomResourceDefinitionContext;
+import io.syndesis.connector.support.verifier.api.ComponentMetadataRetrieval;
+import io.syndesis.connector.support.verifier.api.PropertyPair;
+import io.syndesis.connector.support.verifier.api.SyndesisMetadata;
+import io.syndesis.connector.support.verifier.api.SyndesisMetadataProperties;
+import org.apache.camel.CamelContext;
+import org.apache.camel.component.extension.MetaDataExtension;
 
 public class KafkaMetaDataRetrieval extends ComponentMetadataRetrieval {
+
+    /**
+     * Used to filter which types of connections are we interested in. Right now, only plain connections.
+     */
+    private final Predicate<? super Map<?, ?>>
+        typesAllowed = listener -> listener.get("type").toString().equalsIgnoreCase("plain");
 
     /**
      * TODO: use local extension, remove when switching to camel 2.22.x
      */
     @Override
-    protected MetaDataExtension resolveMetaDataExtension(CamelContext context, Class<? extends MetaDataExtension> metaDataExtensionClass, String componentId, String actionId) {
+    protected MetaDataExtension resolveMetaDataExtension(CamelContext context,
+                                                         Class<? extends MetaDataExtension> metaDataExtensionClass,
+                                                         String componentId, String actionId) {
         return new KafkaMetaDataExtension(context);
     }
 
     @SuppressWarnings("unchecked")
     @Override
-    protected SyndesisMetadata adapt(CamelContext context, String componentId, String actionId, Map<String, Object> properties, MetaDataExtension.MetaData metadata) {
+    protected SyndesisMetadata adapt(CamelContext context, String componentId, String actionId,
+                                     Map<String, Object> properties, MetaDataExtension.MetaData metadata) {
         try {
             Set<String> topicsNames = (Set<String>) metadata.getPayload();
 
             List<PropertyPair> topicsResult = new ArrayList<>();
-            topicsNames.stream().forEach(
+            topicsNames.forEach(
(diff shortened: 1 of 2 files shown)


heiko-braun:
@Delawen for time being you need to backport manually, the bot doesnt work atm

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaMetaDataRetrieval.java
```diff
@@ -15,45 +15,124 @@
...
+        List<PropertyPair> brokers = new ArrayList<>();
+        try (KubernetesClient client = new DefaultKubernetesClient()) {
+
+            //Filter by strimzi resources
+            final CustomResourceDefinitionContext build = new CustomResourceDefinitionContext.Builder()
+                                                              .withGroup("kafka.strimzi.io")
+                                                              .withName("kafkas.kafka.strimzi.io")
+                                                              .withPlural("kafkas")
+                                                              .withScope("Namespaced")
+                                                              .withVersion("v1beta1")
```
christophd: Can we make `version` configurable or make the user supply the version instead of hardcoding it?

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaMetaDataRetrieval.java
```diff
@@ -15,45 +15,124 @@
...
+                                                              .build();
+
+            final List<Map<String, ?>> items = (List<Map<String, ?>>) client.customResource(build).list().get("items");
```
christophd: I would try to introduce a CustomResource model for the Kafka resource and let the Kubernetes client deserialize to the model object. 

Something like
```java
@JsonDeserialize
public class KafkaResource extends CustomResource { // model properties }
```

```java
KubernetesDeserializer.registerCustomKind("kafka.strimzi.io", "Kafka", KafkaResource.class);

CustomResourceDefinition crd = client.customResourceDefinitions().withName("kafkas.kafka.strimzi.io").get();
client.customResources(crd, KafkaResource.class, KafkaResourceList.class, KafkaResourceDoneable.class); 

List<KafkaResource> items = client.list().getItems();
for (KafkaResource item : items) {
  // do something with model resource
}
```

You need to add some boilerplate code for the Kubernetes client but we can then access the custom resource properties via model which should be easier rather than accessing the field names in the maps.

You can use this repo as an example for the boilerplate code to add: https://github.com/fstab/devoxx-be-2019

christophd:
Merge bot already merged it but you can consider to add this in a separate PR when you have time.

Delawen:
Sure!

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaMetaDataRetrieval.java
Delawen: The query requires a version to be supplied. You mean choosing from the UI? That may be not very understandable for the final user.

I am going to check if I can query a list of versions too...

christophd: maybe we do not need the version anymore when refactoring to model deserialization and fetching the custom resource definition with the client.

Delawen: Aaah I left the other suggestion for later because it looked harder.

zregvart: Version is part of the request URL when accessing the CRDs resource via k8s API. We could list and detect what CRDs are present and choose version from that.

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaMetaDataRetrieval.java
zregvart: I'm not sure if depending on [strmizi-api](https://search.maven.org/artifact/io.strimzi/api/0.15.0/jar) would be a good thing; the classes are already defined there. There's also a `Crds` utility class that makes listing existing `Kafka` `CR`s easy:

```java
try (KubernetesClient client = new DefaultKubernetesClient()) {
    Crds.kafkaV1Alpha1Operation(client).list()
        .getItems()
        .forEach(k -> k.getStatus().getListeners()
            .forEach(l -> l.getAddresses()
                .forEach(a -> System.out.println(a.getHost() + " " + a.getPort()))));
}
```

heiko-braun: > depending on strmizi-api 

It would help to avoid this dependency. It creates an undesireable level of coupling

heiko-braun: unless there is a compelling reason to keep it

Delawen: The model used in strimzi is extensive and complex, if we follow this path fully it may make sense to use those model packages.

But maybe we don't need the full packages. At least not for the MVP we are working on, just listing available brokers with not much more info.I am trying to avoid this extra dependency... at least for now.

valdar: Also, remember that the Strimzi package has is own dependencies of Kubernetes client which add a cross product version sync potential source of troubles...

christophd: My idea was that we can use our own KafkaResource with a subset of properties that we are interested in and let the Kubernetes client deserialize the custom resource into that model

zregvart: By referencing a Custom Resource Definition from another project we're already creating a dependency upon that project, this is manifested in us relying on the data being in places we expect it to be. We can pretend that this dependency doesn't exist by creating our own classes to deserialize or we can reuse the classes as _dumb_ data transfer objects, but that doesn't mean we don't depend on the other project in question.
I would depend on the `strimzi-api` but in a way where we're only using the data transfer objects from it. All other dependencies can be excluded and the classes can be shaded within our package limiting our API exposure _only_ to the data transfer object classes.
This makes the dependency explicit, as in we're depending on a specific version of Strimzi and if we change the version we upgrade the data transfer objects and we utilize the Java compiler to enforce the match between the Custom Resource Definition and the classes we use to deserialize.

PR #7410, opened 2019-12-13
ENTESB-12330: Allow for specifying a custom resource on install

phantomjinx:


install/operator/pkg/cmd/internal/install/install.go
```diff
@@ -99,6 +99,7 @@ func New(parent *internal.Options) *cobra.Command {
 		},
 	}
 	app.PersistentFlags().StringVarP(&o.addons, "addons", "", "", "a comma separated list of addons that should be enabled")
```
lgarciaaco: Can't we remove the `addons` option now?

phantomjinx: Probably. I'll put it in on the TODO list.

PR #7428, opened 2019-12-17
chore(jaeger-addon): Remove addon cluster directory

phantomjinx:
* Revert the addon cluster logic to install the jaeger cluster artifacts
  as standard, ie. moving them to the cluster directory

* generator.go
 * Simplify API so that calling Render function, if presented with a
   directory, will delegate to RenderDir function

* install.go
 * Removes renderDir as no longer required (see generator)

* cluster.yml
 * Refactors to a cluster directory since file is becoming too long
 * Adds in jaegar cluster artifacts

* install_cluster.go
 * Removes addon cluster resources logic and installs everything from the
   cluster directory


heiko-braun:
@phantomjinx Does it need to be backported to 1.9.x?

PR #7433, opened 2019-12-17
[ENTESB-12331] Port from master

Delawen:
This contains #7432 and #7404 

Manual port because automatic port does not seem to work.


Delawen:
Closing by now until we know if we want this on 1.9.x.

PR #7441, opened 2019-12-18
[operator] Add backup

lgarciaaco:
# Enable periodical and one time backups.

A backup contains syndesis openshift resources as well as a database backup. The `backup` packages allows for backups to be upload to an external datastore. At the moment, only `S3` is supported, but I plan on adding `rsh` next.

### Backup
To perform a backup from command line just run `syndesis-operator backup`. This will create a backup folder with all backed up resources inside.

Uploading to external datastores is disabled with this mode.

You can set the output directory with `--backup dir`
 ```
syndesis-operator backup --backup /tmp/backup
```

### Restore
To restore, simply run
 ```
syndesis-operator restore --backup /tmp/backup
```

`restore` also takes a zipped backup as an argument.

## Running periodical backups
To enable periodical backups, fill up the backup section of the CR.
```
spec:
  backup:
    schedule: daily
```
`schedule` accepted values are: hourly;daily;midnight;weekly;monthly;yearly

`schedule` is a set once property, changing its value wont yield any result.

To disable periodical backups, leave in blank the backup section in the CR.
### Enable S3 as external datastore
To enable S3 as external datastore you must create a secret as following:
```
kubectl create secret generic syndesis-backup-s3 \
  --from-literal=secret-access-key='replace with access key'  \
  --from-literal=secret-key-id='replace with key id' \
  --from-literal=region='replace with region' \
  --from-literal=bucket-name='replace with bucket name'
```


lgarciaaco:
Fixes https://issues.redhat.com/browse/ENTESB-12437
Fixes https://issues.redhat.com/browse/ENTESB-12440

zregvart:
[Nicer commit messages](https://syndesis.io/community/engineering_guidelines/#commit-messages-should-follow-the-conventional-commits-standardized-commit-message-format)?

heiko-braun:
With this:

```
spec:
  backup:
    schedule: daily
```

Where does the backup go?

lgarciaaco:
> With this:
> 
> ```
> spec:
>   backup:
>     schedule: daily
> ```
> 
> Where does the backup go?

If you specify the secret for S3, to S3, if not, nowhere

install/operator/pkg/syndesis/action/backup.go
```diff
@@ -0,0 +1,93 @@
...
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"k8s.io/client-go/kubernetes"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	cron "github.com/robfig/cron/v3"
```
zregvart: Would it make more sense to have a [k8s CronJob](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/), instead of the backup running within the operator process?

lgarciaaco: no because I need to spin off a docker container for it

zregvart: @lgarciaaco can you elaborate that? You can spin the same image as the one used by the operator with perhaps a different entrypoint/arguments?

lgarciaaco: Imho it is harder to work with pods than this way where the backup is tided up to the running binary. It is harder to develop and to debug. Why don't you tell me the benefits of doing it with a k8s cron job over using a cron withing the operator?

zregvart: I see backup and configuration as two separate concerns, for me it makes more sense to have them separate as any issues with the backup will be limited to the backup `CronJob`, likewise any issues with the operator would not jeopardize the backup. `CronJob` also gives the operators more flexibility than a feature we develop on our own, and matching existing features in our codebase unnecessary brings on more code to maintain.

install/operator/pkg/apis/syndesis/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,6 +1,6 @@
 // +build !ignore_autogenerated
 
-// Code generated by deepcopy-gen. DO NOT EDIT.
+// Code generated by operator-sdk. DO NOT EDIT.
```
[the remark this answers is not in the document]
lgarciaaco: Yea it also bother me ... no idea why

lgarciaaco:
> LGTM.
> May wanna squash the commits then add a single commit message?

Yea I was planning to do so as per Zoran comment

install/operator/pkg/syndesis/action/backup.go
lgarciaaco: Those are arguable ... 
> as any issues with the backup will be limited to the backup CronJob, likewise any issues with the operator would not jeopardize the backup. CronJob

If the operator is jeopardize do I want to continue running backups? maybe syndesis is jeopardize all together? this seems to a bigger decision that requires some further discussion

>  also gives the operators more flexibility than a feature we develop on our own

Right now a feature I developed on my own allows me to do all I need. I don't need to use something different for the promise that in the future I might (or well might not) use some of it features. If I need those feature whatever it might be I can swap then, if the cron package I am using can't provide it.

Either way, my point is that using this library makes my development and test circle easier. I don't see any value using k8s cron over it, but if more engineers think it is better I will gladly change it

avano:
> I plan on adding rsh next.

rsh where?

install/operator/pkg/syndesis/action/backup.go
zregvart: > Those are arguable ...

I'm sorry, they're not, apart from the extend that you're willing to argue about them.

If the operator is failing because, say, it hits memory limits and gets OOMed, the backups are not run. Likewise if the backup ends up crashing the operator any pending configuration changes will not be applied.

Having the operator run the backups functionality that jeopardizes the backups and the rest of the operator functionality.

I'm okay with this getting merged to `master` and further improved upon, I'm not okay with this getting released in 1.9 as I think there will be issues in operating it.

lgarciaaco:
> > I plan on adding rsh next.
> 
> rsh where?

Well I though it might bring some value if you can create a pod with a pvc and scp to it, as an alternative. The method would take the namespace, labels to select the pod, path to copy the backup over

install/operator/pkg/syndesis/action/backup.go
lgarciaaco: I am not willing to argue about anything. As I said, if other see a value right now or If we figure we actually need it, I will gladly change it.

avano:
> > > I plan on adding rsh next.
> > 
> > 
> > rsh where?
> 
> Well I though it might bring some value if you can create a pod with a pvc and scp to it, as an alternative. The method would take the namespace, labels to select the pod, path to copy the backup over

sure, I was just curious. Would this also be a part of upgrade process? Or will the backups be completely separate?

lgarciaaco:
> sure, I was just curious. Would this also be a part of upgrade process? Or will the backups be completely separate?

this is what I plan to use for the upgrade backup and rollback, and actually the `rsh` datastore to failover to the db pod(which has a pvc) in case there is no S3 enabled, as it will be the case for most of the people using 7.5.

PR #7443, opened 2019-12-18
ENTESB-12385: Disable extensions if camelk addon is enabled

phantomjinx:
* If camelk addon has been enabled then extensions are not available so
  specify them as 'false' in the syndesis-ui configmap


heiko-braun:
Great, thanks @phantomjinx !

PR #7444, opened 2019-12-18
TEIIDTOOLS-896 including more information in the dv-info file

shawkins:


shawkins:
@mdrillin see if this makes things clearer in terms of versioning.  If you approve, then we'll also include in 1.9.x to align the exportVersion property name.

PR #7446, opened 2019-12-18
feat(ENTESB-12479) - support for creatable typeahead options in autoform

kahboom:
plus a few package updates.

- ~~update patternfly 4.x  + related across the app (no changes to 3.x and corresponding packages), formik to get `isCreatable` typeahead configuration~~ 
- add typeahead custom autoform widget + support for `type: typeahead` prop in connectors
- update Connection Detail page to reflect changes for typeahead
- minor changes/improvements

Video displaying functionality: https://drive.google.com/file/d/1oRhxjODuF8_LwW8m-v9b1je8dxrAofrd/view?usp=sharing

Proper typing and DRYer code coming soon to a theater near you..

relates to [ENTESB-12479](https://issues.redhat.com/browse/ENTESB-12479) & failing build on https://github.com/syndesisio/syndesis/pull/7366


kahboom:
Some progress. Still testing without `enum` list of brokers, but at least it's loading the right widget. Going to do a test story or mock with the expected payload in a moment.

Screenshot 2019-12-18 18 30 54 copy

gashcrumb:
Did you have a look at how `dataList` can be used for typeahead support on
a text field?


https://github.com/syndesisio/syndesis/blame/master/app/ui-react/packages/auto-form/README.md#L41

And example at:

https://github.com/syndesisio/syndesis/blob/master/app/ui-react/packages/auto-form/stories/SimpleExamples.stories.tsx#L76

Maybe it'd be good to have it as a different type, I dunno.  But anyways
that gives you typeahead on a text field.

kahboom:
@gashcrumb - Hey! Yeah I saw that, but we need PF's option to create a new option, not just typeahead. See [here](https://www.patternfly.org/v4/documentation/react/components/select#typeahead).

We were back and forth whether to support both typeahead + typeahead with a "create  option" from the same `type` property and widget, or maybe adding a flag, but for this POC we decided to just do typeahead with a "create option", especially since we already support typeahead in the places you pointed out. We should consolidate those two in the upcoming sprint.

gashcrumb:
Ah, I see, yeah that makes sense.  Carry on!

app/ui-react/yarn.lock
```diff
@@ -3176,10 +3176,10 @@
     universal-user-agent "^2.0.0"
     url-template "^2.0.8"
 
-"@patternfly/patternfly@2.43.1":
-  version "2.43.1"
-  resolved "https://registry.yarnpkg.com/@patternfly/patternfly/-/patternfly-2.43.1.tgz#9961e331ac78f2cf55a73527656503f5555b053e"
-  integrity sha512-dWw4FeQfLHFXva1VmH8Epusu3VX91VrOH+9FVvRDsCEcajsq5Pc5BCF+H9KyQ4eA6EfTEMsW2F0zVWD/7B4RFA==
+"@patternfly/patternfly@2.40.6", "@patternfly/patternfly@2.43.1":
+  version "2.40.6"
```
zregvart: I thought we were going to need to upgrade the Patternfly version, not downgrade. Am I missing something?

kahboom:
closing temporarily to sort out dependency issues with the PF-react upgrade

app/ui-react/yarn.lock
kahboom: Not touching PF for now, looks like we've gotten the version we need for this feature to work.

PR #7474, opened 2019-12-23
fix(endpoint): all methods for reflective sorting

### LINKED ISSUE #7471: [Usage Problem]The integration api with sort params
Hello guys, I try to use the integration API to fetch list of Integrations which the URL path is `/api/v1/integrations`, But I got a problem

When I use this api with `sort` params like `/api/v1/integrations?sort=version`, then the response is  **Cannot find field %s in %s as int or String field** 

I have already checked the IntegrationHandler, from L185-L194, here's the code

```java

    @Override
    public ListResult<IntegrationOverview> list(final UriInfo uriInfo) {
        final DataManager dataManager = getDataManager();
        final ListResult<Integration> integrations = dataManager.fetchAll(Integration.class,
            new ReflectiveSorter<>(Integration.class, new SortOptionsFromQueryParams(uriInfo)),
            new PaginationFilter<>(new PaginationOptionsFromQueryParams(uriInfo)));

        return ListResult.of(integrations.getItems().stream().map(i -> integrationOverviewHelper.toCurrentIntegrationOverview(i))
            .collect(Collectors.toList()));
    }
```

And I have noticed that the error is relational with `ReflectiveSorter`, and the main code is 

```java

    private Comparator<T> createDelegateComparator(Class<T> modelClass, String fieldName) {
        Comparator<T> delegate = getIntComparator(modelClass, fieldName);
        if (delegate != null) {
            return delegate;
        }

        delegate = getStringComparator(modelClass, fieldName);
        if (delegate != null) {
            return delegate;
        }

        throw new IllegalArgumentException(String.format("Cannot find field %s in %s as int or String field",fieldName, modelClass.getName()));
    }
```

That means the `ReflectiveSorter` will find the field with `get` method to get the value from the `Class<T> ` (aka `Integration.class` here), But the `Integration.class` doesn't have the file with `get` method in the type String or Int.

So, how to use the sort param in integrations fetch api ? feel free to tell me if I have missed some detail in docs or code

zregvart:
This changes the `ReflectionUtils` to include all methods available on a
class rather than the methods that are declared on a class. This will
help in finding the `getVersion` method on `Integration` interface.

Fixes #7471

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/util/ReflectionUtils.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/util/ReflectionUtils.java
@@ -45,7 +45,7 @@ public final class ReflectionUtils {
 
     static Method extractMethod(Class<?> clazz, String fieldName, Class<?>... types) {
         try {
-            Method method = clazz.getDeclaredMethod("get" + fieldName.substring(0, 1).toUpperCase(Locale.US) + fieldName.substring(1));
+            Method method = clazz.getMethod("get" + fieldName.substring(0, 1).toUpperCase(Locale.US) + fieldName.substring(1));
             Class<?> returnType = method.getReturnType();
             for (Class<?> type : types) {
                 if (returnType.isAssignableFrom(type)) {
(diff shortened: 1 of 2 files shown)


Zheaoli:
LGTM，BTW，I wonder if there is any chance to add a hook point that allows people to use the specific Integration class via config file

zregvart:
> LGTM，BTW，I wonder if there is any chance to add a hook point that allows people to use the specific Integration class via config file

@Zheaoli that looks somewhat unrelated to this, perhaps you can create a new issue and explain your use case there.

Issue #6614, opened 2019-09-17
Use Camel upstream


christophd:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

Syndesis community version should use the Camel upstream version. Camel 2.24.2 should be the latest version (Need to check if all Camel components used in Syndesis are available in this version).

Coming from https://github.com/syndesisio/syndesis/wiki/Meeting-Notes

heiko-braun:
@christophd Please talk to @oscerd about this. I htink there was good reason not to align to the most recent 2.x in the past.

Delawen:
Comparing current Syndesis master connectors (folder app/connector) with Camel 2.24.x (folder components):

activemq -> camel-activemq 
amqp -> camel-amqp
api-provider -> ???
aws-ddb -> camel-aws-ddb
aws-s3 -> camel-aws-s3
aws-sns -> camel-aws-sns
aws-sqs -> camel-aws-sqs
box -> camel-box
concur -> ????
dropbox -> camel-dropbox
email -> camel-mail
fhir -> camel-fhir
flow -> ???
ftp -> camel-ftp
gmail -> camel-google-mail
google-calendar -> camel-google-calendar
google-sheets -> camel-google-sheets
http -> camel-http
irc -> camel-irc
jira -> camel-jira
kafka -> camel-kafka
knative -> ???
kudu -> ???
log -> camel-log
mongodb -> camel-mongodb
mqtt -> camel-mqtt
odata -> ???
rest-swagger -> camel-rest-swagger
salesforce -> camel-salesforce
servicenow -> camel-servicenow
sftp -> ???
slack -> camel-slack
sql -> camel-sql
telegram -> camel-telegram
timer -> camel-timer
twitter -> camel-twitter
webhook -> camel-webhook

This is a quick comparison just by name, but it looks that there are several components we should port to Camel.

Delawen:
Do we keep track of any improvement/change on components of Camel-Fuse that should be ported too?

Delawen:
I can start working on porting anything missing, but I need a list of things to port. If there's no list available, I can start checking manually but that's going to take time.

zregvart:
Syndesis connector is not a Camel component, so comparing the list of connectors with Camel components will not show much.

This command:

```shell
[...]

Issue #4769, opened 2019-03-06
Verify Syndesis with Camel 3


heiko-braun:
As a user I want to be able to use Camel 3 with Syndesis to leverage the latest and greatest

rkorytkowski:
> if we could get rid of spring-boot in syndesis meta, then the upgrade to camel 3 is much less problematic
> I think there is already a partial support for discovering meta stuffs using a service loader alike thing but only a few components implement that
> and by getting rid of spring-boot in syndesis meta
> I mean only for the discovery
> syndesis-meta should continue to be packaged as spring-boot
> i.e. https://github.com/syndesisio/syndesis/blob/master/app/meta/src/main/java/io/syndesis/connector/meta/v1/ConnectorEndpoint.java#L53-L73
> this has already a fallback to camel's service loader so it should not be that hard
> 
> on the connector side then we should get rid of things like 
> https://github.com/syndesisio/syndesis/blob/master/app/connector/activemq/src/main/java/io/syndesis/connector/activemq/ActiveMQVerifierAutoConfiguration.java

by @lburgazzoli

rkorytkowski:
@lburgazzoli c38c4aa shows changes I needed to do so far to make the project compile. It currently fails on building the s2i with:

```
[INFO] 
[INFO] --- exec-maven-plugin:1.6.0:exec (rest-builder-image-generator) @ s2i ---
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.springframework.boot.loader.MainMethodRunner.run(MainMethodRunner.java:48)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:87)
	at org.springframework.boot.loader.Launcher.launch(Launcher.java:50)
	at org.springframework.boot.loader.JarLauncher.main(JarLauncher.java:51)
Caused by: java.lang.IllegalArgumentException: Cannot instantiate interface org.springframework.context.ApplicationContextInitializer : org.springframework.boot.context.ConfigurationWarningsApplicationContextInitializer
[...]

PR #7483, opened 2019-12-24
Fixing dependencies for Camel 3.

Delawen:
Finished the first part of migrating to Camel 3: change the dependencies to Camel 3.0.0 and get it successfully compiled! 

Tests still fail, but we can merge to current camel-3.0.x branch to allow others to work over it.

I have moved back and forth some dependency versions that affected all projects (and moved to JUnit 5 on some projects) but I think we are safe to assume inter-dependencies shouldn't change much more.

We can start parallelizing the pending tasks for the migration:

 * A lot of tests are broken 

 * I am working on an old master version. Git pull ¿rebase? is in order.

 * We have to change Syndesis to support only one context at a time: 
https://camel.apache.org/manual/latest/camel-3-migration-guide.html#_multiple_camelcontexts_per_application_not_supported 

* Someone with a better overview on Syndesis should review what else may be a problem on that migration guide

 * After all of the above, build and run and check nothing else broke in the meantime.

I did a huge squash and merged all the commits into one big commit because that way it is easier to rollback when needed. I guess we will squash the full migration into one commit but in the meantime this is the commit that makes camel 3 compile (and shamefully fail on tests).


Delawen:
Related: #4769 #6614

zregvart:
I've rebased the `camel-3.0.x` branch and included this commit on top of that. @Delawen when you get a chance can you have a look at the `camel-3.0.x` branch and see if everything is there.
I'm going to work on fixing the issues tests surfaced.

Delawen:
Already merged by @zregvart manually.

PR #7487, opened 2020-01-03
Add option to install syndesis with options to the syndesis-operator …

claudio4j:
…install app command

### DIFF
--- a/tools/bin/commands/install
+++ b/tools/bin/commands/install
@@ -34,6 +34,8 @@ install::usage() {
                               Use quotes and start with a space before appending the options.
     --maven-mirror            Install Maven Mirror to be used with --maven-mirror when building.
     --custom-resource         Provide a custom resource to be installed by the operator
+    --app-options             Operator parameters when installing the app.
+                              Use quotes and start with a space before appending the options (example: " --addons jaeger").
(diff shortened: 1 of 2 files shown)


tools/bin/commands/install
```diff
@@ -130,6 +132,7 @@ install::run() {
 
     wait_for_deployments 1 syndesis-operator
 
+    local app_options=$(readopt --app-options)
```
phantomjinx: @zregvart 
Sorry, coming to this late. The intention was not to expose app-options as a user switch.
The operator hub has no parameter mechanism other than the custom resource hence the --custom-resource mirrors that use-case.
The app-options var is merely an internal variable to facilitate adding --custom-resource to the existing install command.
To address the 'unbound variable' problem, L135 should probably just be `local app-options=""`
WDYT?

zregvart: As long as the intent that @claudio4j has done here is maintained I'm fine with that. Should be as easy as this to enable addons.

claudio4j: Let the user add parameters to `install app` is easier than changing the custom resource isn't ?  I would like to have this option to install addons.

PR #7488, opened 2020-01-06
fix: multi select tables from diff connection issue

indraraj:
- Can select the multiple tables from any connection available. 

- Screenshot 

image


rareddy:
Looking at the image IMO it would look consistent if we showed the name as {schemaName}.TableName

mdrillin:
> Looking at the image IMO it would look consistent if we showed the name as {schemaName}.TableName

@rareddy can you clarify what you mean?  Are you requesting a change to the list item formats, for example /sampledb/contact?

mdrillin:
@indraraj I pulled down and tested against staging - looks good.  Marked for backport and will approve once the queued backports are in.

PR #7510, opened 2020-01-08
fix(ENTESB-12424): specification of an external database

phantomjinx:
* Removes unnecessary params blob from syndesis-global-config secret


heiko-braun:
@phantomjinx Can you backport this to 1.9.x as well?

PR #7522, opened 2020-01-09
BACKPORT: Commits for jaeger support in the operator

phantomjinx:
Includes:
  ENTESB-12222 - implement custom-resource switch for specifying a customised custom-resource
  ENTESB-12330 - implement jaeger backend in syndesis

Allows for enabling jaeger addon via:
```
kind: Syndesis
metadata:
  name: app
spec:
  addons:
    jaeger:
      enabled: true
```


install/operator/pkg/generator/generator.go
```diff
@@ -30,6 +30,21 @@ func AssetAsBytes(path string) ([]byte, error) {
 	return prometheusRules, nil
 }
 
+func IsDirectory(path string) bool {
```
lgarciaaco: Why making this public when it is used only in this package?

phantomjinx: It was used outside the package in a previous incarnation so left as pulbic as 'might be useful in the future'. Can make private if preferred.

PR #7524, opened 2020-01-09
[operator] Add upgrade package which handles upgrades

lgarciaaco:
### Add upgrade package which handles upgrades

An upgrade is a set of *steps*, each step should have a `run` and a `rollback` functions and the main `upgrade.Upgrade` func will call one or another, sequentially, in case of an upgrade or a rollback(if the upgrade fails). Currently  implemented steps are: `scale down` -> `syndesis backup` -> `database migration` -> `installation` -> `scale up`

#### Steps
**scale down**:  run() -> set available replicas to 0 for `syndesis-meta` and `syndesis-server` `DeploymentConfig`. rolback() -> does the same as run
**syndesis backup**: run() -> performs a syndesis backup (database and openshift resources). rollback() -> does nothing
**database migration**: run() -> performs a database migration. For this, the upgrade pod is used. The upgrade pod is modified so that the only action that it carries out is a database migration. rollback() -> restores the syndesis database from backup
**install**: run() -> does nothing, it is a place holder to rollback the resources in case there is a problem when the operator changes the resources. rollback() -> restore openshift resources from backup.
**scale up**: run() -> set available replicas to 1 for `syndesis-meta` and `syndesis-server` `DeploymentConfig`. rolback() -> does the same as run

fixes https://issues.redhat.com/browse/ENTESB-11430
fixes https://issues.redhat.com/browse/ENTESB-11438
fixes https://issues.redhat.com/browse/ENTESB-12043

Other
* [bugfix] fixes bug that throws and error with custom resources being changed between actions.
* [feat] add `Runner` interface for `pkg.syndesis.backup`. `backup.Runner` should be used instead of `backup.Backup`
* [BREAKING CHANGE]: upgrade pod considerably changed and only preforms one action: database migration


phantomjinx:
This will need rebasing againt latest master to allow for CircleCI to complete correcty.

phantomjinx:
Observation ...
Executing `operator backup --backup ${backup-dir}` on my current install, which happens to include an external-db, returns with the error `error: syndesis-db pod is not running`. Therefore ...

1. Is it the case that running syndesis with an external-db means no backup capability?
2. Could syndesis be backed up from an external-db on the assumption that the db schema in said database was as expected?

lgarciaaco:
> Observation ...
> Executing `operator backup --backup ${backup-dir}` on my current install, which happens to include an external-db, returns with the error `error: syndesis-db pod is not running`. Therefore ...
> 
> 1. Is it the case that running syndesis with an external-db means no backup capability?
> 2. Could syndesis be backed up from an external-db on the assumption that the db schema in said database was as expected?

Yes there is not backup for external-db at the moment, but it should be included in the future. This case will become more relevant when upgrading from 7.6 to 7.7.

My suggestion would be to create a ticket for it, and take it as a separate feature, wdyt?

avano:
just a note: the external db was possible to use also with 7.5 - it was requested and implemented by people from RHMI, but I don't know if they are using it or not

heiko-braun:
> it was requested and implemented by people from RHMI, but I don't know if they are using it or not

I think we need suppport for external DB. RHMI uses a dedicated postgres instance, not the one we ship.

lgarciaaco:
We need to fix https://issues.redhat.com/browse/ENTESB-12435, https://issues.redhat.com/browse/ENTESB-12355 and include backups for external databases (@phantomjinx will create a ticket for it). 

I suggest we should only consider upgrade scenarios where there is no change on the database configs. No localdb -> externaldb or externaldb -> localdb supported.

Even upgrade wont work until the before mentioned issues are addressed, I would like to merge this PR. The change for the external db is a merely change in the `backup` pkg and the ENTESB tickets issues should be solved  after implementing [api migration](https://github.com/operator-framework/operator-sdk/blob/master/doc/user/migrating-existing-apis.md) which is also a different topic.

avano:
@lgarciaaco I tried to upgrade from 1.8.12 to last daily build using this operator, but it got stuck in the loop due to permission errors:

```
"msg":"error while running migrations","type":"upgrade","step":"Database migration","error":"jobs.batch \"upgrade-db-migration\" is forbidden: User \"system:serviceaccount:avano-syndesis:syndesis-operator\" cannot get jobs.batch in the namespace \"avano-syndesis\": no RBAC policy matched"
```

full log is here: https://gist.github.com/avano/8b35a39dff441a64a55ec9c2fb76c36d

What I did:
1) checkout this PR locally
2) copy the config from `1.9.1-20200113` to build/conf/config.yaml
3) `./build.sh --operator-build go --image-build docker --image-name docker.io/avano/syndesis-operator --image-tag upgrade`
4) push to docker hub
5) install 1.8.12 using `docker run --rm --entrypoint syndesis-operator -v $HOME/.kube/config:/tmp/kube/config:z syndesis/syndesis-operator:1.8 install --config /tmp/kube/config --namespace avano-syndesis`
6) wait until the pods are running
7) install new operator using `docker run --entrypoint syndesis-operator -v $HOME/.kube:/tmp/kube:z docker.io/avano/syndesis-operator:upgrade install operator --namespace avano-syndesis --config /tmp/kube/config --image docker.io/avano/syndesis-operator --tag upgrade`
8) check the new operator logs

heiko-braun:
>  I would like to merge this PR.

What about Andrej's comments in  https://github.com/syndesisio/syndesis/pull/7524#issuecomment-574121847 ?

lgarciaaco:
> > I would like to merge this PR.
> 
> What about Andrej's comments in [#7524 (comment)](https://github.com/syndesisio/syndesis/pull/7524#issuecomment-574121847) ?

Ill have a look. The point is that for @phantomjinx to start contributing to this topic, it would be easier for us if we get this bulk merged, so he can work against master and not against my PR. Since this PR is well contained and chances that it will break something else are rather small, I think we should do it.

heiko-braun:
> so he can work against master and not against my PR 

Fine with me. Whatever works best for you. Does @zregvart have an opinion?

PR #7548, opened 2020-01-13
Fixing PMD rules. Reducing complexity and simplifying code.

Delawen:
PMD was complaining. I did a couple refactorings more to avoid unused variable assignments. deprecated functions and such.

### DIFF
--- a/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/IntegrationRouteBuilder.java
+++ b/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/IntegrationRouteBuilder.java
@@ -47,6 +47,7 @@ import io.syndesis.integration.runtime.capture.OutMessageCaptureProcessor;
 import io.syndesis.integration.runtime.logging.IntegrationLoggingConstants;
 import org.apache.camel.CamelContext;
 import org.apache.camel.Processor;
+import org.apache.camel.RuntimeCamelException;
 import org.apache.camel.builder.RouteBuilder;
 import org.apache.camel.impl.DefaultCamelContext;
 import org.apache.camel.model.ExpressionNode;
@@ -127,7 +128,7 @@ public class IntegrationRouteBuilder extends RouteBuilder {
             try {
                 return sourceProvider.getSource(getContext());
             } catch (Exception e) {
-                throw ObjectHelper.wrapRuntimeCamelException(e);
+                throw RuntimeCamelException.wrapRuntimeCamelException(e);
             }
         }
         LOGGER.info("Loading integration from: {}", configurationUri);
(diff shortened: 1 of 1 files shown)


app/integration/runtime/src/main/java/io/syndesis/integration/runtime/IntegrationRouteBuilder.java
```diff
@@ -287,23 +287,18 @@ private void configureFlow(Flow flow, final String flowIndex) throws URISyntaxEx
...
         } else {
             // not in a pipeline so set the step id header to be sure it is there
-            definition = parent.setHeader(IntegrationLoggingConstants.STEP_ID, constant(stepId));
+            parent.setHeader(IntegrationLoggingConstants.STEP_ID, constant(stepId));
         }
 
-        definition = parent.process(OutMessageCaptureProcessor.INSTANCE)
+        return parent.process(OutMessageCaptureProcessor.INSTANCE)
                         .id(String.format("capture-out:%s", stepId));
-        return definition;
```
Delawen: definition variable was not really used. We can avoid assigning and preventing weird behaviours in the future.

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/IntegrationRouteBuilder.java
```diff
@@ -330,7 +325,7 @@ private void configureFlow(Flow flow, final String flowIndex) throws URISyntaxEx
                     }
                 }
             }
-            if (onException.isPresent() && onExceptionStep != null) {
```
Delawen: Kind of a duplicated useless check. If onExceptionStep is not null, then onException is always present.

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/IntegrationRouteBuilder.java
```diff
@@ -287,23 +287,18 @@ private void configureFlow(Flow flow, final String flowIndex) throws URISyntaxEx
             }
         } else {
             // not in a pipeline so set the step id header to be sure it is there
-            definition = parent.setHeader(IntegrationLoggingConstants.STEP_ID, constant(stepId));
+            parent.setHeader(IntegrationLoggingConstants.STEP_ID, constant(stepId));
```
lburgazzoli: `setHeader` returns a `ProcessorDefinition<?> `and it is true that in this case it returns the same instance but reassigning it to a definition is IMHO more idiomatic as the underlying code is the free to return a different instance.

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/IntegrationRouteBuilder.java
```diff
@@ -287,23 +287,18 @@ private void configureFlow(Flow flow, final String flowIndex) throws URISyntaxEx
...
-        definition = parent.process(OutMessageCaptureProcessor.INSTANCE)
+        return parent.process(OutMessageCaptureProcessor.INSTANCE)
                         .id(String.format("capture-out:%s", stepId));
-        return definition;
     }
 
     private ProcessorDefinition<?> configureRouteDefinition(ProcessorDefinition<?> definition, Flow flow, String flowId) {
-        if (definition instanceof RouteDefinition) {
+        if (definition instanceof RouteDefinition
+                && !containsConfiguredActivityTrackingPolicies((RouteDefinition)definition)) {
```
lburgazzoli: You may move the check `definition instanceof RouteDefinition` to the `containsConfiguredActivityTrackingPolicies` method

Delawen: True!

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/IntegrationRouteBuilder.java
lburgazzoli: You still have to check it as at line 321 `onException` is set to Optional.empty();

PR #7554, opened 2020-01-13
refactor: archive deployment configuration

zregvart:
When we build downstream we need the archives from the `s2i` (the
repository) and `ui-react` (distribution archive) need to be deployed to
the Maven repository. This was removed for the `ui-react` in #7022.

This refactors this by reintroducing (opt-in) the archive deployment via
`deploy.archives` Maven property (i.e. `-Ddeploy.archives=true`), or
individually via `deploy.s2i.repository` and `deploy.ui.distribution`.

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -35,6 +35,8 @@
   <properties>
     <!-- Global Syndesis version -->
     <syndesis.version>${project.version}</syndesis.version>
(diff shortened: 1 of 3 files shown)


zregvart:
cc @orpiske

orpiske:
Thanks @zregvart

PR #7571, opened 2020-01-14
[ENTESB-11952] Port from master

Delawen:
Auto Discovery for AMQ/Strimzi/Kafka streams.


heiko-braun:
@Delawen Can you add a description so that we don't need to lookup the ticket to understand what changes you backport?

PR #7591, opened 2020-01-15
fix(c/sql): use standard column names

zregvart:
When fetching the list and input/output parameters of stored procedures
we're relying on `java.sql.DatabaseMetaData`.

For PostgreSQL database instead of the `DatabaseMetaData::getProcedures`
we were using `DatabaseMetaData::getFunctions`, this, by standard should
return `FUNCTION_NAME` and `FUNCTION_TYPE` columns. For other databases
we're using `DatabaseMetaData::getProcedures` returning `PROCEDURE_NAME`
and `PROCEDURE_TYPE` columns.

For all databases we're trying to read `PROCEDURE_NAME` and
`PROCEDURE_TYPE` columns, regardless of the `ResultSet`'s origin:
`DatabaseMetaData::getFunctions` or `DatabaseMetaData::getProcedures`.

The issue reported was caused by the missing `PROCEDURE_NAME` column for
the PostgreSQL database. So seemingly the
`DatabaseMetaData::getFunctions` for PostgreSQL was returning
`PROCEDURE_NAME` column previously.

The database driver version for PostgreSQL did not change, so that
leaves that there was a change in PostgreSQL server that made the
`DatabaseMetaData::getFunctions` return the standard `FUNCTION_NAME` /
`FUNCTION_TYPE` columns vs what we were seeing previously
(`PROCEDURE_NAME` and `PROCEDURE_TYPE`).

This change switches to use the `DatabaseMetaData::getProcedures` method
for PostgreSQL database as well, with the caveat that this might fail on
older versions of the PostgreSQL database.

Ref. https://issues.redhat.com/browse/ENTESB-12415

### DIFF
--- a/app/connector/sql/src/main/java/io/syndesis/connector/sql/SqlSupport.java
+++ b/app/connector/sql/src/main/java/io/syndesis/connector/sql/SqlSupport.java
@@ -95,7 +95,7 @@ public final class SqlSupport {
                     final String name = procedureSet.getString("PROCEDURE_NAME");
                     final StoredProcedureMetadata storedProcedureMetadata = getStoredProcedureMetadata(connection,
                         catalog, schemaPattern, name);
-                    storedProcedureMetadata.setName(procedureSet.getString("PROCEDURE_NAME"));
+                    storedProcedureMetadata.setName(name);
                     storedProcedureMetadata.setType(procedureSet.getString("PROCEDURE_TYPE"));
                     storedProcedureMetadata.setRemark(procedureSet.getString("REMARKS"));
                     storedProcedures.put(storedProcedureMetadata.getName(), storedProcedureMetadata);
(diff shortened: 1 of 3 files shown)


zregvart:
> Just a thought to solve the caveat: Would it be possible to get the Postgres server version and align the logic so we support both newer and older versions at the same time? Or as another option adding a fallback logic when the procedure name can not be fetched?

I've looked into having a fallback or both options for PostgreSQL.

The issue with fallback is that, assuming that whatever made us choose the `DatabaseMetaData::getFunctions` is lack of support for `DatabaseMetaData::getProcedures` in the older version, is that fallback will not work for `DatabaseMetaData::getProcedures`, so we'd have to go with `DatabaseMetaData::getFunctions`, potentially then missing the stored procedures in the newer version.

The issue with the strategy of querying both `DatabaseMetaData::getFunctions` and `DatabaseMetaData::getProcedures` is that that would change the interface for all implementations, for what might be a fluke in an old and unused version, that we would not care about in the long run.

If anything, I would opt for querying both, but since I don't have any understanding on what made us chose `DatabaseMetaData::getFunctions` in the first place I don't have a basis to make that change.

zregvart:
@christophd does the above explanation make sense to you or do we need to rethink the backward compatibility?

PR #7592, opened 2020-01-15
TEIIDTOOLS-837 fix connection display with Create View Wizard

mdrillin:
Resolving issues TTOOLS-837, TTOOLS-835, ENTESB-12364
Modify the SelectSourcesPage (first page of the CreateView wizard) to correct issues with display of connections and refresh of the connection schema.  The connections shown on the createView and importViews wizards were not consistent, and the schema in the createView wizard was not updating correctly
- adds 'useVirtualizationRuntimeMetadata' - allows usage of hook with polling interval
- 'SelectSourcesPage' now utilizing hooks to obtain the connections/dvSourceStatus/schema.  info is supplied to ConnectionSchemaContent
- 'ConnectionSchemaContent' now utilizes the Connection/status info to include the connection status
- 'ConnectionListItem' was improved to display the appropriate connection type icon / status / loading indicators
- updated the relevant stories

### DIFF
--- a/app/ui-react/syndesis/src/modules/data/shared/VirtualizationUtils.ts
+++ b/app/ui-react/syndesis/src/modules/data/shared/VirtualizationUtils.ts
@@ -235,12 +235,12 @@ function getViewDefinition(
  * based on the Virtualization connection status and selection state
  * @param conns the connections
  * @param virtualizationsSourceStatuses the available virtualization sources
- * @param selectedConn name of a selected connection
+ * @param selectedConn name of a selected connection (optional)
  */
 export function generateDvConnections(
   conns: Connection[],
   virtualizationsSourceStatuses: VirtualizationSourceStatus[],
-  selectedConn: string
+  selectedConn?: string
 ): Connection[] {
   const dvConns: Connection[] = [];
   for (const conn of conns) {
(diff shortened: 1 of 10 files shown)


app/ui-react/syndesis/src/modules/data/shared/VirtualizationUtils.ts
```diff
@@ -269,7 +269,7 @@ export function generateDvConnections(
       // loading (true/false)
       schemaLoading = String(virtSrcStatus.loading);
       // selection
-      if (conn.name === selectedConn) {
+      if (selectedConn && conn.name === selectedConn) {
```
indraraj: Hey Mark, I could see in a few cases it's false just because of space in the name, I was wondering is this intentional. like "SadMongo" and "Sad Mongo".

indraraj:
looks good to me. Using `connection` call and `syndesisSourceStatuses` make sense as the same is used in `Import View`.

indraraj:
looks good to me. Using `connection` call and `syndesisSourceStatuses` make sense as the same is used in `Import View`.

app/ui-react/syndesis/src/modules/data/shared/VirtualizationUtils.ts
mdrillin: Indra, you are correct.  I looked into this and the 'sourceName' in our returned 'status' does not match the connection names when there are spaces in it.  I have logged a separate issue to investigate (TEIIDTOOLS-912).  Good catch!

PR #7593, opened 2020-01-15
chore: upgrade operator dependencies

lgarciaaco:
* upgrade operator-sdk to 0.14.0
* upgrade k8s 1.16.2

refactor: some functions had to be changed due to changes, mostly changes introduced when upgrading k8s 1.13 to 1.16

Fixes https://issues.redhat.com/browse/ENTESB-12617


install/operator/pkg/cmd/internal/run/run.go
```diff
@@ -23,13 +23,10 @@ import (
...
 	sdkVersion "github.com/operator-framework/operator-sdk/version"
-	"github.com/syndesisio/syndesis/install/operator/pkg/openshift"
-	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
-
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
-
 	"github.com/syndesisio/syndesis/install/operator/pkg/apis"
 	"github.com/syndesisio/syndesis/install/operator/pkg/controller"
+	"github.com/syndesisio/syndesis/install/operator/pkg/openshift"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
```
davidffrench: This should be updated to `sigs.k8s.io/controller-runtime/pkg/log` according to https://github.com/operator-framework/operator-sdk/blob/master/doc/migration/version-upgrade-guide.md#import-updates.

Would be good to check the other imports listed in the migration doc

install/operator/pkg/cmd/internal/uninstall/uninstall_test.go
```diff
@@ -61,7 +61,7 @@ func TestUninstall(t *testing.T) {
 	// Create a fake client to mock API calls and pass it to the cmd
 	objs := []runtime.Object{&s}
 	cl := fake.NewFakeClient(objs...)
-	cl.List(ctx, client.InNamespace(ns), &sl)
+	cl.List(ctx, &sl)
```
davidffrench: listing from a specific namespace has been removed, is this intentional? You can achieve the same as before with 
```
listOpts := []client.ListOption{
  client.InNamespace(ns),        
}
cl.List(ctx, &sl, listOpts...)
```

install/operator/pkg/cmd/internal/run/run.go
lgarciaaco: Checked and included, thanks for pointing this out!

install/operator/pkg/cmd/internal/uninstall/uninstall_test.go
lgarciaaco: Thanks!

install/operator/pkg/cmd/internal/run/run.go
davidffrench: no problem at all :-)

PR #7601, opened 2020-01-16
[ENTESB-11952] AMQ Streams autodiscovery docs

Delawen:
Documentation on how to configure your Syndesis installation to be able to auto-discover Kafka/AMQ Streams/Strimzi brokers.

### DIFF
--- a/doc/sdh/cli/cmd_install.adoc
+++ b/doc/sdh/cli/cmd_install.adoc
@@ -152,3 +152,59 @@ This example installs the latest Syndesis version of the 1.4 branch to the local
 
 You can see a list of available tags with `git tag`.
 Tags prefixed with `fuse-ignite` are suited for the Fuse Online cluster as those templates do not contain images streams themselves but refer to the image streams installed on this cluster.
+
+
+=== Allowing AutoDiscovery of Kafka Brokers
+
+From version 1.9.0 on, the Kafka connector allow the autodiscovery of brokers. This is achieved by giving privileges to your oc user to see resources on your cluster. Different configurations of privileges of your oc user will let you discover different subsets of brokers.
+
+==== Available Kafka Brokers
+
+First, make sure you have installed some Kafka broker on your cluster. 
+
+For example, if we want to install Strimzi on our minishift, we can do so by using the following commands, based on https://strimzi.io/quickstarts/okd/
+
+* Login to your minishift/openshift instance
++
+----
+oc login -u system:admin
+----
++
+NOTE: Make sure that you run this command with proper user credentials
+
+* Apply Strimzi installation file
++
+----
+oc apply -f https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.15.0/strimzi-cluster-operator-0.15.0.yaml -n myproject
+----
+
+* Provision the Apache Kafka cluster
++
+----
+oc apply -f https://raw.githubusercontent.com/strimzi/strimzi-kafka-operator/0.15.0/examples/kafka/kafka-persistent-single.yaml -n myproject 
+oc wait kafka/my-cluster --for=condition=Ready --timeout=300s -n myproject
+----
+
+* Now you can create some producers
++
+----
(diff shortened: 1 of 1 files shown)


zregvart:
This is great work, I'm not sure it belongs to the Syndesis Developer Handbook, perhaps a better place would be it in `project/designs`?

Perhaps you can get some help from @TovaCohen on the user docs as well.

Delawen:
Where is this ''project/designs' page? On the docs or the developer handbook? I'm confused :confused: :P

zregvart:
@Delawen here it is: https://github.com/syndesisio/syndesis/tree/master/project/designs

TovaCohen:
@zregvart @gaughan 
It sounds like the content that Maria wrote belongs in Installing and Operating Fuse Online on OpenShift Container Platform. You can see the customer portal version of that document for 7.5 here: 
https://access.redhat.com/documentation/en-us/red_hat_fuse/7.5/html/installing_and_operating_fuse_online_on_openshift_container_platform/index 

I would add a section to chapter 2, Managing Fuse Online on OCP, something like: "Configuring Fuse Online to automatically discover Kafka brokers". Does that seem like a good place for the doc that Maria wrote? 

If it is, then be aware that the content for that book is only downstream. 
Does the content for auto-discovery of Kafka brokers also need to be upstream? If yes, what's the situation of the user who needs this content? That is, what is that user's environment? Where would a user expect to find information about this?

Also, the title of this PR has to do with AMQ Streams. What's the relationship between AMQ Streams and discovery of Kafka brokers?

Delawen:
> I would add a section to chapter 2, Managing Fuse Online on OCP, something like: "Configuring Fuse Online to automatically discover Kafka brokers". Does that seem like a good place for the doc that Maria wrote?

I think so.

> Does the content for auto-discovery of Kafka brokers also need to be upstream?

Yes.

> If yes, what's the situation of the user who needs this content? That is, what is that user's environment? Where would a user expect to find information about this?

That's my main doubt. We don't have a section on the docs for each component as in Camel. Maybe we should. And I don't see an "administrator handbook" so I just placed it on installation instructions.

'project/designs' seems too isolated to me. Maybe the solution would be to add it to the docs somewhere? (I couldn't find a link, maybe there is one already?)

> Also, the title of this PR has to do with AMQ Streams. What's the relationship between AMQ Streams and discovery of Kafka brokers?

For this purpose, they are equivalent. Kafka brokers and AMQ Streams are discovered the same way.

TovaCohen:
Is there a good reason to NOT provide this information in the section in the connector guide for connecting to Kafka?  If we do that, then the content would be upstream and downstream in the same hierarchy.  You can see the 7.5 doc for connecting to Kafka here: 

https://access.redhat.com/documentation/en-us/red_hat_fuse/7.5/html/connecting_fuse_online_to_applications_and_services/connecting-to-kafka_connectors

I would add a new section: "Enabling auto-discovery of AMQ Streams". 

The information to be added is similar in nature to what we provide for connecting to Knative resources, and you can see the 7.5 doc for that here: 
https://access.redhat.com/documentation/en-us/red_hat_fuse/7.5/html/connecting_fuse_online_to_applications_and_services/connecting-to-knative-resources_connectors#enabling-connections-to-knative-resources_knative

WDYT?

Delawen:
I still need a place on the community documentation.

TovaCohen:
After a discussion with Maria, I copied this content to a place where I can work on it, and I am going to close this PR. I'm going to add the content to the connector guide.  And we will also work to publish the connector guide, along with the tutorials and the integration guide, on the syndesis.io community site.

Delawen:
@TovaCohen should I copy this to the syndesis.io repo?

TovaCohen:
@Delawen Yes, I think probably you should. In case it is helpful, my edited version of your content is here: 

https://docs.google.com/document/d/1NyNX5golU0Ssa1zY3IJo0RDHZaZIWEazTOhAicYGNz8/edit?usp=sharing

The better alternative is to publish the user doc: 
https://github.com/syndesisio/syndesis/issues/7654

But I have no idea what the priority for doing that is.

PR #7602, opened 2020-01-16
fix(ENTESB-12570): Extends jaeger addon to more complex server configurations

phantomjinx:
New Flags part of jaeger addon configuration (note. Enabled flag must be true for addon to work at all):
* ClientOnly (bool): Only the client configuration in the server and meta pods will be implemented;
* OperatorOnly bool: Client and Operator configuration will be configured BUT no jaeger CRD will be installed;
* QueryUri (string): The address of the query daemon that the client is expected to communicate with (if not set then default remains _http://noauth-syndesis-jaeger-query:443/api_);
* CollectorUri (string): The address of the collector daemon that the client is expected to communicate with (if not set then default remains _http://syndesis-jaeger-collector:14268/api/traces_).

With these settings, it is possible for jaeger to be implemented in 3 different use-cases:
1. Client, Operator & CRD installed by Syndesis in a default memory-storage configuration;
2. Only Client 'hooks' are installed and the query & collector URIs configured to communicate with services totally external to the Syndesis environment, eg. central remote system;
3. Client and Operator are installed hence it is expected that an implementation of the server components will be installed as well. However, their required configuration is far more complex than the default memory-storage CRD hence it is left up to the user to configure their jaeger CRD and apply after Syndesis has been installed, eg. a jaegar server utilising cassandra or elasticsearch as its storage-system (even using a remote elasticsearch instance).

Flags to achieve use-cases
```
          ClientOnly    OperatorOnly     QueryURI         CollectorURI
1             N              N            Empty             Empty
2             Y              N            Remote Addr       Remote Addr
3             N              Y            Depends on name used in jaeger CRD
```


heiko-braun:
Lgtm, please remember to backport it. @phantomjinx Does it still allow for cases where jaeger is not used at all?

install/operator/pkg/generator/assets/infrastructure/03-syndesis-server-config.yml.tmpl
```diff
@@ -65,7 +65,11 @@
         jaeger-activity-tracing:
           enabled: true
       jaeger:
+      {{- if .Syndesis.Addons.Jaeger.QueryUri}}
+        query.api.uri: {{.Syndesis.Addons.Jaeger.QueryUri}}
```
claudio4j: The property name should end with `url` and not `uri`. WDYT to use `uri` as standard, instead of the `url` ?

PR #7603, opened 2020-01-16
TEIIDTOOLS-909 minimizing datasource creation

shawkins:


shawkins:
This together with the teiid spring boot change to implement closable on connection factories should clean up any potential resource leaks due to connection factories.

app/dv/src/main/java/io/syndesis/dv/openshift/SyndesisConnectionSynchronizer.java
```diff
@@ -121,18 +121,34 @@ public void addConnection(DefaultSyndesisDataSource sds, boolean update) {
...
+        boolean create = true;
         if (update) {
             try {
-                this.openshiftClient.deleteDataSource(sds);
+                //ensure that the properties have changed
+                if (sds.getTeiidName() != null) {
+                    TeiidDataSource tds = this.metadataService.findTeiidDatasource(sds.getTeiidName());
+                    if (tds != null) {
+                        DefaultSyndesisDataSource existing = tds.getSyndesisDataSource();
+                        if (existing.getProperties().equals(sds.getProperties())) {
```
rareddy: Is Object equality is what you are going for here?

shawkins: Yes.  Do you have a concern?

Issue #7600, opened 2020-01-16
Connections data not used


ryckowski:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
1. Create API Connector idf-http-connector
2. Create 2 connections idf-http-c1 and idf-http-c2, the second one has different URL
3. Create integration and use idf-http-c2. Click on Configure button just added connection

The attached screenshot shows that the item name changed to connector name so I don't know what connection and thus url  is in use. 
Actually I was trying to replace the connection in my integration but it always used the url from the connector or connection 1 and not the connection I've created later

![c1](https://user-images.githubusercontent.com/10454405/72511520-99aa5100-384b-11ea-86a4-9ab1567c6b03.png)
![c2](https://user-images.githubusercontent.com/10454405/72511521-9a42e780-384b-11ea-908c-05ab1c5fbc81.png)

gashcrumb:
This is the key bit and the real bug:

> Actually I was trying to replace the connection in my integration but it always used the url from the connector or connection 1 and not the connection I've created later

If you use the second connection initially when creating the integration it appears to work, though as you say we're showing the connector name and not the connection name in several cases.  I also suspect some code that tries to preserve the configuration on the step in some cases is coming into play here though as well.

So a couple things I'll look at doing for this:

1)  Update the IVP to show the connection name and not the connector name
2)  Investigate the key being used to figure out if the old step config should be preserved.

gashcrumb:
There's a "descriptor" call that's made after configuring a step, while the connector name is shown it does look like the second connection does indeed get used if you look at the step in the JSON response:

![image](https://user-images.githubusercontent.com/351660/72540237-96649480-384e-11ea-91a0-2e94d46cf350.png)

If you get a chance would you mind verifying if that's the case on your end?

I'll work to get the right name showing up in the meantime.

ryckowski:
I've checked it again and it looks like connection url is not used but on the other hand authentication data entered on the same page are used. My scenario is: 
1. publish integration
2. change url in the connection
3. edit and publish integration
And the old url is used.
But I'm working on branch forked like a month ago so it might be it got fixed in the meantime

PR #7604, opened 2020-01-16
fix: Use the connection name instead of the connector name in t…

gashcrumb:
relates to #7600

Quick example:
image

vs. what it was showing:

image


zregvart:
Perhaps backport to 1.9.x?

gashcrumb:
Yeah, agreed!  Created https://github.com/syndesisio/syndesis/pull/7614 for that

PR #7607, opened 2020-01-16
chore: Update to AtlasMap 1.43.2

igarashitm:
This should fix the problem that data mapper UI is frozen (un-clickable)


zregvart:
Just to note that we need this on `master` also.

igarashitm:
Yep #7606 is for master, seems it failed on ui build though. kicked rerun in circleci now.

Issue #7556, opened 2020-01-13
Failing to run the latest images


zregvart:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

For example when trying to run the meta image:

```
docker run -it --rm docker.io/syndesis/syndesis-meta:latest
Unable to find image 'syndesis/syndesis-meta:latest' locally
latest: Pulling from syndesis/syndesis-meta
ac9208207ada: Pull complete 
1e29086cad24: Pull complete 
76dcefade05b: Pull complete 
543dfa680296: Pull complete 
38c5929163cd: Pull complete 
fa6b7b77c257: Pull complete 
128401db1475: Pull complete 
17f76aceb40a: Pull complete 
d0e3405f2c1f: Pull complete 
fecf74506960: Pull complete 
Digest: sha256:790de0f6b1604640d95253e2e1dea8da91fe7f062cb18054edb2ec2fe92f29d0
Status: Downloaded newer image for syndesis/syndesis-meta:latest
No extensions found
No extensions found
exec java -javaagent:/opt/agent-bond/agent-bond.jar=jolokia{{host=0.0.0.0}},jmx_exporter{{9779:/opt/agent-bond/jmx_exporter_config.yml}} -XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:+ExitOnOutOfMemoryError -cp . -jar /deployments/meta-2.0-SNAPSHOT.jar
I> No access restrictor found, access to any MBean is allowed
Jolokia: Agent started with URL http://172.17.0.2:8778/jolokia/
      _______.                 _               _
    /       |                 | |             (_)
   |   (----`_   _  ____    _ | |  ____   ___  _   ___
    \   \   | | | ||  _ \  / || | / _  ) /___)| | /___)
.----)   |  | |_| || | | |( (_| |( (/ / |___ || ||___ |
|_______/    \__  ||_| |_| \____| \____)(___/ |_|(___/
============ (____/ ===================================
:: Meta :: v2.0-SNAPSHOT


[...]

PR #7626, opened 2020-01-20
chore: update swagger-core

zregvart:
To help with #7556 this updates the `swagger-core` to the latest version
(2.1.1).

This will change the version of the served API documentation from
Swagger 2.0 to OpenAPI 3.0.

For Syndesis Meta I'm not aware of any client using this API
documentation so this should be safe to do.

For the Syndesis Server more extensive changes needed to be done:

 * annotations used needed to change
 * tooling around generating Asciidoc/HTML of the API needed to be
   upgraded / changed
 * endpoints changed from `/api/v1/[internal/]swagger.[yaml|json]` to
   `/api/v1/[internal/]openapi.[yaml|json]`

Changes to the UI build need to be done, as the current tool chain
relies on Swagger 2.0 file.


zregvart:
We're depending on https://github.com/atlasmap/atlasmap/pull/1672 and a release that would contain it.

In the meantime can someone with the UI experience have a look at these changes. Especially c94ee2605e7fd53060623e1b75c38c64c5689929 and d47466d47bd716f143cc30d47c3f4f2cc7931904.

app/ui-react/syndesis/src/modules/integrations/components/IntegrationEditorStepAdder.tsx
```diff
@@ -131,13 +125,8 @@ function getWarningMessage(
...
       <>
         <Link
-          data-testid={
-            'integration-editor-step-adder-add-default-flow-link'
-          }
-          to={props.configureStepHref(
-            idx,
-            props.steps[idx]
-          )}
+          data-testid={'integration-editor-step-adder-add-default-flow-link'}
```
gashcrumb: There's kinda too many formatting only changes like this one, I think you need a `prettier` extension in your IDE to ensure these blocks get formatted properly.  Otherwise subsequent changes after this PR in these areas will contain extra noise.

gashcrumb: Or maybe you have it installed and someone else didn't in which case ignore me :-)

gashcrumb:
That namespace change is gonna be a hard pill to swallow I think TBH, I'll have to pull this down locally to better peek around.

app/ui-react/syndesis/src/modules/integrations/components/IntegrationEditorStepAdder.tsx
zregvart: That was pre-commit hooks doing, I think its using `prettier`.

gashcrumb:
Yep, so interestingly I think the generated file runs afoul of [this guidance](https://www.typescriptlang.org/docs/handbook/modules.html#red-flags) and some of [the text in here](https://www.typescriptlang.org/docs/handbook/namespaces-and-modules.html#needless-namespacing) has me concerned that we're generating 2 files that add to the same namespace

I'm tempted to suggest that we instead change the `--namespace` parameter to `''` and search/replace `declare` with `export` in the generated files with some post-processing step if we can't figure out another way to get `dtsgenerator` to produce what we need.

gashcrumb:
I backed out the 2 commits changing the UI locally and focused on swapping out `sw2dts` only and came up with this relevant snippet of the `package.json` file:

```
  "devDependencies": {
    "cpr": "^3.0.1",
    "dtsgenerator": "^2.3.2",
    "npm-run-all": "^4.1.5"
  },
  "scripts": {
    "build:openapi-internal": "dtsgen --namespace '' --out models-internal.d.ts swagger.internal.json",
    "build:openapi": "dtsgen --namespace '' --out models.d.ts swagger.json",
    "build:sed-openapi": "sed 's/declare interface/export interface/' models.d.ts >./src/models.d.ts",
    "build:sed-openapi-internal": "sed 's/declare interface/export interface/' models-internal.d.ts > ./src/models-internal.d.ts",
    "build:types": "cpr src dist -d",
```

It's not pretty but it does produce typescript definition files that allow the rest of the UI to build without any changes.  @zregvart thoughts?  Probably `sed` isn't the first choice.

I suspect this is exactly why we were using `sw2dts` in the first place, as we only need the interfaces for the parameters, `dtsgenerator` also appears to produce namespaces/types for things like paths etc that we don't care about.

zregvart:
@gashcrumb much better, I think we can work with that. What I would like perhaps is to do a PR against `dtsgenerator` and see where that leads, I see other folk having some of the same issues.

zregvart:
I've created https://github.com/horiuchi/dtsgenerator/pull/386 let's see if that can get merged.

gashcrumb:
Yep, good call.

zregvart:
We have the change we need merged upstream in `dtsgenerator`, I don't think we're in such a hurry not to wait for a release.

zregvart:
@gashcrumb I've the `dtsgenerator` with the change we needed has been released. This now doesn't include the namespace change. Once we get https://github.com/atlasmap/atlasmap/pull/1672 merged and released I think this is good to go.

gashcrumb:
Awesome!

Issue #7616, opened 2020-01-18
CircleCI build issue, not reproducing error when built locally


vhalbert:
<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
DV is having an issue in CircleCI such that a local build issue isn't being seen in CircleCI, its building successfully.  See example [build] (https://circleci.com/gh/syndesisio/syndesis/226084#build-parameters/containers/0)

Run locally:  ./tools/bin/syndesis build --batch-mode --module dv

and it produces exception:  Caused by: java.lang.ClassNotFoundException: com.fasterxml.jackson.core.TSFBuilder

Determined that CircleCI is using maven 3.6.3, so updated my version from 3.5.4 and the issue is still occurring.   Otherwise, have been unable to determine what is being introduced in the CircleCI environment such that the jackson-databind dependency in the app/pom.xml (2.10.2) is overriding the dependency specified in DV (2.9.10.1).

Any suggestions are welcomed.

Also, this may become all mute as the version of  jackson is currently being debated my management.

PR #7627, opened 2020-01-20
Issue #7616  aligned all jackson components for DV to the 2.9.x versions

vhalbert:


zregvart:
Could we use the same version as in Syndesis parent?

https://github.com/syndesisio/syndesis/blob/dbac76401a3c1b165b74a97c93bbf1ba356b1da6/app/pom.xml#L61

vhalbert:
Teiid isn't ready to upgrade, as it breaks the build.

PR #7653, opened 2020-01-22
fix: unable to set memory limits for syndesis components in CR

lgarciaaco:
chore: remove unused constant ReadWriteMany
chore: fix typo in Spec.Components.Database.URL

### DIFF
--- a/install/operator/pkg/apis/syndesis/v1beta1/syndesis_types.go
+++ b/install/operator/pkg/apis/syndesis/v1beta1/syndesis_types.go
@@ -154,9 +154,11 @@ type Resources struct {
 // +kubebuilder:validation:Enum=ReadWriteOnce;ReadOnlyMany;ReadWriteMany
 type VolumeAccessMode string
 
+// Possible VAM can be found here  https://docs.openshift.com/container-platform/4.2/storage/understanding-persistent-storage.html
 const (
 	ReadWriteOnce VolumeAccessMode = "ReadWriteOnce"
 	ReadOnlyMany  VolumeAccessMode = "ReadOnlyMany"
+	// ReadWriteMany VolumeAccessMode = "ReadWriteMany"
 )
 
 type ResourcesWithPersistentVolume struct {
--- a/install/operator/pkg/apis/syndesis/v1beta1/syndesis_types.go
+++ b/install/operator/pkg/apis/syndesis/v1beta1/syndesis_types.go
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 Red Hat, Inc.
+ * Copyright (C) 2020 Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


lgarciaaco:
Fixes https://issues.redhat.com/browse/ENTESB-12618

Checked by setting:
```
  spec:
    components:
      meta:
        resources:
          memory: 600Mi
```
Log output (with `--print-resource-diffs`)
```
resource DeploymentConfig update: syndesis-meta
--- a   00-00-00 00:00:00
+++ b   00-00-00 00:00:00
@@ -98,7 +98,7 @@
           timeoutSeconds: 1
         resources:
           limits:
-            memory: 512Mi
+            memory: 600Mi
           requests:
             memory: 280Mi
         terminationMessagePath: /dev/termination-log
 <nil>
{"level":"info","ts":1579695615.156552,"logger":"action","msg":"resource updated","type":"install","kind":"DeploymentConfig","name":"syndesis-meta","namespace":"syndesis"}
```

install/operator/pkg/apis/syndesis/v1beta1/syndesis_types.go
```diff
@@ -116,7 +116,7 @@ type DatabaseConfiguration struct {
 	Name string `json:"name,omitempty"`
 
 	// Host and port of the PostgreSQL database to access
-	URL string `url:"url,omitempty"`
+	URL string `json:"url,omitempty"`
```
lgarciaaco: @phantomjinx do you think this change can be problematic? I think it was a typo

install/operator/pkg/apis/syndesis/v1beta1/syndesis_types.go
```diff
@@ -157,11 +157,10 @@ type VolumeAccessMode string
 const (
 	ReadWriteOnce VolumeAccessMode = "ReadWriteOnce"
 	ReadOnlyMany  VolumeAccessMode = "ReadOnlyMany"
-	ReadWriteMany VolumeAccessMode = "ReadWriteMany"
```
phantomjinx: This is a possible Volume Access Mode so why is it being removed?
eg. https://docs.openshift.com/container-platform/4.2/storage/understanding-persistent-storage.html

install/operator/pkg/apis/syndesis/v1beta1/syndesis_types.go
phantomjinx: Is this also a typo problem in v1alpha1?

install/operator/pkg/apis/syndesis/v1beta1/syndesis_types.go
lgarciaaco: Because it is not used anywhere and my go-linter doesn't like it. We can comment it out maybe?

install/operator/pkg/apis/syndesis/v1beta1/syndesis_types.go
lgarciaaco: We don't have it in `v1alpha1`

PR #7656, opened 2020-01-22
fix(ENTESB-12570): Extends jaeger addon to more complex server config…

phantomjinx:
This was wrongly merged onto the jaeger branch (which should probably be deleted).

Also, fixed typo as [mentioned](https://github.com/syndesisio/syndesis/pull/7602/files#r368179598) 

…urations

* syndesis_type.go
* configuration.go
 * Adds flags for configs
 1. Enabled: All jaeger resources installed
 2. Enabled + ClientOnly: Only client config installed - no operator or crd
 3. Enabled + OperatorOnly: Client and Operator config installed - no crd
 - QueryUri / CollectorUri allows modifiying client specified URIs for
   alternative configurations

* grant.go
 * To allow for user to apply alternative CRD, they must be a member of the
   jaeger-operator role so need cluster-admin to grant them this on install
 * Moves grant-related resources to their own directory

* Templates
 * Adds if conditionals to jaeger templates to only install if requisite
   flags are true

*


claudio4j:
I did install jaeger-all-in-one on a new namespace following these commands
https://gist.github.com/claudio4j/6ec87d655e0a85e5a3b634936e04d2b3

Then had built syndesis-operator and installed syndesis as
https://gist.github.com/claudio4j/2edfe2a7546f8a7017a48ff65aa1bf12

There are no errors, but I noticed `syndesis-jaeger` and `jaeger-operator` was installed in syndesis namespace, but as I want to use an external jaeger (from observability namespace), the `syndesis-jaeger` and `jaeger-operator` should not be installed, right ?

```
$ oc get pods -w
NAME                              READY     STATUS    RESTARTS   AGE
jaeger-operator-7c97cf697-5kpb8   2/2       Running   0          6m
syndesis-db-1-dzgsr               2/2       Running   0          6m
syndesis-jaeger-fd5f84484-b79t2   2/2       Running   0          6m
syndesis-meta-1-mcdvl             1/1       Running   0          6m
syndesis-oauthproxy-1-nkgdx       1/1       Running   0          6m
syndesis-operator-1-vsfnf         1/1       Running   0          9m
syndesis-prometheus-1-4xdjd       1/1       Running   0          6m
syndesis-server-1-zfbsn           1/1       Running   2          6m
syndesis-ui-1-z87mg               1/1       Running   0          6m
```

Also, I noticed the syndesis app doesn't show the new jaeger parameters `operatorOnly, clientOnly`.
Maybe am I missing some command/parameter to properly install syndesis-operator or syndesis-operator is not using these new parameters.

```
$ oc describe syndesis app
Name:         app
Namespace:    syndesis
Labels:       <none>
Annotations:  syndesis.io/applicationUrl=https://syndesis-syndesis.192.168.42.160.nip.io
API Version:  syndesis.io/v1beta1
Kind:         Syndesis
Metadata:
  Creation Timestamp:  2020-01-24T17:38:10Z
  Generation:          1
  Resource Version:    1304547
  Self Link:           /apis/syndesis.io/v1beta1/namespaces/syndesis/syndesises/app
  UID:                 49cb90f0-3ed0-11ea-a380-5254007f23d6
Spec:
  Addons:
    Camelk:
    Dv:
      Resources:
    Jaeger:
      Enabled:  true
    Knative:
    Ops:
    Public API:
    Todo:
  Backup:
  Components:
    Database:
      Resources:
    Grafana:
      Resources:
    Meta:
      Resources:
        Volume Capacity:  1Gi
    Oauth:
    Prometheus:
      Resources:
        Volume Capacity:  1Gi
    Server:
      Features:
      Resources:
    Upgrade:
      Resources:
  Force Migration:  false
Status:
  Backup:
  Phase:    Installed
  Version:  latest
Events:     <none>
```

heiko-braun:
I think you are right. If an an external jeager is configured, we should not create the internal components.

heiko-braun:
@phantomjinx What do you think?

phantomjinx:
@claudio4j 
This is really odd.
I have run installs some 10-15 times and at no point has it failed to not observe the CR settings. I have tried varying the permutations and ordering but they seem to be heeded correctly.

Screenshot_20200129_211842

In both the artifacts installed and the CR representation, the configuration is as per requested.

Any chance you could double-check your setup?

Just to reiterate:
* Having **clientOnly=true** does NOT install the operator, syndesis' default jaeger CR or jaeger backend
* Having **operatorOnly=true** (& **clientOnly=false**) installs the operator but NOT the jaeger CR or jaeger backend

cc @heiko-braun

claudio4j:
I figured out the problem was the syndesis-operator updated itself from dockerhub and it doesn't contain the new jaeger attributes. To fix that, run `oc set env dc/syndesis-operator DEV_SUPPORT=true` to prevent self-updating.

PR #7658, opened 2020-01-22
ENTESB-12598 The order of steps in the activity log is not in order in which the steps are defined. 1.9.x

claudio4j:
https://issues.redhat.com/browse/ENTESB-12598

### DIFF
--- a/app/server/logging/jaeger/src/test/java/io/syndesis/server/logging/jaeger/service/JaegerActivityTrackingServiceTest.java
+++ b/app/server/logging/jaeger/src/test/java/io/syndesis/server/logging/jaeger/service/JaegerActivityTrackingServiceTest.java
@@ -26,14 +26,26 @@ import javax.ws.rs.WebApplicationException;
 import io.syndesis.common.util.json.JsonUtils;
 import io.syndesis.server.endpoint.v1.handler.activity.Activity;
 import org.junit.Test;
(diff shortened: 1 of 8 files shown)


app/server/logging/jaeger/src/test/java/io/syndesis/server/logging/jaeger/service/JaegerActivityTrackingServiceTest.java
```diff
@@ -52,8 +63,9 @@ public void shouldRetainLastRetainActivityLogs() throws IOException {
         List<Activity> activities = service.getActivities("test", null, null);
         assertThat(activities).isNotNull();
 
-
         String activitiesJson = JsonUtils.writer().withDefaultPrettyPrinter().writeValueAsString(activities).trim();
+        // print the activitiesJson to replace the content of expected-activities.json
+        // System.out.println(activitiesJson);
         String expectedActivitiesJson = resource("expected-activities.json").trim();
         assertThat(activitiesJson).isEqualTo(expectedActivitiesJson);
```
zregvart: // nit
I'd use JSONAssert here instead not to rely on formatting changes or property order in the JSON.

claudio4j: Done. Thanks.

PR #7685, opened 2020-01-24
Upgrade `@storybook/*` to 5.3.8

zregvart:
I went through all `package.json` files and manually changed the version
of anything `@storybook/*` to 5.3.8 (current latest). I've also removed
`@types/storybook__react` since it's deprecated due to
`@storybook/react` providing type definitions and updated
`@dump247/storybook-state` to latest version.


zregvart:
Missed one, I'll check for more...

PR #7687, opened 2020-01-24
chore: add metrics for product and operator version

lgarciaaco:
Also creates service monitor. Change configuration variable name to avoid collition with package name.


lgarciaaco:
Fixes https://issues.redhat.com/browse/ENTESB-12333

PR #7689, opened 2020-01-24
[ENTESB-12785] Align MongoDB versions

Delawen:
We had two different versions of MongoDB plus a third one coming from Camel dependencies.

### DIFF
--- a/app/dv/pom.xml
+++ b/app/dv/pom.xml
@@ -123,7 +123,6 @@
     <!-- this is simply a version no dependencies involved right now -->
     <version.springboot.teiid>1.3.0</version.springboot.teiid>
     <version.mysql>8.0.17</version.mysql>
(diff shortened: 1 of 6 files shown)


app/dv/pom.xml
```diff
@@ -354,7 +353,7 @@
       <dependency>
         <groupId>org.mongodb</groupId>
         <artifactId>mongo-java-driver</artifactId>
-        <version>${version.org.mongodb}</version>
+        <version>${mongodb.version}</version>
```
zregvart: cc @rareddy

Delawen: More cc of people mentioned in pom, just in case @rareddy is busy: @blafond @shawkins @mdrillin

Delawen: Solved in #7731

PR #7695, opened 2020-01-26
TEIIDTOOLS-908: Validating the connection before doing any VDB creati…

rareddy:
…on, such that weed out any bad connections to begin with. Along with connection close fix this should handle any bad connections

### DIFF
--- a/app/dv/src/main/java/io/syndesis/dv/openshift/SyndesisConnectionSynchronizer.java
+++ b/app/dv/src/main/java/io/syndesis/dv/openshift/SyndesisConnectionSynchronizer.java
@@ -15,14 +15,19 @@
  */
 package io.syndesis.dv.openshift;
 
+import java.sql.Connection;
 import java.util.Collection;
 import java.util.Map;
 import java.util.stream.Collectors;
+
+import javax.sql.DataSource;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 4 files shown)


rareddy:
@shawkins The only thing I can think of here is, remove the update of the metadata since UI is going to take care of that, instead, check for the existence of the source vdbs against each data source every 5mins or even 1 min. If (sourceVDB.notExists && dataSource.valid) {createsourcevdb} 

That way we still have the background thread keep checking for the validity of connections, WYDT?

shawkins:
Just wanted to circle back to the original issue.  There's some logging due to mongodb internals that will happen as long as there's an invalid mongodb data source defined.  The changes to close and/or not create connection factories should not resolve that.  And if it's not removed as you originally had, then the logging would still happen.  This could be addressed by filtering the info level for the org.mongodb.driver.cluster context.

The other is an exception when we can't load the metadata, which is what we now trying to determine what is the best approach.

> That way we still have the background thread keep checking for the validity of connections, WYDT?

As long as the data source still exists, then we can just have the one background job that does the metadata refresh.  Since we aren't deleting the invalid data source the user can always manually attempt a reload once those changes are available.  If you want to add an isValid check prior to attempting the metadata load to not even attempt the load and therefore not incur the logged exception.

rareddy:
@shawkins 
- Added logging property to remove the "org.mongodb" to WARN level
- Just removed the delete logic, but kept the connection validation as is, where only on valid connection the source vdb is created.

PR #7696, opened 2020-01-26
TEIIDTOOLS-912: When data source name in Syndesis have spaces, the Sy…

rareddy:
…ndesis removes the spaces but while reporting it does not restore thus the matching fails etween what is in the metadata source and syndesis source, this patch will fix to send back correct name


rareddy:
@shawkins it is now simplified by adding the syndesis name into TeiidDataSource

rareddy:
@shawkins is this good to go?

PR #7700, opened 2020-01-26
TEIIDTOOLS-923: generating a pseduo SQL code for two or more tables a…

rareddy:
…s agree upon as the commented code with joins, single table will same as before


rareddy:
@shawkins updated PR by removing the column types and replacing tabs with spaces.

PR #7704, opened 2020-01-27
fix: Increase default message timeout for integration tests

christophd:
Integration tests are flaky on CircleCi because of message receive timeouts
- Use default timeout setting for all mocked server components
- Make default timeout settable via system properties
- Add default message timeout setting on syndesis CLI
- Increase timeout setting for CircleCI build jobs to 180s

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -564,7 +564,7 @@ jobs:
             mkdir -p ~/integrations/
             mkdir -p ~/junit/
             mkdir -p ~/logs/
-            ./tools/bin/syndesis integration-test --output /home/circleci/integrations | tee ~/logs/test_log.txt
+            ./tools/bin/syndesis integration-test --output /home/circleci/integrations --default-timeout 180 | tee ~/logs/test_log.txt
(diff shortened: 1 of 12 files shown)


zregvart:
That seems like a long timeout, especially for a simulated environment, are we sure the cause for test failure is slowness?

christophd:
@zregvart I can not reproduce these timeouts on a local environment. The timeout happens when the Syndesis integration runtime Docker container takes a long time to start. Probably pulling some images or artifacts or something.

The simulated http server in tests then raises a request timeout because it is missing the 1st request from the running Syndesis integration.

I would try with that long timeout setting and see if tests are green. They have not been successful for a long time on CircleCi and I want to have a green build for once in while.

zregvart:
@christophd yeah, let's try this. We could also try pulling images before running tests. I'm afraid of adding timeouts, as they seem like a quick solution to the problem but might end up hiding the real issue, I'd much rather invest in test stabilisation and performance rather than resorting to quick solutions that might fail in the future again.

christophd:
@zregvart that's true. Usually you do not have much time to investigate on flaky tests though and this is something that gets postponed all the time because people got other things to do

Issue #7419, opened 2019-12-16
Syndesis step extension fails when included in an integration with an API endpoint


sigreen:
I am able to successfully import a custom step extensions and use it in an integration that does not include an API endpoint e.g.

Timer -> custom step -> log

If I import a Swagger doc and setup and API-based integration then try to include the custom step, the UI hangs and I see the following stacktrace in syndesis-server.log:

```
2019-12-16 16:36:10.422 ERROR [-,9c5044b9210ee566,9c5044b9210ee566,false] 1 --- [  XNIO-3 task-6] .s.e.v.h.e.EntityNotFoundExceptionMapper : Entity Not Found Exception null

javax.persistence.EntityNotFoundException: null
	at io.syndesis.server.endpoint.v1.operations.Getter.get(Getter.java:39) ~[server-endpoint-1.7.1-20190617.jar!/:1.7.1-20190617]
	at io.syndesis.server.endpoint.v1.handler.connection.ConnectorHandler.get(ConnectorHandler.java:117) ~[server-endpoint-1.7.1-20190617.jar!/:1.7.1-20190617]
	at io.syndesis.server.endpoint.v1.handler.connection.ConnectorHandler.getConnectorIcon(ConnectorHandler.java:136) ~[server-endpoint-1.7.1-20190617.jar!/:1.7.1-20190617]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_151]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_151]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_151]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[na:1.8.0_151]
	at org.jboss.resteasy.core.ResourceLocatorInvoker.createResource(ResourceLocatorInvoker.java:69) ~[resteasy-jaxrs-3.6.1.Final.jar!/:3.6.1.Final]
	at org.jboss.resteasy.core.ResourceLocatorInvoker.createResource(ResourceLocatorInvoker.java:48) ~[resteasy-jaxrs-3.6.1.Final.jar!/:3.6.1.Final]
[...]

PR #7715, opened 2020-01-28
ENTESB-12544 Syndesis step extension fails when included in an integration with an API endpoint

claudio4j:
https://issues.jboss.org/browse/ENTESB-12544

Fixes https://github.com/syndesisio/syndesis/issues/7419

### DIFF
--- a/app/ui-react/syndesis/src/shared/EntityIcon.tsx
+++ b/app/ui-react/syndesis/src/shared/EntityIcon.tsx
@@ -39,9 +39,13 @@ function getConnectionIcon(
     connection.icon.toLowerCase().startsWith('db:') ||
     connection.icon.toLowerCase().startsWith('extension:')
   ) {
-    return `${apiUri}/connectors/${connection.connectorId}/icon?${
-      connection.icon
-    }`;
+    if (connection.connectorId) {
+      // on Connections, Integrations screen, when there is an existing Connection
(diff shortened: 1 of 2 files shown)


app/ui-react/syndesis/src/shared/EntityIcon.tsx
```diff
@@ -39,9 +39,11 @@ function getConnectionIcon(
     connection.icon.toLowerCase().startsWith('db:') ||
     connection.icon.toLowerCase().startsWith('extension:')
   ) {
-    return `${apiUri}/connectors/${connection.connectorId}/icon?${
-      connection.icon
-    }`;
+    if (connection.connector) {
```
gashcrumb: Could you maybe throw a comment in here what edge condition this `if`.`else` statement covers?  This function for sure needs plenty of supporting info IMO, each time I come in here (and probably most other folks) I've gotta remember what's going in here :-)

claudio4j: The previous `connection.connectorId` property is `undefined` when connection is an extension of type `Connection`, this happens on *Create Connection* and *Connections* screen. When the `Connection` is from an `Extension` use the `connection.connector.id` to retrieve it from the `ConnectorHandler` endpoint.

claudio4j: I commented on the code, see if that helps. Thanks for the review.

PR #7730, opened 2020-01-28
TEIIDTOOLS-922 deleting the single service

shawkins:


shawkins:
This should be tacked onto 922 - the delete wasn't updated to the new naming convention either.

PR #7732, opened 2020-01-28
ENTESB-12645: Extend db backup to cover external databases

phantomjinx:
* cmd/.../backup.go
 * operator-config switch required for command line execution
 * requires installed syndesis CR to be passed in

* install_forge.go
 * Tidy up with centralising of code into util functions

* cmd/.../restore.go
 * operator-config switch required for command line execution
 * custom-options switch made available for command line execution if
   alternative custom options to pg_restore are preferred
 * requires installed syndesis CR to be passed in

* backup/syndesis...yml.tmpl
 * Templates for Jobs to be executed by backup and restore

* backup/[backup|restore].go
 * Major refactoring to generate a Job which creates a delegate Pod for
   running pg_dump/pg_restore which both connect remotely to the db

* configuration.go
 * BackupConfiguration provides conduit between code and templates


heiko-braun:
@phantomjinx Looks like this needs to be rebased.

install/operator/pkg/syndesis/backup/backup.go
```diff
@@ -34,30 +35,54 @@ import (
...
+	backupPath      string
+	Namespace       string
+	BackupDir       string
+	Delete          bool
+	LocalOnly       bool
+	Context         context.Context
+	Client          *client.Client
+	Syndesis        *v1beta1.Syndesis
+	CustomOptions   string
+	synConfig       *configuration.Config
```
lgarciaaco: I believe `sysConfig` doesn't need to use pointer semantic?

install/operator/pkg/syndesis/backup/backup.go
```diff
@@ -193,35 +218,273 @@ func (b *Backup) ensureDir() (err error) {
...
+// Execute a task within a Job
+//
+func (b *Backup) executeJob(jobName string, jobTask jobTaskFunc) error {
+	//
+	// TODO
+	//
+	// Since the job keeps dying after 8-12 seconds regardless of the pod
+	// still running (maybe a 3.11 issue?), it is necessary to monitor the
+	// target pod instead. Is the job is more trouble than its worth??
+	//
```
lgarciaaco: Is this really the case? Very fishy, did you try 4.x?

install/operator/pkg/util/kube.go
```diff
@@ -173,3 +180,33 @@ func Exec(o ExecOptions) error {
...
+			CONTROLLER_UID_LABEL: controllerUid,
+			JOB_NAME_LABEL:       job.Name,
+		})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return podList, nil
+}
```
lgarciaaco: The method is called only once from `backup,go`, no reason for having it here instead of in the package where it is called from, actually it is totally fine not to have a method for this since it is called once.

install/operator/pkg/util/kube.go
```diff
@@ -173,3 +180,33 @@ func Exec(o ExecOptions) error {
...
+
+/**
+ * Refreshes a pod with its latest state on the server
+ */
+func RefreshPod(ctx context.Context, c client.Client, pod *corev1.Pod) error {
```
lgarciaaco: Similar comment to previous one

install/operator/pkg/syndesis/backup/backup.go
```diff
@@ -193,35 +218,273 @@ func (b *Backup) ensureDir() (err error) {
...
+			//
+			return true, nil
+		}
+
+		//
+		// Still waiting on pods being created by job
+		//
+		return false, nil
+	})
+
```
lgarciaaco: The wait into a wait approach is confusing. Can't we really use the job to figure out these stuff? If we can't use the job for this case, then it is probably better to spin up a pod directly because using a job that 'keeps dying' turns from being handy to being a trouble.

install/operator/pkg/syndesis/backup/restore.go
```diff
@@ -52,6 +65,10 @@ func (b *Backup) Restore() (err error) {
 			return err
 		}
 
+		nfi, err := os.Stat(dir)
+		if !nfi.IsDir() {
+			return fmt.Errorf("Unzipped backup directory does not exist: %s", dir)
+		}
 		b.BackupDir = dir
```
lgarciaaco: Why is this needed, dir is surely created few lines above or?
```
		dir, err := ioutil.TempDir("/tmp", "restore-")
		if err != nil {
			return err
		}
		defer os.RemoveAll(dir)
		os.Chmod(dir, 0755)
```

install/operator/pkg/syndesis/backup/restore.go
```diff
@@ -63,9 +80,9 @@ func (b *Backup) Restore() (err error) {
 		return
 	}
 
-	if err = b.RestoreResources(); err != nil {
-		return
-	}
+	// if err = b.RestoreResources(); err != nil {
+	// 	return
+	// }
```
lgarciaaco: Did you forget to uncomment this?

install/operator/pkg/syndesis/backup/backup.go
```diff
@@ -290,15 +548,15 @@ func (b *Backup) clientConfig() *rest.Config {
...
 		cl, err := client.New(b.clientConfig(), client.Options{})
 		if err != nil {
-			return nil, err
+			util.ExitOnError(err)
 		}
 		b.Client = &cl
 	}
-	return *b.Client, nil
+	return *b.Client
 }
```
lgarciaaco: I don't think this is a good approach and this is why. If the operator is scheduled to run a backup(and we can afford that a backup fails, because it is a backup right) and this method fails for whatever reason, then this will kill the operator.

Rather wrap the error and return I would say.

install/operator/pkg/util/syndesis_object.go
```diff
@@ -0,0 +1,87 @@
+/*
```
lgarciaaco: I don't think this file should exist at all

install/operator/pkg/syndesis/backup/restore.go
phantomjinx: Existing code for handling a backup directory that has been zipped up. On the command line it is possible to restore by specifying an aforementioned zip archive.

phantomjinx:
Hey @zregvart @heiko-braun 
In Luis' absence, would you mind giving this update a once-over, please?

As far as tests go, many more could be added but I really need to talk to Luis first in order to understand how to approach unit testing functions such as Backup.Run(). Of course, if you have any input on that I'd be really interested.

Thanks

PGR

PR #7733, opened 2020-01-28
fix(ENTESB-12325): mv edit action to int list row, rm check for canEdit

kahboom:
- fixes [ENTESB-12325](https://issues.redhat.com/browse/ENTESB-12325)

Screenshot 2020-01-28 21 03 53


amysueg:
CC: @gaughan @dongniwang

PR #7738, opened 2020-01-29
chore: fix permissions for batch.jobs

lgarciaaco:
Fixes https://issues.redhat.com/browse/ENTESB-12846


heiko-braun:
@lgarciaaco Does it need to be backported?

lgarciaaco:
> @lgarciaaco Does it need to be backported?

yes I will make sure it is backported after it is merged

PR #7741, opened 2020-01-29
[ENTESB-12785] Align MongoDB version. Now also using the same version Camel uses.

Delawen:
Same as #7689 but for master


christophd:
@Delawen having some merge conflicts here

PR #7782, opened 2020-01-31
chore: Update pf4 version and migrate some components

gashcrumb:
relates to [ENTESB-12896](https://issues.redhat.com/browse/ENTESB-12896)

heads up @syndesisio/qe so far the biggest change was migrating the integration list, some tests could require adjustments as part of this.

Note, a bunch of changes are also related to refactoring components to function components, which doesn't affect the structure.

This chart was updated:
image

Recent items:
image

Integration list, buttons are in a separate component and additional polish will come later:
image

This angle bracked on the detail page:
image

Connection detail form:
image

The tooltip that shows up in the collapsed IVP and I can't seem to get a screenshot easily.

### DIFF
--- a/app/ui-react/packages/auto-form/package.json
+++ b/app/ui-react/packages/auto-form/package.json
@@ -77,7 +77,7 @@
     }
   },
   "dependencies": {
-    "@patternfly/react-core": "^3.120.5",
+    "@patternfly/react-core": "^3.122.4",
     "@patternfly/react-icons": "^3.14.18",
     "@patternfly/react-styles": "^3.6.5",
     "formik": "^1.5.4"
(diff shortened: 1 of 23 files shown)


gashcrumb:
BTW I'd be fine with leaving this unmerged for now as it could cause undue hassle when backporting bug fixes.

heiko-braun:
@gashcrumb let's keep it for 7.7

gashcrumb:
@heiko-braun oh yeah, that was never the plan, this effort targets 7.7.  I'm saying I'd be fine keeping it out of master too until it's less likely other UI changes need backporting to 1.9.x due to some components requiring restructuring etc.

gashcrumb:
As I think about it, I was hoping to do this in batches of files, like this has 10 files, but I could probably break out a separate PR to do the pf4 version update + chart addition and focus on getting that on master first, then focus on a PR per file or something to keep it more manageable.  How does that sound?

gashcrumb:
I'll go that route I think, easy enough to just cherry-pick changes from this branch to start off with.

PR #7801, opened 2020-02-03
fix: If a step has no configurable properties, show that and don't error out

gashcrumb:
fixes [ENTESB-12888](https://issues.redhat.com/browse/ENTESB-12888)

image

This also adds the same message for the split step as it also doesn't currently have configuration either.

@claudio4j FYI


gashcrumb:
Wonder if it's time to bump the version of yarn on CI:

image

double-checking if I see the same error locally.

gashcrumb:
Builds ok here:

image

gashcrumb:
Well shoot the maven build works fine for me.

gashcrumb:
3rd time was the trick apparently.

heiko-braun:
@gashcrumb Do you backport it manually?

gashcrumb:
yeah, will do that in a few...

PR #7804, opened 2020-02-04
TT-928: correcting the connection synchronization logic to syndesis d…

rareddy:
…ata sources, and removing unsed source schema files

### DIFF
--- a/app/dv/src/main/java/io/syndesis/dv/RepositoryManager.java
+++ b/app/dv/src/main/java/io/syndesis/dv/RepositoryManager.java
@@ -61,7 +61,7 @@ public interface RepositoryManager {
 
     SourceSchema createSchema(String id, String name, String contents);
 
-    List<String> findAllSchemaNames();
+    List<String> findAllSourceIds();
 
 
     DataVirtualization createDataVirtualization(String virtualizationName);
--- a/app/dv/src/main/java/io/syndesis/dv/openshift/SyndesisConnectionSynchronizer.java
+++ b/app/dv/src/main/java/io/syndesis/dv/openshift/SyndesisConnectionSynchronizer.java
@@ -17,8 +17,7 @@ package io.syndesis.dv.openshift;
 
 import java.sql.Connection;
 import java.util.Collection;
-import java.util.Map;
-import java.util.stream.Collectors;
+import java.util.List;
 
 import javax.sql.DataSource;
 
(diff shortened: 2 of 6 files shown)


shawkins:
To answer the first question from the JIRA the situation you are trying to address is when the dv pod restarts and not all sources still exist in syndesis.  In that case any source that successfully created a schema that no long exists is an orphan.  My question is that seems to be a pretty narrow case - are we sure that is what is occurring?

The rest of the questions/comments above should still remain valid.

rareddy:
>when are you seeing that you have a source schema that you need to remove, but no datasource >present?

Yes. The syndesisSourceStatus, walks over the schema lists and finds the schema that do not have a data source assosiated with it. Since previous logic always assumed that for every source schema there is a datasource avaialble, it ends up in NPE, thus not completing call.

> This also seems to have removed the deletion of a datasource based upon this synchronization.

I only separated `deleteConnection` and `synchronizeConnections` use separate methods. The 
datasource delete should work as is.

>And it's possible to have a datasource that does not have a schema (sad mongo), so we can't base >the deletion of the datasource solely on existing schema.

`handleDelecteConnection` is not modified

> Finally completely eliminate timing issues, the iteration in getSyndesisSourceStatuses should be >over the TeiidDataSources

Sure, I can correct that one

> are we sure that is what is occurring?

Not sure how the orphan schema got accrued with no backing datasources, that was just guess. But that is what I have seen when I looked through debugger for staging. I think, you can verify that looking at `sourceschema` table and distinct `sourceid` columns aginst the number of connections that DV supports in the connections tab.

shawkins:
> I only separated deleteConnection and synchronizeConnections use separate methods. The
datasource delete should work as is.
> handleDelecteConnection is not modified

I'm not saying handleDeleteConnection needs modified, I'm saying that you removed/altered a call to handleDeleteConnection from synchronizeConnections.  With the change you are proposing if we loose connection to syndesis when a delete event is broadcast, then the schema will be deleted when we later synchronize, but the datasource will be orphaned until teiid-syndesis restarts.

> Not sure how the orphan schema got accrued with no backing datasources, that was just guess.  But that is what I have seen when I looked through debugger for staging. 

I don't see a debugging port on the dv pod, how are you attaching a debugger?

It would be great to know what is happening.  Given that this is not one of the race/startup conditions mentioned on the JIRA, there's only a couple of explanations.  It's either we were down when the once successful source was removed or a logic error.  The only place a source schema is created is as a side effect of getUniqueKomodoName, which is called by TeiidOpenShiftClient.createDataSource.  In the current code it's assumed that the following call to registerDataSource will succeed and thus a datasource will also be created.  If there is some path by which registerDataSource is not being called / completed that would create an orphaned source schema - for example would be an unexpected exception from DataSourceDefinition.createDatasource.

rareddy:
> but the datasource will be orphaned until teiid-syndesis restarts.
oh, I see. I guess I was thinking connectionSynchronization only happens once during start, I will fix that

> I don't see a debugging port on the dv pod, how are you attaching a debugger?
staging has port open.

rareddy:
updated to only run at startup

shawkins:
What do you mean to only run at startup?

rareddy:
When `update = false` which only happens during the startup, so modified `handleConnections` to work in new way only when `update=false`

PR #7823, opened 2020-02-05
backport: Adapt DV UI to publishing/deployment backend changes - TEIIDTOOLS-914

mdrillin:
Backport for 1.9.x.  These changes adapt the UI to changes on the DV backend to resolve virtualization publishing issues.
- model changes to add deployedState, deployedRevision, deployedMessage
- VirtualizationVersionsPage changes - the version table can now show published virtualization version and in-progress virtualization simultaneously.
- PublishStatusWithProgress changed to utilize status label same as virtualizationListItem

### DIFF
--- a/app/ui-react/packages/api/src/useVirtualization.tsx
+++ b/app/ui-react/packages/api/src/useVirtualization.tsx
@@ -16,6 +16,7 @@ export const useVirtualization = (
 ) => {
   const { read, resource, ...rest } = useApiResource<Virtualization>({
     defaultValue: {
+      deployedState: 'NOTFOUND',
       description: '',
       empty: true,
       id: '',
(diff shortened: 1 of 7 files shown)


shawkins:
I think this looks good.  The only comment would be that with the second commit on TEIIDTOOLS-914 you won't see the COMPLETE state.  If the build is complete, then the backend will set the published state / message to the deployed state / message.

rareddy:
@shawkins @indraraj can we merge this now that backend changes have been merged?

shawkins:
See the previous comment.  I did make a second commit that remove the need for the UI to consider the build COMPLETE state, so that it would be more consistent with the old UI logic.  

However the changes here still account for COMPLETE (see line 109), so I'm thinking we should get on the same page - either simplify these to remove COMPLETE, or change the backend back to sending COMPLETE as a terminal build state.

indraraj:
I can look to incorporate the changes required form ui side on Monday.

indraraj:
raised diff pr for this issue

PR #7829, opened 2020-02-06
fix: failing operator backup test

phantomjinx:
Test passes locally and on circleci but fails on jenkins so shoring up to be more robust.

### DIFF
--- a/install/operator/pkg/syndesis/backup/backup_test.go
+++ b/install/operator/pkg/syndesis/backup/backup_test.go
@@ -3,6 +3,8 @@ package backup
 
 import (
 	"context"
+	"io/ioutil"
+	"os"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


install/operator/pkg/syndesis/backup/backup_test.go
```diff
@@ -48,11 +51,15 @@ func TestBackupRun_InvalidDirectory(t *testing.T) {
 	ns := "syndesis"
 	s, _ := v1beta1.NewSyndesis(ns)
 
-	b, err := NewBackup(ctx, cl, s, "/foo")
+	dirpath := "/tmp/foo" + strconv.FormatInt(time.Now().Unix(), 10)
```
zregvart: use [ioutil.TempDir](https://golang.org/pkg/io/ioutil/#TempDir)

PR #7835, opened 2020-02-06
[ENTESB-11725] For Master: Google Sheets Validation was broken

Delawen:
New Google API version returns a 404 error when the spreadsheet ID is not valid. Now we make sure that is not a validation error. A validation error will pop up when credentials are wrong (or some other unexpected issue).

### DIFF
--- a/app/connector/google-sheets/pom.xml
+++ b/app/connector/google-sheets/pom.xml
@@ -29,6 +29,9 @@
   <packaging>jar</packaging>
 
   <properties>
+    <!-- To review when api client gets updated if 1.34.0 is available for this -->
+    <!-- This dependency is aligned with the camel version -->
+    <google-api-camel-client-version>1.22.0</google-api-camel-client-version>
     <google-api-client-version>1.34.0</google-api-client-version>
     <google-api-services-sheets-version>v4-rev551-1.22.0</google-api-services-sheets-version>
   </properties>
@@ -44,6 +47,18 @@
             <groupId>org.mortbay.jetty</groupId>
             <artifactId>servlet-api</artifactId>
           </exclusion>
+          <exclusion>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava-jdk5</artifactId>
+          </exclusion>
+        </exclusions>
+      </dependency>
+      <!-- This also could be removed once we align versions up -->
+      <dependency>
+        <groupId>com.google.api-client</groupId>
+        <artifactId>google-api-client</artifactId>
+        <version>${google-api-camel-client-version}</version>
+        <exclusions>
           <exclusion>
             <groupId>com.google.guava</groupId>
             <artifactId>guava-jdk5</artifactId>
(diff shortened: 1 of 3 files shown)


app/connector/google-sheets/src/main/java/io/syndesis/connector/sheets/GoogleSheetsVerifierExtension.java
```diff
@@ -0,0 +1,63 @@
...
+                Integer code = e.getStatusCode();
+                //We may have a "404 spreadsheet no found" which is fine
+                //What we don't want is a bad credential error or similar
+                if (code != 404) {
+                    //Move to outer error
+                    throw e;
+                }
+            }
+        } catch (Exception var6) {
+            var6.printStackTrace();
```
christophd: `printStackTrace()` writes to `System.out` I would rather log via logging framework. In general I doubt that we need to log the stacktrace here (maybe some leftover from debugging session?).

app/connector/google-sheets/src/main/java/io/syndesis/connector/sheets/GoogleSheetsVerifierExtension.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public GoogleSheetsVerifierExtension(String defaultScheme, CamelContext context) {
+        super(defaultScheme, context);
+    }
+
+    @Override
+    protected Result verifyConnectivity(Map<String, Object> parameters) {
+        ResultBuilder builder = ResultBuilder.withStatusAndScope(Result.Status.OK, Scope.CONNECTIVITY);
+
+        try {
```
christophd: We can avoid the nested try-catch when adding a private helper method that constructs the error result from generic exception. Both catch clauses can then call this private helper. Something like:

```
try {
...
} catch (GoogleJsonResponseException e) {
    ...
    if (code != 404) {
        builder.error(getErrorFromException(e));    
    }
} catch (Exception ex) {
    builder.error(getErrorFromException(e));    
}
```

app/connector/google-sheets/src/main/java/io/syndesis/connector/sheets/GoogleSheetsVerifierExtension.java
```diff
@@ -0,0 +1,63 @@
...
+            } catch (GoogleJsonResponseException e) {
+                Integer code = e.getStatusCode();
+                //We may have a "404 spreadsheet no found" which is fine
+                //What we don't want is a bad credential error or similar
+                if (code != 404) {
+                    //Move to outer error
+                    throw e;
+                }
+            }
+        } catch (Exception var6) {
```
christophd: we should use a better name for `var6`

app/connector/google-sheets/src/main/java/io/syndesis/connector/sheets/GoogleSheetsVerifierExtension.java
Delawen: Right, leftover :)

app/connector/google-sheets/src/main/java/io/syndesis/connector/sheets/GoogleSheetsVerifierExtension.java
Delawen: Sounds better

app/connector/google-sheets/src/main/java/io/syndesis/connector/sheets/GoogleSheetsVerifierExtension.java
Delawen: Changed!

Delawen:
Wow, the merge button was disabled for me :) Thanks!

PR #7837, opened 2020-02-06
feat(dv): teiidtools-904 prototype teiid ddl language server framework

blafond:
Contains tokenizing, parsing framework and initial code-completion language server capability for Teiid DDL data virtualization editing feature


blafond:
You are correct on AbstractLanguageServerRunnable.  Hold over from one of my earlier versions.

Correct on  moving TeiidDdlLanguageServerRunner and TeiidDdlWebSocketRunner to test. The first one is used through eclipse to set up a local language server and probably convert it to be generic for the test classes ... or not if we tie these tests to a Teiid server instance.

The threading model is from a language server example.  I don't yet have the chops to clean it up for our server/dv environment, so I'd welcome your input on how this should look.

The logging is currently for initial development.  Before this PR would be merged, I'll be cleaning that up.

shawkins:
Starting to work this more I see that there are at least 3 execution models:

1. in process - you hand the LSP server some streams like TeiidDdlLanguageServer
2. remote as a single process per server - this is where the exit method and checking of the parent process is necessary.  I don't see anything in the current code what would seem to launch instances this way.  This seems to be needed to support cases where the jre of the lsp server could be different than that of the parent process.
3. remote with a server instance per websocket connection

So it would appear that for this to run in the teiid-syndesis vm, we'd need to do #3 and launch a separate websocket.  Alternatively you could run the lsp in it's own process and make service calls into teiid-syndesis.

rareddy:
Going with (3)  is there a "State" that LSP server maintaining to keep a separate connection? Can we not introduce some kind of "client-id" and process the request in async fashion in backend?

shawkins:
> Going with (3) is there a "State" that LSP server maintaining to keep a separate connection? 

The websocket layer has it's own notion of sessions.  Each session gets its own lsp server instance.

> Can we not introduce some kind of "client-id" and process the request in async fashion in backend?

I don't entirely follow you.  Are you asking about how websocket lsp client requests are processed?  Or does the first answer address this.

shawkins:
I've opened a pr against your branch @blafond.  My only change is https://github.com/blafond/syndesis/pull/3/commits/a6a494f502ba1e753b8e63dd5a74b0e8726a4be9 the rest are rebasing you to current master.  The changes start the websocket lsp server in syndesis using all defaults for host/port/context.  It's either this approach or running the lsp in it's own process and making remote calls into teiid-syndesis.

rareddy:
> The websocket layer has it's own notion of sessions. Each session gets its own lsp server instance.

Basically, I am questioning the need for the instance per WebSocket connection? What kind of state does LSP server has where this is required? This seems like the potential for OOM issues based on number of sessions, I was thinking if there is any way to make it into a single LSP server instance.

shawkins:
> Basically, I am questioning the need for the instance per WebSocket connection?

That's the way the framework is implemented. The server instance holds the reference to the client, so yes that instance object is needed.

> What kind of state does LSP server has where this is required? 

You are jumping the gun a little.  Until we have things a little more fleshed out I wouldn't start optimizing.  The other state that the lsp server instance is holding includes LSP4j constructs like TeiidDdlTextDocumentService - we'd have to look into that and other classes a good bit to understand the thread safety / usage pattern.  Just a quick search shows examples of the document service holding a reference to it's LSP server - https://github.com/angelozerr/lsp4xml/blob/master/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLTextDocumentService.java

Ultimately without profiling it would just be a stab in the dark at how much transitive state is being held...

blafond:
mostly.. I vote for (1) above and "expose a route to the websocket that we start in our pod"

blafond:
> > Basically, I am questioning the need for the instance per WebSocket connection?
> 
> That's the way the framework is implemented. The server instance holds the reference to the client, so yes that instance object is needed.
> 
> > What kind of state does LSP server has where this is required?
> 
> You are jumping the gun a little. Until we have things a little more fleshed out I wouldn't start optimizing. The other state that the lsp server instance is holding includes LSP4j constructs like TeiidDdlTextDocumentService - we'd have to look into that and other classes a good bit to understand the thread safety / usage pattern. Just a quick search shows examples of the document service holding a reference to it's LSP server - https://github.com/angelozerr/lsp4xml/blob/master/org.eclipse.lsp4xml/src/main/java/org/eclipse/lsp4xml/XMLTextDocumentService.java
> 
> Ultimately without profiling it would just be a stab in the dark at how much transitive state is being held...

At least for the completion calls, there should be minimal held state. The Document Service owned by our language server supports multiple documents, but in our case it'll be 1 per DdlEditor. 

The document is updated on each edit and the completion call with Cursor ends up parsing and returning completion items response.  The parsing process should be garbage collected after service call.  Diagnostics will use the same basic framework and occur on any open/change/close document event.

shawkins:
I failed to make the websocket be more spring managed - I'm hitting what looks like https://github.com/spring-projects/spring-framework/issues/22131 when changing TeiidDdlWebSocketEndpoint into an annotated ServerEndpoint.  The issue seems to indicate that this is fixed with the latest tomcat, but upgrading our dependency still did not resolve this.  So I'll give up on making that code tighter for now.  I'll add a commit tomorrow to update the operator assets to change the dv pod to use named ports, and add a websocket port to our service.  We can then manually a route to expose it.

rareddy:
I did not understand the need to expose WebSocket port as route? I do not see similar requirement with syndesis-server's WebSocket implementation having similar limitation, what is different in our case? If this is because of VS Code stuff, IMO we should not bring those concerns just yet into this.

shawkins:
> I did not understand the need to expose WebSocket port as route?

See the previous comments.  It's a question of where the lsp server lives.

> If this is because of VS Code stuff, IMO we should not bring those concerns just yet into this.

No, this is because of https://issues.redhat.com/browse/TEIIDTOOLS-917 which expects an lsp backend.

shawkins:
Closing as a new pr has been opened.

PR #7839, opened 2020-02-06
Teiidtools 914

shawkins:


app/dv/src/main/java/io/syndesis/dv/server/endpoint/DataVirtualizationService.java
```diff
@@ -159,19 +161,38 @@
 
     private RestDataVirtualization createRestDataVirtualization(final DataVirtualization virtualization) throws KException {
         RestDataVirtualization entity = new RestDataVirtualization(virtualization);
```
rareddy: Is there any advantage in having "VirtualizationStatus" and as well as "RestDataVirtualization" feels like we can do away with single class?

shawkins: I wouldn't make that kind of change.  One defines the contract with the client which includes additional state from the repository, the other is internal and simply conveys the latest build/deployment status.

PR #7842, opened 2020-02-07
bug(dv) TEIIDTOOLS-938 validation should not cache view definitions

shawkins:


shawkins:
> can we use this technique for "live" vdb?

To clarify here the way we were validating was confusing to the metadata layer.

Are you asking if we can not cache anything related to metadata, and inject new view definitions on save?  

The code is already setup to replace view definitions if the projected columns don't change - that's just our standard alter (although it is non-standard in implementation because it does not respect transaction boundaries).

The issue we have is when the columns change, then there's more stuff that we need to prevent from being cached (mainly system/pg materialization - which isn't that much of an issue here), concurrency issues with accessing the columns and other associated metadata of a modified table, and if we are going to support multiple server instances the problem of broadcasting metadata changes or directly wiring the definitions to the database - currently the code checks to see if vdb has been altered since it was last loaded, then reloads if needed. 

Cached results and prepared plans should automatically invalidate based upon the existing alter logic.

The only downside for the live vdb would be constantly re-parsing/resolving/validating, but presumably that is less expensive that fully reloading the vdb - and you'd eventually mitigate with more intelligent, rather than fully disabled metadata caching.

rareddy:
>currently the code checks to see if vdb has been altered since it was last loaded, then reloads if needed.

Curious, wherein the code this happens?

> constantly re-parsing/resolving/validating, but presumably that is less expensive that fully reloading the vdb

Last time when we are looking into this we didn't have a decent dependency graph of the view to invalidate, now maybe this seems possible? No, I am not thinking in the immutable environments, more possibilities may be with our Microservices version updates.

shawkins:
> Curious, wherein the code this happens?

We are tracking by the dv version number that gets incremented with every change:

https://github.com/syndesisio/syndesis/blob/b69c147022dd80cd84c488bc4a0ff2521d052e94/app/dv/src/main/java/io/syndesis/dv/server/endpoint/MetadataService.java#L216

> Last time when we are looking into this we didn't have a decent dependency graph of the view to invalidate, now maybe this seems possible? 

We do have the dependencies and we do track result set and prepared plans by them.  Currently a view definition query expression change invalidates those entries, but we don't have (or yet need) logic that transitively removes anything from lower level metadata caches, and of course we don't have anything tied into materialization.  The simple thing that I thought you were proposing is just to prevent any metadata caching.  Having the transitive logic helps address the constantly re-parsing / resolving / validating issue.

There's actually two cases for structural view changes:
- for the development time (teiid-syndesis) we need to tolerate a view being invalid.  That's actually where the problem comes into play of needing to re-resolve / validate all of the references to the changed view.
- in normal usage you should only allow view modifications that are fully valid (which of course we do currently validate for altering just the query expression), so you don't need to re-resolve / validate everything, but you do need to have better concurrency control around the metadata modification and some system materialization may need reloaded.

> No, I am not thinking in the immutable environments, more possibilities may be with our Microservices version updates.

We're probably getting our wires crossed a little.  I'm thinking that you are asking for teiid-syndesis and the usage scenario of changing views without reloading the vdb.  In that case you do get throw out a lot of "real" considerations of caching / transactions.  In general scenarios actual modifications to our metadata require more work.

PR #7845, opened 2020-02-07
chore: Fix integration tests to not wait for other containers to start

christophd:
Nightly integration tests were failing for quite some time because of a Http request timeout in the Http-to-AMQ tests. I boiled this down to the integration runtime container wait strategy not being able to wait for the AMQ testcontainer to be pulled and started. Therefore the test logic started way too early (while container images still being pulled) and a Http request timeout causes the tests to fail.

Now using a static initializer for AMQ testcintainer instead of the @ClassRule

### DIFF
--- a/app/test/integration-test/src/test/java/io/syndesis/test/itest/amq/AMQToHttp_IT.java
+++ b/app/test/integration-test/src/test/java/io/syndesis/test/itest/amq/AMQToHttp_IT.java
@@ -40,7 +40,6 @@ import org.springframework.test.context.ContextConfiguration;
 import org.springframework.util.SocketUtils;
 import org.testcontainers.Testcontainers;
 import org.testcontainers.containers.GenericContainer;
-import org.testcontainers.containers.wait.strategy.Wait;
 
 /**
  * @author Christoph Deppisch
@@ -59,8 +58,12 @@ public class AMQToHttp_IT extends SyndesisIntegrationTestSupport {
     @Autowired
     private JmsEndpoint todoJms;
 
-    @ClassRule
-    public static JBossAMQBrokerContainer amqBrokerContainer = new JBossAMQBrokerContainer();
+    private static final JBossAMQBrokerContainer amqBrokerContainer;
+
+    static {
+        amqBrokerContainer = new JBossAMQBrokerContainer();
+        amqBrokerContainer.start();
+    }
 
     /**
      * Integration waits for messages on AMQ queue and maps incoming tasks to Http service. Both AMQ and Http connections use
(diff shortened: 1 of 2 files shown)


zregvart:
AFAIK ClassRules are run before any test methods. Perhaps instead of making the `JBossAMQBrokerContainer` static field we should try to [wait](https://www.testcontainers.org/features/startup_and_waits/) for it to startup.

christophd:
The AMQ Testcontainer already has a wait strategy on the log message in order to be sure that the message broker is started. 

Unfortunately the integration runtime Testcontainer is not waiting for this to happen. It also starts right away and fails because the AMQ images has to be pulled first. As a result of this the wait strategy on the integration runtime container also fails to hold the test from starting.

christophd:
@gashcrumb what's wrong with the UI build?

gashcrumb:
@christophd am looking at that now, think the bit installing prerequisites needs cleaning up.

gashcrumb:
This is what I'm trying btw https://github.com/syndesisio/syndesis/pull/7848

zregvart:
@christophd can we wait on a port being available, would that wait strategy work better? Perhaps we could create an issue at the testcontainers project and have this fixed there?

bsideup:
@zregvart FTR Testcontainers supports and uses the port-based strategy by default, it can also be explicitly enabled with `waitingFor(Wait.forListeningPort())`

bsideup:
> Therefore the test logic started way too early (while container images still being pulled)

`Startable#start` (the same API as `@ClassRule` calls) is blocking and won't return until the image is pulled and container either started or failed (will throw an error in case of a startup failure)

I see that you're using networks, and, since JUnit does not guarantee the order of rules, probably were hitting the issue where `integrationContainer` was attempted to start before `amqBrokerContainer` (just a guess). If it is the case, you can use `dependsOn(amqBrokerContainer)` to ensure that the broker always starts before the `amqBrokerContainer`.

bsideup:
> Unfortunately the integration runtime Testcontainer is not waiting for this to happen. It also starts right away and fails because the AMQ images has to be pulled first.

it will indeed pull the images first, but then the waiting strategy will be applied and it won't be considered started until it signals a success.

christophd:
@bsideup thanks for the explanation! I will try to use `dependsOn` then and see if this does the trick. I think I remember to have used this before and have had some issues with it but I can't really remember exactly. So I will try

zregvart:
Thanks @bsideup!

christophd:
@bsideup `dependsOn` worked for me! Many thanks!

PR #7846, opened 2020-02-07
fix(deps): exclude javax.annotation-api

zregvart:
Seems that there are duplicate but different classes in
`javax.annotation:javax.annotation-api:1.3.2` vs
`org.jboss.spec.javax.annotation:jboss-annotations-api_1.2_spec:1.0.2.Final`.

This excludes the `javax.annotation:javax.annotation-api` as the JBoss
one seems to be used more.

Caveat, on `master` we have excluded both the javax ones and the JBoss
ones in favor of the Jakarta versions.


zregvart:
Retest this please

PR #7847, opened 2020-02-07
Upgrade to Camel 3

zregvart:
I'm going to create a draft PR based on the `camel-3.0.x` branch to enable discussion around this.


zregvart:
Bare in mind that this branch now tracks Camel 3.1.0-SNAPSHOT, the snapshot releases that are available at [ASF snapshot repository](https://repository.apache.org/content/groups/snapshots/) do not necessarily reflect what's latest on the [`master`](https://github.com/apache/camel/) branches of camel or [`camel-spring-boot`](https://github.com/apache/camel-spring-boot/). 

So usually you need to pull and build those projects locally, the easiest way is to run `./mvnw -Pfastinstall install` in the root of those projects.

PR #7850, opened 2020-02-07
chore: clean up UI prerequisite commands to minimize naming explicit packages (on 1.9.x)

gashcrumb:
Since I see the same error is occuring for that branch of course.


zregvart:
Retest this please

PR #7851, opened 2020-02-07
TEIIDTOOLS-928: correcting the connection synchronization logic to sy…

rareddy:
…ndesis data sources, and removing unsed source schema files (#7804)


rareddy:
@shawkins this is the you approved already for master, trying to pull into 1.9.x

PR #7856, opened 2020-02-10
Update to productised Jaeger 1.13 - ENTESB-12909

claudio4j:
https://issues.redhat.com/browse/ENTESB-12909


heiko-braun:
@claudio4j I think we should keep the upstream 1.9.x clean from product references and instead bring this change to https://github.com/jboss-fuse/syndesis/tree/1.9.x 

@zregvart What do you think?

heiko-braun:
Sorry, I just realised that it's not using the prod images at all. I guess it's fine then.

But for https://github.com/jboss-fuse/syndesis/tree/1.9.x we should set it to https://access.redhat.com/containers/#/registry.access.redhat.com/distributed-tracing/jaeger-rhel7-operator

PR #7859, opened 2020-02-10
fix: Adds in GOPROXY environment variable for operator building with go

phantomjinx:


phantomjinx:
> With this change, it will always use a proxy right? (https://proxy.golang.org by default) But I don't want to use a proxy if I don't need to.

No it won't. If you leave the default it will but you can specify --go-proxy direct which bypasses the use of a proxy (see https://arslan.io/2019/08/02/why-you-should-use-a-go-module-proxy)

PR #7860, opened 2020-02-10
fix: Removes deficient backup test

phantomjinx:
* Test is too fragile depending on the platform it is executed on.


phantomjinx:
> If the test makes sense, why not adding
> 
> ```
> 	if testing.Short() {
> 		t.Skip("skipping app install tests in short mode")
> 	}
> ```
In hindsight, the test makes little sense and trying to develop a situation where it does is so contorted and error-prone, it makes the test not worth it. It ends up being a test for the sake of another test. More tests on the backup are needed but I really want to do them properly.

PR #7870, opened 2020-02-11
Fix syndesis minishift install and dev --cleanup

claudio4j:
Fix syndesis minishift install scripts to download syndesis-operator 2.x
when a tag is not specified
Enhance dev --cleanup removes dangling docker images
Enhance --full-reset also removes ~/.syndesis/bin/syndesis-operator

### DIFF
--- a/tools/bin/commands/util/operator_funcs
+++ b/tools/bin/commands/util/operator_funcs
@@ -88,16 +88,25 @@ download_operator_binary() {
     esac
 
     if [[ "${pattern}" != "unknown" ]]; then
-        release_uri="https://api.github.com/repos/syndesisio/syndesis/releases?per_page=1"
         release_tag="$(readopt --tag)"
+        url=""
         if [[ -n "$release_tag" ]]; then
             release_uri="https://api.github.com/repos/syndesisio/syndesis/releases/tags/${release_tag}"
+            url=$(curl -s $release_uri \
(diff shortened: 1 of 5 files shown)


tools/bin/commands/util/operator_funcs
```diff
@@ -88,17 +88,26 @@ download_operator_binary() {
...
         if [[ -n "$release_tag" ]]; then
             release_uri="https://api.github.com/repos/syndesisio/syndesis/releases/tags/${release_tag}"
+            url=$(curl -s $release_uri \
+                | grep browser_download_url \
+                | grep ${pattern} \
+                | cut -d '"' -f 4)
+        else
+            # retrieve the last 4 releases and get the latest 2.0 release
+            url=$(curl -s "https://api.github.com/repos/syndesisio/syndesis/releases?per_page=4" \
+                | grep "download/2" \
```
zregvart: This will blow up on us when we switch to 3.x.y, not sure if anyone will remember this change at that point...

claudio4j: When it downloads syndesis-operator it prints the exact version, so we are going to see it, at least. Currently it downloads the last tag, which was 1.9 yesterday, today is 2.0, unpredictable. Later we may add an option to specify the major version which we want to download.

PR #7873, opened 2020-02-12
[operator] Fix for install option --dev

lgarciaaco:
When installing the syndesis with `--dev` the syndesis-operator DC should point to the syndesis-operator IS with latest tag.


install/operator/pkg/generator/assets/install/operator.yml.tmpl
```diff
@@ -99,6 +99,10 @@
         - syndesis-operator
         from:
           kind: ImageStreamTag
-          name: syndesis-operator:{{.Tag}}
+          {{- if .DevSupport}}
+          name: syndesis-operator:latest
+          {{else}}
+          syndesis-operator:{{.Tag}}
```
phantomjinx: Shouldn't this line begin wih name too?

PR #7890, opened 2020-02-14
fix(ENTESB-12960): Missing version placeholder.

Delawen:
When building the image for an integration with Mongo, missing versino made it fail.


Delawen:
Trying again... UI is what is failing, which shouldn't be related to this.

PR #7900, opened 2020-02-17
[ENTESB-12846] Fix permissions for jobs

avano:
https://issues.redhat.com/browse/ENTESB-12846 master


avano:
the fail in the test is unrelated to this one (failed in ui module)

PR #7904, opened 2020-02-17
ENTESB-12645: Fixes issues with scheduled backups of external database

phantomjinx:
* object.go
 * Ensure all properties of syndesis are copied, ie. the ExternalDB
   attribute.

* backup.go
 * ExternalDatabase function must be called from configuration in order to
   update the ...Database.URL attribute
 * Fix incorrect constants used in backup that should be dependent on
   target host
 * Perform checks on backup file dump to see if it is empty

### DIFF
--- a/install/operator/pkg/syndesis/backup/backup.go
+++ b/install/operator/pkg/syndesis/backup/backup.go
@@ -542,6 +542,9 @@ func (b *Backup) backupDatabase() error {
 	if err != nil {
 		return err
 	}
+	if err = sc.ExternalDatabase(b.context, b.client, b.syndesis); err != nil {
+		return err
+	}
 
 	dbURL, err := url.Parse(sc.Syndesis.Components.Database.URL)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


phantomjinx:
@heiko-braun @zregvart 
With @lgarciaaco unavailable, would you be available to at least give this a look over, pls?

install/operator/pkg/syndesis/backup/backup.go
```diff
@@ -542,6 +542,9 @@ func (b *Backup) backupDatabase() error {
 	if err != nil {
 		return err
 	}
+	if err = sc.ExternalDatabase(b.context, b.client, b.syndesis); err != nil {
```
heiko-braun: @phantomjinx  Did it fail before because the errors havn't been caught?

phantomjinx: It failed for a couple of reasons.
The ExternalDatabase was not being set in the job pod environment so it was trying to backup syndesis-db (default), which was of course not there. Secondly, the pod was being 'successful' as I had not added checks on the backup dump file being empty, ie. erroneous success status. Both addressed by this and fixed.

PR #7905, opened 2020-02-17
ENTESB-12645: Fixes issues with scheduled backups of external database

phantomjinx:
* object.go
 * Ensure all properties of syndesis are copied, ie. the ExternalDB
   attribute.

* backup.go
 * ExternalDatabase function must be called from configuration in order to
   update the ...Database.URL attribute
 * Fix incorrect constants used in backup that should be dependent on
   target host
 * Perform checks on backup file dump to see if it is empty


phantomjinx:
@heiko-braun @zregvart 
With @lgarciaaco unavailable, would you be available to at least give this a look over, pls?

PR #7913, opened 2020-02-18
chore: bump yarn and add terser workaround

gashcrumb:
Bumps yarn to the latest 1.x (1.22.0)

Also adds a workaround for https://github.com/facebook/create-react-app/issues/8320 to hopefully sort the terser error on CI.

Also skips the ui-doc job for now as it performs a second UI build but we don't currently do anything with the output.


gashcrumb:
yay, 2 successful builds :smile:

PR #7923, opened 2020-02-19
chore: remove javax/org.jboss.spec dependency declarations

zregvart:
This removes the javax/org.jboss.spec dependencies replaced with Jakarta
ones.


gashcrumb:
rebase should fix that UI build issue

PR #7952, opened 2020-02-21
Fix for ENTESB 12972 - Todo pod does not deploy on OCP 4.4

KurtStam:
Fix for https://issues.redhat.com/browse/ENTESB-12972: Todo pod does not deploy on OCP 4.4

The reason of it not building is due to the version of PHP being too old for the OCP 4.4 php imagestream in openshift namespace. With this fix we reference the docker image it was referencing directly.


KurtStam:
@gashcrumb any ideas why the UI module is failing?

heiko-braun:
Thanks @KurtStam. Can you backport this to 1.9.x?

install/operator/pkg/generator/assets/addons/todo/04-todo-example.yml.tmpl
```diff
@@ -75,9 +75,8 @@
     strategy:
       sourceStrategy:
         from:
-          kind: ImageStreamTag
-          name: 'php:7.0'
-          namespace: openshift
+        kind: DockerImage
+        name: 'docker.io/centos/php-71-centos7'
       type: Source
```
lgarciaaco: isn't indentation wrong?

install/operator/pkg/generator/assets/addons/todo/04-todo-example.yml.tmpl
```diff
@@ -75,9 +75,8 @@
     strategy:
       sourceStrategy:
         from:
-          kind: ImageStreamTag
-          name: 'php:7.0'
-          namespace: openshift
+        kind: DockerImage
+        name: 'docker.io/centos/php-71-centos7'
```
asmigala: For product, this should probably be a different image (probably `rhscl/php-71-rhel7` though not 100 % sure). For images that need to be overwritten, we generally put those into `config.yaml`, I think that's what we need to do here as well, so that productization can override it. 

cc @orpiske

heiko-braun:
@lgarciaaco Can the todo PHP image be overriden using `config.yaml` [1]. or does have to be done in the `04-todo-example.yml.tmpl`?

[1]  https://github.com/jboss-fuse/fuse-online-operator/blob/private-devel-fuse-7.6-openshift-rhel-7/build/conf/config.yaml

lgarciaaco:
> @lgarciaaco Can the todo PHP image be overriden using `config.yaml` [1]. or does have to be done in the `04-todo-example.yml.tmpl`?
> 
> [1] https://github.com/jboss-fuse/fuse-online-operator/blob/private-devel-fuse-7.6-openshift-rhel-7/build/conf/config.yaml

@heiko-braun curently it is not possible. The image is not part of any configuration.

heiko-braun:
@lgarciaaco ok, no problem. So if we need to change it for product, should modify this file [1] ?

[1] https://github.com/jboss-fuse/syndesis/blob/master/install/operator/pkg/generator/assets/addons/todo/04-todo-example.yml.tmpl

lgarciaaco:
> @lgarciaaco ok, no problem. So if we need to change it for product, should modify this file [1] ?
> 
> [1] https://github.com/jboss-fuse/syndesis/blob/master/install/operator/pkg/generator/assets/addons/todo/04-todo-example.yml.tmpl

That is right, that file needs to be modified before building the operator

heiko-braun:
See also https://github.com/jboss-fuse/syndesis/pull/18/files

install/operator/pkg/generator/assets/addons/todo/04-todo-example.yml.tmpl
KurtStam: Thanks & fixed

> isn't indentation wrong?

Thanks & fixed

KurtStam:
I think the indentation was the only thing that needed fixing in the end, so this is taken care of now, so it should be good to go.

heiko-braun:
Thanks @KurtStam !

PR #7970, opened 2020-02-26
feat(operator): determine PostgreSQL from image

zregvart:
This adds logic to determine the version of the PostgreSQL from Docker
image metadata. Usually the metadata contains environment variable
`POSTGRESQL_VERSION` (e.g. `POSTGRESQL_VERSION=9.5`); I've also noticed
that there could be a `version` label also present, so that can be used
as a fallback if `POSTGRESQL_VERSION` environment variable is not
present. Caveat with using the `version` label is that some images have
the value set to `1`, go figure.

The logic performs as follows:
 1. image name (pull spec) is taken from the `DeploymentConfig`'s
    container template
 2. image name is parsed to extract the ID (sha hash)
 3. `Image` is fetched by the extracted ID
 4. from the `Image` data Docker image metadata is parsed
 5. `POSTGRESQL_VERSION` environment variable is looked up in the Docker
    image metadata, failing that
 6. `version` label is looked up in the Docker image metadata, failing
     that

The logic to determine if the database upgrade is needed now considers
this the above determined version and the constant
`wantedPostgreSQLVersion` corresponding to the version that the operator
would install.

Determining the version from just the image name, lacking the complete
pull spec (e.g. no ID) would mean contacting the image registry to
resolve the full pull spec and fetch the ID needed to fetch the `Image`.
This was discarded as an option for it would introduce additional
complexity and in it's place simple constant was added.

Consequence of that is that the `wantedPostgreSQLVersion` needs to be
updated with each upgrade of the PostgreSQL database.

Ref. https://issues.redhat.com/browse/ENTESB-13073


install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -291,7 +298,11 @@ func GetProperties(file string, ctx context.Context, client client.Client, synde
 		if err := client.Get(ctx, types.NamespacedName{Namespace: syndesis.Namespace, Name: "syndesis-db"}, databaseDeployment); err == nil {
 			for _, c := range databaseDeployment.Spec.Template.Spec.Containers {
 				if c.Name == "postgresql" {
-					configuration.DatabaseNeedsUpgrade = c.Image != configuration.Syndesis.Components.Database.Image
+					currentPostgreSQLVersion, err := postgresqlVersionFromImage(client, c.Image)
+					if err != nil {
+						return nil, err
```
phantomjinx: As a general point, we return and log errors like this, & for the most part these sort of errors never get thrown. However, on the rare instance it does, what do we expect to happen? Operators are supposed to be self-healing etc... so how much mitigation should we put in place over time to try and isolate and kill off any possible corner cases that may arise?

@heiko-braun @lgarciaaco

heiko-braun: Does an operator have a somehting like a preflight phase? I hitnk checks like these should be performed, before any modifications are made.

zregvart:
I have another approach for this, so I'll close this for now.

PR #7986, opened 2020-02-27
chore(operator): determine version of postgresql

zregvart:
In upgrades we need to know if we should set the `POSTGRESQL_UPGRADE`
environment variable based on the currently running (source) and the
to-be (target) versions of PostgreSQL.

This determines the version of the source PostgreSQL database by
connecting to the database and executing `SHOW server_version`. Since
we're interested only in major and minor version (e.g. 9.6, not 9.6.12)
parses the version and reduces it to a `float64` containing the version.

To determine the target database version the `syndesis-operator` pod as
defined using the OpenShift `DeploymentConfig` now contains an init
container running from the same image configured for the `syndesis-db`
database. This init container runs `postgres -V` which outputs to the
log the version. The operator parses the log output of the init
container to determine the version, also reduced to `float64` containing
only major and minor version.

Ref. https://issues.redhat.com/browse/ENTESB-13073

There is an additional test failure that I'm too lazy to track down, I don't think the failure is related to the changes I made. These are non-short tests that we forget to run. Like in 946c681

I'm fairly certain that I made mistakes in 9e9a2ea, the use case and the code is very much unknown to me.

As usual try to ignore the vendored dependencies in d0b03b5.

This is how I tested:
 1. started fresh minishift
 2. run this with fuse-online-install from 1.8.x branch
```shell
oc login -u admin -p admin
oc delete project syndesis
oc new-project syndesis
./install_ocp.sh --setup --grant developer
oc adm policy add-cluster-role-to-user admin -z syndesis-operator # I forgot the issue why this was needed
oc login -u developer -p x
./install_ocp.sh
```
 3. built the operator from this branch with `./build.sh`
 4. patched the operator deployment config to include the changes and point to the image version built by s2i:
```shell
oc patch dc syndesis-operator -p '{"spec": {"template": {"spec": {"initContainers": [{"name": "postgres-version", "image": "centos/postgresql-96-centos7", "command": ["sh", "-c", "postgres -V"]}]}}, "triggers": [{"type": "ConfigChange"}, {"type": "ImageChange", "imageChangeParams": {"automatic": true, "containerNames": ["syndesis-operator"], "from": {"kind": "ImageStreamTag", "name": "syndesis-operator:latest", "namespace": "syndesis"}}}]}}'
```

Ideally the operator logs should show a line like:
```json
{"level":"info","ts":1582821799.265115,"logger":"configuration","msg":"PostgreSQL upgrade summary","source_postgres_version":"9.50","target_postgres_version":"9.60","perform_upgrade":"true"}
```

I'm not sure about other places where the operator deployment config needs to be modified.


install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -281,26 +292,74 @@ func GetProperties(file string, ctx context.Context, client client.Client, synde
...
+	binaryLog, err := ioutil.ReadAll(stream)
+	if err != nil {
+		return 0, err
+	}
+
+	versionString := string(binaryLog)
+	extracted := postgresVersionRegex.FindStringSubmatch(versionString)
+	if len(extracted) < 2 {
+		return 0, fmt.Errorf("Unable to parse PostgreSQL version from version string: `%s`", versionString)
+	}
```
phantomjinx: Seems more elegant that including Docker metadata objects

avano:
@zregvart looks ok when upgrading from pg 9.5 (db migration happens) and also 9.6 (no db migration) :+1:

zregvart:
Okay, this has baked enough, to have it backported in 1.9.x I need to merge it now.

PR #7990, opened 2020-02-28
chore(operator): determine version of postgresql

zregvart:
This the backport of #7986 for 1.9.x

PR #7991, opened 2020-02-28
[ENTESB-13103] Add patch permission to syndesis resource

avano:
Standard user can create and delete syndesis, but he can't edit it due to missing `patch` permission


zregvart:
We have to be careful with the generated assets_vfsdata.go we don’t overwrite changes from #7990 or vice versa

avano:
@zregvart We both changed a different file so that should be ok

PR #8013, opened 2020-03-03
fix(operator): provide default database image

### LINKED ISSUE #8009: Bake in PostgreSQL image at build time

zregvart:
When the operator is run as CLI there is no `/conf/config.yml` to be
read, as it is only put in the container image. This leads to:

```
Error: open /conf/config.yaml: no such file or directory
```

This reads the `install/operator/build/conf/config.yaml` fetches the
database image and creates a `install_defaults.go` file at build time
containing the `defaultDatabaseImage` constant with the value of the
read database image.

At runtime, if a config file was specified via `--config` parameter the
database image is read from that file and the default is replaced.

1974610  contains only vendoring of go-yaml, a8a0eca contains the changes described above.

Fixes #8009


zregvart:
@lgarciaaco @phantomjinx can you have another look, would a766f9b07c606795fe450f3ac3581540bd81658c fix this?

install/operator/pkg/cmd/internal/install/install.go
```diff
@@ -121,10 +120,12 @@ func New(parent *internal.Options) *cobra.Command {
 	cmd.PersistentFlags().StringVarP(&o.tag, "tag", "", pkg.DefaultOperatorTag, "sets operator tag that gets installed")
 	cmd.PersistentFlags().BoolVarP(&o.wait, "wait", "w", false, "waits for the application to be running")
 	cmd.PersistentFlags().BoolVarP(&o.devSupport, "dev", "", false, "enable development mode by loading images from image stream tags.")
+	cmd.PersistentFlags().StringVarP(&configuration.TemplateConfig, "	", "", "/conf/config.yaml", "Path to the operator configuration file.")
```
lgarciaaco:
```suggestion
	cmd.PersistentFlags().StringVarP(&configuration.TemplateConfig, "operator-config", "", "/conf/config.yaml", "Path to the operator configuration file.")
```

zregvart: Nice catch! I typoed over that by accident.

lgarciaaco:
Just to clarify, now we have to always have a config file at hand if we want to use the cli. I am not too clear on how this will look like for `syndesis minishift --install` / `syndesis crc --install` and for [fuse-online-install](https://github.com/syndesisio/fuse-online-install)

zregvart:
> Just to clarify, now we have to always have a config file at hand if we want to use the cli. I am not too clear on how this will look like for `syndesis minishift --install` / `syndesis crc --install` and for [fuse-online-install](https://github.com/syndesisio/fuse-online-install)

Currently yes, when this change is merged no. The only command that requires the config file should be `install forge`, the requirement to have the config file was introduced in #7986 for other `install` sub-commands and this should (hopefully) remove it.

heiko-braun:
For what it's worth, fuse-online-install use the forge command, and it includes a config.yaml:

https://github.com/syndesisio/fuse-online-install/blob/master/release.sh#L340

phantomjinx:
> > Just to clarify, now we have to always have a config file at hand if we want to use the cli. I am not too clear on how this will look like for `syndesis minishift --install` / `syndesis crc --install` and for [fuse-online-install](https://github.com/syndesisio/fuse-online-install)
> 
> Currently yes, when this change is merged no. The only command that requires the config file should be `install forge`, the requirement to have the config file was introduced in #7986 for other `install` sub-commands and this should (hopefully) remove it.

Backup and restore and both available from the operator CLI and they too require the config file.

install/operator/pkg/cmd/internal/install/install.go
```diff
@@ -121,10 +120,12 @@ func New(parent *internal.Options) *cobra.Command {
 	cmd.PersistentFlags().StringVarP(&o.tag, "tag", "", pkg.DefaultOperatorTag, "sets operator tag that gets installed")
 	cmd.PersistentFlags().BoolVarP(&o.wait, "wait", "w", false, "waits for the application to be running")
 	cmd.PersistentFlags().BoolVarP(&o.devSupport, "dev", "", false, "enable development mode by loading images from image stream tags.")
+	cmd.PersistentFlags().StringVarP(&configuration.TemplateConfig, "operator-config", "", "/conf/config.yaml", "Path to the operator configuration file.")
```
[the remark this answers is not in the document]
lgarciaaco: This makes it global for all install subcommands, we still need to keep it for other commands.

zregvart:
@phantomjinx I'd keep the `--operator-config` separate for `install`, `backup` and `restore`, at least for now. I don't think having it global to all commands is helpful.

heiko-braun:
@phantomjinx Does @zregvart last https://github.com/syndesisio/syndesis/pull/8013#issuecomment-594448096 resolve the need for changes you requested before?

PR #8022, opened 2020-03-04
feat(server): autodiscovery circuit breaker call

squakez:
* Added circuit breaker instead of plain REST call to meta connection properties
* Refactored common-model to add dynamic metadata modeling
* Refactored MetadataCommand to be used as abstract base for MetadataCommandAction and MetadataConnectionPropertiesCommand
* Refactored unit test to use DynamicConnectionPropertiesMetadata instead of raw Response
* Removed useless pom dependencies

Closes ENTESB-12498

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/connection/ConfigurationProperty.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/connection/ConfigurationProperty.java
@@ -21,7 +21,7 @@ import java.util.Optional;
 
 import io.syndesis.common.model.Ordered;
 import io.syndesis.common.model.WithTags;
-import io.syndesis.common.model.connection.DynamicActionMetadata.ActionPropertySuggestion;
+import io.syndesis.common.model.connection.WithDynamicProperties.ActionPropertySuggestion;
 
 import org.immutables.value.Value;
 
--- a/app/common/model/src/main/java/io/syndesis/common/model/connection/DynamicConnectionPropertiesMetadata.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/connection/DynamicConnectionPropertiesMetadata.java
@@ -19,23 +19,18 @@ import java.util.List;
 import java.util.Map;
 
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import io.syndesis.common.model.DataShape;
 import org.immutables.value.Value;
 
 @Value.Immutable
-@JsonDeserialize(builder = DynamicActionMetadata.Builder.class)
-public interface DynamicActionMetadata extends WithDynamicProperties {
+@JsonDeserialize(builder = DynamicConnectionPropertiesMetadata.Builder.class)
+public interface DynamicConnectionPropertiesMetadata extends WithDynamicProperties {
 
-    DynamicActionMetadata NOTHING = new DynamicActionMetadata.Builder().build();
+    DynamicConnectionPropertiesMetadata NOTHING = new DynamicConnectionPropertiesMetadata.Builder().build();
 
-    final class Builder extends ImmutableDynamicActionMetadata.Builder {
-        // make ImmutableDynamicActionMetadata.Builder accessible
+    final class Builder extends ImmutableDynamicConnectionPropertiesMetadata.Builder {
+        // make ImmutableDynamicConnectionPropertiesMetadata.Builder accessible
     }
 
     @Override
     Map<String, List<ActionPropertySuggestion>> properties();
-
-    DataShape inputShape();
-
-    DataShape outputShape();
 }
--- a/app/common/model/src/main/java/io/syndesis/common/model/connection/WithDynamicProperties.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/connection/WithDynamicProperties.java
@@ -19,23 +19,24 @@ import java.util.List;
 import java.util.Map;
 
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
-import io.syndesis.common.model.DataShape;
 import org.immutables.value.Value;
 
-@Value.Immutable
-@JsonDeserialize(builder = DynamicActionMetadata.Builder.class)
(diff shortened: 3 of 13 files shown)


app/common/model/src/main/java/io/syndesis/common/model/connection/ConfigurationProperty.java
```diff
@@ -37,15 +37,15 @@
      */
     String MULTIPLE_SEPARATOR = ",";
 
-    class Builder extends ImmutableConfigurationProperty.Builder {
+    class Builder extends io.syndesis.common.model.connection.ImmutableConfigurationProperty.Builder {
```
zregvart: nit

```suggestion
    class Builder extends ImmutableConfigurationProperty.Builder {
```

app/common/model/src/main/java/io/syndesis/common/model/connection/ConfigurationProperty.java
```diff
@@ -37,15 +37,15 @@
...
         // make ImmutableConfigurationProperty.Builder accessible
     }
 
     @Value.Immutable
     @JsonDeserialize(builder = PropertyValue.Builder.class)
     interface PropertyValue {
 
-        class Builder extends ImmutablePropertyValue.Builder {
+        class Builder extends io.syndesis.common.model.connection.ImmutablePropertyValue.Builder {
```
zregvart: nit

```suggestion
        class Builder extends ImmutablePropertyValue.Builder {
```

app/common/model/src/main/java/io/syndesis/common/model/connection/DynamicActionMetadata.java
```diff
@@ -24,29 +24,15 @@
...
-            }
-        }
-
-        String displayValue();
-
-        String value();
-    }
-
-    final class Builder extends ImmutableDynamicActionMetadata.Builder {
+    final class Builder extends io.syndesis.common.model.connection.ImmutableDynamicActionMetadata.Builder {
```
zregvart: nit

```suggestion
    final class Builder extends ImmutableDynamicActionMetadata.Builder {
```

app/common/model/src/main/java/io/syndesis/common/model/connection/DynamicConnectionPropertiesMetadata.java
```diff
@@ -0,0 +1,36 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonDeserialize(builder = DynamicConnectionPropertiesMetadata.Builder.class)
+public interface DynamicConnectionPropertiesMetadata extends WithDynamicProperties {
+
+    DynamicConnectionPropertiesMetadata NOTHING = new DynamicConnectionPropertiesMetadata.Builder().build();
+
+    final class Builder extends io.syndesis.common.model.connection.ImmutableDynamicConnectionPropertiesMetadata.Builder {
```
zregvart: nit

```suggestion
    final class Builder extends ImmutableDynamicConnectionPropertiesMetadata.Builder {
```

app/common/model/src/main/java/io/syndesis/common/model/connection/WithDynamicProperties.java
```diff
@@ -0,0 +1,42 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import org.immutables.value.Value;
+
+public interface WithDynamicProperties {
+
+    @Value.Immutable
+    @JsonDeserialize(builder = ActionPropertySuggestion.Builder.class)
+    interface ActionPropertySuggestion {
+
+        final class Builder extends io.syndesis.common.model.connection.ImmutableActionPropertySuggestion.Builder {
```
zregvart: nit
```suggestion
        final class Builder extends ImmutableActionPropertySuggestion.Builder {
```

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectorHandler.java
```diff
@@ -129,7 +127,7 @@ public Connector get(final String id) {
 
         // Retrieve dynamic properties, if connector is dynamic
         if (connector.getTags().contains("dynamic")) {
-            connector = enrichWithDynamicProperties(connector);
+            connector = enrichConnectorWithDynamicProperties(connector);
```
zregvart: I'd remove this from the `GET /connectors/{id}`. I'd like the retrieval of dynamic metadata, i.e. invoking of the syndesis-meta service to be tied to an explicit request from the UI for it, not in every case when the connector is fetched.

`GET` requests should be idempotent and should not have side effects, this will endanger that.

app/common/model/src/main/java/io/syndesis/common/model/connection/ConfigurationProperty.java
squakez: For some reason my IDE forces that repeatedly. I've tried to remove but with no luck. I'll have a further look to have it cleaner, I don't like it either.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectorHandler.java
squakez: I agree, though it would be out the scope of the request for this PR (circuit breaker). I have created  further issue to take care of this refactory #8023 separately. Are you okey to proceed like this?

Also, from performance perspective, right now the "enriching" call is conditioned to the presence of a tag that we are configuring on the connectors which use this feature.

app/common/model/src/main/java/io/syndesis/common/model/connection/ConfigurationProperty.java
squakez: Fixed

PR #8029, opened 2020-03-04
backport: autodiscovery circuit breaker call (#8022)

squakez:
Backport feature #8022


zregvart:
I don't think we should backport this. This hasn't been scheduled for 1.9.

heiko-braun:
@zregvart is right. At this point we only backport blocking issues.

PR #8055, opened 2020-03-09
Allowing for the absence of the pg_up metric

david-martin:
The FuseOnlineDatabaseInstanceDown alert doesn't trigger if syndesis-db is down.
However, it does look like the `pg_up{job="syndesis-db"}` metric is no longer being scraped.
The problem is how prometheus does the `sum`.
If it tries to sum an empty result (no data), it also returns an empty result.
Therefore, we need to add a conditional for the metric being absent and alert in that scenario.

Steps to reproduce from upstream (private) issue https://issues.redhat.com/browse/INTLY-5990

`oc project redhat-rhmi-fuse-operator`
scale down fuse operator - `while true; do echo "check" && if ! oc get deployments | grep 0/0 &>/dev/null; then echo "scaling fuse operator down" && oc scale --replicas 0 deployment syndesis-operator; fi; done`
scale down syndesis-db
check alerts


zregvart:
@david-martin you need to run the code generation to have the `pkg/generator/assets_vfsdata.go` updated, run `go generate ./pkg/...` and update the PR with the resulting `assets_vfsdata.go`.

Delawen:
This looks a bit stalled. @david-martin any news here?

david-martin:
Thanks @zregvart @Delawen ,
I've pushed the changes with the regenerated assets_vfsdata.go.

PR #8057, opened 2020-03-09
chore: upgrade pf for OAuth components

kahboom:
drops support for patternfly-react OAuth components, replace with @patternfly/react-core

related epic: [ENTESB-12896](https://issues.redhat.com/browse/ENTESB-12896)

Screenshot 2020-03-09 18 54 11

### DIFF
--- a/app/ui-react/packages/ui/src/Settings/OAuth/OAuthAppExpanderBody.tsx
+++ b/app/ui-react/packages/ui/src/Settings/OAuth/OAuthAppExpanderBody.tsx
@@ -1,4 +1,4 @@
-import { Alert, Button, Form, Grid, GridItem } from '@patternfly/react-core';
+import { Alert, Button, Form, Stack, StackItem } from '@patternfly/react-core';
 import * as React from 'react';
 
 export interface IOAuthAppExpanderBodyProps {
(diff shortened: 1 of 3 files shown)


app/ui-react/packages/ui/src/Settings/OAuth/OAuthAppExpanderBody.tsx
```diff
@@ -15,52 +14,59 @@ export interface IOAuthAppExpanderBodyProps {
...
+              {i18nAlertDetail}
+            </Alert>
+          </GridItem>
+        </Grid>
+      )}
+      <Grid sm={12} md={8}>
+        <GridItem>
+          <Form isHorizontal={true}>{children}</Form>
+        </GridItem>
+      </Grid>
```
gashcrumb: I think you could probably combine these two grids and set the colspan to 12 for these?  Either that or I wonder if just using a Stack would work just as well here.

app/ui-react/packages/ui/src/Settings/OAuth/OAuthAppExpanderBody.tsx
```diff
@@ -15,52 +14,59 @@ export interface IOAuthAppExpanderBodyProps {
...
+    i18nSaveButtonText,
+    onRemove,
+    onSave,
+    showSuccess
+  }) => {
+  return (
+    <>
+      {showSuccess && (
+        <Grid sm={11}>
+          <GridItem>
```
gashcrumb: This probably could also be stuck in the same grid as the other stuff, or maybe a Stack and then showSuccess could hide/show the StackItem that holds this alert?

kahboom:
@gashcrumb I like the concept of swapping `Grid` for `Stack` in `OAuthAppExpanderBody`. It makes it cleaner and actually looks better on mobile because of the gutter. See screenshots with changes:

Screenshot 2020-03-10 10 46 17

**With alert..**

Screenshot 2020-03-10 10 46 33

**On mobile..**

Screenshot 2020-03-10 10 48 34

PR #8061, opened 2020-03-10
Add install_defaults.go for postgresql default image - Backport 1.9.x

claudio4j:
The file `install/operator/pkg/cmd/internal/install/install_defaults.go` is generated when using the `go` build mode, but it doesn't generate when running in `docker` mode, which is the CI mode runs.


claudio4j:
Closing as adding the generated files is not the right fix.

PR #8062, opened 2020-03-10
Add install_defaults.go for postgresql default image

claudio4j:
The file `install/operator/pkg/cmd/internal/install/install_defaults.go` is generated when using the `go` build mode, but it doesn't generate when running in `docker` mode, which is the CI mode runs.

(cherry picked from commit 4a6f0e0c291d19520d891f2a3f14fe9e78a35c66)


lgarciaaco:
I wonder is https://github.com/syndesisio/syndesis/blob/master/install/operator/pkg/constants.go isnt a better place

claudio4j:
Closing as adding the generated files is not the right fix.

PR #8063, opened 2020-03-10
chore: migrate action select list in the editor

gashcrumb:
relates to [ENTESB-12896](https://issues.redhat.com/browse/ENTESB-12896)

image

I think we can save the text alignment for when we've removed patternfly-react and it's related stylesheet.

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/choice/SelectChoiceConfigModePage.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/choice/SelectChoiceConfigModePage.tsx
@@ -24,6 +24,7 @@ import {
 } from '@syndesis/ui';
 import { WithRouteData } from '@syndesis/utils';
 import * as React from 'react';
+import { useTranslation } from 'react-i18next';
 import { PageTitle } from '../../../../../shared';
 import { IEditorSidebarProps } from '../EditorSidebar';
 import {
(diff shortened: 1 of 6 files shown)


app/ui-react/syndesis/src/modules/integrations/components/editor/choice/SelectChoiceConfigModePage.tsx
```diff
@@ -58,23 +59,27 @@ export interface ISelectChoiceConfigModePageProps extends IPageWithEditorBreadcr
...
   public render() {
     return (
       <WithRouteData<ISelectConfigModeRouteParams, ISelectConfigModeRouteState>>
         {(params, state) => {
           const positionAsNumber = parseInt(params.position, 10);
           const options = [
             {
-              description: 'Provides preselected properties and condition operators for simple expressions.',
+              description:
+                'Provides preselected properties and condition operators for simple expressions.',
```
kahboom: Should we not be using i18n here? Or are we kind of putting that on the back burner for now?

gashcrumb: well sheesh I didn't even spot that, was just updating the usage of the component.

But now I can't unsee it, lemme tidy up what I'm doing and I'll add that change to this PR.

gashcrumb: K, updated that file in a separate commit.

PR #8066, opened 2020-03-10
Call Go generator when using docker to build operator - 1.9.x

claudio4j:
CI builds the operator with docker mode and the code generator `./pkg/cmd/internal/install/generator` should be called at this point to generate the `pkg/cmd/internal/install/install_defaults.go`

### DIFF
--- a/install/operator/.lib.sh
+++ b/install/operator/.lib.sh
@@ -149,6 +149,7 @@ FROM golang:1.13.7
 WORKDIR /go/src/${OPERATOR_GO_PACKAGE}
 ENV GO111MODULE=on
 COPY . .
+RUN go generate ./pkg/...
 RUN go test -test.short -mod=vendor ./cmd/... ./pkg/...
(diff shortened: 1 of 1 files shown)


install/operator/.lib.sh
```diff
@@ -149,6 +149,9 @@ FROM golang:1.13.7
 WORKDIR /go/src/${OPERATOR_GO_PACKAGE}
 ENV GO111MODULE=on
 COPY . .
+RUN go mod vendor
```
phantomjinx: This line is problematic & should not be necessary since the vendor directory is already resident as part of the src code directory as identified by WORKDIR (L149). It would also cause issues since it would try downloading stuff as not be able to necessarily have access to get it.

install/operator/.lib.sh
```diff
@@ -149,6 +149,9 @@ FROM golang:1.13.7
...
+RUN go generate ./pkg/...
+RUN go mod tidy
```
phantomjinx: Not clear that this should be here either if `go mod vendor` is removed.

install/operator/.lib.sh
claudio4j: done, thanks.

install/operator/.lib.sh
claudio4j: removed.

PR #8087, opened 2020-03-16
Add cleanup to delete old prereleases

lgarciaaco:
Cleanup prereleases older than 40 days, one page at a time.

### DIFF
--- a/tools/bin/commands/release
+++ b/tools/bin/commands/release
@@ -151,6 +151,9 @@ release::run() {
     # Release the binaries
     publish_artifacts "${topdir}" "$release_version" $prerelease
 
+    # Cleanup old prereleases
+    delete_prereleases_older_than 40
+
     # Create release description based on commit between releases
     # if check_for_command gren; then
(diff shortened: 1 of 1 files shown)


zregvart:
I had a different take on this, I was going to delete pre-releases but leave last 10. I had a change that I didn't create a pull request for. Seemed a bit easer as I started from git tags rather than from github releases (that need to be paged and all). Anyway good work :)

Since we're still running pre-release builds for 1.9.x do we need to backport this to 1.9.x branch?

lgarciaaco:
> I had a different take on this, I was going to delete pre-releases but leave last 10. I had a change that I didn't create a pull request for. Seemed a bit easer as I started from git tags rather than from github releases (that need to be paged and all). Anyway good work :)
> 
> Since we're still running pre-release builds for 1.9.x do we need to backport this to 1.9.x branch?

I guess we could, or we wait until we don't do any more releases and we run a one time cleanup. I am fine either way

Delawen:
Leaving some of them doesn't sound bad.

PR #8092, opened 2020-03-16
fix(meta): Prometheus resource name change

### LINKED ISSUE #8089: [Camel 3.x] Prometheus failure when calling meta REST API
## This is a...

<pre><code>
[ ] Feature request
[ ] Bug report  
</code></pre>

## Description

Probably introduced by the new `camel` 3.x libraries upgrade. When calling a rest API it returns:
```
2020-03-16 10:39:46.657 ERROR 1 --- [  XNIO-1 task-5] io.undertow.request: UT005023: Exception handling request to /api/v1/verifier/kafka
...
2020-03-16 10:33:23.560 ERROR 1 --- [  XNIO-1 task-4] io.undertow.request: UT005023: Exception handling request to /api/v1/connectors/kafka/properties/meta
java.lang.IllegalArgumentException: Prometheus requires that all meters with the same name have the same set of tag keys. There is already an existing meter named 'http_server_requests_seconds' containing tag keys [method, status, uri]. The meter you are attempting to register has keys [exception, method, outcome, status, uri].
--
```
See full [trace.log](https://github.com/syndesisio/syndesis/files/4337599/trace.log)

squakez:
Using a different name in order not to collide with the ones used by `server` pod.

Closes #8089

With the latest upgrade to Camel 3.1, probably because a new library update, it's not possible to register a meter with the [same name and different tags](https://github.com/syndesisio/syndesis/blob/master/app/server/runtime/src/main/java/io/syndesis/server/runtime/EndpointMetricsFilter.java#L66-L72). Actually `server` pod is already registering a meter, but with a different set of tags. Changing the name solves the problem, **but** I don't have visibility on the impact of such change. I don't know if those metrics are collected and if the name change is going to produce any disruptive change in any consumer.

### DIFF
--- a/app/meta/src/main/java/io/syndesis/connector/meta/EndpointMetricsFilter.java
+++ b/app/meta/src/main/java/io/syndesis/connector/meta/EndpointMetricsFilter.java
@@ -67,7 +67,7 @@ public class EndpointMetricsFilter implements ContainerRequestFilter, ContainerR
             tags.add(Tag.of("method", getMethod(requestContext)));
             tags.add(Tag.of("status", getStatus(responseContext)));
             tags.add(Tag.of("uri", getUri()));
-            sample.stop(registry.timer("http.server.requests", tags));
+            sample.stop(registry.timer("http.meta.requests", tags));
         }
     }
 


squakez:
Thanks @oscerd! Can you confirm that changing that name should not have any disruptive impact? I miss the whole picture on who is consuming those meters.

oscerd:
It should be possibile to change the name through the following sb property I believe
So it should be safe to do this. +1 for me.

PR #8093, opened 2020-03-16
Add ability to manually refresh DV source schema

mdrillin:
This adds ability in DV for the user to initiate a source schema refresh, rather than wait for the backend (which does an automatic refresh at pre-defined intervals).  In the import views and create view wizards, refresh buttons were added which kick off the refresh.  Once the refresh is complete, the 'last updated' time will change.
- adds 'refreshConnectionSchema' function in useVirtualizationHelpers to initiate the backend refresh.
- 'lastLoad' attribute added to the status model
- modified components to add the refresh button, status and last updated display
- schema refresh is triggered in the component based on the lastLoad times.
- Updated and resolved issues with stories


app/ui-react/packages/ui/src/Data/Virtualizations/Views/ViewInfoList.tsx
```diff
@@ -1,23 +1,72 @@
...
+                  onClick={handleRefreshClick}
+                  isDisabled={props.connectionLoading}
+                >
+                  {<SyncIcon />}&nbsp;
+                  {props.i18nRefresh}
+                </Button>
+               </FlexItem>
+            </Flex>
+          </Flex>
+        </div>{' '}
```
indraraj: can remove the embedded expression with the empty string.

indraraj:
Hey Mark,

I have tested locally, looks good to me. Just added one comment on Jira for "LastLoad" alignment.

app/ui-react/packages/ui/src/Data/Virtualizations/Views/ViewInfoList.tsx
mdrillin: removed the expression

mdrillin:
Changed layout in ConnectionSchemaListItem for better alignment of the 'Last updated' text (added comment to JIRA), and removed the extra expression.

PR #8094, opened 2020-03-17
[WIP

Delawen:
Fixing versions of camel resources. Try to get rid of static references so next camel version change will be automatic.


Delawen:
Ah damn, this was a work in progress.

PR #8096, opened 2020-03-17
Fixing some integration tests

Delawen:
Fixing versions of camel resources. Try to get rid of static references so next camel version change will be automatic.


Delawen:
Still WIP, can't properly test the changes you approved if I don't fix other things before.

Delawen:
#8174 merged so now we can proceed :)

PR #8121, opened 2020-03-23
ENTESB-12649 disabling db healthcheck to prevent session/query stats

shawkins:


app/dv/src/main/resources/s2i/application.properties
```diff
@@ -26,4 +26,5 @@ spring.main.allow-bean-definition-overriding=true
 teiid.jdbc-enable=true
 teiid.pg-enable=true
 
-management.health.mongo.enabled=false
\ No newline at end of file
+management.health.mongo.enabled=false
+logging.level.org.teiid=DEBUG
```
rareddy: Why logging in debug for needed for this? Is an oversite?

shawkins:
Sorry that came from the wrong branch - that was for debugging the web socket.  Here's the correct one line change.

Issue #8034, opened 2020-03-05
Integration fails to start: prometheus config cannot be found


KurtStam:
```
Starting the Java application using /opt/run-java/run-java.sh ...
exec java -javaagent:/opt/jolokia/jolokia.jar=config=/opt/jolokia/etc/jolokia.properties -javaagent:/opt/prometheus/jmx_prometheus_javaagent.jar=9779:/tmp/src/prometheus-config.yml -XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:+ExitOnOutOfMemoryError -cp . -jar /deployments/project-0.1-SNAPSHOT.jar
OpenJDK 64-Bit Server VM warning: If the number of processors is expected to increase from one, then you should configure the number of parallel GC threads appropriately using -XX:ParallelGCThreads=N
Exception in thread "main" java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at sun.instrument.InstrumentationImpl.loadClassAndStartAgent(InstrumentationImpl.java:386)
	at sun.instrument.InstrumentationImpl.loadClassAndCallPremain(InstrumentationImpl.java:401)
Caused by: java.io.FileNotFoundException: /tmp/src/prometheus-config.yml (No such file or directory)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at java.io.FileReader.<init>(FileReader.java:72)
	at io.prometheus.jmx.shaded.io.prometheus.jmx.JmxCollector.<init>(JmxCollector.java:74)
	at io.prometheus.jmx.shaded.io.prometheus.jmx.JavaAgent.premain(JavaAgent.java:47)
	... 6 more
FATAL ERROR in native method: processing of -javaagent failed
#
# A fatal error has been detected by the Java Runtime Environment:
#
[...]

PR #8127, opened 2020-03-25
fix 8034 Integration fails to start: prometheus config cannot be found

KurtStam:
#8034
https://issues.redhat.com/browse/ENTESB-13085

### DIFF
--- a/app/server/runtime/src/test/java/io/syndesis/server/runtime/integration/ProjectDSLGeneratorTest.java
+++ b/app/server/runtime/src/test/java/io/syndesis/server/runtime/integration/ProjectDSLGeneratorTest.java
@@ -1,0 +1,53 @@
+package io.syndesis.server.runtime.integration;
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


KurtStam:
Yes I totally agree and I opened a new issue: #8128

zregvart:
Aren't we creating a test that should be (and probably is) in Camel already?

KurtStam:
@zregvart the issue we that we excluded all dependencies. I'm not sure how a test in Camel is going to help. Or maybe you mean something else?

app/server/runtime/src/test/java/io/syndesis/server/runtime/integration/ProjectDSLGeneratorTest.java
```diff
@@ -0,0 +1,53 @@
...
+        final Document openApiDoc = Library.readDocumentFromJSONString(new String(openApiBytes, StandardCharsets.UTF_8));
+
+        final StringBuilder code = new StringBuilder();
+        RestDslGenerator.toAppendable((OasDocument) openApiDoc)
+            .withClassName("RestRoute")
+            .withPackageName("io.syndesis.example")
+            .withoutSourceCodeTimestamps()
+            .generate(code);
+
+        Assert.assertTrue(code.length() > 0);
```
zregvart: This test here, doesn't test much about dependencies or anything within Syndesis, right?

That kind of testing is already done in [Camel](https://github.com/apache/camel/tree/master/tooling/openapi-rest-dsl-generator/src/test/java/org/apache/camel/generator/openapi).

PR #8129, opened 2020-03-25
Install forge updates

phantomjinx:


phantomjinx:
Goes in hand with https://github.com/syndesisio/fuse-online-install/pull/235

phantomjinx:
Generated [template](https://gist.github.com/phantomjinx/797de72e6f4386809ae1349e85c19c95)

heiko-braun:
@phantomjinx Cool, thanks. I will take a look tomorrow.

heiko-braun:
@phantomjinx Two comments: 

Knowing the registry ip (or resolving it from the outside of a cluster) might not be possible, but I think `docker-registry.default.svc.cluster.local:5000` could work. Can we use this as a default value for the parameter?

The other thing aI am not sure about is the need to specify the image names/version in paramters. But it looks as if these are optional? Does the templest generator fetch it from `config.yaml`?

heiko-braun:
@phantomjinx Hang on, looks my step are not working as expected

heiko-braun:
Here's an alternative:

```
    triggers:
    - imageChangeParams:
        automatic: true
        containerNames:
        - syndesis-ui
        from:
          kind: ImageStreamTag
          name: <IMAGE>:<VERSION>
          namespace: ${IMAGE_STREAM_NAMESPACE}
```

phantomjinx:
As discussed,
- docker-registry.default.svc:5000 is the internal registry used by openshift and is specified as a service. When the template is applied this registry will be visible to the template so images in that registry can be found. However, only if this registry is exposed (openshift route) would this registry be available to the likes of the import script;
- Looking into the SYNDESIS_REGISTRY default value, ie. "docker.io". This is a hardcoded default added [here](https://github.com/syndesisio/syndesis/pull/8129/files#diff-26292d6d1943cf9bef44dd64f6e128a1L132) (sorry! thought it came from config.yml). Since this is a default installed openshift service installed, actually see no reason in making this the default value.
- Image credentials are directly fetched from the config.yml specified on the cli when running `<operator> install forge --operator-config <this file here> ....`. So the current generated template uses the product config.yml & therefore injects into the template the product image credentials. By making these parameters it ensures a degree of flexibility should alternative images needed to be inputted using these templates.

phantomjinx:
Updated SYNDESIS_REGISTRY default value. See [also](https://github.com/syndesisio/syndesis/pull/8129/files#diff-26292d6d1943cf9bef44dd64f6e128a1R141)

Issue #8125, opened 2020-03-25
Google Sheet skips empty values


SvenC56:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
Hey,  
I'm not 100% sure if it's a bug or a documentation issue. I'm exporting a table from Oracle (JDBC) to Google Sheets. The base table has some `null` values which get skipped by Google Sheets if I run the integration. Hence the next value gets inserted in this slot instead of it's original position. In between I run a Data Mapping step to connect the db columns with the spreadsheet ones.
Because of this the data is not usefull for analysis anymore.

Is there any additional step to be done or is this a bug?

zregvart:
Looks like a bug to me not sure if it's in Syndesis or in Atlasmap. Could you post an export of an integration (if you can) as an example just to check if this is an issue with the Google Sheets connector or the way mapping is used.

Export is a ZIP file that contains JSON dump of the Integration model, the secrets in there are encrypted with per instance unique encryption key, you might still wish to redact any of those or remove any other identifying data (IPs and such).

SvenC56:
Here you find the Export:

https://svencieslok-my.sharepoint.com/:u:/g/personal/mail_svencieslok_de/EUjt4gECOs1DhzEk6J_RohcB18YQFdL5tlLFZ-Cq_e8d8Q?e=djxTEm

I added also some pictures as example:

![alt text](https://i.imgur.com/4GqC4G2.png)

![alt text](https://i.imgur.com/CJOISwN.png)

![alt text](https://i.imgur.com/mlgB1Tw.png)

zregvart:
Fixed via #8130

SvenC56:
@christophd hey I created a fresh installation with `1.9.0-20200402` but the error seems to persist for me. 😕

zregvart:
I'll reopen this, let's see if @christophd can have another look.

christophd:
I had a 2nd look and can confirm that the issue has not been resolved yet. Classic situation where unit tests where green but integration tests would have seen the error is still present.

Now provided a fix and tested with a running integration using Google sheets.

SvenC56:
With the latest release `1.9.0-20200403` it is working for me. The issue can be closed from my side. 🙂

christophd:
@SvenC56 many thanks for verifying!

PR #8130, opened 2020-03-25
fix(#8125): Fix empty values handling in Google sheets

christophd:
Null values are skipped according to the Google docs. This leads to invalid data in value ranges for update and append operation. Using empty String values to fill in gaps that were not mapped in data mapper or mapped as Null value.


zregvart:
@christophd what do you think about backporting this to `1.9.x`?

christophd:
sure I can backport. is backporting to 1.9.x still an option @heiko-braun

heiko-braun:
Sure, why not. Especially if want to cut a Community release from that branch

PR #8131, opened 2020-03-27
fix download more releases to catch a 2.x release

KurtStam:


zregvart:
With #8124 we would not need this as at most 10 pre-/snapshot releases of 1.9.x would remain. Merged it now regardless of the QE tests failure.

PR #8132, opened 2020-03-27
[Backport-1.9.x]: Fix empty values handling in Google sheets

christophd:
Manual backport of #8130 

Null values are skipped according to the Google docs. This leads to invalid data in value ranges for update and append operation. Using empty String values to fill in gaps that were not mapped in data mapper or mapped as Null value.


zregvart:
retest this please

PR #8147, opened 2020-03-30
[operator] generate olm bundle

lgarciaaco:
This PR introduces a new command for syndesis-operator cli, which generates all bundle assets needed for OLM installation in Openshift 4.5. Previously this was done manually. As the metadata to create the bundle gets  more complicated, it makes sense that the operator generates it since the operator binary knows about this metadata (role, role bindings, images, deployment, relatedImages, version, whether productized or not). As result, we might have to change our build pipeline for downstream adding steps to generate the bundle files and update the product version.
- https://github.com/syndesisio/syndesis/commit/9381060e9e5a498c7cd909ac0b93b5288f940317 introduces a new field *Version* to the operator config file
- https://github.com/syndesisio/syndesis/commit/1b47488cffd9b3ef383903e15694362ce528cf25 introduces the olm package in charge of generating the bundle
- https://github.com/syndesisio/syndesis/commit/2d06cd1151c3be3b5f9f3a6256a435f511c8f41e introduces the olm command

The operator binary includes `olm` command, which can be used as following: `./syndesis-operator olm --operator-config ./conf/config.yaml --operator-image syndesis/syndesis-operator:1.9 --path ./syndesis-bundle`, where:

```
--operator-config string       Path to the operator configuration file. (default "/conf/config.yaml")
-o, --operator-image string    Syndesis operator docker image (default "syndesis/syndesis-operator:latest")
-p, --path string              Path where bundle files should be saved (default ".")
```
This will then generate the bundle files that can be used for OLM installation in Openshift 4.5+

Fixes https://issues.redhat.com/browse/ENTESB-12918
Fixes https://issues.redhat.com/browse/ENTESB-13207


lgarciaaco:
@claudio4j https://github.com/syndesisio/syndesis/commit/9381060e9e5a498c7cd909ac0b93b5288f940317 introduces the changed we spoke about and that you also need

PR #8152, opened 2020-03-30
feat(integration): Enable Flows dependencies

squakez:
* Providing the Flow the possibility to include a list of Dependencies
* Include the list of Flows dependencies into the Project generated from an Integration

Closes ENTESB-13246

### DIFF
--- a/app/integration/api/src/main/java/io/syndesis/integration/api/IntegrationResourceManager.java
+++ b/app/integration/api/src/main/java/io/syndesis/integration/api/IntegrationResourceManager.java
@@ -216,7 +216,20 @@ public interface IntegrationResourceManager {
     }
 
     default Collection<Dependency> collectDependencies(Integration integration) {
-        return collectDependencies(integration.getFlows().stream().flatMap(flow -> flow.getSteps().stream()).collect(Collectors.toList()), true);
+        final List<Dependency> dependencies = new ArrayList<>();
+        dependencies.addAll(integration.getDependencies());
+        dependencies.addAll(collectDependencies(integration.getFlows()));
+        return dependencies;
+    }
+
+    default Collection<Dependency> collectDependencies(List<Flow> flows){
+        final List<Dependency> dependencies = new ArrayList<>();
+        for(Flow flow : flows){
+            dependencies.addAll(flow.getDependencies());
+            Collection<Dependency> stepsDependencies = collectDependencies(flow.getSteps(), true);
+            dependencies.addAll(stepsDependencies);
+        }
+        return dependencies;
     }
 
     default Collection<Dependency> collectDependencies(Collection<? extends Step> steps, boolean resolveExtensionTags) {
--- a/app/integration/project-generator/src/test/java/io/syndesis/integration/project/generator/ProjectGeneratorTest.java
+++ b/app/integration/project-generator/src/test/java/io/syndesis/integration/project/generator/ProjectGeneratorTest.java
@@ -611,6 +611,39 @@ public class ProjectGeneratorTest {
     }
 
     @Test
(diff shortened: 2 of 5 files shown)


app/integration/api/src/main/java/io/syndesis/integration/api/IntegrationResourceManager.java
```diff
@@ -216,7 +216,19 @@ default String sanitize(String name) {
...
+        collectedDependencies.addAll(
+            collectDependencies(integration.getFlows().stream().flatMap(flow -> flow.getSteps().stream()).collect(Collectors.toList()), true)
+        );
+        return collectedDependencies;
+    }
+
+    default Collection<Dependency> collectDependencies(List<Flow> flows){
+        final List<Dependency> dependencies = new ArrayList<>();
+        for(Flow flow : flows){
+            dependencies.addAll(flow.getDependencies());
```
lburgazzoli: Wonder if collecting dependencies of steps should be done here instead of in `collectDependencies(Integration)` to respect the hierarchy

squakez: Good point

squakez: I've amended with the suggestion, please have a look now

app/integration/project-generator/src/test/java/io/syndesis/integration/project/generator/ProjectGeneratorTest.java
```diff
@@ -61,6 +61,7 @@
 import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.assertj.core.util.Lists;
```
zregvart: I think you're looking for `java.util.Collections` here

app/integration/project-generator/src/test/java/io/syndesis/integration/project/generator/ProjectGeneratorTest.java
```diff
@@ -610,6 +611,29 @@ public void testDependencyCollection() throws Exception {
...
+
+        Dependency mavenDependency = Dependency.maven("io.syndesis:something:1.0.0");
+        Dependency libraryExtension = Dependency.libraryTag("someLibraryTag");
+
+        Integration integration = new Integration.Builder()
+            .id("test-integration")
+            .name("Test Integration")
+            .description("This is a test integration!")
+            .addFlow(new Flow.Builder()
+                .steps(Lists.emptyList())
```
zregvart:
```suggestion
                .steps(Collections.emptyList())
```

squakez:
> Why flows and not integrations?

The idea is that every flow can have its own dependencies. We've discussed about this and we decided to apply this change at flow instead that at integration.

lburgazzoli:
> > Why flows and not integrations?
> 
> The idea is that every flow can have its own dependencies. We've discussed about this and we decided to apply this change at flow instead that at integration.

I think Zoran is right, we should do it also at integration level

zregvart:
If we plan on publishing (running) flows in separate processes then dependency per flow has some merit. Otherwise I would prefer having dependencies attached to the element that brought the need for the dependency (say connector/step/action) or if this dependency is result of human adding the dependency then that dependency should be at the deployment unit level: and that's integration.

We need to consider the ease of managing these dependencies, again if the human added then management is delegated to the human, if the dependency was added by code the dependency needs to reside near the cause of the addition (connector/step/action) in order to ease the management of it.

Issue #8150, opened 2020-03-30
Cannot build the ReactUI


KurtStam:
The dailybuild  on March 25 fails to build the UI React:

[INFO] Server :: Builder :: Image Generator ............... SUCCESS [  9.996 s] 
[INFO] S2I ................................................ SUCCESS [ 42.997 s] 
[INFO] Meta ............................................... SUCCESS [ 23.636 s] 
[INFO] UI React ........................................... FAILURE [20:33 min]

KurtStam:
March 28 adds a bit more detail:

[ERROR] Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.9.1:yarn (yarn-test) on project ui-react: Failed to run task: 'yarn test' failed. org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1) -> [Help 1] 
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.9.1:yarn (yarn-test) on project ui-react: Failed to run task 
- show quoted text -
Caused by: org.apache.maven.plugin.MojoFailureException: Failed to run task 
    at com.github.eirslett.maven.plugins.frontend.mojo.AbstractFrontendMojo.execute (AbstractFrontendMojo.java:100) 
- show quoted text -
Caused by: com.github.eirslett.maven.plugins.frontend.lib.TaskRunnerException: 'yarn test' failed. 
    at com.github.eirslett.maven.plugins.frontend.lib.YarnTaskExecutor.execute (YarnTaskExecutor.java:64) 
    at com.github.eirslett.maven.plugins.frontend.mojo.YarnMojo.execute (YarnMojo.java:65) 
    at com.github.eirslett.maven.plugins.frontend.mojo.AbstractFrontendMojo.execute (AbstractFrontendMojo.java:94) 
- show quoted text -
Caused by: com.github.eirslett.maven.plugins.frontend.lib.ProcessExecutionException: org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1) 
    at com.github.eirslett.maven.plugins.frontend.lib.ProcessExecutor.execute (ProcessExecutor.java:82) 
    at com.github.eirslett.maven.plugins.frontend.lib.ProcessExecutor.executeAndRedirectOutput (ProcessExecutor.java:64) 
[...]

PR #8159, opened 2020-03-31
chore: add babel corejs2 runtime as dev dependency for app test

gashcrumb:
for #8150

:man_shrugging:


gashcrumb:
Nope, that's not it

PR #8162, opened 2020-03-31
feat(dv): language server and monaco editor implementations

blafond:
Feature branch for implementation of DV language server and monaco editor based client.

- ENTESB-12900
- ENTESB-13051

This initial commit contains a server-side language server with incremental DDL parsing and validation (diagnostics) as well as first set of code-completion items and templates.

- TEIIDTOOLS-917
- TEIIDTOOLS-904
- TEIIDTOOLS-935
- TEIIDTOOLS-936


shawkins:
See https://github.com/blafond/syndesis/pull/5

blafond:
> To help scope this better I would like to remove anything that is not needed. There are still a couple of the original old classes for making web sockets that can be consolidated / removed. I'd go further to remove all currently unused code - including partially implemented features, such as hover, diagnostic helper, bracket analysis, etc. When those features are polished and ready to commit, then let's bring them back.
> 
> A lot of the logic that could or should be driven from the parser is hardcoded:
> DdlTokenAnalyzer.getNextWordsByKind, a lot of the parser.statement package, completion.providers, etc. What we talked about initially would have been driven by the expected tokens array, such that you'd know exactly what was allowed for the next token. A lot of the structural analysis is based upon assumptions of the query structure - as select ... from ... where ...; and does not seem to account for things like the with clause, unions, or subqueries.

The completion item implementation currently reflects a context based approach, which is basically a token with some additional info. I also added filtering to pare down lists if the context identifies existing tokens that can't be repeated. We can drop that part and rely on parsing/diagnostics.

I haven't spent much time on the standard SELECT query expression items other than allowing functions in SELECT. 

> eventually we'll want to metadata data drive some of that json creation - types and functions especially

That would be nice

> 
> The use of System.out should be minimized. Pass printstreams to your lower level objects and/or have them build strings / utilize toString - which helps with debugging as well. Then the bulk of that logic can also be testable.
> 
> Minimize the use of info logging.
> 
> A few places have catch exception.
> 
> I'll submit a pr that incorporates most of this. That should be easier to deal with than just review items here.

PR #8180, opened 2020-04-02
[Backport-1.9.x](#8125): Fix empty values handling in Google sheets

christophd:
The getOrDefault is not working when there is a key in the map with actual null value. So use Optional.ofNullable and make sure to handle null values and use empty String values instead.

Manual backport of #8178


zregvart:
I'd merge when the Circle CI check passes, the QE Jenkins PR build failed with something unrelated.

christophd:
ok will merge now

PR #8183, opened 2020-04-02
Fix: schema list issue

indraraj:
- Jira Issue https://issues.redhat.com/browse/TEIIDTOOLS-962

- Changes in ViewInfosContent component using schemaList state for populating the list of schema from the selected connection.

- updating the schema value only when schema(array) fetched from API call have one element.

- screenshot 
image


mdrillin:
This is not quite going to work due to differences in the 2 schema service calls.  I made a couple changes - will close this PR and open a new one for your review.

PR #8219, opened 2020-04-07
chore(migration): camel 3.1 changes

Delawen:
Fixing migration scripts for camel 3.

Truth is: I am not 100% sure how to test this to be sure it is foolproof.

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/Schema.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/Schema.java
@@ -20,7 +20,7 @@ package io.syndesis.common.model;
  */
 public final class Schema {
     // changing this will reset all the DB data.
-    public static final int VERSION = 40;
+    public static final int VERSION = 41;
 
     private Schema() {
       // quasi utility class
--- a/app/server/runtime/src/main/resources/migrations/up-40.js
+++ b/app/server/runtime/src/main/resources/migrations/up-40.js
@@ -14,12 +14,15 @@
  * limitations under the License.
  */
 
+console.log("Migrating to schema 40 ...");
+
+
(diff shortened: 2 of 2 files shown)


app/common/model/src/main/java/io/syndesis/common/model/Schema.java
```diff
@@ -20,7 +20,7 @@
  */
 public final class Schema {
     // changing this will reset all the DB data.
-    public static final int VERSION = 40;
+    public static final int VERSION = 41;
```
zregvart, line 23: I don't think this needs changing, not without an `up-41.js` script, otherwise we might have an issue when someone adds `up-41.js` it will not be executed as the schema version is already at 41.

app/server/runtime/src/main/resources/migrations/up-40.js
```diff
@@ -68,6 +68,13 @@ migrate("integrations", "/integrations", function(integration) {
                             if (Array.isArray(step.connection.connector.actions)) {
                                 step.connection.connector.actions.forEach(migrateAction);
                             }
+
+                            if(Array.isArray(step.connection.connector.dependencies)) {
```
zregvart, line 72: Nit
```suggestion
                            if (Array.isArray(step.connection.connector.dependencies)) {
```

app/common/model/src/main/java/io/syndesis/common/model/Schema.java
Delawen: But there is an up-41.js script.

zregvart: Missed that.

Delawen:
it is still running on an old version, didn't get the latest commit.

PR #8222, opened 2020-04-07
user doc: Emphasize that Query field accepts standard SQL only

roldanbob:


doc/modules/connecting/p_adding-db-connection-start.adoc
```diff
@@ -41,6 +41,13 @@ the connection invokes the specified statement or stored procedure every
 five minutes.
 . Click *Next*.
 
+.Limitations on the use of SQL extensions
+SQL statements that you enter in the *Query* field must be standard SQL constructs. 
+Fuse Online does not recognize or parse SQL extensions such as Procedural Language/PostgreSQL (PL/pgSQL)
```
TovaCohen, line 46:
```suggestion
{prodname} does not recognize or parse SQL extensions such as Procedural Language/PostgreSQL (PL/pgSQL) 
```

PR #8224, opened 2020-04-08
Fix Camel Version to 3.1.0 on some diverging files

claudio4j:
Current Camel version in app/pom.xml is 3.1.0 but
there are several project files that uses a different
camel version.


install/operator/build/conf/config-test.yaml
```diff
@@ -32,7 +32,7 @@ Syndesis:
             Enabled: false
             Tag: "fabric8/s2i-java:3.0-java8"
             SHA: "fabric8/s2i-java:3.0-java8@sha256:4d13fcc6eda389d4d679602171e11593eadae9b9"
-            CamelVersion: "2.23.2.fuse-760024"
+            CamelVersion: "3.1.0"
             CamelKRuntime: "0.3.4.fuse-740008"
```
zregvart, line 36: I don't think this version of Camel K supports this version of Camel. We need to change the Camel K runtime to 1.1.0 or newer.

heiko-braun: I think @zregvart is right and @nicolaferraro can comment on this. But I believe it's as simple as bumping camel-k version.

claudio4j: Just updated the camel-k-runtime, thanks for the review.

PR #8225, opened 2020-04-08
feat(server): extension filter

squakez:
* Overriding the extension API list to include an extensionType filter
* Add unit test to verify the new filter

Closes ENTESB-13247

Please, consider that this PR does not target a code refactory. This is in parallel [under discussion](https://groups.google.com/forum/#!topic/syndesis/PGJ1A6meKoQ)

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/extension/ExtensionHandler.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/extension/ExtensionHandler.java
@@ -45,6 +45,7 @@ import java.util.stream.Collectors;
 import io.swagger.annotations.Api;
 import io.swagger.annotations.ApiImplicitParam;
 import io.swagger.annotations.ApiImplicitParams;
+import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 import io.syndesis.common.model.Dependency;
@@ -75,7 +76,6 @@ import io.syndesis.server.endpoint.v1.SyndesisRestException;
 import io.syndesis.server.endpoint.v1.handler.BaseHandler;
 import io.syndesis.server.endpoint.v1.operations.Deleter;
 import io.syndesis.server.endpoint.v1.operations.Getter;
-import io.syndesis.server.endpoint.v1.operations.Lister;
 import io.syndesis.server.endpoint.v1.operations.PaginationOptionsFromQueryParams;
 import io.syndesis.server.endpoint.v1.operations.SortOptionsFromQueryParams;
 import io.syndesis.server.endpoint.v1.util.PredicateFilter;
@@ -90,7 +90,7 @@ import org.springframework.stereotype.Component;
 @Api(value = "extensions")
 @Component
 @ConditionalOnBean(FileDAO.class)
-public class ExtensionHandler extends BaseHandler implements Lister<Extension>, Getter<Extension>, Deleter<Extension> {
+public class ExtensionHandler extends BaseHandler implements Getter<Extension>, Deleter<Extension> {
 
     private final FileDAO fileStore;
     private final ExtensionActivator extensionActivator;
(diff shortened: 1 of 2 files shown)


app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/extension/ExtensionHandler.java
```diff
@@ -247,7 +250,26 @@ public void install(@NotNull @PathParam("id") final String id) {
...
+            name = "page", value = "Page number to return", paramType = "query", dataType = "integer", defaultValue = "1"),
+        @ApiImplicitParam(
+            name = "per_page", value = "Number of records per page", paramType = "query", dataType = "integer", defaultValue = "20"),
+        @ApiImplicitParam(
+            name = "query", value = "The search query to filter results on", paramType = "query", dataType = "string"),
+        @ApiImplicitParam(
+            name = "extensionType", value = "The type to filter results on", paramType = "query", dataType = "string"),
+    })
+    public ListResult<Extension> list(@Context UriInfo uriInfo) {
+        String extensionType = uriInfo.getQueryParameters().getFirst("extensionType");
```
zregvart, line 272:
```suggestion
    })
    public ListResult<Extension> list(@Context UriInfo uriInfo, @QueryParam("extensionType") @DefaultValue("Steps") @ApiParam(required = false) Extension.Type extensionType) {
```

squakez: Thanks @zregvart. Please, notice that the method still overrides from Lister. Your suggestion is what I'd like to do as an additional task once we reach consensus on [this thread](https://groups.google.com/forum/#!topic/syndesis/PGJ1A6meKoQ).

zregvart: Lister is not a contract, Lister is a mixin. Make it not override from Lister, if you look at the API the end result is the same. If you look at the code this option is much better. Though I think we should not use reflection for filtering this suggestion has no relevance toward reflection. This suggestion is about making the API contract explicit and type safe.

squakez: I've applied the suggestion. I wanted to work separately on this feature and later on cleaning everything that was depending by that List interface. As that was born by a discussion with you, please feel free to chime in the thread: https://groups.google.com/forum/#!topic/syndesis/PGJ1A6meKoQ

zregvart: Actually, you did not apply the suggestion.

squakez: Hey @zregvart. If I read correctly you proposed a code change and  "Make it not override from Lister". This was provided by this commit: 34bd4d09cdbb1abb72f4a9030f3d95782fc179e7. Were you suggesting something else?

zregvart: Did you notice the suggestion diff above?

squakez: I've added another commit (34bd4d09cdbb1abb72f4a9030f3d95782fc179e7) with the change requested, [this is the final result](https://github.com/syndesisio/syndesis/blob/master/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/extension/ExtensionHandler.java#L268). A part the default value, the rest is the same. Or maybe your're talking about something else?

zregvart: @squakez if you carefully examine the suggestion diff displayed above this conversation, you might notice that the extension type parameter is of `Extension.Type` and not fetched from from `uriInfo.getQueryParameters()`

squakez: Thanks, I see that. I've provided a `String` parameter instead. I blindly assumed it had to be a String and to be converted. I'll log a new issue to fix that. The thread I've opened is still idle, I have bandwidth to work on a general improvement if I got some positive feedback though.

PR #8243, opened 2020-04-14
feat: ENTESB-12882 Add metering labels to Fuse Online operator

claudio4j:
Add com.redhat.product labels to pod only when building with
productized=true in operator level.
There is an assumption about the versioning scheme
For operator tag version x.y the product version is x+6.y
Example:
Operator Tag: 1.6, Product Version: 7.6
Operator Tag: 1.7-12, Product Version: 7.7

https://issues.redhat.com/browse/ENTESB-12882


app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java
```diff
@@ -75,10 +74,21 @@
 
     private final NamespacedOpenShiftClient openShiftClient;
     private final OpenShiftConfigurationProperties config;
+    private Map<String, String> productizedLabelsTemplate = new HashMap<>();
 
     public OpenShiftServiceImpl(NamespacedOpenShiftClient openShiftClient, OpenShiftConfigurationProperties config) {
         this.openShiftClient = openShiftClient;
         this.config = config;
+        // retrieve the productized labels: com.redhat, they exists only when syndesis-operator is built with productized=true
+        DeploymentConfig dc = openShiftClient.deploymentConfigs() != null  ?
```
phantomjinx, line 83: Just a note to say this will be modified when the kubernetes branch gets merged since all deploymentconfigs will become deployments.
Will rebase the k8 branch to include this then modify. Will be couple of days yet before merging.

app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java
```diff
@@ -75,10 +74,21 @@
...
+            openShiftClient.deploymentConfigs().withName("syndesis-server").get() : null;
+        if (dc != null) {
+            Map<String, String> labels = dc.getSpec().getTemplate().getMetadata().getLabels();
+            productizedLabelsTemplate = Collections.unmodifiableMap(
+                labels.entrySet().stream()
+                    .filter(entry -> entry.getKey().contains("com.redhat"))
```
phantomjinx, line 89: This doesn't feel right...
The PR is going technically into upstream master yet we are specifically looking for company-centric product-only labels. Is there a way perhaps to add these to a properties file that is only included in product?

@zregvart WDYT?

app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java
```diff
@@ -246,6 +256,11 @@ protected void ensureDeploymentConfig(String name, DeploymentData deploymentData
         if (jaegerCollectorUri == null) {
             jaegerCollectorUri = "http://syndesis-jaeger-collector:14268/api/traces";
         }
+        Map<String, String> productizedLabels = new HashMap<>(4);
+        if (!productizedLabelsTemplate.isEmpty()) {
+            productizedLabels.putAll(productizedLabelsTemplate);
+            productizedLabels.put("com.redhat.component-name", name);
```
phantomjinx, line 262: If I was another dev working on this upstream am not sure I'd be comfortable with certain product label being added. Maybe if we can generalise this so that any interesting labels that have been added to the DeploymentConfig could then be handled like this. Then we don't have to call it 'productized' but, dunno, linking-labels or something...

app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java
claudio4j: ok, once you branch is merged I will change this impl.

install/operator/build.sh
```diff
@@ -36,13 +37,17 @@ where options are:
   --s2i-stream-name <name>                s2i image stream name (default: syndesis-operator)
   --go-options <name>                     additional build options to pass to the go build
   --go-proxy <url>                        proxy url for finding go dependencies (default: https://proxy.golang.org)
+  --productized <true|false>              activate settings only for product builds (default: false)
```
phantomjinx, line 40: This should not be necessary. We already have a productized flag in [config.yaml](https://github.com/syndesisio/syndesis/blob/master/install/operator/build/conf/config.yaml#L3). This is modified in the product version of that file.

install/operator/build.sh
```diff
@@ -36,13 +37,17 @@ where options are:
...
 ENDHELP
 	exit 0
 fi
 
 if [ $OPERATOR_BUILD_MODE != "skip" ] ; then
-  build_operator $OPERATOR_BUILD_MODE "$SOURCE_GEN" "$GO_PROXY_URL" -ldflags "-X github.com/syndesisio/syndesis/install/operator/pkg.DefaultOperatorImage=$OPERATOR_IMAGE_NAME -X github.com/syndesisio/syndesis/install/operator/pkg.DefaultOperatorTag=$OPERATOR_IMAGE_TAG" $GO_BUILD_OPTIONS
+  LDFLAGS="-X github.com/syndesisio/syndesis/install/operator/pkg.DefaultOperatorImage=$OPERATOR_IMAGE_NAME"
+  LDFLAGS=$LDFLAGS" -X github.com/syndesisio/syndesis/install/operator/pkg.DefaultOperatorTag=$OPERATOR_IMAGE_TAG"
+  LDFLAGS=$LDFLAGS" -X github.com/syndesisio/syndesis/install/operator/pkg.Productized=$PRODUCTIZED"
+  build_operator $OPERATOR_BUILD_MODE "$SOURCE_GEN" "$GO_PROXY_URL" -ldflags "$LDFLAGS" $GO_BUILD_OPTIONS
```
phantomjinx, line 50: Need to understand why the productized flag is required at compile time rather than runtime. The config file is read at runtime and takes care of the Productized flag for all the infrastructure components. Therefore, the only thing that is not subjected to the config file is the operator install itself which uses RenderScope rather than configuration.
See below for further thoughts on this bit.

install/operator/pkg/cmd/internal/install/install.go
```diff
@@ -245,16 +250,23 @@ func (o *Install) render(fromFile string) ([]unstructured.Unstructured, error) {
...
-		EnabledAddons: addons,
-		DatabaseImage: o.databaseImage,
+		Namespace:      o.Namespace,
+		Image:          o.image,
+		Tag:            o.tag,
+		DevSupport:     o.devSupport,
+		Role:           RoleName,
+		Kind:           "Role",
+		Productized:    productized,
+		ProductVersion: productVersion,
```
phantomjinx, line 267: So the Productized flag is being read in to RenderScope from a boolean set at compile-time not at runtime. This seems wrong since it is possible to read the Productized flag from config. Although we don't at the moment call [configuraiton.GetProperties](https://github.com/syndesisio/syndesis/blob/master/install/operator/pkg/syndesis/configuration/configuration.go#L308) in this install.Render function we do call it in the [before](https://github.com/syndesisio/syndesis/blob/master/install/operator/pkg/cmd/internal/install/install.go#L154) function. Therefore, it should be possible to extract the Productized flag from that configuration and add it to Renderscope (maybe using the o.Productized as the vehicle to hold the value migrating it to o.productized).

install/operator/pkg/constants.go
```diff
@@ -4,3 +4,4 @@ var DefaultOperatorImage = "docker.io/syndesis/syndesis-operator"
 var DefaultOperatorTag = "latest"
 var JobNameLabel = "job-name"
 var ControllerUidLabel = "controller-uid"
+var Productized = "false"
```
phantomjinx, line 7: This should be unnecessary.

install/operator/pkg/generator/assets/addons/dv/addon-dv-server.yml.tmpl
```diff
@@ -42,6 +42,12 @@
       metadata:
         labels:
           app: syndesis
+{{if .Productized}}
+          com.redhat.component-name: syndesis-dv
+          com.redhat.component-version: "{{.Tag}}"
+          com.redhat.product-name: "Red_Hat_Fuse"
+          com.redhat.product-version: "{{.ProductVersion}}"
+{{end}}
```
phantomjinx, line 50: These components should work in the same way as [amq-example](https://github.com/syndesisio/syndesis/blob/master/install/operator/pkg/generator/assets/infrastructure/04-amq-example.yml.tmpl#L1)

install/operator/pkg/generator/assets/addons/jaeger/server_operator.yml.tmpl
```diff
@@ -19,6 +19,12 @@
       metadata:
         labels:
           name: jaeger-operator
+{{if .Productized}}
+          com.redhat.component-name: jaeger-operator
+          com.redhat.component-version: "{{.Tag}}"
+          com.redhat.product-name: "Red_Hat_Fuse"
```
phantomjinx, line 25: Really think we need to discuss the difference between product and upstream as putting these into upstream just seem wrong.
Appreciate opinions please @zregvart @Delawen @heiko-braun @lgarciaaco @.....

install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -61,14 +62,16 @@ var log = logf.Log.WithName("configuration")
...
-	OpenShiftConsoleUrl        string         // The URL to the OpenShift console
+	Version                    string // Syndesis version
+	DevSupport                 bool   // If set to true, pull docker images from imagetag instead of upstream source
+	Scheduled                  bool   // Legacy parameter to set scheduled:true in the imagestreams, but we dont use many imagestreams nowadays
+	ProductName                string // Usually syndesis or fuse-online
+	PrometheusRules            string // If some extra rules for prometheus need to be specified, they are defined here
+	OpenShiftProject           string // The name of the OpenShift project Syndesis is being deployed into
+	OpenShiftOauthClientSecret string // OpenShift OAuth client secret
+	OpenShiftConsoleUrl        string // The URL to the OpenShift console
+	Tag                        string
```
phantomjinx, line 73: Tag need a comment?

install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -320,6 +323,12 @@ func GetProperties(file string, ctx context.Context, client client.Client, synde
 		}
 	}
 	configuration.generatePasswords()
+	v, err := strconv.ParseBool(pkg.Productized)
```
phantomjinx, line 326: configuration already has a Productized flag that is set by loadFromFile(...)

install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -701,3 +716,42 @@ func getSyndesisEnvVarsFromOpenShiftNamespace(secret *corev1.Secret) (map[string
...
+For operator tag version x.y the product version is x+6.y
+Example:
+input 1.6, returns: 7.6
+input 1.7-12, returns: 7.7
+
+If the input parameter is only a string "latest", it returns the same value "latest".
+It discards the patch level version number.
+- operatorTagVersion: A string value: example 1.6, 1.7-12
+*/
+func ProductVersionOfTag(ver string) string {
```
phantomjinx, line 732: This feels like an addon function that should only exist downstream.

zregvart:
For changes to the server, they seem to be overly complex, as demonstrated by additional code complexity check suppression. Changing this many files at this many points for a elementary feature such this is too much.

I'd go with creating a `DeploymentDataCustomizer` to add labels and modifying the `syndesis-server` config map to put these additional labels when this is built as product.

We already have some references to "redhat" in project generation, not sure if that's relevant any more or it can be removed as well.

claudio4j:
> I'd go with creating a `DeploymentDataCustomizer` to add labels and modifying the `syndesis-server` config map to put these additional labels when this is built as product.

I didn't know about this `DeploymentDataCustomizer` thanks for pointing it.
 
> We already have some references to "redhat" in project generation, not sure if that's relevant any more or it can be removed as well.

Yeah, there is `app/integration/project-generator/src/main/resources/application-redhat.properties` isn't used for product builds ?

zregvart:
> > I'd go with creating a `DeploymentDataCustomizer` to add labels and modifying the `syndesis-server` config map to put these additional labels when this is built as product.
> 
> I didn't know about this `DeploymentDataCustomizer` thanks for pointing it. But implementing with custom redhat labels, will just move the product feature to a different file in the same project.

This is about having separate concerns separate, what is in this pull request is mixing the data generation with data application, those are two separate concerns and should not be mixed.

The key part is the use of config map, the `DeploymentDataCustomizer` should be configured via the config map to add arbitrary labels configured there. Then it's up to anyone to add whatever labels they need in addition to the default operational labels we need in Syndesis.

> > We already have some references to "redhat" in project generation, not sure if that's relevant any more or it can be removed as well.
> 
> Yeah, there is `app/integration/project-generator/src/main/resources/application-redhat.properties` isn't used for product builds ?

I have no idea how and where is that used.

claudio4j:
As this is a product specific feature, I will rework it in generic way to generate the labels, as a result I am closing this PR.

PR #8249, opened 2020-04-15
feat(integration-test): Add support for extensions

### LINKED ISSUE #8248: [Test] Integration test support for extensions
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description
We are missing the upload of extensions in the integration test. In order to properly test any integration with provided extensions, we should include them in the `integration test` runtime.

squakez:
* Added support for extensions by adding the "extensions" folder under the "deployments" S2I and "LOADER_HOME" environment variable used to locate the correct classpath
* IntegrationDependencies_IT test to verify extension is uploaded and its dependencies are available at classpath
* Additional unit test to verify the S2I fat jar is included as expected

Closes #8248, closes ENTESB-13383


app/test/integration-test/pom.xml
```diff
@@ -295,17 +295,6 @@
     <testResources>
       <testResource>
         <directory>src/test/resources</directory>
-        <filtering>true</filtering>
```
squakez, line 298: @christophd Are these filters needed? for some reason, with these filters, any jar I am using to test is converted by maven in a zip and the integration is failing.

christophd: I think we need to add an exclude for `*.jar` files when those live in the `src/test/resources` folder. Initially we have used integration exports as `.zip` but now we are using the pure `model.json`. So I think we could also remove that part from the `pom.xml`

app/test/integration-test/src/test/java/io/syndesis/test/itest/extensions/IntegrationDependencies_IT.java
```diff
@@ -0,0 +1,115 @@
...
+import org.springframework.test.context.ContextConfiguration;
+import org.testcontainers.containers.wait.strategy.Wait;
+
+@ContextConfiguration(classes = IntegrationDependencies_IT.EndpointConfig.class)
+public class IntegrationDependencies_IT extends SyndesisIntegrationTestSupport {
+
+    @Test
+    public void shouldVerifyDependecyClasspath() throws URISyntaxException {
+        // We need an S2I to load extensions in classpath
+        System.setProperty("syndesis.s2i.build.enabled", "true");
```
christophd, line 43: I would prefer to add this as a new setting on the `SyndesisIntegrationRuntimeContainer.Builder`. The default value for this setting would come from the System property but the user can overwrite this on the builder with something like:

```
new SyndesisIntegrationRuntimeContainer.Builder()
            .name("bean-to-log")
            .enableS2iBuild()
```

app/test/integration-test/src/test/java/io/syndesis/test/itest/extensions/IntegrationDependencies_IT.java
```diff
@@ -0,0 +1,115 @@
...
+        System.setProperty("syndesis.debug.enabled", "false");
```
christophd, line 44: This flag could be set on the `SyndesisIntegrationRuntimeContainer.Builder`. There is already a possibility to set `enableDebug()`

```
new SyndesisIntegrationRuntimeContainer.Builder()
            .name("bean-to-log")
            .enableDebug()
```

So we need to refactor/add something to disable debug on the builder.

PR #8255, opened 2020-04-16
[Backport 1.10.x] Kafka connector TLS support

squakez:
Porting few commits and changing some import to be compliant with Camel 2.x


squakez:
I am having UI failures even after merging to the latest `1.10.x`

@syndesisio/ui any idea how to fix this?

kahboom:
I've kicked off a new build (seemed to fix it the last time this happened), but this theoretically should have been fixed with this: https://github.com/syndesisio/syndesis/pull/7913

squakez:
> I've kicked off a new build (seemed to fix it the last time this happened), but this theoretically should have been fixed with this: #7913

Thanks @kahboom , it seems that those commits have not yet ported to `1.10.x`. I can pick them into this PR too if there is no other cascade commits they need.

gashcrumb:
Also I found when working with the DV folks changing that setting from 2 to
1 improved reliability even more.

PR #8257, opened 2020-04-16
1.10.x

shawkins:
back porting master work into 1.10.x

### DIFF
--- a/app/common/util/src/main/java/io/syndesis/common/util/EventBus.java
+++ b/app/common/util/src/main/java/io/syndesis/common/util/EventBus.java
@@ -44,6 +44,7 @@ public interface EventBus {
     /**
      * Callback interface clients implement to receive events.
      */
+    @FunctionalInterface
     interface Subscription {
 
         /**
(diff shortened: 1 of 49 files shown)


shawkins:
@mdrillin this is my best attempt at pulling back master changes with commit history for just the backend.  

I did not include:

468604b6684470c2935c17144d10363692503456 combine similar dv schema services
b61c81f98f2f46c86adfe7750cb4ca748ffa7e27 TEIIDTOOLS-960 Resolve issue with teiid source naming

because the overlap with ui changes.

I also reduced the scope of f2fc423c4617fd2bb687f2a9e329995f083d762e to just the dv pom change.

Most of the backports had a conflict, so I did not yet attempt anything on the ui side.  Since you have a longer list if you see similar issues due to intervening commits / backports, you may want to try just working off of a diff.  Can you pull this locally and see how far you get with your change list?  Then we can coordinate on those two commits above that cross both sides.

mdrillin:
@shawkins sure, I will pull it down.  thanks

PR #8279, opened 2020-04-20
fix(test): maven debug option fix

### LINKED ISSUE #8265: [itest] Can't debug container
## This is a...

<pre><code>
[ ] Feature request
[X] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
</code></pre>

## Description

If I try to use the debugging method posted [here](https://github.com/syndesisio/syndesis/blob/master/app/test/integration-test/README.md#debugging%20syndesis), it doesn't suspend the integration container to let me debug. I tried both adding the `.enableDebug()` programatically and adding a `-Dsyndesis.debug.enabled=true` parameter to the `mvn clean verify` command. On both cases (or combined) it just run the test normally.

squakez:
We were using `exec:java` debugging option instead of `spring-boot:run` option.

Fixes #8265


Delawen:
Makes sense :)

Issue #8215, opened 2020-04-07
failed[integration-test]: TodoOpenApiV3_IT


christophd:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

TodoOpenApiV3_IT is failing on master. Here is what I see in the logs when running the test locally:

```
07:15:38.496 [ream-1082378213] INFO  INTEGRATION_RUNTIME_CONTAINER  - STDOUT: io.syndesis.common.util.SyndesisConnectorException: SQL SELECT did not SELECT any records
07:15:38.496 [ream-1082378213] INFO  INTEGRATION_RUNTIME_CONTAINER  - STDOUT: 	at io.syndesis.connector.sql.customizer.SqlConnectorCustomizer.doAfterProducer(SqlConnectorCustomizer.java:124) ~[connector-sql-2.0-SNAPSHOT.jar:2.0-SNAPSHOT]
07:15:38.497 [ream-1082378213] INFO  INTEGRATION_RUNTIME_CONTAINER  - STDOUT: 	at org.apache.camel.support.AsyncProcessorConverterHelper$ProcessorToAsyncProcessorBridge.process(AsyncProcessorConverterHelper.java:67) ~[camel-support-3.1.0.jar:3.1.0]
07:15:38.497 [ream-1082378213] INFO  INTEGRATION_RUNTIME_CONTAINER  - STDOUT: 	at org.apache.camel.processor.Pipeline$PipelineTask.run(Pipeline.java:90) ~[camel-base-3.1.0.jar:3.1.0]
07:15:38.503 [ream-1082378213] INFO  INTEGRATION_RUNTIME_CONTAINER  - STDOUT: 	at org.apache.camel.impl.engine.DefaultReactiveExecutor$Worker.schedule(DefaultReactiveExecutor.java:153) [camel-base-3.1.0.jar:3.1.0]
07:15:38.503 [ream-1082378213] INFO  INTEGRATION_RUNTIME_CONTAINER  - STDOUT: 	at org.apache.camel.impl.engine.DefaultReactiveExecutor.scheduleMain(DefaultReactiveExecutor.java:60) [camel-base-3.1.0.jar:3.1.0]
[...]

PR #8287, opened 2020-04-21
Fix for TodoOpenApiV3_IT #8215

KurtStam:
Fixing #8215, failing integration tests due to upgrade to Camel 3


christophd:
@KurtStam do we need to backport this to 1.10.x?

KurtStam:
No need to backport as this is caused by the camel-3 upgrade.

christophd:
totally, thanks for having a look

PR #8298, opened 2020-04-22
Replaced old references of CreateConnection translation

mkralik3:
After this change https://github.com/syndesisio/syndesis/commit/d2375570b29048b08b80e0e0e162292d948dbef8#diff-38d54cc1ec6b270563c94153f356a568L49 , some references to CreateConnection translation property were out of date.
image
I have replaced all occurrence `shared:CreateConnection` to `shared:linkCreateConnection` in the `ui-react` package.

### DIFF
--- a/app/ui-react/syndesis/src/modules/connections/components/ConnectionsWithToolbar.tsx
+++ b/app/ui-react/syndesis/src/modules/connections/components/ConnectionsWithToolbar.tsx
@@ -94,7 +94,7 @@ export class ConnectionsWithToolbar extends React.Component<
                   sortTypes={sortTypes}
                   resultsCount={filteredAndSortedConnections.length}
                   {...helpers}
-                  i18nLinkCreateConnection={t('shared:CreateConnection')}
+                  i18nLinkCreateConnection={t('shared:linkCreateConnection')}
                   i18nResultsCount={t('shared:resultsCount', {
                     count: filteredAndSortedConnections.length,
                   })}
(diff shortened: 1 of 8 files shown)


kahboom:
is this also in master btw? if not, it should go in there first and then get backported into 1.10.x

mkralik3:
@kahboom The change https://github.com/syndesisio/syndesis/commit/d2375570b29048b08b80e0e0e162292d948dbef8#diff-38d54cc1ec6b270563c94153f356a568L49  (PR - https://github.com/syndesisio/syndesis/pull/8263) which caused this problem is not in the master yet. Due to that, I have not created the PR with this fix to the master yet either.

mdrillin:
I probably made a mistake on our master ---> 1.10.x DV backports.  Looking into it.

gashcrumb:
My fault I think, sorry!  I didn't care for the key name that was used for that text string and changed it when refactoring the create connection pages but could have probably communicated that better.

mdrillin:
@gashcrumb I think two of the translation files were missed on the master commit.  I'd propose that we change the it and ru translations to 'CreateConnection' on master as intended.  Then get the 1.10.x in sync with that.  wdyt?

gashcrumb:
:+1:

mdrillin:
Ok, I'll do that.  I think I just messed up the 1.10.x backport when I resolved conflicts :-)

mdrillin:
@gashcrumb I've submitted the PR against master to make the translation keys consistent.  Then another PR against 1.10.x to bring it into sync with master.  The commit histories will be a little be confusing, but I think they would have been regardless...

gashcrumb:
Yep, that's perfect, thanks for taking the time to straighten out my leftovers :smile:

Issue #8296, opened 2020-04-22
`syndesis-operator grant --cluster` error


zregvart:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

If I run `syndesis-operator grant --cluster` I get:

```
error: Object:
  apiVersion: rbac.authorization.k8s.io/v1
  kind: ClusterRoleBinding
  metadata:
    name: syndesis-installer-syndesis/syndesis-operator
    namespace: default
  roleRef:
    apiGroup: rbac.authorization.k8s.io
    kind: ClusterRole
    name: syndesis-installer
  subjects:
  - apiGroup: rbac.authorization.k8s.io
    kind: User
    name: syndesis/syndesis-operator
: ClusterRoleBinding.rbac.authorization.k8s.io "syndesis-installer-syndesis/syndesis-operator" is invalid: metadata.name: Invalid value: "syndesis-installer-syndesis/syndesis-operator": may not contain '/'
```

This is with [1.9.0](https://github.com/syndesisio/syndesis/releases/tag/1.9.0)

phantomjinx:
You have missed a --user switch out ...
`syndesis-operator grant --user developer --cluster`

Although, the lack of a `--user <user>` should be better handled.

zregvart:
Not much documentation on this, is there any?

lgarciaaco:
This is probably a bug in [cobra](github.com/spf13/cobra) or something we are missing, because the `user` flag is marked as mandatory:

https://github.com/syndesisio/syndesis/blob/f477ec1b3bd8b8aaaf2ccce12aba775a31f7324e/install/operator/pkg/cmd/internal/grant/grant.go#L55

phantomjinx:
> Not much documentation on this, is there any?

I would say no for 2 reasons:

1. The switches and functionality is the operator has been so fluid, documentation rarely keeps up
2. The operator has tended to be wrapped in other scripts rather than called directly hence direct user interaction has been uncommon.

If we are going to put the operator front and centre then no reason this situation shouldn't change.

PR #8302, opened 2020-04-22
[operator] --user is mandatory when calling grant

lgarciaaco:
If `syndesis-operator grant` command is called, the `--user` flag is mandatory, but it was ignoring this fact. Also, remove the default value from `--user` which made no sense at all.

Fixes: https://github.com/syndesisio/syndesis/issues/8296


lgarciaaco:
Now when running  `./syndesis-operator grant --cluster`

The output is:
```
Error: required flag(s) "user" not set
Usage:
  grant grant [flags]

Flags:
      --cluster                          add the permission for all projects in the cluster(requires cluster admin privileges)
.
.
.
```

PR #8303, opened 2020-04-22
chore(dv) ENTESB-13076 adding a hook to reset dv

shawkins:
adding truncates for dv tables in the existing test support method seems
brittle. trying to work this through a db function is very postgresql
specific, so adding a new dv endpoint seemed like the best option


shawkins:
I don't know if it's due to aa187bbcb9fbd3246d64893cffcde2a18ae74bfc but I can't build the operator to regenerate the assets for the dv template change.  I get:

F0422 09:29:15.192403 3049306 config.go:57] Failed loading boilerplate: open /home/shawkins/go/src/k8s.io/kube-openapi/boilerplate/boilerplate.go.txt: no such file or directory

PR #8309, opened 2020-04-23
fix 8034 Integration fails to start: prometheus config cannot be found

KurtStam:
Backporting fix to 1.10.x


KurtStam:
No need to backport this fix as 1.10.x is not on Camel-3

PR #8310, opened 2020-04-23
refactor(ci): PR number from CI_PULL_REQUEST env

zregvart:
Seems that `CIRCLE_PR_NUMBER` is not getting set on pull requests that
originate from branches, `CI_PULL_REQUEST` seems to be set on both so
this changes to using that instead.


avano:
retest this please

PR #8314, opened 2020-04-23
fix(ENTESB-13518): Fix upgrading syndesis installed with an external db

phantomjinx:
* configuration.go
 * GetProperties is the essential function called for initialising a
   configuration object. However, ExternalDatabase is crucial for checking
   if the db is external. Instead of calling this separately, include it as
   part of GetProperties

* install.go
 * ExternalDatabase function removed as now internal
 * No need to try and call removePostgresUpgradeTrigger() since this is
   only for internal db

* .lib.sh
 * Fixes naming/tagging of operator imagestream

### DIFF
--- a/install/operator/pkg/generator/generator_test.go
+++ b/install/operator/pkg/generator/generator_test.go
@@ -6,14 +6,43 @@ import (
 	"strings"
 	"testing"
 
+	osappsv1 "github.com/openshift/api/apps/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/syndesisio/syndesis/install/operator/pkg/apis/syndesis/v1beta1"
 	"github.com/syndesisio/syndesis/install/operator/pkg/generator"
 	"github.com/syndesisio/syndesis/install/operator/pkg/syndesis/configuration"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
+
+//
+// Registers the DeploymentConfig type and adds in a
+// mock syndesis-db deployment config runtime object
+//
+func fakeClient() client.Client {
+	s := scheme.Scheme
+	osappsv1.AddToScheme(s)
+
+	synDbDeployment := &osappsv1.DeploymentConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "syndesis-db",
+		},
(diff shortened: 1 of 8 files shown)


avano:
retest this please

install/operator/pkg/generator/generator_test.go
```diff
@@ -6,15 +6,44 @@ import (
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
+//
+// Registers the DeploymentConfig type and adds in a
+// mock syndesis-db deployment config runtime object
+//
+func fakeClient() client.Client {
```
phantomjinx, line 27: This has to be added so that the fakeclient knows about v1.DeploymentConfig. This is not so surprising given this client is a basic fake and not aware of openshift APIs.

install/operator/pkg/generator/generator_test.go
```diff
@@ -78,7 +107,7 @@ func TestGenerator(t *testing.T) {
 		},
 	}
 
-	configuration, err := configuration.GetProperties("../../build/conf/config.yaml", context.TODO(), fake.NewFakeClient(), syndesis)
+	configuration, err := configuration.GetProperties("../../build/conf/config.yaml", context.TODO(), fakeClient(), syndesis)
```
phantomjinx, line 110: This fake.NewFakeClient() is not enough to allow for DeploymentConfig() to be handled.

install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -329,47 +329,72 @@ func GetProperties(file string, ctx context.Context, client client.Client, synde
...
 
-	databaseDeployment := &appsv1.DeploymentConfig{}
-	if err := client.Get(ctx, types.NamespacedName{Namespace: syndesis.Namespace, Name: "syndesis-db"}, databaseDeployment); err == nil {
+	//
+	// Only required if not using an external database and need to check
+	// on whether the deployment config is already is upgrade mode
+	//
+	if client != nil && len(syndesis.Spec.Components.Database.ExternalDbURL) <= 0 {
+		databaseDeployment := &appsv1.DeploymentConfig{}
+		if err := client.Get(ctx, types.NamespacedName{Namespace: syndesis.Namespace, Name: "syndesis-db"}, databaseDeployment); err != nil {
```
phantomjinx, line 347: This seemed an innocent change but in fact uncovers a problem with the runtime client. It doesn't support DeploymentConfig. Looking at [L337](https://github.com/syndesisio/syndesis/pull/8314/files#diff-8f63b1158cd0c8188207de68b0013e69L337) the condition was `if err == null` so if the `client.Get(...)` had no error then do stuff. The change turns it around so that if there is an error then flag it. Therefore, the following error was being swallowed and ignored:
```
error: no kind is registered for the type v1.DeploymentConfig in scheme "k8s.io/client-go/kubernetes/scheme/register.go:69"
```
At some point, this started happening because this code did its job of maintaining an upgrade. So at some point it stopped working??

@lgarciaaco

PR #8318, opened 2020-04-23
chore(dv) ENTESB-13076 adding a hook to reset dv

shawkins:
adding truncates for dv tables in the existing test support method seems brittle. trying to work this through a db function is very postgresql specific, so adding a new dv endpoint seemed like the best option


shawkins:
retest this please

PR #8344, opened 2020-04-27
fix(atlasmap+itest): previous mappings were lost on following steps

Delawen:
So this is a possible fix on the bug I was hunting on the integration tests. I'm not quite convinced this is the proper solution. Let's discuss :)

I reduced the problem to the usage of a `org.apache.camel.StreamCache` when extracting data from previous steps. First time you extract the data, no problems. Further attempts to extract the data lead to an empty/exhausted StreamCache returning an empty value/string.

So when you use the data coming from the immediate previous step, everything is fine. But if you try to reuse older data, empty strings.

I am not sure if we used StreamCache previously or why it is failing now. 

So, my solution is: read once on the `io.syndesis.integration.runtime.capture.OutMessageCaptureProcessor` and pass that read value to AtlasMap. No StreamCache on AtlasMap.

This leads to issues because I suspect:
 a) Not all InputStreamCache can be easily converted to String.
 b) What about the rest of the StreamCache types? I know this one usecase, but there has to be more for sure!

Also added a very dummy integration test for this usecase.

### DIFF
--- a/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/capture/OutMessageCaptureProcessor.java
+++ b/app/integration/runtime/src/main/java/io/syndesis/integration/runtime/capture/OutMessageCaptureProcessor.java
@@ -17,11 +17,11 @@ package io.syndesis.integration.runtime.capture;
 
 import java.util.HashMap;
 import java.util.Map;
-
 import io.syndesis.integration.runtime.logging.IntegrationLoggingConstants;
 import org.apache.camel.Exchange;
 import org.apache.camel.Message;
 import org.apache.camel.Processor;
+import org.apache.camel.converter.stream.InputStreamCache;
 import org.apache.camel.support.MessageSupport;
 
 /**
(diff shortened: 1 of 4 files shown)


Delawen:
Funny note: this was quite fun to debug, because everytime I inspected the value of the `getBody(String.class)` and checked it was right, then the function that extracted that value always returned "" which was very annoying :)

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/capture/OutMessageCaptureProcessor.java
```diff
@@ -44,6 +44,13 @@ public void process(Exchange exchange) throws Exception {
                 ((MessageSupport) copy).setExchange(message.getExchange());
             }
 
+            //If it is a stream cache, it can only be read once
+            //We are going to read it more than once
+            //to reuse variables from previous steps
+            if(copy.getBody() instanceof InputStreamCache) {
+                copy.setBody(copy.getBody(String.class));
```
zregvart, line 51: Bodies could be quite large streams, say files from S3. Or even binary and converting to and from Strings might corrupt them.

Delawen:
Not like this :)

PR #8367, opened 2020-04-29
refactor(server): remove query parameteres using reflection

### LINKED ISSUE #8286: Reduce reflection usage in API filtering
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description

As [originally discussed with the community](https://groups.google.com/forum/#!topic/syndesis/PGJ1A6meKoQ), we can think to refactor and reduce the usage of reflection in `server` API.

The generic [Lister](https://github.com/syndesisio/syndesis/blob/master/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/operations/Lister.java) class has a default implementation that expose a few implicit parameters that are not used by the `UI`. The implementation makes a massive use of reflection.

We use it to server 3 features:

- Filtering
- Pagination
- Sorting

_Pagination_ is fine, as it does not use reflection and can provide a general purpose definition of how we paginate list contents. We may remove that default implementation and all the reflection used for _filtering_ and _sorting_ and leave it as an explicit requirements for those entities that have such need (at the moment just a bunch of them).

squakez:
As specified in #8286, the goal of this PR is to specify explicitly any API parameter instead of leaving it implicit and wide open to any possibility. The UI is not making usage of sorting parameters, and in a very few occasions is using generic `query` filtering parameters.

I've worked over 3 related commits to ease the refactor and the PR review:

6d45721 : changed the operation interface List which is inherited by many entities API. The list method now have only pagination query parameters. I've changed also all classes overriding the default implementation to adhere the new signature.

312c295 : I've removed any usage of the generic UriInfo class. Now all APIs have a fixed contract.

ba4d017 : the old Filter and Sorter utility classes are no longer used and can be removed from the base code.

As this change is removing certain parameters, I feel we should upgrade API version as well from `v1` to `v2`.

Closes #8286


squakez:
Any feedback @syndesisio/backend - should I move to be ready for review?

PR #8392, opened 2020-05-04
TEIIDTOOLS-821, 917, 959

blafond:
DV language server and integrated monaco editor in UI

Language server includes
-  initial code-completion items including keywords, functions and snippets
- diagnostics


app/dv/src/main/java/io/syndesis/dv/lsp/parser/statement/FromClause.java
```diff
@@ -0,0 +1,79 @@
...
+        return null;
+    }
+
+    private void processFromTokens() {
+        // Look for SELECT and FROM tokens
+        Token fromToken = findTokenByKind(FROM);
+        if(fromToken != null ) {
+            setFirstTknIndex(getTokenIndex(fromToken));
+        } else {
+            this.analyzer.addException(new DdlAnalyzerException("Parsing Exception:  there is no 'FROM' keyword in your query expression"));
```
shawkins, line 54: It's perfectly valid to use select without from.  People are more familiar with that than using VALUES

Generally it won't define a view, but some customers like to use it for query factoring, such that a common table defines literals:

with constants as (select 1 as val ...)
select ...

app/dv/src/main/java/io/syndesis/dv/lsp/parser/DdlAnalyzerConstants.java
```diff
@@ -0,0 +1,537 @@
...
+     */
+    int[] CREATE_VIEW_STATEMENT = { CREATE, VIEW };
+
+    /*
+     * Array of tokens that match the start of a CREATE VIRTUAL VIEW
+     * statement
+     */
+    int[] CREATE_VIRTUAL_VIEW_STATEMENT = { CREATE, VIRTUAL, VIEW };
+
+    int[] NON_DATATYPE_KEYWORDS = {
```
shawkins, line 143: Can we remove this list and just drive it off of SQLContants?  We add keywords more often than datatypes, so if we we need the datatypes hardcoded for now that's fine.  Also if there is ever consideration for domain types - which can use non-reserved keyword names - then this may need some refinement.

blafond:
Just updated DdlEditor.tsx to include Indra's using memo changes

indraraj:
thanks for incorporating DdlEditor changes.

app/dv/src/main/java/io/syndesis/dv/lsp/parser/statement/FromClause.java
blafond: Due to it's complexity, the query expression completion logic is incomplete and we can work it under [TEIIDTOOLS-992](https://issues.redhat.com/browse/TEIIDTOOLS-992)

The different use case logic for different/simpler forms of the expression can also be added as well to diagnostics as well as completion item framework

app/dv/src/main/java/io/syndesis/dv/lsp/parser/DdlAnalyzerConstants.java
blafond: Since some of the datatypes required SNIPPET definition, it made sense to define them like other completion items (*.json) and lazily load them when needed. If you have an idea on how to generate the completion items from SQLConstants, then log a jira and we'll replace the array.

Note that the "int" value of all of the SQLConstants, defines the token KIND which is used extensively during the completion item collection and diagnostics process via the TokenContext.

shawkins: Please see https://github.com/blafond/syndesis/pull/6 - if there are specific exclusions you want, that could be hard-coded.

Separating the keywords to datatype and non-datatype may not be what you need to do - you will still need datatypes in the completion list for cast/conversion functions - cast, convert, xmlcast, etc.

app/ui-react/packages/api/package.json
```diff
@@ -70,4 +80,4 @@
       }
     }
   }
-}
+}
```
mdrillin, line 83: Why did 'immer' version go back - is it required for monaco?  Same for ts-jest version.  Are the 'patternfly-react' and '@types/patternfly-react' dependencies required here?

app/ui-react/packages/ui/package.json
```diff
@@ -46,21 +46,24 @@
...
+    "monaco-editor-webpack-plugin": "^1.8.2",
     "react": "^16.6.0",
     "react-content-loader": "^4.0.1",
     "react-docgen-typescript-loader": "^3.1.0",
     "react-dom": "16.9.0",
     "react-router": "^5.0.0",
     "react-router-dom": "^5.0.0",
     "rimraf": "^3.0.2",
-    "ts-jest": "^25.2.1",
+    "ts-jest": "^24.0.0",
```
mdrillin, line 58: Was ts-jest going backwards required for monaco editor? are all of the other added dependencies still required after moving some of the code to api?

app/ui-react/packages/api/package.json
```diff
@@ -11,29 +11,34 @@
...
+    "monaco-editor": "^0.19.3",
+    "monaco-editor-webpack-plugin": "^1.8.2",
+    "patternfly-react": "^2.29.2",
     "react": "^16.6.0",
     "react-docgen-typescript-loader": "^3.1.0",
     "react-dom": "^16.6.0",
     "@testing-library/react": "^9.3.2",
     "rimraf": "^3.0.2",
-    "ts-jest": "^25.2.1",
+    "ts-jest": "^24.0.0",
```
gashcrumb, line 33: Is this downgrade needed?

app/ui-react/packages/api/package.json
```diff
@@ -11,29 +11,34 @@
...
     "@types/react": "^16.4.18",
     "@types/react-dom": "^16.0.9",
     "expect": "^24.5.0",
     "jest": "^25.1.0",
     "jest-cli": "^25.1.0",
+    "jest-dom": "^3.1.3",
     "microbundle": "^0.11.0",
+    "monaco-editor": "^0.19.3",
+    "monaco-editor-webpack-plugin": "^1.8.2",
+    "patternfly-react": "^2.29.2",
```
gashcrumb, line 27: This should come out, I'd rather avoid having this dependency as it brings in patternfly 3 which is a step backwards.

app/ui-react/packages/api/package.json
```diff
@@ -11,29 +11,34 @@
   "license": "MIT",
   "devDependencies": {
     "@babel/core": "^7.1.5",
-    "@testing-library/jest-dom": "^4.0.0",
     "@types/expect": "^24.3.0",
     "@types/file-saver": "^2.0.0",
     "@types/jest": "^24.0.11",
+    "@types/patternfly-react": "*",
```
gashcrumb, line 17: This shouldn't be needed?

app/ui-react/packages/api/package.json
```diff
@@ -11,29 +11,34 @@
   "license": "MIT",
   "devDependencies": {
     "@babel/core": "^7.1.5",
-    "@testing-library/jest-dom": "^4.0.0",
```
gashcrumb, line 14: I don't think this should be removed.

app/ui-react/packages/api/package.json
```diff
@@ -52,8 +58,12 @@
     "@syndesis/utils": "*",
     "deepmerge": "^4.2.2",
     "file-saver": "^2.0.1",
-    "immer": "^6.0.3",
-    "react-fast-compare": "^2.0.2"
+    "immer": "^5.0.0",
```
gashcrumb, line 61: Why the downgrade here?

app/ui-react/syndesis/craco.config.js
```diff
@@ -15,13 +24,17 @@ module.exports = function({ env, paths }) {
           ],
         },
       }),
-    },
-    jest: {
-      configure: (jestConfig, { env, paths, resolve, rootDir }) => ({
-        ...jestConfig,
-
-        moduleFileExtensions: ['ts', 'tsx', 'js'],
-      }),
```
gashcrumb, line 24: I don't think we want to remove this block.

app/ui-react/syndesis/package.json
```diff
@@ -11,7 +11,6 @@
   },
   "dependencies": {
     "@allpro/react-router-pause": "^1.1.3",
-    "@craco/craco": "^5.0.2",
```
gashcrumb, line 14: It's a nitpick, but usually all the `@` packages are first.

app/ui-react/syndesis/src/modules/data/locales/data-translations.en.json
```diff
@@ -15,7 +15,6 @@
     "RUNNING": "Running",
     "SUBMITTED": "Submitted"
   },
-  "connectionNotFound": "Connection not found",
```
gashcrumb, line 18: Just to double-check do you intend for these lines to be removed?

app/ui-react/packages/api/package.json
blafond: reverted dependencies and removed patternfly

app/ui-react/packages/ui/package.json
blafond: reverted dependencies and removed patternfly

app/ui-react/packages/api/package.json
blafond: reverted dependencies and removed patternfly

app/ui-react/packages/api/package.json
blafond: restored it

app/ui-react/packages/api/package.json
blafond: restored

app/ui-react/syndesis/craco.config.js
blafond: restored

app/ui-react/syndesis/src/modules/data/locales/data-translations.en.json
blafond: reverted all but cursor related entries

app/ui-react/syndesis/package.json
blafond: moved up in the fil

blafond:
retest this please

rareddy:
@blafond please see if you can submit 1.10.x backport on this

PR #8399, opened 2020-05-05
Fixes ENTESB-13717: Initial version of integration test for SOAP Connector

dhirajsb:


dhirajsb:
I added WS-Security Username Token and PasswordText checks as well. It makes sense to validate the response in the integration via a database entry. I'll add that tomorrow.

dhirajsb:
Having some trouble getting atlastmap to map SOAP response to DB SQL. Will need @igarashitm to look into what's going on.

PR #8402, opened 2020-05-05
BACKPORT fix(ENTESB-13518): Fix upgrading syndesis installed with an external db

phantomjinx:
* configuration.go
     * GetProperties is the essential function called for initialising a
       configuration object. However, ExternalDatabase is crucial for checking
       if the db is external. Instead of calling this separately, include it as
       part of GetProperties
    
* install.go
 * ExternalDatabase function removed as now internal 
 * No need to try and call removePostgresUpgradeTrigger() since this is
       only for internal db


heiko-braun:
Was the backport intended for 1.9, rather than 1.10?

heiko-braun:
I see, it on both. Nevermin

PR #8407, opened 2020-05-06
fix(test): explicit copy of S2I assembly project

### LINKED ISSUE #8403: S2I based integration tests are failing
## This is a...

<pre><code>
[ ] Feature request
[ ] Bug report  
</code></pre>

## Description

Within the [latest changes](https://github.com/syndesisio/syndesis/commit/1b3a19c003576caffa25202543dc92fedf4fa6b7#diff-870d232685fd3bdfd76497ab9f65ff39R42) introduced to make testcontainers work under CI/CD it was not taken in consideration the S2I based integration which are executing the following statement to startup the test:
```
exec java -javaagent:/opt/jolokia/jolokia.jar=config=/opt/jolokia/etc/jolokia.properties -javaagent:/opt/prometheus/jmx_prometheus_javaagent.jar=9779:/opt/prometheus/prometheus-config.yml -XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:+ExitOnOutOfMemoryError -cp . -jar /deployments/integration-runtime.jar
```
We need to apply the same privilege setting also when using this configuration.

squakez:
Added an explicit step to copy S2I assembly phase. As of 1b3a19c003576caffa25202543dc92fedf4fa6b7 we're no longer sharing the filesystem so we need to copy the result of a container to be used by other containers.

Fixes #8403

Locally all tests are passing again:
```
[INFO] Results:
[INFO] 
[INFO] Tests run: 54, Failures: 0, Errors: 0, Skipped: 0
```
Looking forward to see if CI/CD are also working.

### DIFF
--- a/app/test/test-support/src/main/java/io/syndesis/test/container/s2i/S2iProjectBuilder.java
+++ b/app/test/test-support/src/main/java/io/syndesis/test/container/s2i/S2iProjectBuilder.java
@@ -16,6 +16,9 @@
 
 package io.syndesis.test.container.s2i;
 
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Objects;
 import java.util.Optional;
@@ -48,7 +51,19 @@ public class S2iProjectBuilder implements ProjectBuilder {
         SyndesisS2iAssemblyContainer syndesisS2iAssemblyContainer = new SyndesisS2iAssemblyContainer(integrationName, projectDir, imageTag);
         syndesisS2iAssemblyContainer.start();
 
-        Path fatJar = projectDir.resolve("target").resolve("project-0.1-SNAPSHOT.jar");
+        // The S2I assembly container result need to be copied to the local host
+        // to be used by S2I integration containers
+        Path target = projectDir.resolve("target");
+        Path fatJar = target.resolve("project-0.1-SNAPSHOT.jar");
+        try {
+            Files.createDirectories(target);
+            syndesisS2iAssemblyContainer.copyFileFromContainer(
+                "/tmp/src/target/project-0.1-SNAPSHOT.jar",
+                fatJar.toAbsolutePath().toString()
+            );
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
 
         return new Project.Builder()
             .projectPath(projectDir)


app/test/test-support/src/main/java/io/syndesis/test/container/s2i/S2iProjectBuilder.java
```diff
@@ -48,7 +50,21 @@ public Project build(IntegrationSource source) {
         SyndesisS2iAssemblyContainer syndesisS2iAssemblyContainer = new SyndesisS2iAssemblyContainer(integrationName, projectDir, imageTag);
         syndesisS2iAssemblyContainer.start();
 
-        Path fatJar = projectDir.resolve("target").resolve("project-0.1-SNAPSHOT.jar");
+        // The S2I assembly container result need to be copied to the local host
+        // to be used by S2I integration containers
+        Path target = projectDir.resolve("target");
+        Path fatJar = target.resolve("project-0.1-SNAPSHOT.jar");
+        String s2iAssemblyContainerSourcePath = "/tmp/src/target/project-0.1-SNAPSHOT.jar";
```
Delawen, line 57: This is very Linux dependent. Maybe use one of the Files.createTmpDir utilities out there?

(Probably not problematic for our continuous delivery, but bugging to some developers?)

christophd: +1

app/test/test-support/src/main/java/io/syndesis/test/container/s2i/S2iProjectBuilder.java
```diff
@@ -48,7 +50,21 @@ public Project build(IntegrationSource source) {
...
+        String localHostTargetDestination = fatJar.toAbsolutePath().toString();
+        try {
+            Files.createDirectories(target);
+            syndesisS2iAssemblyContainer.copyFileFromContainer(s2iAssemblyContainerSourcePath, localHostTargetDestination);
+        } catch (IOException e) {
+            throw new CannotCopyFromAssemblyContainerException(
+                String.format("Impossible to copy assembly fat jar from container path %s to local host path %s.", s2iAssemblyContainerSourcePath, localHostTargetDestination),
```
christophd, line 64: I would suggest to have the error message as part of `CannotCopyFromAssemblyContainerException` and use container path and host path as constructor arguments there. Otherwise I do not see the benefit of having a dedicated exception class for this very specific case.

zregvart: I'd use `java.io.UncheckedIOException`

app/test/test-support/src/main/java/io/syndesis/test/container/s2i/S2iProjectBuilder.java
zregvart: From what I understand this is the path within the container, doesn't represent the local filesystem path. And it's fixed as Java S2I build runs Maven within `/tmp/src`. Not sure why does this have to be a variable, could be fed directly to `copyFileFromContainer`.

Delawen: Aah, didn't notice it runs inside the container. I just have an automatic alarm everytime I see an absolute path on the code.

Should be fine then.

squakez: > From what I understand this is the path within the container, doesn't represent the local filesystem path. And it's fixed as Java S2I build runs Maven within `/tmp/src`. Not sure why does this have to be a variable, could be fed directly to `copyFileFromContainer`.

As I am using the path again in the exception I prefer not to hard copy the path and rely on a variable instead

app/test/test-support/src/main/java/io/syndesis/test/container/s2i/S2iProjectBuilder.java
squakez: Yes, I do prefer wrapping the exception into a standard one instead, thanks @zregvart.

PR #8421, opened 2020-05-07
Refactor of database upgrade

zregvart:
This moves the database upgrade into the operator upgrade life cycle.
Now the database upgrade will only be attempted if the version of
Syndesis changes and the operator kicks of the upgrade.

The upgrade is performed by creating a Deployment of the PostgreSQL
target image with the `POSTGRESQL_UPGRADE=copy` environment variable.
Presence of this variable triggers the execution of `pg_upgrade` in the
container startup. After the deployment scales up to one replica,
meaning that the database upgrade has completed, as the container is now
ready, the Deployment is removed, shutting down the container.

Upgrade needs to be done after migrations, this is less optimal, the
migrations need a running database, but the database needs to be scaled
down to be upgraded. If the database is not scaled down during upgrade
the database instance spun up for the upgrade can't obtain the lock on
the database and will fail. After upgrade the database can't be scaled
up again as the data files have been migrated to the new version and the
database deployment points to the old version image. Running the old
database version (image) with new data files will fail.

Also fixes an issue in error reporting (6998c85) and takes the opportunity to upgrade to PostgreSQL v10 (6ef2033).

Ref. https://issues.redhat.com/browse/ENTESB-13110


zregvart:
retest this please

zregvart:
@phantomjinx @lgarciaaco I'd like to merge this before we start to conflict, can you have a look?

install/operator/pkg/syndesis/upgrade/database_upgrade.go
```diff
@@ -0,0 +1,414 @@
...
+// variable
+type databaseUpgrade struct {
+	step
+	syndesis *v1beta1.Syndesis
+	target   func() (float64, error) // target version of PostgreSQL as detected at runtime from the file left by the init container
+	current  func() (float64, error) // current version of PostgreSQL as detected at runtime by querying the running database
+	cleanup  func() error            // how to perform cleanup, that is what to do in case of rollback or when we're done with the upgrade
+}
+
+func newDatabaseUpgrade(base step, s *v1beta1.Syndesis) stepRunner {
```
lgarciaaco, line 70: We are implementing pointer semantic for all steps yet this returns a value. Also I think it is clearer if the function return the data type(databaseUpgrade) and not the interface it implements, since it makes it more flexible to use.

install/operator/pkg/syndesis/upgrade/database_upgrade.go
```diff
@@ -0,0 +1,414 @@
...
+	one := int32(1)
+	memory, err := resource.ParseQuantity(config.Syndesis.Components.Database.Resources.Memory)
+	if err != nil {
+		return err
+	}
+
+	metadata := upgradeMetadata.DeepCopy()
+	metadata.SetNamespace(u.namespace)
+
+	return u.client().Create(u.context, &appsv1.Deployment{
```
lgarciaaco, line 212: Isn't better to have this in `pkg/syndesis/generator` and parse it here?

install/operator/pkg/syndesis/upgrade/database_upgrade.go
zregvart: Actually it's other way around, read about encapsulation and Liskov substitution principle.

install/operator/pkg/syndesis/upgrade/database_upgrade.go
zregvart: To say that I dislike everything about `pkg/syndesis/generator` would be an understatement. Using format-fragile templating for data structures that are more succinct and easier to manage in code to an end result of compressed binary blobs that can't be diffed and are often left stale or result in conflicts. I fail to see what benefit does it give us over creating structs in code.

PR #8427, opened 2020-05-08
TEIIDTOOLS-996 updating to latest teiid / teiid spring boot

shawkins:
making the assumptino that we'll want to use the same teiid/tsb version
in syndesis as in the generated image.

with this we can further simplify the version management in the template
pom and the project pom


shawkins:
retest this please

shawkins:
I had asked about using snapshot versions and was given an initial green light, but unfortunately the connector-support build https://circleci.com/gh/syndesisio/syndesis/321410?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link is not setup for that.  So we can either update whatever settings may be in place or just wait on this until we have released versions ready.  Teiid 14.0 should be available by end of week.

PR #8442, opened 2020-05-11
TEIIDTOOLS-986 removing empty role assignments

shawkins:


shawkins:
from the call today, this will remove the role assignment if everything is revoked.

shawkins:
@mdrillin do you want this in 1.10.x as well?

mdrillin:
@shawkins yes, I think it should be backported

PR #8451, opened 2020-05-12
S2I and integration test fixes

zregvart:
Originally motivated to help with backward compatibility with OpenShift 3.11 this was implemented:

refactor(s2i): replace `ADD --chown` in Dockerfile (609208c)

Seems that the Dockerfile build on OpenShift 3.11 fails if the
`ADD --chown` syntax is used, this reverts to using `ADD` + `RUN chown`
instead to be backward compatible.

A race condition in test-support for integration tests was found and fixed in:

fix(test): copy project jar race condition (e1fa833)

The S2I test container is considered started when the output of the
container shows `... done`, that also coincides with the
`/usr/local/s2i/assemble` script finishing and the process stopping,
stopping the container with it. Using
`GenericContainer::copyFileFromContainer` on a stopped container fails
with `IllegalStateException: copyFileFromContainer can only be used when the Container is created`
so the container needs to be running at the point we reach to copy the
files outside of it.

This adds a `&& sleep infinity`, so that after the `assemble` script
finishes the `sh` process in the container remains running so we can
guarantee that `copyFileFromContainer` can be performed afterwards.

The container is stopped when test finishies and/or ryuk garbage
collects spawned containers.

### DIFF
--- a/app/test/test-support/src/main/java/io/syndesis/test/container/s2i/S2iProjectBuilder.java
+++ b/app/test/test-support/src/main/java/io/syndesis/test/container/s2i/S2iProjectBuilder.java
@@ -48,8 +48,9 @@ public class S2iProjectBuilder implements ProjectBuilder {
                 .map(Objects::toString)
                 .orElse("s2i-assembly");
 
-        SyndesisS2iAssemblyContainer syndesisS2iAssemblyContainer = new SyndesisS2iAssemblyContainer(integrationName, projectDir, imageTag);
-        syndesisS2iAssemblyContainer.start();
+        final SyndesisS2iAssemblyContainer s2i = new SyndesisS2iAssemblyContainer(integrationName, projectDir, imageTag);
+        s2i.setCommand("sh", "-c", SyndesisS2iAssemblyContainer.S2I_ASSEMBLE_SCRIPT + " && sleep infinity");
+        s2i.start();
 
         // The S2I assembly container result need to be copied to the local host
         // to be used by S2I integration containers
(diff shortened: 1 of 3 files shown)


zregvart:
@dhirajsb can you perhaps check on your end if this helps on OpenShift 3.11

app/test/test-support/src/main/java/io/syndesis/test/container/s2i/S2iProjectBuilder.java
```diff
@@ -48,16 +48,17 @@ public Project build(IntegrationSource source) {
                 .map(Objects::toString)
                 .orElse("s2i-assembly");
 
-        SyndesisS2iAssemblyContainer syndesisS2iAssemblyContainer = new SyndesisS2iAssemblyContainer(integrationName, projectDir, imageTag);
-        syndesisS2iAssemblyContainer.start();
+        final SyndesisS2iAssemblyContainer s2i = new SyndesisS2iAssemblyContainer(integrationName, projectDir, imageTag);
+        s2i.setCommand("sh", "-c", SyndesisS2iAssemblyContainer.S2I_ASSEMBLE_SCRIPT + " && sleep infinity");
```
squakez, line 52: The container is setting the command already in https://github.com/syndesisio/syndesis/pull/8451/files#diff-870d232685fd3bdfd76497ab9f65ff39R50 I wonder if we can remove that or instead move the `sleep` inside the `CMD` statement. About the `sleep` statement I'm always scary to have something turning potentially in a zombie.

zregvart: This is about changing the command from `/usr/local/s2i/assemble` to `sh -c /usr/local/s2i/assemble && sleep infinity`, I don't want to change this for every use case of this container I do want to change it for this one. As discussed in the [description](https://github.com/syndesisio/syndesis/pull/8451#issue-416957215) the container is stopped by Testcontainers and it's removal is guaranteed by Testcontainers/ryuk.

zregvart:
retest this please

avano:
@zregvart 

```
DeploymentConfig.apps.openshift.io "syndesis-operator" is invalid: spec.template.spec.initContainers[0].image: Required value
```

seems like this may break it: https://github.com/syndesisio/syndesis/pull/8441 @lgarciaaco

zregvart:
@avano seems unrelated to this

PR #8461, opened 2020-05-13
fix(build): don't attempt to delete zero tags

zregvart:
[skip ci]

When we try to delete old (11th and older) snapshot tag we should not
try to delete if there aren't any tags to delete, i.e. when there are
less than 10 snapshot tags in the repository. This adds a check before
`git push --delete origin` to make sure we're not trying to pass an
empty list of tags to that command.


zregvart:
Duplicates #8459, didn't see it :)

PR #8467, opened 2020-05-14
[operator] Prepare the changes for CLOUDBLD-26

lgarciaaco:
**Change env variable names for replacing operand images**

With CLOUDBLD-26 at sight, it is a requirement that operand images are replaceable using environment variables, and such environment variables must follow a name convention (RELATED_IMAGE_*). We have in place the mechanisms to overwrite operand images.

[This commit](https://github.com/syndesisio/syndesis/commit/63a87e04658b4eeea1bbb11016223619466cc23a) changes the name of the environment variables to meet CLOUDBLD-26 requirements.

For sure, some adjustments in QE are needed. << @avano @mcada 

**Reflect required changes from CLOUDBLD-26 when generating bundle image files**

As a requirement from CLOUDBLD-26, the syndesis operator deployment must specify operand images using environment variables.

We don't need the `relatedImage` section in the csv anymore, since this will be generated on the fly based on the RELATED_IMAGE_* env variables.

**Minor changes**
- Remove version from path when generating bundle image files
- Make possible to call `GetProperties()` without connection to an openshift/kubernetes clster


install/operator/pkg/cmd/internal/olm/olm.go
```diff
@@ -50,7 +50,7 @@ func New(parent *internal.Options) *cobra.Command {
 }
 
 func (o *Olm) generate() (err error) {
-	conf, err := configuration.GetProperties(context.TODO(), configuration.TemplateConfig, o.ClientTools(), &v1beta1.Syndesis{})
+	conf, err := configuration.GetProperties(context.TODO(), configuration.TemplateConfig, nil, &v1beta1.Syndesis{})
```
phantomjinx, line 53: For my understanding (sure it makes sense), can you explain why this change?

install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -302,7 +302,9 @@ func GetProperties(ctx context.Context, file string, clientTools *clienttools.Cl
 	}
 
 	if ac, err := capabilities.ApiCapabilities(apiClient); err != nil {
-		return nil, err
+		if apiClient != nil {
+			return nil, err
+		}
```
phantomjinx, line 307: For my understanding (sure it makes sense), can you explain why this change?

install/operator/pkg/cmd/internal/olm/olm.go
lgarciaaco: Because when I run command I dont need to connect to any cluster. This command is run offline.

install/operator/pkg/syndesis/configuration/configuration.go
lgarciaaco: Same as previous comment

avano:
thanks for letting us know @lgarciaaco, if I understand it correctly, we only need to change the env var names right (we are adding it by ourselves to the deployment config) and if the operator is installed from operator hub and has the related images in the csv, then the variables are added automatically to the DC, right?

lgarciaaco:
> thanks for letting us know @lgarciaaco, if I understand it correctly, we only need to change the env var names right (we are adding it by ourselves to the deployment config) and if the operator is installed from operator hub and has the related images in the csv, then the variables are added automatically to the DC, right?

That is right @avano, changing the variable names on your end should be sufficient.

PR #8490, opened 2020-05-18
[ResetDB] Check if syndesis-dv svc is present before trying to reset it

avano:


zregvart:
Thoughts for a potential refactor, not sure if it was a good idea to depend on dv in this part, though if `syndesis-dv` service is not present would not we get a `UnknownHostException` which we could separately handle from the `IOException`.

avano:
It was requested by QE, but it seems there is no particular reason that it the dv stuff reset should be part of syndesis server reset and we can handle it in our test suite.

So there are two options, either to remove the DV reset from here, or catch `UnknownHostException`, I assume you prefer the first one, right? :-)

zregvart:
@avano I think it would be better if the client invoked two endpoints, not sure if there's an issue doing that from tests.

avano:
@zregvart no issue at all, I'll do it in the testsuite first and then I'll edit this PR to remove the dv stuff

zregvart:
@avano also no hurry, we can merge this and work on refactoring later...

zregvart:
Thanks!

PR #8496, opened 2020-05-18
[ENTESB-13527] returning dynamic autogenerated properties

Delawen:
Autogenerated values when getting connector details where not autogenerated.

Example: webhook:

https://-------------.nip.io/api/v1/connections/webhook

```
          "propertyDefinitionSteps": [
            {
              "name": "configuration",
              "properties": {
                "contextPath": {
                  "tags": [
                    "context-path"
                  ],
                  "order": 0,
                  "componentProperty": false,
                  "defaultValue": "Mn3KDtSqI7cFJ76Yec8c5Vg1MzNKXkbosPMBo7TC4rFVxyHqzh",
                  "deprecated": false,
                  "description": "The Webhook token that will be set as final part of the URL",
                  "displayName": "Webhook Token",
                  "generator": "alphanum:50",
                },```

### DIFF
--- a/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectionHandler.java
+++ b/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectionHandler.java
@@ -15,6 +15,13 @@
  */
 package io.syndesis.server.endpoint.v1.handler.connection;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
 import javax.persistence.EntityNotFoundException;
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
@@ -28,22 +35,18 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.SecurityContext;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Date;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.tags.Tag;
 import io.syndesis.common.model.Kind;
 import io.syndesis.common.model.ListResult;
+import io.syndesis.common.model.action.ConnectorAction;
+import io.syndesis.common.model.action.ConnectorDescriptor;
 import io.syndesis.common.model.bulletin.ConnectionBulletinBoard;
 import io.syndesis.common.model.connection.ConfigurationProperty;
 import io.syndesis.common.model.connection.Connection;
 import io.syndesis.common.model.connection.ConnectionOverview;
 import io.syndesis.common.model.connection.Connector;
+import io.syndesis.common.util.RandomValueGenerator;
 import io.syndesis.server.credential.CredentialFlowState;
 import io.syndesis.server.credential.Credentials;
 import io.syndesis.server.dao.manager.DataManager;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/connection/ConnectionHandler.java [hunk 1]
```diff
@@ -133,15 +136,43 @@ public ConnectionOverview get(final String id) {
...
+        enrichWithDynamicProperties(c).ifPresent(builder::connector);
 
         // set the board
         DataManagerSupport.fetchBoard(dataManager, ConnectionBulletinBoard.class, id).ifPresent(builder::board);
 
         return builder.build();
     }
 
+    private static Optional<Connector> enrichWithDynamicProperties(Optional<Connector> conn) {
+        if (conn != null && conn.isPresent()) {
```
zregvart, line 149: Prefer early exit:
```suggestion
        if (conn == null || !conn.isPresent()) {
            return Optional.empty();
        }
```

Delawen: Make sense. Waiting for @gashcrumb to take a look at the UI side and check if this PR is really needed. If it is, I'm moving that return upwards.

gashcrumb: I think this PR is definitely needed, on staging the call to fetch the actions doesn't return the `defaultValue` property filled in, just verified that.  Still getting up to speed as to why it may not show up in the UI.

gashcrumb: In fact, I would propose if this gets the `defaultValue` set properly in the response, let's get it in and we can follow up with an additional PR with a fix to the UI if needed.

gashcrumb: After looking at this further I'm a bit concerned that your fix is being masked by another issue that's crept into the codebase.  I've raised https://github.com/syndesisio/syndesis/issues/8498 to split it out so I stop commenting in your PR :-)

gashcrumb: nope, the problem is right here in front of my face.  [This bit](https://github.com/syndesisio/syndesis/blob/master/app/ui-react/syndesis/src/modules/integrations/components/editor/endpoint/WithConfigurationForm.tsx#L142) is ensuring that for webhook specifically we don't query the BE, because we hit that case, override the definition and then return the cached descriptor vs fetch it from the BE.  @kahboom @KurtStam hey, aren't you guys looking at an issue in this area also?

Delawen: But anyway, we should push this, right?

zregvart: One more thing, check for `x != null` should not be done on a method that receives `Optional`, that's the whole point of `Optional`...

Delawen: I also wonder that, I just don't trust the optional can't be null :D

Removing the null check then.

gashcrumb: Hey, just to follow up, that line I pointed to is also the cause of ENTESB-13527, because normally when that form is shown the UI *always* fetches the descriptor for the action as part of showing the form.  Adding the extra conditional for webhook skipped that.  So yeah, this change isn't needed as the UI currently only uses the API response in this case to list the available actions for the user to pick one.

Delawen: Then closing and not merging :)

Let me know if something else is needed!

PR #8500, opened 2020-05-18
DV permission component stories

mdrillin:
changes related to TEIIDTOOLS-979

- Adds stories for a few of the permission components
- other minor DV story updates
- a few i18n additions
- minor layout adjustments

### DIFF
--- a/app/ui-react/syndesis/src/modules/data/locales/data-translations.en.json
+++ b/app/ui-react/syndesis/src/modules/data/locales/data-translations.en.json
@@ -107,6 +107,8 @@
   "permissionSelectAll": "Select all ({{allListLength}} items)",
   "permissionSet": "Set permissions",
   "permissionSeletedViews": "Grant or Revoke permissions for the selected Views: ",
+  "permissionShowLess": "Show less",
+  "permissionShowMore": "Show more",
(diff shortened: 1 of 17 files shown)


app/ui-react/syndesis/src/modules/data/locales/data-translations.en.json
```diff
@@ -107,6 +107,8 @@
   "permissionSelectAll": "Select all ({{allListLength}} items)",
   "permissionSet": "Set permissions",
   "permissionSeletedViews": "Grant or Revoke permissions for the selected Views: ",
+  "permissionShowLess": "Show less",
```
indraraj, line 110: these can be moved to shared and can remove permission from the key.

indraraj:
not sure why we removed withNotes

mdrillin:
> not sure why we removed withNotes

They were not showing up in storybook.  They were not removed, just changed the implementation which needed updating.

app/ui-react/syndesis/src/modules/data/locales/data-translations.en.json
mdrillin: agree - there are probably others that could be in shared too

Issue #8466, opened 2020-05-14
Connection verifier is not using configuredProperties parameters


squakez:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

I am trying to add a `configuredProperties` into my connector, something like the following:

```
  "componentScheme": "kafka",
  "configuredProperties": {
    "transportProtocol": "PLAINTEXT"
...
  "name": "Debezium CDC over Kafka",
  "properties": {
    "brokers": {
      "componentProperty": true,
....
    }
  },
```
I need to use the `transportProtocol` constant value both in `Verifier` and `Connection`. It seems that the verification POST is not adding those values in their body when called for a verification:

![verifier_properties](https://user-images.githubusercontent.com/23169723/81922666-bb8ac900-95dc-11ea-85c1-de4b8c6552c0.png)

It results in a verification failure when it should not be.

squakez:
There is still a failing scenario. When a property is used as `configuredProperty` in an old version of the connector, and later the same property is used as a user property; in such situation the verifier stick with the old configured property instead of picking the user provided one. We need to validate such scenario as well.

PR #8502, opened 2020-05-19
fix(server): merge existing and new configured properties on server load

squakez:
* Added a merge method to include existing and new configured properties instead of replacing the one coming from previous connector's version
* Added a couple of unit test to validate the merge

Ref #8466

### DIFF
--- a/app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
+++ b/app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -167,9 +168,9 @@ public class DataManager implements DataAccessObjectRegistry {
                         final Connector existing = fetch(Connector.class, id);
                         if (existing != null) {
                             // the only mutable part of the Connector
-                            final Map<String, String> configuredProperties = existing.getConfiguredProperties();
-
-                            connector = connector.builder().configuredProperties(configuredProperties).build();
+                            final Map<String, String> existingConfiguredProperties = existing.getConfiguredProperties();
+                            final Map<String, String> mergedConfiguredProperties = merge(existingConfiguredProperties, connector.getConfiguredProperties());
+                            connector = connector.builder().configuredProperties(mergedConfiguredProperties).build();
                         }
                         store(connector, Connector.class);
                     }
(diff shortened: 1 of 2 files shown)


app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
```diff
@@ -167,9 +168,9 @@ private void loadData() {
                         final Connector existing = fetch(Connector.class, id);
                         if (existing != null) {
                             // the only mutable part of the Connector
-                            final Map<String, String> configuredProperties = existing.getConfiguredProperties();
-
-                            connector = connector.builder().configuredProperties(configuredProperties).build();
+                            final Map<String, String> existingConfiguredProperties = existing.getConfiguredProperties();
+                            final Map<String, String> mergedConfiguredProperties = merge(existingConfiguredProperties, connector.getConfiguredProperties());
+                            connector = connector.builder().configuredProperties(mergedConfiguredProperties).build();
```
zregvart, line 173: I've upgraded Immutables for #8269 and there's a `putAllConfiguredProperties` method on `Builder` classes, not sure if it's there in the current version. Might want to wait for that as it makes the implementation simpler:

```java
connector = connector.builder().putAllConfiguredProperties(existing.getConfiguredProperties())
```

app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
```diff
@@ -182,6 +183,20 @@ private void loadData() {
         }
     }
 
+    private static Map<String, String> merge(Map<String, String> existingProps, Map<String, String> newProps) {
+        Map<String, String> mergedProps = new HashMap<>();
+        mergedProps.putAll(existingProps);
+        // New properties will override any existing property with same name
+        mergedProps.putAll(newProps);
+        LOGGER.info("Merging {} new configured properties with {} existing configured properties: {} merged configured properties",
```
zregvart, line 191: Not sure if this merits an INFO level log

app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
```diff
@@ -182,6 +183,20 @@ private void loadData() {
         }
     }
 
+    private static Map<String, String> merge(Map<String, String> existingProps, Map<String, String> newProps) {
```
zregvart, line 186: There's a `io.syndesis.common.util.CollectionsUtils::aggregate` you can use instead.

app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
squakez: Unfortunately it fails the `shouldOverrideExistingConfiguredProperties` test. Basically it does not override any new configured properties introduced in the new connector version.

app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
squakez: I could lower to `DEBUG`. However I feel like it is something we should notify to a user: in my case, if the info was present, I had not to go deep in code to understand that there was a merge of properties that was the root cause of the failure.

app/server/dao/src/main/java/io/syndesis/server/dao/manager/DataManager.java
squakez: Many tests failing ans it does not allow properties duplication: `java.lang.IllegalStateException: Duplicate key smtp` when calling the `loadData()` method.

PR #8505, opened 2020-05-19
fixes issue with webhook error loading

kahboom:
fixes [ENTESB-13527](https://issues.redhat.com/browse/ENTESB-13527) as well

Screenshot 2020-05-20 17 13 34

### DIFF
--- a/app/ui-react/syndesis/src/modules/integrations/components/editor/endpoint/WithConfigurationForm.tsx
+++ b/app/ui-react/syndesis/src/modules/integrations/components/editor/endpoint/WithConfigurationForm.tsx
@@ -93,18 +93,17 @@ export interface IWithConfigurationFormProps {
 /**
  * A really specific helper function to apply collected error keys to
  * an error mapping form
- * @param action
+ * @param definition
  * @param errorKeys
  */
-function applyErrorKeysToForm(action: Action, errorKeys: ErrorKey[]) {
-  const definition = {
-    ...action.descriptor!.propertyDefinitionSteps![0],
-  } as any;
-  const errorResponseCodes = definition!.properties!
-    .errorResponseCodes as IConfigurationProperty;
+function applyErrorKeysToForm(
+  definition: IConfigurationProperties,
+  errorKeys: ErrorKey[]
+) {
+  const errorResponseCodes = definition.errorResponseCodes as IConfigurationProperty;
   // TODO compatibility, remove when this property is defined
   if (!errorResponseCodes) {
-    return definition.properties;
+    return definition;
   }
   const extProperties =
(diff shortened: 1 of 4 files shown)


app/ui-react/syndesis/src/modules/integrations/components/editor/endpoint/WithConfigurationForm.tsx
```diff
@@ -156,6 +160,11 @@ export const WithConfigurationForm: React.FunctionComponent<IWithConfigurationFo
       </ConfigurationForm>
     );
   }
+
+  const webhookDefinitionOverride = isWebHook
+    ? applyErrorKeysToForm(action, props.errorKeys!)
```
gashcrumb, line 165: So this won't avoid ENTESB-13527 I think, as the `action` here won't have the generated `defaultValue` property.  I'll suggest what you may need to do below.

app/ui-react/syndesis/src/modules/integrations/components/editor/endpoint/WithConfigurationForm.tsx
```diff
@@ -181,7 +190,12 @@ export const WithConfigurationForm: React.FunctionComponent<IWithConfigurationFo
           }
         >
           {() => (
-            <ConfigurationForm action={action} descriptor={data} {...props}>
+            <ConfigurationForm
+              action={action}
+              definitionOverride={webhookDefinitionOverride}
```
gashcrumb, line 195:
```suggestion
              definitionOverride={isWebHook
    ? applyErrorKeysToForm(data, props.errorKeys!) : null}
```

I think would ensure the generated token `defaultValue` for "Context Path" will be supplied to the form.

gashcrumb: Check that `data` there is actually the definition, I might be wrong and this might be an object with the data shapes and `propertyDefinitionSteps` etc, in which case maybe an option is to pass a function down to `ConfigurationForm` that would allow you to do the `isWebHook` check and apply the definition override to the current page of the property definition, i.e. something like

```
definitionCustomizer={(definition) => isWebHook ? applyErrorKeysToForm(definition, props.errorKeys!) : definition}
```

where `definitionCustomizer` could be defined in `ConfigurationForm`'s property interface as `(definition: IConfigurationProperties) => IConfigurationProperties`

gashcrumb: You'll know you have the right behavior when this field has a hash in it:

image

that's the symptom of ENTESB-13527

PR #8516, opened 2020-05-20
Code checks upgrade and cleanup

zregvart:
This started as a support work needed for #8269 and spiralled a bit out of control. Should get us to a point where we can configure the ErrorProne for Java 11 in #8269 and rebase it on top of this. It does require a version of ErrorProne that hasn't been released and that's the reason why I split this up in a separate pull request. More details in the commit messages.


app/common/model/src/main/java/io/syndesis/common/model/integration/step/template/VelocityTemplatePreProcessor.java
```diff
@@ -61,6 +61,7 @@ protected boolean isText(String token) {
     }
 
     @Override
+    @SuppressWarnings("JdkObsolete") // j.u.Matcher accepts only StringBuffer
```
Delawen, line 64: Maybe we should add some issue on github to remind us to check about this at some point?

zregvart: Some of this can't be fixed easily, i.e. when using JDK classes that work with StrinBuffer and not StringBuilder...

PR #8522, opened 2020-05-21
Cleanup code

Delawen:
I run some of the code analysis on IntellijIDEA and cleaned up and improved the code a bit. 

Cleaned up many of the exceptions throws that were not really throwing any exception and changed some expressions using Java 8 to simplify the code and make it less error prone.

There's still many warnings pending but didn't want to do a bigger PR.

No bug fixes, no features, just rewriting code.

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/ModelData.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/ModelData.java
@@ -16,7 +16,6 @@
 package io.syndesis.common.model;
 
 import java.io.IOException;
-
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonRawValue;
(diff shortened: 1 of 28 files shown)


app/common/model/src/main/java/io/syndesis/common/model/support/Equivalencer.java
```diff
@@ -175,10 +172,6 @@ public String failureMessage() {
             return equivalent((StepDescriptor) one, (StepDescriptor) another);
         }
 
-        if (tgtClass.equals(ConnectorAction.class)) {
```
Delawen, line 178: This was duplicated four lines upwards. Was not reachable anyway.

app/common/model/src/main/java/io/syndesis/common/model/ModelData.java
```diff
@@ -73,7 +72,7 @@ public void setDataFromJson(JsonNode json) throws JsonProcessingException {
     public T getData() throws IOException {
         if (data == null && kind != null && json != null) {
             @SuppressWarnings("unchecked")
```
squakez, line 74: maybe we can also get rid of the `@SuppressWarning` now

Delawen: True. I'm working on the next round already, adding that :)

PR #8532, opened 2020-05-22
Fix for Entesb 13816 - The defaultValue should be object not string

KurtStam:
This to avoid quotes being put around the values during the maven-support-plugin validation step.


KurtStam:
retest this please

avano:
retest this please

PR #8572, opened 2020-05-26
fix(ui): [ENTESB-13841] - fix API provider step not working

kahboom:
fixes [ENTESB-13841](https://issues.redhat.com/browse/ENTESB-13841) where the API provider step stopped working.

cc @mcada @mmuzikar:
There is a change to a couple of class names you might want to be aware of. I have taken the `OpenApiSelectMethod` component that was previously being used for the Select Method page of both the API Client Connector & API Provider step (in the Integration Editor) and split it into their own respective components, to accommodate for diverging changes going into the future.

So, basically:
- `open-api-select-method` => `api-connector-creator-select-method` + `api-provider-select-method`
- `open-api-select-method__dnd-container` => `api-connector-creator-select-method__dnd-container` + `api-provider-select-method__dnd-container`
- `open-api-select-method__url-container` => `api-connector-creator-select-method__url-container` + `api-provider-select-method__url-container`

I've left the `data-testid` in tact in both components as `openapi-select-method`. If you'd like for me to update these to reflect the new name of the components (e.g. `api-connector-creator-select-method`, `api-provider-select-method`) then let me know and I'd be happy to. Otherwise it will stay like this. Test IDs for things like `data-testid={'method-scratch-2x'}` and `data-testid={'method-scratch-3x'}` are left the same for both as well.

api-provider-fix


mmuzikar:
Thank you for letting us know about this change. For now I think we can keep the test ids the same and make any changes when they are needed after those components start diverging.

PR #8573, opened 2020-05-26
Fixes ENTESB-13911, added default value of authenticationType and passwordType properties of SOAP connector

dhirajsb:


dhirajsb:
@kahboom any idea why UI module build is failing for this PR? The error output doesn't _speak_ to me.

kahboom:
@dhirajsb I think you need to rebase with the latest changes in master

PR #8580, opened 2020-05-27
Fix for ENTESB-13812 - NPE when integration contains webhook step and an error

KurtStam:
This was caused by changed API in Camel-3


heiko-braun:
Thanks @KurtStam. Can you create the backport as well?

KurtStam:
@heiko-braun this is caused by the upgrade to camel-3. The 1.10.x branch is on camel-2.x, so it should not get back ported.

PR #8597, opened 2020-05-29
fix(server): skip connection properties comparison when dynamic

### LINKED ISSUE #8519: [Connection autodiscovery] Always showing a "connection modified" warning
## This is a...

<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
When we use the connection autodiscovery feature (available only for Kafka connector at the moment), we always have a warning notifying the user that the connection changed:

image

I guess that the warning simply matches the `connection` json configuration stored in the database against the `integration` json configuration enriched with dynamic properties. This is a bit misleading as the connection did not really change.

squakez:
* Added a condition to skip the connection properties comparison if we are marking the connector as dynamic. A dynamic connector will retrieve connection properties and fail the comparison as they meant to be read at runtime.
* Added a unit test to validate the fix.

Fixes #8519

### DIFF
--- a/app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/IntegrationUpdateHandler.java
+++ b/app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/IntegrationUpdateHandler.java
@@ -402,7 +402,7 @@ public class IntegrationUpdateHandler extends AbstractResourceUpdateHandler<Inte
                             dataManager.update(dbConnection);
                         }
 
-                        if (connection.getId().isPresent()) {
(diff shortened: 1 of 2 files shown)


squakez:
@zregvart, @Delawen as you were involved in the feature design would you mind having a look? do you think it is enough to skip the check when the connection details are retrieved dynamically?

squakez:
@syndesisio/backend any feedback please?

Delawen:
So if it is dynamic it will just skip the comparison? Is that a regression of some kind?

Besides that, LGTM.

squakez:
> So if it is dynamic it will just skip the comparison? Is that a regression of some kind?
> 
> Besides that, LGTM.

That's correct, with this PR we'll skip the comparison as it's the intended way the dynamic connection has to work.

Delawen:
Then go for it!

PR #8617, opened 2020-06-01
TEIIDTOOLS-765 adding a lookup for views in source paths

shawkins:


shawkins:
I'm starting with 1.10.x as master doesn't seem to be working - at least for me.  Also it looks like our unit tests aren't running on 1.10.x when you run maven in app/dv.  I suspect this means they aren't running at all as there is at least one test failing for me after all the backport stuff - but we are still building successfully in ci.  Since that is not related to these changes, I'll address that separately.

PR #8618, opened 2020-06-02
fix(dv) reverting backport junit changes to allow unit tests to run

shawkins:


shawkins:
It looks like the junit related pom changes were incomplete without other changes from the parent, so we need to revert those out - otherwise the unit tests aren't running as expected.

PR #8623, opened 2020-06-02
ENTESB-13074: Adjust path of pg_ctl

phantomjinx:
Modified the path as discussed with @avano & tested with upgrade from 1.9.x to 1.10.x (while sampledb integration implemented).


heiko-braun:
Does this work with the base images we use downstream? They might not have the same PATH setup as the upstream ones

avano:
@heiko-braun that was actually the issue, in upstream image there is `pg_ctl` in `/usr/bin`, but in the productized image, there isn't, see [here](https://issues.redhat.com/browse/ENTESB-13074?focusedCommentId=14108498&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14108498)

PR #8634, opened 2020-06-03
fix: TEIIDTOOLS-1040 Fix wizard issue removing tables with same name

mdrillin:
Resolves issue in the CreateView wizard - add/remove tables with same name from different connections was not working properly.

modified the index lookup to include both the table name and connection name.


indraraj:
Retest this please.

PR #8635, opened 2020-06-03
fix: TEIIDTOOLS-1036 Add Publish messages

mdrillin:
Adding tooltip message on the Publish button to better describe 
- what happens upon publish
- why the publish button may be disabled


indraraj:
Retest this please.

PR #8624, opened 2020-06-02
Backport: Cross console navigation

dlabaj:
Backport for PR #8561

PR #8561, opened 2020-05-26
Cross console updates for fuse

dlabaj:
Added code for cross-console updates for fuse.  This feature will be enabled when RHMI server has been updated. 

When RHMI server is not available the product will work exactly like it currently does:

Screen Shot 2020-06-02 at 12 14 05 PM

If the RHMI server is discovered and the minimum number of integration applications are installed in the Open Shift environment the UI will display the cross navigation drop down:

Screen Shot 2020-06-02 at 12 13 27 PM

PR #8639, opened 2020-06-04
chore: adjust java & node heap setting for ui

gashcrumb:
in support of #8561 and #8624 

And because the UI build seems to be generally failing all the time at the moment.


gashcrumb:
@cunningt out of curiosity would node 12.x be an OK LTS version to use?

gashcrumb:
retest this please

PR #8642, opened 2020-06-04
fix(dv) addressing a regression from the refactoring

shawkins:
the last load may be null if the source hasn't loaded


shawkins:
please retest this

shawkins:
retest this please

PR #8648, opened 2020-06-05
ENTESB-13652 deleting build pods on stop

shawkins:


shawkins:
Build pods are not currently cleaned up on a virtualization stop/delete.  The completed pods are garbage collected some time later by openshift.  One rationale for letting completed build pods linger is to get the logs in the event of a failure. 

For this fix I've just added explicit deletion for the build pods.  This prevents the build pod already exists error, but I wasn't able to reproduce anything about logs not existing - which this of course could make worse.

When you are working with a virtualization and don't explicitly stop it, you just keep incrementing the build config and creating more builds / build pods.  While that leaves logs around longer, it seems like there may eventually be a resource issue.

rareddy:
Typically the Build pods are owned by the builds, when Build get deleted they get removed automatically, I am not sure what is happening in this case

shawkins:
> Typically the Build pods are owned by the builds, when Build get deleted they get removed automatically, I am not sure what is happening in this case

That's what I would assume too.  However that is not what is happening - and it may be some kind of change in openshift as this functionality seemed to work fine months ago.  When the build is deleted the completed build pod sticks around just with a null owner.  It will eventually be "garbage collected" by openshift, but if another build is attempted that uses the same name / increment in the meantime we get hit with the pod exists error.

claudio4j:
Can this fix be merged ?

shawkins:
> Can this fix be merged ?

I think so.  @mdrillin do you want to test this first?

mdrillin:
> I think so. @mdrillin do you want to test this first?
 
@shawkins  I can pull it down and test, but if you want to merge go ahead

mdrillin:
retest this please

PR #8656, opened 2020-06-05
fix(ui): SOAP Security fixes

kahboom:
Some fixes for SOAP API client connector, specifically for the security step.

Kapture 2020-06-05 at 19 37 06

- refactoring to pass i18n and configuration props in single object
- add support for multiple password types (text, none, digest)
- additional soap-y stuff (e.g. timestamp, addUsernameTokenNonce)
- initial setup for validation (upcoming PR to improve this)
- fixes item number 4 in [ENTESB-13977](https://issues.redhat.com/browse/ENTESB-13977)
- fix for security stories

missing improvements to potentially abstract SOAP security options and of course, tests.


app/ui-react/packages/ui/src/Customization/apiClientConnectors/create/ApiConnectorCreatorSecurity.tsx
```diff
@@ -1,109 +1,252 @@
...
+export interface IDropdownOption {
   value?: string;
   label?: string;
 }
 
+export interface IDropdownOptions {
+  [key: string]: IDropdownOption[];
+}
+
+export interface II18n {
```
gashcrumb, line 26: I don't think this shortcut is going to be worthwhile, it can introduce NPEs and makes the interface vague between this component and it's consumers.

app/ui-react/packages/models/src/extra.d.ts
```diff
@@ -53,6 +53,34 @@ export interface IApiSummarySoap extends APISummary {
   connectorTemplateId?: string;
 }
 
+/**
+ * Customizable properties in API Client Connector wizard
+ */
+export interface ICreateConnectorProps {
```
gashcrumb, line 59: I'm on the fence about putting this interface here.  Is this specific to a component or API function?  Maybe it should stay alongside the function/component where it's used.  Usually the `extra.d.ts` file is for creating "massaged" API types vs. component property interfaces.

app/ui-react/packages/ui/src/Customization/apiClientConnectors/create/ApiConnectorCreatorSecurity.tsx
```diff
@@ -1,109 +1,252 @@
...
+  [key: string]: string;
```
gashcrumb, line 27: If you do want to use this interface, I'd suggest adding all the keys that are used in the component so that the consumer of this component has type info.

app/ui-react/packages/ui/src/Customization/apiClientConnectors/create/ApiConnectorCreatorSecurityForm.tsx
```diff
@@ -1,92 +1,84 @@
...
-  /**
-   * Used specifically for determining the default type, mostly used
-   * for None and Basic types.
-   */
-  initialAuthenticationType?: string;
-  /**
-   * Authorization URL, required for OAuth 2.0.
-   */
-  initialAuthorizationUrl?: string;
+  defaultValues?: any;
```
gashcrumb, line 40: No chance we could use a concrete interface or [union](https://www.tutorialspoint.com/typescript/typescript_union.htm) here?

app/ui-react/packages/api/src/useApiConnectorSummary.tsx
```diff
@@ -6,8 +6,7 @@ import { callFetch } from './callFetch';
 export function useApiConnectorSummary(
   specification: string,
   connectorTemplateId?: string,
-  portName?: string,
-  serviceName?: string
+  configured?: any
```
gashcrumb, line 9: What's the type for this?  Is it just a dictionary or something?  Be good to use something more helpful than any if possible.

app/ui-react/syndesis/src/modules/apiClientConnectors/pages/create/SecurityPage.tsx
```diff
@@ -29,60 +35,49 @@ export const SecurityPage: React.FunctionComponent = () => {
...
+  const defaultValues = {
+    accessTokenUrl:
+      properties!.tokenEndpoint && properties!.tokenEndpoint.defaultValue,
+    authenticationType: properties!.authenticationType.defaultValue,
+    authorizationUrl:
+      properties!.authorizationEndpoint &&
+      properties!.authorizationEndpoint.defaultValue,
+    passwordType:
+      properties!.passwordType && properties!.passwordType.defaultValue,
+  };
```
gashcrumb, line 52: I think you can use [optional chaining](https://www.valentinog.com/blog/chaining/) to clean this up considerably.

app/ui-react/syndesis/src/modules/apiClientConnectors/pages/create/SecurityPage.tsx
```diff
@@ -29,60 +35,49 @@ export const SecurityPage: React.FunctionComponent = () => {
...
+
+  const dropdowns = {
+    authenticationTypes:
+      properties!.authenticationType &&
+      (properties!.authenticationType.enum || []).sort((a, b) =>
+        a.value!.localeCompare(b.value!)
+      ),
+    passwordTypes:
+      properties!.passwordType &&
+      (properties!.passwordType.enum || []).sort((a, b) =>
```
gashcrumb, line 62: And this stuff too

kahboom:
Thanks @gashcrumb , all good suggestions. I can squash all of these commits into one and address your concerns in a second commit to keep things a bit cleaner, if that's okay with you.

app/ui-react/packages/models/src/extra.d.ts
kahboom: It's currently used in `@syndesis SecurityPage,  DetailsPage, ReviewActionsPage` and `@api useApiConnectorCreator`. There was need for a very similar interface in the `@ui` package as well, but we can't use `@models` in there, so I created `ICreateConnectorPropsUi`. I was trying to avoid was duplicating code that is indirectly based on an existing model, but I'm happy to provide different interfaces for the api, syndesis, and UI packages if you think that's the best thing, and keeping this interface where it was. Let me know what you think?

app/ui-react/packages/api/src/useApiConnectorSummary.tsx
kahboom: It's quite similar to the interface we were discussing [here](https://github.com/syndesisio/syndesis/pull/8656#discussion_r437563722), actually not sure why I didn't use that one depending on what we decide to do with it.

app/ui-react/packages/models/src/extra.d.ts
gashcrumb: Yeah, I'd keep separate models for the UI component, even if you create a `models.ts` if it's shared by a few files.

PR #8659, opened 2020-06-05
fix(ui): [ENTESB-13977] - list all api client connectors

kahboom:
fix for second item in [ENTESB-13977](https://issues.redhat.com/browse/ENTESB-13977), to list all API client connectors.

cc @dhirajsb 

depends on https://github.com/syndesisio/syndesis/pull/8650


dhirajsb:
I've added this commit to #8663 so closing this duplicate PR.

PR #8660, opened 2020-06-05
Add the loading component

christiemolloy:
Adds the loading component to Fuse.

@dlabaj , I have this manually set for 2 seconds like discussed in the meeting. Should the spinner always show now? Maybe I could add a conditional for if the cross-console is added then show the spinner.

Also, currently it is not pulling in the styles set on the LoadingPage component, but I added in the core package.

These are the styles that are missing but are set in `core.css`
```
.app__loading-container {
  background-color: var(--pf-global--BackgroundColor--200);
  position: absolute;
  top: 0;
  right: 0;
  bottom: 0;
  left: 0;
  z-index: 700 !important;
}
```


claudio4j:
@christiemolloy there are some conflicts, probably occurred after your PR, so can you rebase this PR ?

gashcrumb:
Hmm, looks like this introduced some test failures, do they run locally?  Can use `syndesis ui --run-tests` to check.

christiemolloy:
closing this PR in favor of #8766

Issue #8704, opened 2020-06-16
[Operator] Installing with --dev does not use s2i:latest image


squakez:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
I've run a fresh installation from the latest `master` operator and altough using the `--dev`, the `s2i` image configured is not the latest as it should be:
```
...
openshift:
  apiBaseUrl: 'https://localhost:8443/oapi/v1'
  namespace: 'syndesis'
  imageStreamNamespace: 'syndesis'
  builderImageStreamTag: syndesis-s2i:2.0.0-20200608
...
```
As a workaround I am changing the value to `syndesis-s2i:latest` but I'd expect this to be in the configuration.

@claudio4j @lgarciaaco @phantomjinx do you mind having a look?

phantomjinx:
How about this?

https://github.com/syndesisio/syndesis/pull/8706

PR #8706, opened 2020-06-16
#8704: Uses 'latest' builder image stream if running with dev

phantomjinx:


claudio4j:
@phantomjinx I think you should add `assets_vfsdata.go`

PR #8717, opened 2020-06-17
move + rename api connector wizard-specific components to create dir, combine stories

kahboom:
depends on https://github.com/syndesisio/syndesis/pull/8701 (because that PR contains a fix for storybook)

This PR is to prepare for an upcoming PR (and to make it easier to review) which will involve splitting the Step 4 (Detail/Review) components of Create API Client Connector page and the components from the API Connector Detail page.

The purpose of this task is to make reusing these components easier, for example, in a future version of Syndesis that might involve a lot of change.

Ultimately, the existing components will be used strictly for the editor, and in a subsequent PR I'll be creating components specific to the Detail page.

Here I have moved all of those existing components to a `create` directory used for the wizard, and combined all the existing API Connector Detail stories that use those components. I've also renamed them to be easier to understand, e.g. `ApiConnectorDetailsForm` -> `ApiConnectorCreatorDetails` which matches the naming structure of the other editor-specific components.

Side note: I moved the `icon` strings from the stories into a separate file to make working with stories easier.


app/ui-react/packages/ui/src/Customization/apiClientConnectors/create/securityValidation.ts
```diff
@@ -0,0 +1,27 @@
...
+export default function validateSecurity(values: ICreateConnectorPropsUi) {
+  const errors: IErrorValidation = {};
+
+  if (
+    values.authenticationType === 'basic' ||
+    values.authenticationType === 'ws-security-ut'
+  ) {
+    if (values.passwordType !== 'PasswordNone') {
+      if (!values.username) {
+        errors.username = 'Username is required';
```
gashcrumb, line 17: How would you localize this string, is it shown to the user?

kahboom: Good point. Yes, this does get shown to the user. Actually it's not such a large file that I can't just put it in the existing component, or I can create a separate component for it entirely?

gashcrumb:
retest this please

PR #8761, opened 2020-06-25
TEIIDTOOLS-1059 Modify DV roles logic for empty privileges

mdrillin:
Modified the roles service - if an empty set of privileges is provided, the role is removed.  Updated the IntegrationRolesTest for the scenario.


mdrillin:
retest this please

PR #8763, opened 2020-06-26
initial lsp quick fix framework

blafond:
Initial dv lsp quick fix support - TEIIDTOOLS-970


blafond:
This PR is just to get something in for quick fixes. The process involves capturing the incremental errors during parsing, constructing a diagnostics object including an error code which can be used to return applicable quick fixes. 

Both the  table body and query expression clauses are missing key parsing, diagnostic and quick fix support.

Targeting the **query-primary** structure as the next step for query expression support is good and I'll log that specifically.

KurtStam:
@dependabot rebase

KurtStam:
@blafond Can you also open a GH issue for this, so it's easier to understand what's going on? It's a rather large PR..

claudio4j:
Closing, as DV was removed from syndesis.

PR #8766, opened 2020-06-26
feat(loadingpage): add loading page

christiemolloy:
Adds the loading spinner before any content is loaded on the page.

Test by opening your console and changing "Network" to "Slow 3G" to test the order of components that are loaded.


app/ui-react/syndesis/src/app/App.tsx
```diff
@@ -58,6 +58,11 @@ export const App: React.FunctionComponent<IAppBaseProps> = ({
     }
   };
 
+  const spinner = document.getElementById('loading-page-spinner');
```
riccardo-forina, line 61: This block would run at every render, you can run it once like this

```
React.useEffect(() => {
  const spinner = document.getElementById('loading-page-spinner')!; //  ! because we know the element will be there
  spinner.setAttribute('hidden', 'true');
}, []);
```

christiemolloy: Great point, thanks for the suggestion -- updated

PR #8785, opened 2020-07-03
fix: Downloading platform-detect correctly

phantomjinx:


Delawen:
Syndesis CI still failing...

PR #8791, opened 2020-07-05
Fix: More changes to get the release build working

phantomjinx:
* pom.xml
 * Detected unused defined dependencies

* .lib.sh
 * platform-detect must be built with the docker option as well as go

* openshift_funcs
 * Syntax error on checking for command

### DIFF
--- a/app/server/cli/pom.xml
+++ b/app/server/cli/pom.xml
@@ -65,11 +65,6 @@
     </dependency>
 
     <dependency>
-      <groupId>io.syndesis.common</groupId>
-      <artifactId>common-model</artifactId>
-    </dependency>
-
-    <dependency>
       <groupId>io.syndesis.server</groupId>
       <artifactId>server-runtime</artifactId>
       <classifier>classes</classifier>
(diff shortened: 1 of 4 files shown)


app/server/cli/pom.xml
```diff
@@ -64,11 +64,6 @@
       <artifactId>common-util</artifactId>
     </dependency>
 
-    <dependency>
-      <groupId>io.syndesis.common</groupId>
```
Delawen, line 68: This needed to be removed?

Delawen:
Maybe you left some dependency out of the push?

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project server-monitoring: Compilation failure

[ERROR] /workspace/app/server/monitoring/src/main/java/io/syndesis/server/monitoring/PublishingStateMonitor.java:[39,36] package io.syndesis.server.openshift does not exist

[ERROR] -> [Help 1]
```

PR #8796, opened 2020-07-06
[WiP] Integration tests fix

Delawen:


app/test/integration-test/src/test/java/io/syndesis/test/itest/mail/SendMail_IT.java
```diff
@@ -129,7 +129,7 @@ public void testSendMailError(@CitrusResource TestCaseRunner runner) {
 
         runner.variable("first_name", "Joanne");
         runner.variable("company", "Red Hat");
-        runner.variable("email", "joanne@syndesis.org");
+        runner.variable("email", "joan");
```
christophd, line 132: do we know where that change came from in Syndesis implementation?

Delawen: This is still WiP partly because there are errors I don't know where they come from.

Most of the errors fixed are just reimporting the model.json, redoing the data mapping and exporting again the model.json. I can reformat to diff, sure.

christophd: I had a look at this particular SendMail_IT test and the problem here is that the error coming from the mail server is not propagated to the webhook as a response. 

The test simulates an error when the integration is sending the mail. So the integration should respond with a Http 500. The error is ignored and 200 OK is returned as Webhook response instead.

I think we should revert the change made in the test here and open a separate issue. WDYT?

Delawen: Sure, go for it!

My guess is that half of this errors comes from data mappings broken on new code and the other half comes from the api behaving differently than before (missing configuration parameters?).

Feel free to redo all this if you find a better way.

christophd: in case the data mappings are broken because of code changes we made we need to add migration steps for auto fixing those data mappings in existing integrations. Otherwise we may cause a lot of frustration for users.

Can you say something about what in particular is broken in the data mappings?

christophd: same for missing configuration parameters. we need to have auto migration steps for these.

Delawen: Not really... I was in the middle of investigating it. But as Kurt took another approach I decided to wait and see what happens.

christophd:
I think we should merge all 204 vs. 200 fixes right now (maybe as a separate PR?)

Delawen:
> I think we should merge all 204 vs. 200 fixes right now (maybe as a separate PR?)

I can separate it. But do we know why that happened? The default api behaviour changed?

christophd:
@Delawen yes, the 200 is the new default for Webhook connector responses. I have discussed this with @KurtStam and we felt that the integration tests just need an update here. Sorry, should have mentioned that.

Delawen:
Ok, Then cherry-picking!

Delawen:
#8815

KurtStam:
@Delawen going to close this since the tests are fixed now.

PR #8803, opened 2020-07-08
fix(8801): address availability of consolev1 on all platforms

phantomjinx:
* consolev1 api is part of openshift but not available on minikube or
  minishift. Therefore, need to caveat the use of consolev1.consolelink
  to ensure it does not hurt those other platforms.

* run.go
 * Moves addition of scheme to where other schemes are installed

* syndesis_controller.go
 * removeConsoleLink should only be performed if the platform supports the
   api in the first place

* grant_olm_cluster_role.yml.tmpl
* olm_cluster_role.yml.tmpl
 * Adds the extra cluster role and binding that allows permissions on the
   consolelink.console.openshift.io api permissions
 * Caveats with ApiServer.ConsoleLink, ie. api being present on the cluster

* capabilities
 * Checks for consolelinks.console.openshift.io api and sets flag if present

* configuration
 * SetConsoleLink is only performed if ConsoleLink is true


Delawen:
The build failed. Re-running it.

PR #8810, opened 2020-07-14
chore: Fix build CLI on MacOS

christophd:
Check for existing ggrep binary was not working properly on MacOS. 

The command hit the another error before actually doing the check:

`common_funcs: line 415: OK: command not found`

Fixed check to work on MacOS


christophd:
@phantomjinx Do you happen to know why we explicitly need ggrep on MacOS for the build?

christophd:
@phantomjinx can you please take a look?

Issue #8811, opened 2020-07-14
Errors in mail connector are not propagated anymore


christophd:
## This is a...

<pre><code>
[ ] Feature request
[x] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

This can be seen as part of #8776 for fixing the `SendMail_IT` integration test in particular.

The integration test is failing because the simulated error on the mail server is not propagated anymore. The Syndesis integration is triggered with a Webhook request and sends a mail message as part of the integration logic. The mail server simulates an error and this should be reflected in a Http 500 in the Webhook response to the calling client.

At the moment the mail server error is ignored and the Webhoook returns a 200 OK status code. This is why the test fails.

PR #8815, opened 2020-07-17
fix(integration-tests): Fixing tests so they pass. HTTP changed from 204 to 200.

Delawen:
Updating integration tests to match new default.


app/test/integration-test/src/test/java/io/syndesis/test/itest/mail/SendMail_IT.java
```diff
@@ -129,7 +129,7 @@ public void testSendMailError(@CitrusResource TestCaseRunner runner) {
 
         runner.variable("first_name", "Joanne");
         runner.variable("company", "Red Hat");
-        runner.variable("email", "joanne@syndesis.org");
+        runner.variable("email", "joan");
```
christophd, line 132: Can you please revert this particular change. It should be addressed and properly fixed in #8811

PR #8825, opened 2020-07-21
Bump camel base to version 3.3

### LINKED ISSUE #8764: Bump camel base version to 3.3
## This is a...

<pre><code>
[x] Feature request
[ ] Bug report  
</code></pre>

## Description

Required to target camel k 1.3 runtime.

squakez:
Major challenges are related to the requirement since camel version 3.2 to [include a contextPath into each endpoint URI](https://camel.apache.org/manual/latest/camel-3x-upgrade-guide-3_2.html#_endpoint_uris_without_context_path). This has been addressed in f169b8b, together with the change of component initialization into `init()` component proxy instead of `start()`.

Closes #8764, closes ENTESB-14319


app/common/model/src/main/java/io/syndesis/common/model/integration/step/template/AbstractTemplatePreProcessor.java
```diff
@@ -231,7 +231,11 @@ public void reset() {
 
     @Override
     public Map<String, Object> getUriParams() {
-        return Collections.emptyMap();
+        // Since Camel 3.3 we must declare explicitly the usage of templates
+        // that are stored in headers
+        Map<String, Object> defaultMap = new HashMap<>();
+        defaultMap.put("allowTemplateFromHeader","true");
+        return defaultMap;
```
lburgazzoli, line 238: This may be replaced by `Collections.singletonMap("allowTemplateFromHeader","true")` ?

squakez: It was my first choice, but that would return an immutable map and the returned map will be updated by the caller.

lburgazzoli: wonder if those parameter should be set at construction time since it seems they are constants

squakez: I see. What if we have a static singleton map and we return in that method a new collection filled with the singleton?

lburgazzoli: up to you

heiko-braun:
Good work @squakez !

Delawen:
As long as the integration tests that fail are the same as in master, I would say we can merge this.

app/common/model/src/main/java/io/syndesis/common/model/integration/step/template/AbstractTemplatePreProcessor.java
```diff
@@ -231,7 +236,7 @@ public void reset() {
 
     @Override
     public Map<String, Object> getUriParams() {
-        return Collections.emptyMap();
+        return new HashMap<>(HEADER_TEMPLATE_OPTIONS);
```
lburgazzoli, line 239: Wonder if we can change the constructor of AbstractTemplatePreProcessor to accept a map, so specific implementation can provide their own uri params when the instance is created ora have a protected `Map<String, Object> uriParams` that is amended by the subclasses so then we can have something like:

```java
public Map<String, Object> getUriParams() {
    return Collections.unmodifiableMap(uriParams);
}
```

claudio4j:
@squakez there are some additional files in install/operator to change Camel Version
```
install/operator/build/conf/config-test.yaml
install/operator/pkg/syndesis/configuration/configuration_test.go
install/operator/pkg/generator/assets/addons/camelk/camel-catalog-3.1.0.yaml.tmpl
```
Then regenerate `install/operator/pkg/generator/assets_vfsdata.go` by 
```
cd install/operator
GOOSLIST=linux ./build.sh --operator-build go
```

squakez:
> @squakez there are some additional files in install/operator to change Camel Version
> 
> ```
> install/operator/build/conf/config.yaml
> install/operator/build/conf/config-test.yaml
> install/operator/pkg/syndesis/configuration/configuration_test.go
> install/operator/pkg/generator/assets/addons/camelk/camel-catalog-3.1.0.yaml.tmpl
> ```
> 
> Then regenerate `install/operator/pkg/generator/assets_vfsdata.go` by
> 
> ```
> cd install/operator
> GOOSLIST=linux ./build.sh --operator-build go
> ```
> 
> The operator build will generate additional files in vendor directory, they are part of the rebuild and should be pushed too.

Thanks @claudio4j very good point!

app/common/model/src/main/java/io/syndesis/common/model/integration/step/template/AbstractTemplatePreProcessor.java
squakez: I think the problem is caused by MustacheTemplatePreProcessor. I'm bringing back the abstract method to return an final map and then do the same on the Mustache subclass, so, all returned maps will be final.

squakez:
Please, @claudio4j I need your help to confirm the changes applied to the operator are good. I am not entirely sure about the introduction of the new camel-catalog-3.3.0 file. I took the content of the generated file in the runtime integration and created a new file with the templated version took from camel-catalog.3.1.0 previously there - is that good!?

claudio4j:
I am unsure too about the content of `camel-catalog-3.3.0.yaml.tmpl`. @lburgazzoli can you help review its content ?
I see there is a property `camel.version: 3.0.1` in `spec.runtime.metadata`, shouldn't this be 3.3.0 ?

lburgazzoli:
> I am unsure too about the content of `camel-catalog-3.3.0.yaml.tmpl`. @lburgazzoli can you help review its content ?
> I see there is a property `camel.version: 3.0.1` in `spec.runtime.metadata`, shouldn't this be 3.3.0 ?

this is an auto generated file that re-structure the camel-catalog to be usable by camel-k and yes it should be 3.3.0

squakez:
I've picked the file generated by runtime-camelk-integration as specified in the 3.1.0 catalog:

[camel-catalog-3.3.0.yaml.txt](https://github.com/syndesisio/syndesis/files/4954724/camel-catalog-3.3.0.yaml.txt)

Issue #8882, opened 2020-08-03
[Camel K] Integration health checks


squakez:
Verify how `camel k` integrations are exposing their health checks.

squakez:
cc @christophd

squakez:
The health checks are managed by adding an health customizer through application properties:
```
customizer.health.enabled=true
customizer.health.path=/actuator/health
```
the [logic executed](https://github.com/apache/camel-k-runtime/blob/master/camel-k-main/camel-k-runtime-health/src/main/java/org/apache/camel/k/health/HealthContextCustomizer.java) will provide a new route to the Vert.x http server launched by camel k runtime. Therefore, we must carefully watch out the presence of any additional `platform-http` configuration that could be present in the same integration, ie:
```
customizer.platform-http.path=/webhook/
customizer.platform-http.enabled=true
```
That will result in having the actuator endpoint exposed in `/webhook/actuator/health` as one can expect.

Issue #8881, opened 2020-08-03
[Camel K] Integration exposing services


squakez:
There may be some possible issue for an integration exposing certain services (ie, webhooks, api-provider).

squakez:
cc @christophd

squakez:
Replaced by #8951 and #8952

PR #8884, opened 2020-08-03
fix(#8855): Enable Camel K runtime for integration tests

### LINKED ISSUE #8855: [Camel K] support runtime in integration tests
By default the automated integration tests use the S2i integration runtime in combination with Spring Boot when verifying integrations. We should make it possible to also use the Camel K integration runtime for these tests in order to have a test automation for the Syndesis Camel K runtime.

There has already been some steps towards choosing the integration runtime. Users can set the integration runtime via System property or environment setting:

```
-Dsyndesis.integration.runtime=spring-boot
SYNDESIS_INTEGRATION_RUNTIME=spring-boot
```

If you want to switch to Camel K runtime you just need to set `camel-k` as value. This will then use the Camel K project builder to generate the integration resources and run the integrations: https://github.com/syndesisio/syndesis/blob/master/app/test/test-support/src/main/java/io/syndesis/test/integration/project/CamelKProjectBuilder.java

The Camel K project builder tries to get all resources in place for the runtime (Maven POM, application.properties, customizers, etc.). 

Users can enable various Camel K customizers via System property or environment setting

```
-Dsyndesis.camel.k.customizers=health,logging,syndesis
```

The value is a comma separated list of customizer names that will be automatically enabled for the integration test runtime.

The project builder in its current state is not completed yet and must be verified with the latest Camel K version used in Syndesis. All tests failing with Camel K runtime enabled at the moment.

christophd:
Fixes #8855 

Integration tests do not use the Camel K operator but instead need to prepare a proper S2i image that can be used in the pure Docker infrastructure based on Testcontainers. 

This PR creates a new S2i image that is based on the traditional Syndesis Spring Boot S2i image (syndesis:syndesis-s2i). The new Camel K S2i image (syndesis:syndesis-s2i-camelk) adds Camel K related artifacts that get baked into the image. Users can switch the runtime (spring-boot vs. camel-k) via command line option when running the integration tests.

The PR fixes the basic Camel K runtime for the integration tests so that all tests are able to start the integration runtime container with Camel K. There are some tests still failing though because of misconfigured exposed services such as webhooks and api-provider. These problems are addressed by separate follow-up issues (see #8881 #8882).

The approach in this PR has a downside as we need to use JDK 11 in Syndesis S2i base image. This is because Camel K runtime executes with Java 11 only.

The Camel K integration runtime was based on an older version of Camel K. This PR fixes several configurations and properties.

Runtime should be using
- org.apache.camel.k/camel-k-runtime-main (instead of camel-k-runtime-jvm)
- org.apache.camel.k/camel-k-runtime-http (instead of camel-k-runtime-servlet)
- customizer.platform-http.enabled property
- customizer.platform-http.bind-port property
- customizer.health.enabled (not enabled by default, but only when a service is exposed)
- customizer.health.path

Obsolete properties
- customizer.servlet.enabled
- customizer.servletregistration.enabled


squakez:
I'd move this commit as a separate PR, outside the scope of camel k: 1ebd060d4b3160a4d4eecca3112eba56489b9616

app/integration/project-generator/src/test/resources/io/syndesis/integration/project/generator/testGenerateProject/integration.json
```diff
@@ -53,7 +53,8 @@
               "kind": "ENDPOINT",
               "entrypoint": "direct:extension"
             },
-            "actionType": "step"
+            "actionType": "step",
+            "tags":["expose"]
```
squakez, line 57: What this `expose` will do for the integration?

app/s2i/src/main/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fabric8/s2i-java:3.0-java8
+FROM fabric8/s2i-java:3.1-java11
```
squakez, line 1: This implies that also standard S2I (springboot) runtime will change to java 11. We could either have a different base image for camel-k runtime, but I guess we don't want to maintain two different base images.

app/test/integration-test/README.md
```diff
@@ -83,36 +85,45 @@ The following system properties (or environment variables) are known to the proj
...
-mvn verify -Dit.test=MyTestClassName
+tools/bin/syndesis integration-test -t MyTestClassName
 ```
 
 ```bash
-mvn verify -Dit.test=MyTestClassName#mytestMethodName
+tools/bin/syndesis integration-test -t MyTestClassName#mytestMethodName
 ```
 
-Please, notice that the integration test may have been disabled by default through the setting of the variable `skip.integration.tests` to `true`. If that is the case, ensure to override the value to `false` instead:
```
squakez, line 104: I undestand that property will be implicit in the `syndesis integration-test` command execution, right?

app/integration/project-generator/src/test/resources/io/syndesis/integration/project/generator/testGenerateProject/integration.json
christophd: this marks that the integration exposes a service. only when integrations expose some service the project generator will add service related properties (such as camel.component.servlet.header-filter-strategy) to the application.properties. Camel K runtime is more restrictive compared to Spring Boot runtime when it comes to having properties in application.properties that do not match the components used in the integration.

app/s2i/src/main/docker/Dockerfile
christophd: yes the image has JDK11 then also for Spring Boot runtime, but the Maven POM that runs the integration explicitly sets the source and target Java version to 1.8.

app/test/integration-test/README.md
christophd: Yep

PR #8895, opened 2020-08-05
Dvremoved

vhalbert:


install/operator/pkg/syndesis/versions/versions_test.go
```diff
@@ -35,34 +35,34 @@ import (
...
+// 			if (err != nil) != tt.wantErr {
+// 				t.Errorf("unstructuredToV1Alpha1() error = %v, wantErr %v", err, tt.wantErr)
+// 				return
+// 			}
+// 			if !reflect.DeepEqual(gotS, tt.wantS) {
+// 				t.Errorf("unstructuredToV1Alpha1() gotS = %v, want %v", gotS, tt.wantS)
+// 			}
+// 		})
+// 	}
+// }
```
phantomjinx, line 65: Why is this test commented out?

vhalbert: Ah, that's a carry over from 1.11.x change, once all this finally builds, can work to get that turned back on

PR #8913, opened 2020-08-11
fix(build): Allows pushing of images to quay.io

phantomjinx:
* commands/build
* build.sh
* .lib.sh
 * Removes operator-image-name parameter as not necessary
 * Adds image-namespace parameter so can be modified, eg. set namespace as
   user-id -> quay.io/phantomjinx/syndesis-operator
 * Splits image-namespace from image-name so former can be modified
 * Full operator image name: registry/namespace/image-name
 * Passes the image.tag into maven builds so correct tag is built and pushed

* app/pom.xml
 * Modifies image-name properties to expose both namespace & tag

* s2i/pom.xml
 * Extra config needed to ensure fabric8 pushes the correct named image

### DIFF
--- a/app/test/s2i-camelk/pom.xml
+++ b/app/test/s2i-camelk/pom.xml
@@ -356,7 +356,52 @@
               </execution>
             </executions>
           </plugin>
+        </plugins>
+      </build>
+    </profile>
 
+    <!-- Profile for pushing a compiled image to a registry defined by the system property docker.push.registry -->
+    <profile>
+      <id>image-push</id>
+      <activation>
+        <property>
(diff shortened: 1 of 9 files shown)


app/test/s2i-camelk/pom.xml
```diff
@@ -356,7 +356,52 @@
               </execution>
             </executions>
           </plugin>
+        </plugins>
+      </build>
+    </profile>
 
+    <!-- Profile for pushing a compiled image to a registry defined by the system property docker.push.registry -->
```
squakez, line 363: Do we need to push also the test image?

PR #8919, opened 2020-08-11
imp error message

lina2358:
Illegal Argument on Call Given parameter is of unknown type/format: array/null
vs
Illegal Argument on Call: Given parameter is of unknown type/format: array/null

one might think "call-given-parameter"


claudio4j:
Thank you for the patch.

PR #8922, opened 2020-08-12
fix: use productized db image for initPod

lgarciaaco:
When we build in fuse-online we dont generate the assets, therefore this file needs to contains the final db productized image. Source: https://github.com/jboss-fuse/fuse-online-operator/blob/private-devel-fuse-7.7-openshift-rhel-7/Dockerfile#L14

Remove padding

Fixes https://issues.redhat.com/browse/ENTESB-14436?filter=12341128

### DIFF
--- a/install/operator/pkg/syndesis/action/install.go
+++ b/install/operator/pkg/syndesis/action/install.go
@@ -195,12 +195,10 @@ func (a *installAction) Execute(ctx context.Context, syndesis *v1beta1.Syndesis)
 			continue
 		}
 
-		a.log.Info("Installing addon", "Name", addonInfo.Name())
-
 		if config.ApiServer.OlmSupport && addonInfo.GetOlmSpec() != nil && addonInfo.GetOlmSpec().Package != "" {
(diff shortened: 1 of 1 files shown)


claudio4j:
I think the productized postgresql should be in the internal productized 1.11.x branch.

lgarciaaco:
A bit more of information. That docker image used for the init pod, is set in the generated resource install_defaults.go in a constant, and it is set from the value found in the configuration config.yaml. This not only goes against how we do things in general (those values that can change are either set in the configuration file OR we use a constant if the scenario where that variable is used takes place before the configuration file is read) but it is a bad practice because we dont have a way to overwrite it for production other than changing the code. As you know we don't generate resources files when we build production and that's why this value is never changed there

Any other solution would need a refactor of how this value is read .. and I am not comfortable doing big changes for a bugfix at this stage.

lgarciaaco:
Backporting https://github.com/syndesisio/syndesis/pull/8929 also in this PR

lgarciaaco:
Replaced by https://github.com/jboss-fuse/syndesis/pull/43

PR #8923, opened 2020-08-12
fix: use productized db image for initPod

lgarciaaco:
When we build in fuseonline we dont generate the assets, therefore this file needs to contains the final db productized image. Source: https://github.com/jboss-fuse/fuse-online-operator/blob/private-devel-fuse-7.7-openshift-rhel-7/Dockerfile#L14


lgarciaaco:
Related with #8922

lgarciaaco:
Replaced by https://github.com/jboss-fuse/syndesis/pull/43

PR #8924, opened 2020-08-13
fix(ENTESB-13779): Add odata-v2 connector

christophd:
Adds connector for OData services in version 2.0 with create, read, update, delete operations as well as connector metadata retrieval.

The odata v2 connector is basically derived from the odata v4 connector with following adjustments:
- Using Odata V2 EdmEntity model
- Using MERGE operation instead of PATCH (Http patch is fully supported from v3 onwards, v2 uses custom Http merge operation)
- Not able to use client entity markers e.g. ('Address') when formatting key predicates
- Preserving String literal quote marks in key predicates even if predicate is a number (e.g. '1')
- Using OData V2 sample tutorial (http://olingo.apache.org/doc/odata2/tutorials/OlingoV2BasicClientSample.html) in unit tests as a test server
- Using `$inlinecount=allpages` as count parameter

Known issues:
- Result count is not propagated to the results when using splitResults option on the Camel olingo2 component (https://issues.apache.org/jira/browse/CAMEL-15405)
- Camel olingo2 component adds OData metadata to all requests by default. This leads to errors when OData service (e.g. https://services.odata.org/V2/OData/OData.svc/) does not support metadata being added to the request. EntityProviderWriteProperties has options to enable `contentOnly` settings for these kind of OData services.  EntityProviderWriteProperties are not accessible on the camel-olingo2 component when writing content. (https://issues.apache.org/jira/browse/CAMEL-15406)
- Merge operations update all fields - also those fields that are not mentioned in the request body. (see https://issues.apache.org/jira/browse/CAMEL-15407)
- Unit tests (odata-v2 and odata-v4) do make use of external reference services (e.g. https://services.odata.org/V2/OData/OData.svc/) and connect to those services with each test run. Unit tests should only use local services.


app/connector/odata-v2/src/main/java/io/syndesis/connector/odata2/component/ODataComponent.java
```diff
@@ -0,0 +1,311 @@
...
+        component.setConfiguration(configuration);
+        return Optional.of(component);
+    }
+
+    private void configureResourcePath(Olingo2AppEndpointConfiguration configuration, Map<String, Object> options) {
+        //
+        // keyPredicate is not supported properly in 2.21.0 but is handled
+        // in 2.24.0 by setting it directly on the configuration. Can modify
+        // this when component dependencies are upgraded.
+        //
```
phantomjinx, line 262: This probably ranks as some technical debt that we may or may not ever get around to sorting out?

christophd: I have tried to use key predicate header directly but soon hit unexpected behavior. So I kept the workaround as it is.

PR #8941, opened 2020-08-18
refactor: implement multi-stage build

blaggacao:
- enables skaffold based workflows
- the multi-arch control flow needed to be extracted out of the container, which blowed up the changeset

Unblocks https://github.com/xoe-labs/syndesis/pull/1

/cc @phantomjinx 

For speed go generate and got test should be taken out of the container, however we need to guarantee existence of a go in path, which we can do with shell.nix - Once that PR is accepted a whole lot of current complexities evaporate into joyful in-existence.

### DIFF
--- a/install/operator/.lib.sh
+++ b/install/operator/.lib.sh
@@ -44,6 +44,12 @@ build_container_operator()
 {
     local container_cmd="${1:-}"
     shift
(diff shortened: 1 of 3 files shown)


syndesisqe:

Issue #8950, opened 2020-08-20
[Camel K] Camel K operator installation


squakez:
## This is a...

<pre><code>
[x] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description
As mentioned in #8784 we need to provide camel k operator during installation procedure if this is not yet installed.

PR #8996, opened 2020-09-01
#8950: Enable OLM installation of camel-k

phantomjinx:
* First step towards default installation of camel-k
 ** Set Enabled flag to ALWAYS install camel-k addon
 ** Adds Olm stanza that will create a subscription to the
    community-operator camel-k 1.1.0.
 ** If a later version is required then first need to create a catalog
    source that targets appropriately

### DIFF
--- a/install/operator/build/conf/config.yaml
+++ b/install/operator/build/conf/config.yaml
@@ -30,7 +30,10 @@ Syndesis:
             Resources:
                 Memory: "1024Mi"
         CamelK:
-            Enabled: false
+            Enabled: true
+            Olm:
+                Package: "camel-k"
+                Channel: "stable"
             Image: "fabric8/s2i-java:3.0-java8"
             CamelVersion: "3.3.0"
             CamelKRuntime: "1.3.0"


squakez:
I have a question. How this behave if I previously have a camel k operator installed? will it just skip the OLM installation or produce any error?

phantomjinx:
> I have a question. How this behave if I previously have a camel k operator installed? will it just skip the OLM installation or produce any error?

It will install a second operator based on the created subscription. If you have already installed a camel-k operator from OLM, ie. existing subscription, then the current framework does take account of this.

At this point, I made the minimum changes to get this to work but since camel-k is no longer going to be an addon then its install can be refactored to become a component and at that point checks for existing camel-k operators (as well as OLM subscriptions) can be better implemented.

WDYT?

squakez:
Yeah, I feel that's the way we should proceed. However, since we're not yet completely ready to promote `camel-k` runtime, I think we should pause this PR for the moment.

christophd:
adding WIP label as this PR is paused for the moment

PR #9000, opened 2020-09-02
[ENTESB-13846] Adding extra options to Kafka

Delawen:
Now Kafka connector has dynamic <key, value> pairs that allows us to add either additional properties defined explicitly in https://camel.apache.org/components/latest/kafka-component.html or even add additional properties as in "additionalProperties"[1] on the Kafka Camel connector.

[1] https://camel.apache.org/components/latest/kafka-component.html#_options

### DIFF
--- a/app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaConnectionCustomizer.java
+++ b/app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaConnectionCustomizer.java
@@ -15,14 +15,19 @@
  */
 package io.syndesis.connector.kafka;
 
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.List;
+import java.util.Locale;
 import java.util.Map;
-
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.syndesis.connector.support.util.ConnectorOptions;
 import io.syndesis.connector.support.util.KeyStoreHelper;
 import io.syndesis.integration.component.proxy.ComponentProxyComponent;
 import io.syndesis.integration.component.proxy.ComponentProxyCustomizer;
 import org.apache.camel.component.kafka.KafkaConfiguration;
-
 import org.apache.camel.support.jsse.KeyManagersParameters;
 import org.apache.camel.support.jsse.KeyStoreParameters;
 import org.apache.camel.support.jsse.SSLContextParameters;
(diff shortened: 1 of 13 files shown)


squakez (commented):
I see you're making use of Reflection in order to set Configuration values. We discussed in the past about the convenience not to use Reflection in any new development, so I wonder if there is any alternative way of achieving that.

Delawen:
> I see you're making use of Reflection in order to set Configuration values. We discussed in the past about the convenience not to use Reflection in any new development, so I wonder if there is any alternative way of achieving that.

I don't like reflection either, but the other two options I could think about were:
 * Always assume all aditionalProperties don't have a specific configuration, which removes a lot of the shiny interesting part of this PR (being able to add Camel options even if they are not on the Syndesis UI explicitly)
 * Very long boring fail-prone if/else checking one by one the property.

If someone can hint me some other way to do this, I'm eager :) Because, yes, I hate reflection :D

squakez:
Using `beanutils` does not solve the reflection problem as it uses under the hood reflection, see https://github.com/apache/commons-beanutils/blob/master/src/main/java/org/apache/commons/beanutils2/PropertyUtilsBean.java#L2059

I wonder if you can achieve the same result skipping the setting of configuration at all, as all the `KafkaConfiguration` parameters are marked as `@UriParam` (ie, [offset](https://github.com/apache/camel/blob/master/components/camel-kafka/src/main/java/org/apache/camel/component/kafka/KafkaConfiguration.java#L112)). In theory, if you add those configuration setting in the `customize() options map` you should have the same result as Camel should be able to map the options value to the underling component:
```
           for (Map<String, String> attribute : attributes) {
                final String key = attribute.get("key").trim();
                if (!key.isEmpty()) {
                    final String value = attribute.get("value");
                    final String property = getCanonicalPropertyName(key);
                    options.put(property, value);
```
You can give it a try, not sure how much it must be adjusted to work.

Delawen:
> You can give it a try, not sure how much it must be adjusted to work.

Not working out of the box, but trying to iterate in that direction.

I chosed BeanUtils because I saw that it was already used on other parts of Syndesis.

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaConnectionCustomizer.java
```diff
@@ -62,29 +63,18 @@ public void customize(ComponentProxyComponent component, Map<String, Object> opt
...
+                    //So the PropertyBindingSupport does its work below
+                    options.put(property, value);
+
+                    //In case this is an unknown attribute on KafkaConfiguration
+                    configuration.getAdditionalProperties().put("additionalProperties." + property,
+                        value);
                 }
             }
+
+            PropertyBindingSupport.bindProperties(new DefaultCamelContext(), configuration, options);
```
squakez, line 77: +1 for using PropertyBindingSupport, but there is no need to create a new Camel Context. You should instead implement in this class the `CamelContextAware` interface and put there `this.getCamelContext()`. Just as reference see https://github.com/syndesisio/syndesis/blob/master/app/connector/mongodb/src/main/java/io/syndesis/connector/mongo/MongoClientCustomizer.java#L35

squakez (changes requested):
Just implement the CamelContextAware interface and you should be ready to go ;)

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaConnectionCustomizer.java
Delawen: Solved!

squakez (approved):
:+1:

PR #9015, opened 2020-09-07
Generate Dockerfile and container.yaml

lgarciaaco:
Generate `container.yaml` and `Dockerfile` as part of the `syndesis-operator olm` command. Those two files were missing but including them will ease the automation of this process.


install/operator/pkg/cmd/internal/olm/olm.go
```diff
@@ -50,7 +50,7 @@ func New(parent *internal.Options) *cobra.Command {
 }
 
 func (o *Olm) generate() (err error) {
-	conf, err := configuration.GetProperties(context.TODO(), configuration.TemplateConfig, o.ClientTools(), &v1beta1.Syndesis{})
+	conf, err := configuration.GetProperties(context.TODO(), configuration.TemplateConfig, nil, &v1beta1.Syndesis{})
```
phantomjinx, line 53: Why is this necessary?

Is it to stop the ApiCapabilities being analysed?

install/operator/pkg/syndesis/configuration/configuration.go
```diff
@@ -353,12 +354,12 @@ func GetProperties(ctx context.Context, file string, clientTools *clienttools.Cl
...
 
-	if ac, err := capabilities.ApiCapabilities(clientTools); err != nil {
-		return nil, err
-	} else {
-		configuration.ApiServer = *ac
+		if ac, err := capabilities.ApiCapabilities(clientTools); err != nil {
+			return nil, err
+		} else {
+			configuration.ApiServer = *ac
+		}
```
phantomjinx, line 362: Is there something causing problems with this?

install/operator/pkg/syndesis/olm/annotations_test.go
```diff
@@ -36,4 +36,6 @@ func Test_annotation_build(t *testing.T) {
 	err = a.build()
 	assert.NoError(t, err)
 	assert.NotNil(t, a.body)
+
+	// fmt.Print(string(a.body))
```
phantomjinx, line 40: Do you still want these?

install/operator/pkg/syndesis/olm/docker.go
```diff
@@ -0,0 +1,62 @@
...
+      summary="Operator which manages the lifecycle of the Fuse Online application." \
+      description="Operator which manages the lifecycle of the Fuse Online application." \
+      com.redhat.component="fuse-online-operator-metadata-container" \
+      io.k8s.description="Operator which manages the lifecycle of the Fuse Online application." \
+      io.k8s.display-name="Red Hat Fuse Online Operator" \
+      io.openshift.tags="fuse"
+`
+	m = fmt.Sprintf(m, channel, channel, c.config.SupportedOpenShiftVersions, c.config.Version)
+	c.body = []byte(m)
+	return
```
phantomjinx, line 61: Do you need this "return"?

install/operator/pkg/syndesis/olm/docker_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/syndesisio/syndesis/install/operator/pkg/apis/syndesis/v1beta1"
+	"github.com/syndesisio/syndesis/install/operator/pkg/syndesis/configuration"
+	syntesting "github.com/syndesisio/syndesis/install/operator/pkg/syndesis/testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_docker_build(t *testing.T) {
+	clientTools := syntesting.FakeClientTools()
+	conf, err := configuration.GetProperties(context.TODO(), "../../../build/conf/config-test.yaml", clientTools, &v1beta1.Syndesis{})
```
phantomjinx, line 32: You've included a clientTools in your _conf_ here but not used in that actual implementation. Probably makes little difference if d.build() doesn't use it but might be problematic in the future?
Might be safer for them to be consistent since the test would at least flag up the same result as the implementation.

install/operator/pkg/cmd/internal/olm/olm.go
lgarciaaco: I don't need to any client in the context of the `olm` command. I think the best way to express this in a declarative way is setting the clientool to nil. The change below is to make this code work.

install/operator/pkg/syndesis/olm/docker_test.go
lgarciaaco: You mean I dont need the clienttool and it is enough with nil?

install/operator/pkg/syndesis/olm/docker.go
lgarciaaco: Yes, the function returns an error

install/operator/pkg/syndesis/configuration/configuration.go
lgarciaaco: Yes. As commented above I don't need a client tool for `olm` command. this command will run in CI systems most likely. The block of code I put inside the condition throws an error if the clienttool is nil.

install/operator/pkg/syndesis/olm/annotations_test.go
lgarciaaco: I can remove

lgarciaaco:
Fixes https://issues.redhat.com/browse/ENTESB-13607

avano:
retest this please

PR #9030, opened 2020-09-11
fix(ui): Disable/enable array fields in Kafka Connector

kahboom:
Also add some styling and hides buttons for adding/removing array fields when in disabled mode.

cc @Delawen 

Screen Shot 2020-09-11 at 13 28 55

Screen Shot 2020-09-11 at 13 44 57


Delawen (approved):
Can you migrate it to 1.11.x?

PR #9041, opened 2020-09-14
Expose the cpu and memory limit/request settings for infrastructure components

phantomjinx:


install/operator/pkg/syndesis/versions/versions.go
```diff
@@ -165,13 +199,135 @@ func APIMigrator(ctx context.Context, c client.Client, n string) (r SyndesisAPIM
...
+	}
+
+	if err = api.updateApis(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Migrates from old v1alpha1 api to v1beta2. This overwrite v1alpha1
```
lgarciaaco, line 218: This scenario is not possible. v1alpha1 is only present for 7.6 and we only support sequential upgrades

lgarciaaco (approved):

install/operator/pkg/syndesis/versions/versions.go
phantomjinx: Covered all possibilities.

PR #9043, opened 2020-09-15
fix(ui-react): change oauth cookie name

### LINKED ISSUE #9040: Logout does not clean cookies
## This is a...

<pre><code>
[ ] Feature request
[ ] Bug report  
</code></pre>

## Description
It seems that when we force the logout, we don't explicitly clean the `oauth_proxy` cookies, keeping the session open indefinetely:

syndesis-logout

squakez:
Using the proper cookie name, `oauth_proxy` instead `_oauth_proxy` solved the issue.

Fix #9040

### DIFF
--- a/app/ui-react/docker/nginx-syndesis.conf
+++ b/app/ui-react/docker/nginx-syndesis.conf
@@ -25,7 +25,7 @@ location ~ /.+ {
 location = /logout {
     set $cookie "";
     if ($http_syndesis_xsrf_token = "awesome") {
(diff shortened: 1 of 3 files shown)


squakez:
@syndesisio/ui folks, I am not sure when that regression was introduced but it seems that changing properly the name fixes the problem. However I still see a warning because apparently there is some duplication when logging out:
```
Cookie “oauth_proxy” has been rejected because it is already expired.
```
If you want to take the opportunity and have a second look to have a more reliable logout process.

riccardo-forina (approved):
LGTM 👀

squakez:
> LGTM eyes

Thanks Riccardo. What do you think of that warning message? I have the feeling that there is some duplication in the cleaning but with my limited knowledge of the `ui` module I can't really say which it is.

PR #9047, opened 2020-09-16
BACKPORT ENTESB-14015: Exposes memory and cpu requests & limits in the CR

phantomjinx:
* For infrastructure elements server, db, meta & prometheus

* Since the CR structure is modified, the CR version is upgraded to v1beta2


heiko-braun:
Thanks @phantomjinx !

PR #9063, opened 2020-09-22
doc-remove-refs-to-camel-k-and-knative

MelissaFlinn:
Hi again @christophd - in this pull request, I am removing references to knative and Camel K from the doc. Kindly verify that this is correct. Thank you!


christophd:
Hey Melissa, I think this should not go into master as the Camel K runtime will remain in the Syndesis 2.x

Instead this should go into 1.11.x branch only. I remember that I already have removed some docs related to Camel K in that 1.11.x branch. Can you please have a look?

christophd (changes requested):
should not go into master but into 1.11.x branch only

MelissaFlinn:
Hi @christophd - I also sent you a message in Chat - I looked at the 1.11x branch and I see that you already made the changes there. However, those are old doc files. I have always made my changes in master.

PR #9067, opened 2020-09-25
chore: Update to AtlasMap 2.0.8 

igarashitm:
Includes: atlasmap/atlasmap#2372 XML schema doesn't load properly when anonymous complex type exists


igarashitm:
Ugh forgot to push release button on Nexus, hold on for 30 mins...

PR #9085, opened 2020-09-29
[ENTESB-13846] Kafka custom properties

Delawen:
Removing exception thrown when it shouldn't (re-casting something already casted) and backporting #8597


squakez (approved):
:+1:

PR #9088, opened 2020-09-30
fix(ui): ENTESB-13846 - kafka validate + add button fix

kahboom:
Fix for [ENTESB-13846](https://issues.redhat.com/browse/ENTESB-13846) - make validation + add buttons use the same class for styling.

Screen Shot 2020-09-30 at 11 23 17

cc @Delawen

### DIFF
--- a/app/ui-react/packages/auto-form/src/widgets/FormArrayComponent.tsx
+++ b/app/ui-react/packages/auto-form/src/widgets/FormArrayComponent.tsx
@@ -123,8 +123,7 @@ export const FormArrayComponent: React.FunctionComponent<IFormArrayControlProps>
       {!props.form.isSubmitting && !props.property.disabled && (
         <div className={'form-array-control__array-add'}>
           <Button
-            variant={'link'}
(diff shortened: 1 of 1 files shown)


Delawen (approved):

Delawen:
@gashcrumb if you don't see this impacting other parts, I think we can merge.

app/ui-react/packages/auto-form/src/widgets/FormArrayComponent.tsx
```diff
@@ -124,6 +124,7 @@ export const FormArrayComponent: React.FunctionComponent<IFormArrayControlProps>
         <div className={'form-array-control__array-add'}>
           <Button
             variant={'link'}
+            className={'pf-m-secondary'}
```
gashcrumb, line 127: I think this actually should be used like `<Button variant={'secondary'}/>` and not by adjusting the classname.

gashcrumb (changes requested):

PR #9090, opened 2020-09-30
ENTESB-14768: Stop the migration routine creating a custom resource

phantomjinx:
* At the end of the migration CRD, if a migration has been performed then
  a new CR is created. Avoid doing this if there was no migration.


claudio4j:
installing operator continues to install syndesis/app

lgarciaaco (approved):

PR #9094, opened 2020-10-01
Ability to configure the server DB connection pool from the custom resource

claudio4j:
https://issues.redhat.com/browse/ENTESB-14654


install/operator/pkg/generator/assets/infrastructure/03-syndesis-server-config.yml.tmpl
```diff
@@ -34,6 +34,25 @@
           username: '{{.Syndesis.Components.Database.User}}'
           password: '{{.Syndesis.Components.Database.Password}}'
           driver-class-name: org.postgresql.Driver
+          hikari:
```
squakez, line 37: Shouldn't we add a check here? something like `if .Syndesis.Components.Server.ConnectionPool`, otherwise skip the `hikari` property at all

squakez (approved):
Just a minor comment for your consideration, the rest looks good.

install/operator/pkg/generator/assets/infrastructure/03-syndesis-server-config.yml.tmpl
claudio4j: The `.Syndesis.Components.Server.ConnectionPool` object is always non null as there are other objects under it. To not set the hikari parameter, I would need to add a confusing template function to check all 6 parameters, which I found not worthy, as there is no harm in adding the empty hikari parameter. Thanks for your review.

PR #9099, opened 2020-10-05
fix(kafka): [ENTESB-14751] Refresh kafka topics on step configuration

Delawen:
Now topic name is a text box with autocomplete using the lastest list of kafka topics.


squakez (commented):
Is this new `datalist` type open to include any new topic name? Right now the connector gives you the possibility to create a new topic on the fly.

Delawen:
The dataList is an auto-complete text field with the list of topics. And the list of topics gets updated everytime you try to edit the step.

squakez:
Then, it looks okey for me.

squakez (approved):

PR #9103, opened 2020-10-06
[fix] ENTESB-14258 overriding the postgres driver version from spring boot

KurtStam:
https://issues.redhat.com/browse/ENTESB-14258

The postgres driver version is brought in my spring-boot. With the current pull request we are overriding the 42.2.11 to 42.2.16 (currently the  latest)


zregvart (approved):
:+1: LGTM

PR #9107, opened 2020-10-06
fix(ui) - ENTESB-14905 - SOAP address not showing in detail page

kahboom:
fix for [ENTESB-14905](https://issues.redhat.com/browse/ENTESB-14905):

- hide all fields on Detail page Edit view that don't contain an existing value
- show the Address field if a value for it exists

Screen Shot 2020-10-06 at 15 15 12

Screen Shot 2020-10-06 at 15 15 19

### DIFF
--- a/app/ui-react/packages/ui/src/Customization/apiClientConnectors/detail/ApiConnectorDetailConfigEdit.tsx
+++ b/app/ui-react/packages/ui/src/Customization/apiClientConnectors/detail/ApiConnectorDetailConfigEdit.tsx
@@ -8,6 +8,7 @@ export interface IApiConnectorDetailConfigEdit {
   /**
    * Property labels
    */
+  i18nLabelAddress: string;
   i18nLabelBaseUrl: string;
   i18nLabelDescription: string;
   i18nLabelHost: string;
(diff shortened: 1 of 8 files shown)


app/ui-react/packages/ui/src/Customization/apiClientConnectors/detail/ApiConnectorDetailConfigEdit.tsx
```diff
@@ -56,68 +58,92 @@ export const ApiConnectorDetailConfigEdit: React.FunctionComponent<IApiConnector
...
+            id="connector-name"
+            aria-describedby="horizontal-form-name-helper"
+            data-testid={'api-connector-name-field'}
+            name="name"
+            onChange={onChange}
+            validated={isValid}
+          />
+        </FormGroup>
+      )}
+      {properties.description && (
```
kahboom, line 82: What a pain that the formatting has caused the diff to look like this.. all I did was add a check for the property like `{properties.description && ( // display the field )}`, and of course add a form group for the Address property.

gashcrumb: np :smile:

gashcrumb (approved):
:+1:

gashcrumb:
@kahboom looks like some of the test code needs to be updated with that interface change.

gashcrumb (approved):
:+1

PR #9109, opened 2020-10-06
Expose JAVA_OPTIONS for syndesis server/meta components

claudio4j:
https://issues.redhat.com/browse/ENTESB-14869

Removes the ConnectionPool assignments as there is no need, because the mergo.merge operation already sets the values from syndesis custom resource to the configuration object.


claudio4j:
Ignore app/test/integration-test/src/test/java/io/syndesis/test/itest/apiconnector/SoapConnector_IT.java temporarily, as it is having unrelated issues to the PR change itself.

PR #9113, opened 2020-10-06
Fixes weird failures in Soap integration test

dhirajsb:


Delawen (approved):
One of this classic "but how could this have worked before :D"

PR #9118, opened 2020-10-07
Update dependencies ENTESB-10454

claudio4j:
This is related to "ENTESB-10454 Syndesis alignment problems"
https://issues.redhat.com/browse/ENTESB-10454

Changes to the tests: It should be @BeforeEach to run properly
Changes to maven-plugin, the AbstractMojo from spring-boot-maven-plugin already provides a
project attribute, but once the plugin runs, the project attribute is
not set. So, renaming just works.

Update go vendor as there was a recent update to github.com/spf13/afero-1.4.0


Delawen:
Integration tests failing but I don't see why your changes affect this.

squakez (commented):
The PR itself looks okey, a part some integration error maybe driven by the google libraries changes. However, if this is targeting a backport to 1.11, I think there will be quite a lot of more work to do. Unless there are CVEs to fix, I'd personally park this activity. We introduce more risk than benefits IMO.

PR #9139, opened 2020-10-15
ENTESB-15027 role permissions - Operator Hub

phantomjinx:
* csv.go
 * Permissions incorrectly assigned to syndesis-operator when they should
   be assigned to both the kafka & public-api service accounts
 * Ensure the `synop olm ...` command is executed in the operator directory

* csv_test.go
 * Modifies tests accordingly


install/operator/pkg/syndesis/olm/csv.go
```diff
@@ -207,16 +208,50 @@ func (c *csv) build() (err error) {
 	}
 	c.setVariables()
 
+	//
+	// Ensure the following files can be located
+	//
+	if _, err := os.Stat("pkg"); os.IsNotExist(err) {
+		return fmt.Errorf("This needs to be executed from the syndesis/install/operator directory.")
+	}
```
zregvart, line 216: Ideally, if this is important, embedding the data needed in the executable binary might be a better option.

For quick stab at robustness as `pkg` is somewhat widespread, we might go with a marker file, or try to access the required files and error out if they can't be read.

Or if it makes make sense to have a look at the current working directory instead?

```suggestion
	if cwd, err := os.Getwd(); err != nil {
		if strings.HasSuffix(cwd, "/install/operator") == false {
			return fmt.Errorf("This needs to be executed from the syndesis/install/operator directory.")
		}
	} else {
		return err
	}
```

install/operator/pkg/syndesis/olm/csv_test.go
```diff
@@ -32,9 +33,14 @@ func Test_csv_build(t *testing.T) {
 	assert.NoError(t, err)
 
 	c := &csv{config: conf, operator: "operator-image"}
+
+	pkgPath := "pkg"
+
+	_ = os.Mkdir(pkgPath, os.ModeDir)
+	defer os.Remove(pkgPath)
```
zregvart, line 40: Sad, this bit, seems we took on some debt by necessitating the `pkg` directory.

install/operator/pkg/syndesis/olm/csv_test.go
```diff
@@ -108,44 +114,60 @@ func Test_csv_loadDeploymentFromTemplate(t *testing.T) {
...
 
-	assert.True(t, ok, "Expected return value should be a map[interface{}interface{}")
-}
-
-func Test_csv_loadClusterRoleFromTemplate(t *testing.T) {
-	clientTools := syntesting.FakeClientTools()
-	conf, err := configuration.GetProperties(context.TODO(), "../../../build/conf/config-test.yaml", clientTools, &v1beta2.Syndesis{})
-	assert.NoError(t, err)
+			assert.True(t, len(r) > 0)
+			assert.Equal(t, tc.expect, len(r))
```
zregvart, line 159: Perhaps both asserts are not needed?
```suggestion
			assert.Equal(t, tc.expect, len(r))
```

zregvart (commented):
Ideally I'd like to see a bit more robustness, I think this might lead to issues, i.e running from the wrong directory and including unexpected YAML files. We embed some YAML files already perhaps that would be a better approach?

Other than that straightforward, and works in the short run.

avano:
retest this please

install/operator/pkg/syndesis/olm/csv_test.go
phantomjinx: gone.

phantomjinx:
> Ideally I'd like to see a bit more robustness, I think this might lead to issues, i.e running from the wrong directory and including unexpected YAML files. We embed some YAML files already perhaps that would be a better approach?
> 
> Other than that straightforward, and works in the short run.

olm asset files not embedded in the same manner as generator yaml files.

zregvart (approved):
Thanks!

zregvart:
@phantomjinx I've looked at the QE test suite, it fails with this error:

```
+ docker run --entrypoint syndesis-operator -v /home/hudson/.kube:/tmp/kube:z quay.io/syndesispr/syndesis-operator:9139-b83d19adcbf6a4e6139c36bdf38d3c52d658edfe install cluster --config /tmp/kube/config -n syndesis
panic: could not find the root module directory

goroutine 1 [running]:
github.com/syndesisio/syndesis/install/operator/pkg/build.init.0()
	/home/circleci/project/install/operator/pkg/build/pkg.go:45 +0x249
```

the `9139-b83d19adcbf6a4e6139c36bdf38d3c52d658edfe` tag is built on CircleCI and pushed to quay.io here: https://quay.io/repository/syndesispr/syndesis-operator?tag=latest&tab=tags

zregvart (changes requested):
We probably need to take a look at that failure in QE test

zregvart (commented):
@phantomjinx I think https://github.com/phantomjinx/syndesis/pull/1 should fix this, can you take a look?

zregvart (approved):
I think we should be okay now, let's wait for the QE tests to finish.

PR #9175, opened 2020-11-03
ENTESB-14949: Revert jaeger-operator role-bindings to cluster-role-bi…

phantomjinx:
…ndings

* Since the requirement is binding to a ClusterRole with the potential for
  listing cluster-level resources such as ClusterRoleBindings, only a
  cluster-role-binding provides the necessary permission.

* This is in line with the other grant resources such as kafka and
  the public-api-connector.

* This is only relevant in a non-OperatorHub use-case such as running on
  minishift 3.11 since later versions install jaeger using the OperatorHub.


lgarciaaco (approved):

mmuzikar:
retest this please

PR #9176, opened 2020-11-04
fix: Missing patch body in FHIR Patch operation

rkorytkowski:
ENTESB-14237


claudio4j:
@rkorytkowski can you backport to 1.11.x ?

PR #9178, opened 2020-11-05
Cache the apache license locally

claudio4j:
When building the s2i image, it downloads all licenses
however when building in jenkins there are many network errors
which we believe are related to apache infra avoiding
many request in a short time.

This way the apache license is cached locally and will speed up
the build process.


zregvart (approved):
Awesome! Thanks!

PR #9179, opened 2020-11-05
fix(ui) ENTESB-15273 - API Client Connector

kahboom:
Fixes [ENTESB-15273](https://issues.redhat.com/browse/ENTESB-15273) - When the fields are empty during editing API Connector, they disappear.

Since we are not using Autoform for the API Client Connector wizard and we want to make as few drastic changes as possible, I grabbed the keys from the `properties` object to determine which fields should be shown/hidden in the respective UI components. Fields that have no value set are hidden from the View mode, but available in Edit mode. Name and description fields are always shown. It's not my favorite solution to this problem, but it's lower risk, because it ensures that the fields are only shown if they are specified within the `properties` property.

I also discovered another bug where you are not able to edit the Address field of a SOAP Connector, verified by @mkralik3, which is fixed in this PR.

- [Video/GIF of OpenAPI connector fix](https://issues.redhat.com/secure/attachment/12501471/Kapture%202020-11-05%20at%2013.41.08.gif)
- [Video/GIF of SOAP connector fix](https://issues.redhat.com/secure/attachment/12501481/Kapture%202020-11-05%20at%2014.50.05.gif)

### DIFF
--- a/app/ui-react/packages/api/src/WithApiConnectorHelpers.tsx
+++ b/app/ui-react/packages/api/src/WithApiConnectorHelpers.tsx
@@ -11,6 +11,7 @@ export interface IWithApiConnectorHelpersChildrenProps {
     apiConnector: Connector,
     name?: string,
     description?: string,
+    address?: string,
     host?: string,
     basePath?: string,
     icon?: string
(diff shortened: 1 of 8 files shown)


gashcrumb:
Wonder, is this test related to this change:

image

gashcrumb:
retest this please

gashcrumb (approved):
:+1:

PR #9191, opened 2020-11-10
[fix] backport ENTESB-15274 API Provider response is always 200 for 2xx responses

KurtStam:
Regression


mkralik3:
This PR https://github.com/syndesisio/syndesis/pull/9194 resolved the issue ENTESB-15274, verified on the latest version 1.11.0-20201113. In that case, this issue can be closed.

PR #9194, opened 2020-11-11
[Backport] fix(connector/api): status code setting

squakez:
Integration tests were showing a different status code, because we only updated it if an error was raised. Moving the status code update outside the error check solves the issue.

Ref ENTESB-15131


mkralik3:
@squakez Shouldn't be this backported also to the master?

PR #9195, opened 2020-11-12
fix(version-test): This 1.10 was was missing from the 1.11.x change

Delawen:
The configuration for the tests was using an old version.


zregvart (approved):

zregvart:
The QE tests fail, but I don't think they can fail because of this change.

PR #9196, opened 2020-11-12
fix(s2i): make sure cached artifacts are used

zregvart:
When the image for the integration is build via S2I we need to use the
artifacts that have been cached in the `/tmp/artifacts/m2` in the
Syndesis S2I image.

Previously passing `-Dmaven.repo.local=/tmp/artifacts/m2` in the Build
seemed to have done just that, but for _whatever_ reason, perhaps[1],
this started failing recently. Or we noticed it failing recently.

So if the specifying `maven.repo.local` is no longer helpful, the
solution here is to specify the `/tmp/artifacts/m2` repository within
the `settings.xml` placed in `configuration` directory of the generated
project.

[1] https://lists.apache.org/thread.html/f8790f0dcc7195c841fd2717605c96036c82a01d5a45842ac9bfe2fd%40%3Cusers.maven.apache.org%3E

### DIFF
--- a/app/integration/project-generator/src/main/resources/io/syndesis/integration/project/generator/templates/settings.xml
+++ b/app/integration/project-generator/src/main/resources/io/syndesis/integration/project/generator/templates/settings.xml
@@ -23,6 +23,16 @@
       <id>additional-repositories</id>
       <repositories>
         <repository>
+          <id>_local</id>
+          <url>file:///tmp/artifacts/m2</url>
+          <releases>
+            <enabled>true</enabled>
+          </releases>
+          <snapshots>
+            <enabled>true</enabled>
+          </snapshots>
+        </repository>
+        <repository>
           <id>central</id>
           <name>Maven Central</name>
           <url>https://repo.maven.apache.org/maven2/</url>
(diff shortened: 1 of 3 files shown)


Delawen:
Some tests are failing but not for this reason.

zregvart:
> Some tests are failing but not for this reason.

I've rerun the integration tests just to make sure...

Delawen (approved):

zregvart:
One thing that I'm a bit concerned is that the S2I image will have the artifacts twice now. Once in the `/tmp/artifacts/m2` and once in `$HOME/.m2` with this change. Which will make the S2I image bigger in size and impact the performance.

Delawen:
Can we make the image copy the repository directly on `$HOME/.m2`? Would that work?

zregvart:
> Can we make the image copy the repository directly on `$HOME/.m2`? Would that work?

IIRC there was a permission issue in doing that (when run on OpenShift).

zregvart:
I've found that changing `settings.localRepository` to `maven.repo.local` here would help also. Perhaps this is not needed but a backport of https://github.com/syndesisio/syndesis/pull/7426

christophd:
@zregvart how did you find that? I totally forgot about it 😅

zregvart:
I'm going to close this due to side-effects and work on backport of #7426 to 1.11

PR #9197, opened 2020-11-12
chore: Use no-transfer-progress option on integration-test Maven buil…

zregvart:
…d by default [ci skip]

This should shrink the amount of log output on CircleCI significantly. We have reached log output limits of CircleCI before.

(cherry picked from commit 18d9d0283c2295ed5437821e9f7c6c13247fdb00)

### DIFF
--- a/tools/bin/commands/integration-test
+++ b/tools/bin/commands/integration-test
@@ -24,7 +24,7 @@ integration-test::run() {
 }
 
 maven_args() {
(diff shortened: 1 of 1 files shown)


zregvart:
I think it's okay to merge this. There are issues with the integration tests (that might get fixed with #9198) and with the QE tests, but it's hard for me to see how those are influenced by this change.

PR #9198, opened 2020-11-13
fix: fix integration tests on 1.11.x

zregvart:
Use maven.repo.local property instead of settings.localRepository to set the local Maven repository for integration runtime container.

(cherry picked from commit 66067878c12fd5fa178cef98585280fc99af51f5)


oscerd (approved):

christophd:
thanks! 👍

zregvart:
This needed a rebase to include #9195, hopefully the integration tests will pass now.

Delawen:
Maybe it is not enough and we need the previous patch?

To me it is still not clear when/why this maven change happened. Maybe in JDK 8 (as opposed to JDK 11 in master) this still is not enough?

zregvart:
No idea, locally in my IDE (Eclipse) tests pass on this branch, they fail from command line. Logs:

```
13:26:32.291 [ream-1133381918] INFO  INTEGRATION_RUNTIME_CONTAINER  - STDOUT: Starting the Java application using /opt/run-java/run-java.sh ...
13:26:32.291 [ream-1133381918] INFO  INTEGRATION_RUNTIME_CONTAINER  - STDOUT: exec java -javaagent:/opt/jolokia/jolokia.jar=config=/opt/jolokia/etc/jolokia.properties -javaagent:/opt/prometheus/jmx_prometheus_javaagent.jar=9779:/opt/prometheus/prometheus-config.yml -XX:+UseParallelGC -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:+ExitOnOutOfMemoryError -cp . -jar /deployments/integration-runtime.jar
13:26:32.291 [ream-1133381918] INFO  INTEGRATION_RUNTIME_CONTAINER  - STDERR: Error: Invalid or corrupt jarfile /deployments/integration-runtime.jar
```

And indeed `/deployments/integration-runtime.jar` is 0 bytes. I've looked in the S2I build and there `/deployments/project-0.1-SNAPSHOT.jar` looks good. I think we might be missing another commit from master here. Looking into it.

zregvart:
So the integration container image is trying to copy `.../app/test/integration-test/target/integrations/i-timer-to-log-*/target/project-0.1-SNAPSHOT.jar` to `/deployments/integration-runtime.jar`, but the `.../app/test/integration-test/target/integrations/i-timer-to-log-*/target/` directory doesn't exist.

zregvart:
Yeah, I think we need to include #8403 and #8451 as well.

zregvart:
We also need #7673

zregvart:
And #8815

zregvart:
And #7391

zregvart:
I think I'll remove the S2I Dockefile change, I don't think it's needed, to keep the changes within integration tests only.

zregvart:
Retest this please

PR #9204, opened 2020-11-16
Change imagePullPolicy from ifNotPresent to Always

lgarciaaco:
As noticed in ENTESB-15296, images are not always pulled due to the floating tag. This is a problem where there are CVS released and an upgrade of the images is required


lgarciaaco:
Fixes https://issues.redhat.com/browse/ENTESB-15296

PR #9216, opened 2020-11-27
fix: configured vs default values in API connector

zregvart:
On the custom API Client Connector's detail page we want to show the
configured property value or if it doesn't exist the default value of
the property.

Similarly on the configuring a connection we want to use the
configured property first and then the the default value.

ref. https://issues.redhat.com/browse/ENTESB-15292


kahboom (approved):
Great, thanks for the test as well @zregvart . 👍

PR #9224, opened 2020-12-03
fix for ENTESB-12104 - demoData flag used to add sampledb

KurtStam:


install/operator/pkg/generator/assets/database/syndesis-db.yml.tmpl
```diff
@@ -29,6 +29,17 @@
...
+    delete-sample-db.sh: |
+      #!/bin/bash
+      until bash -c "psql -h 127.0.0.1 -U $POSTGRESQL_USER -q -d $POSTGRESQL_DATABASE -c 'SELECT 1'"; do
+        echo "Waiting for Postgres server..."
+        sleep 1
+      done
+      echo "***** deleting sampledb"  
+      psql <<EOF
+        DROP DATABASE sampledb;
+        DROP USER sampledb;
```
zregvart, line 41: Might be a bit safer

```suggestion
        DROP DATABASE IF EXISTS sampledb;
        DROP USER IF EXISTS sampledb;
```

zregvart:
@KurtStam you need to regenerate the operator assets, i.e. run `go generate ./pkg/generator/` from `install/operator`.

PR #9248, opened 2020-12-24
chore(sql): support UUID column type in PostgreSQL

zregvart:
When we encounter a column of JDBC type `OTHER` we now look at the type
name to see if it equals `UUID` (regardless of case), and report the
column having `VARCHAR` JDBC type instead.

Ref. issues.redhat.com/browse/ENTESB-13689

(cherry picked from commit 96a2189)

Also includes:

chore(build): explicit annotation processors

Makes annotation processor configuration explicit and removes Lombok
from SQL connector. This way we can depend on only the annotations of
Immutables and this clears a path for configuring ErrorProne/Immutables
on Java 9+.

(cherry picked from commit 8dd3dbdb7dc08692962d3fd2969181b82ca031b7)

```
# Conflicts:
#	app/connector/debezium/pom.xml
#	app/connector/fhir/pom.xml
#	app/connector/kafka/pom.xml
#	app/connector/knative/pom.xml
#	app/connector/webhook/pom.xml
#	app/extension/maven-plugin/pom.xml
#	app/server/builder/image-generator/pom.xml
#	app/server/builder/maven-plugin/pom.xml
#	app/server/update-controller/pom.xml
```

### DIFF
--- a/app/common/model/pom.xml
+++ b/app/common/model/pom.xml
@@ -44,7 +44,7 @@
     <dependency>
       <groupId>org.immutables</groupId>
       <artifactId>value</artifactId>
-      <scope>compile</scope>
+      <classifier>annotations</classifier>
     </dependency>
 
     <dependency>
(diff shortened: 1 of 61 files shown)


zregvart:
retest this please

PR #9257, opened 2021-01-06
ENTESB-14875 Reload ConfigMaps to avoid manual pod respin

claudio4j:
When the user changes properties on syndesis custom resource
that changes values in ConfigMaps, the spring context or spring app
should be restarted to take the changes into effect, otherwise the
user must manually respin the pod.


zregvart (approved):
:+1: Thanks!

PR #9275, opened 2021-01-15
fix: consider 'none' data shape present

### LINKED ISSUE #3807: API Provider Data Type Mismatch

zregvart:
For cases where action's descriptor has data shape of type 'none' we
would assert that the data shape does not exist. This would cause us not
to recommend adding a mapping step between a input step with data shape
of 'none' and a step containing a data shape other than 'none'. In that
case we do wish to suggest adding a mapping step as mapper can operate
without inputs, i.e. it can add constant values and map those to the
output.

Fixes #3807
Ref https://issues.redhat.com/browse/ENTESB-11568


kahboom (approved):
looks great, thanks for the tests 👍

PR #9276, opened 2021-01-15
fix: make sure closable resources are closed

zregvart:
When investigating ENTESB-15548[1] I've found several instances of
streams not being closed. Some of these streams are streams that are
obtained from the PostgreSQL JDBC driver and might lead to connection
leak. I haven't investigated if that's the case, other than what's
already noted on ENTESB-15548. I think this refactoring is safe.

[1] https://issues.redhat.com/browse/ENTESB-15548

### DIFF
--- a/app/extension/converter/pom.xml
+++ b/app/extension/converter/pom.xml
@@ -77,6 +77,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.assertj</groupId>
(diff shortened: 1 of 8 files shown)


KurtStam (approved):
Looks good to me.

zregvart:
Needs to be rebased on #9282

PR #9284, opened 2021-01-18
fix(ui) [ENTESB-11603] prevent configured properties from being cleared in integration + API connector editors

kahboom:
fixes [ENTESB-11603](https://issues.redhat.com/browse/ENTESB-11603)

- for the Create Connection fix, include `configuredProperties` in `backHref` of the `ReviewPage`, since there is already a check in place for configured fields in `ConfigurationPage`
- went ahead and fixed this for the Security step of the API client connector as well, which, conversely, did _not_ check for configured fields in `SecurityPage`, so this check was added

Kapture 2021-01-18 at 15 12 45


zregvart (approved):

gashcrumb (approved):
:+1:

PR #9288, opened 2021-01-20
fix: make sure closable resources are closed

zregvart:
When investigating ENTESB-15548[1] I've found several instances of
streams not being closed. Some of these streams are streams that are
obtained from the PostgreSQL JDBC driver and might lead to connection
leak. I haven't investigated if that's the case, other than what's
already noted on ENTESB-15548. I think this refactoring is safe.

[1] https://issues.redhat.com/browse/ENTESB-15548

(cherry picked from commit 54fd7ff931aafe7445a03daae4a9528d9baa9ad7)

```
# Conflicts:
#	app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/extension/DefaultVerifierExtensionUploader.java
#	app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/support/IntegrationSupportHandler.java
```

### DIFF
--- a/app/extension/converter/pom.xml
+++ b/app/extension/converter/pom.xml
@@ -77,6 +77,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+    </dependency>
+
+    <dependency>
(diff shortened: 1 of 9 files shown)


zregvart:
retest this please

PR #9305, opened 2021-01-29
ENTESB-15631: Creates github action to release kustomize archive

phantomjinx:
* Creates new github action

* Pairs down Makefile - still called from action script

* Removes release.sh script as no longer required.


zregvart:
We do everything via the syndesis scripts, should this go into `syndesis release`?

PR #9315, opened 2021-02-03
chore: Update to AtlasMap 2.1.4

igarashitm:
Release Note: https://github.com/atlasmap/atlasmap/releases/tag/atlasmap-2.1.4


Delawen:
I re-runned the failed jobs.

The integration test one fails on `{"message":"No such image: testcontainersofficial/ryuk:0.3.0"}` which is weird because I don't see that change in this PR.

igarashitm:
@Delawen yep, I think the same error is seen in other PRs as well

zregvart:
Try re-basing to include #9323

PR #9320, opened 2021-02-04
feat(integration): support pollEnrich

zregvart:
This adds support for Connector Actions with `PollEnrich` pattern placed
other than at first step of the Integration Flow by using Poll Enrich
EIP.

To utilize this support, Connector Action needs to declare the `pattern`
as `PollEnrich` and the connector's Camel component needs to support
`PollingConsumer`.

To support before/after consumers for poll enrich, we're not using the
`pollEnrich` Route DSL, but we're adding `PollEnricher` processor
directly to the route via `process` Route DSL method. This allows us
to add a pipeline of `before-PollEnricher-after` processors.

User interface now allows actions with the `PollEnrich` pattern in all
but the first step.


gashcrumb (approved):
UI changes :+1:

KurtStam (approved):
Looks good to me

PR #9324, opened 2021-02-08
fix: tolerate repeated close of extension stream

zregvart:
The stream we pass from the database to the HTTP client will be closed
by the HTTP client, it might be closed only for happy day cases or it
might be closed for both happy and error cases, we don't have control
over that. If we use try-with-resources, as we do currently, we might
try to close the same stream twice: once by the HTTP client once by the
try-with-resources.

This allows the stream to be closed multiple times.

Ref. https://issues.redhat.com/browse/ENTESB-15671


Delawen (approved):
LGTM but tests passing...

PR #9326, opened 2021-02-09
feature(sonarcloud.io):  Adding automatic code analysis per PR

Delawen:
This adds another github action per PR (and per push to monitor main branches like 1.*.x) that executes verify tasks like test coverage, findbugs,... and pushes to https://sonarcloud.io/dashboard?id=syndesis-parent

It does consider only unit tests as test coverage. It does not execute integration tests.

It also adds a couple of badges to the README to show test coverage and quality in general.

In time, we could force some quality assurance on each PR according to the analysis that compares it to the main branch, like on https://sonarcloud.io/dashboard?id=syndesis-parent&pullRequest=6 (this same branch) But let's start by just measuring it and see if it is useful.


Delawen:
The pom file is painfully big because we can't compile in JDK 11. Once we can compile in JDK 11 (if ever), it will be severelly reduced.

app/pom.xml
```diff
@@ -388,20 +389,6 @@
...
-              <execution>
-                <id>basepom.default</id>
-                <phase />
-              </execution>
-              <execution>
-                <id>basepom.default-it</id>
-                <phase />
-              </execution>
-            </executions>
-          </plugin>
```
zregvart: We don't want jacoco in `flash` profile, or do we?

app/pom.xml
```diff
@@ -653,6 +640,197 @@
...
+        <basepom.check.skip-coverage>false</basepom.check.skip-coverage>
+      </properties>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <!-- Fix "There was a timeout or other error in the fork" errors -->
+              <forkedProcessTimeoutInSeconds>120</forkedProcessTimeoutInSeconds>
```
zregvart: This reduces the timeout from configured [`150` via `basepom.test.timeout`](https://github.com/syndesisio/syndesis/blob/7996e55ec874eab689a790c96fcac041ffc1f0ad/app/pom.xml#L117), see https://github.com/basepom/basepom/blob/31581ff6dfb631a5ee36939a0552665de0b3dc90/foundation/pom.xml#L371

app/pom.xml
```diff
@@ -653,6 +640,197 @@
...
+            <artifactId>jacoco-maven-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>basepom.default</id>
+                <goals>
+                  <goal>prepare-agent</goal>
+                </goals>
+                <configuration>
+                  <!-- Sets the VM argument line used when unit tests are run. -->
+                  <propertyName>surefireArgLine</propertyName>
```
zregvart: Default `argLine` should do

app/pom.xml
```diff
@@ -653,6 +640,197 @@
...
+      </properties>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+              <!-- Fix "There was a timeout or other error in the fork" errors -->
+              <forkedProcessTimeoutInSeconds>120</forkedProcessTimeoutInSeconds>
+              <argLine>@{surefireArgLine}</argLine>
```
zregvart: We loose [test memory configuration and file encoding](https://github.com/basepom/basepom/blob/31581ff6dfb631a5ee36939a0552665de0b3dc90/foundation/pom.xml#L367) with this approach; `argLine` set by `jacoco:prepare-agent` should be already present.

app/pom.xml
```diff
@@ -653,6 +640,197 @@
...
+                  <propertyName>surefireArgLine</propertyName>
+                </configuration>
+              </execution>
+              <execution>
+                <id>report</id>
+                <goals>
+                  <goal>report</goal>
+                </goals>
+                <phase>verify</phase>
+              </execution>
```
zregvart: I think this bit is the only bit we need.

app/pom.xml
```diff
@@ -653,6 +640,197 @@
       </properties>
     </profile>
 
+    <!-- Sonar Cloud and Coverage -->
+    <profile>
+      <id>sonarcloud</id>
+      <properties>
+        <basepom.check.fail-coverage>false</basepom.check.fail-coverage>
+        <basepom.check.skip-coverage>false</basepom.check.skip-coverage>
```
zregvart: This should be `true` only if `skip-checks` or `flash` profiles are active, so by default it's `false`, do we need to set it here?

app/pom.xml
```diff
@@ -653,6 +640,197 @@
...
+              <execution>
+                <id>basepom.invoker-integration-testing.default</id>
+                <phase>none</phase>
+              </execution>
+            </executions>
+          </plugin>
+          <plugin>
+            <groupId>org.sonarsource.scanner.maven</groupId>
+            <artifactId>sonar-maven-plugin</artifactId>
+            <version>3.8.0.2131 </version>
```
zregvart:
```suggestion
            <version>3.8.0.2131</version>
```

app/pom.xml
```diff
@@ -653,6 +640,197 @@
...
+      <dependencies>
+        <!--  JDK dependencies  -->
+        <!-- Needed until we are able to compile in JDK 11+ -->
+        <dependency>
+          <groupId>com.sun</groupId>
+          <artifactId>tools</artifactId>
+          <version>8</version>
+          <scope>system</scope>
+          <systemPath>${java.home}/../../../jdk-8-hotspot/1.0.0/x64/lib/tools.jar</systemPath>
+        </dependency>
```
zregvart: Should this be in the plugin configuration? I'd trigger this based on `project.build.targetJdk` property, i.e. a separate profile that adds this as a dependency of the plugin activated based on `project.build.targetJdk=1.8`.

app/pom.xml
zregvart: Perhaps have the plugin coordinates declared in `pluginManagement` section. That way the version is used when invoked from command line (not just in the profile).

.github/workflows/sonarcloud.io.yml
```diff
@@ -0,0 +1,50 @@
...
+          key: ${{ runner.os }}-sonar
+          restore-keys: ${{ runner.os }}-sonar
+      - name: Cache Maven packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
+      - name: Maven Verify
+        run: cd app; mvn verify install -Psonarcloud -pl \!ui-react -fn
```
zregvart:
```suggestion
        run: cd app; ./mvnw verify install -Psonarcloud -pl \!ui-react -fn
```

.github/workflows/sonarcloud.io.yml
```diff
@@ -0,0 +1,50 @@
...
+          version: '11'
+      - name: Remove Problematic BOM project
+        run: rm -r app/extension/bom
+      - name: Remove Problematic BOM project reference
+        run: sed -i "/bom/d" app/extension/pom.xml
+      - name: Send analysis to SonarCloud
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # Needed to get PR information, if any
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        run: cd app; mvn verify -P\!non-staging -Dsonar.login=${{secrets.SONAR_TOKEN}} -Psonarcloud_send -pl \!ui-react
```
zregvart: Perhaps we can simplify this a bit to:

```suggestion
        run: cd app; ./mvnw -Dsonar.login=${{secrets.SONAR_TOKEN}} sonar:sonar
```

if we declare all the `sonar.*` properties outside of the `Psonarcloud_send` profile and set the plugin/dependency on `tools.jar` based on `project.build.targetJdk` property.

.github/workflows/sonarcloud.io.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          java-version: 11
+      - name: Install Java 11
+        uses: AdoptOpenJDK/install-jdk@v1
+        with:
+          version: '11'
+      - name: Remove Problematic BOM project
+        run: rm -r app/extension/bom
+      - name: Remove Problematic BOM project reference
+        run: sed -i "/bom/d" app/extension/pom.xml
```
zregvart: Seems rather drastic, perhaps exclude via profile?

app/pom.xml
```diff
@@ -174,6 +174,7 @@
     <module>s2i</module>
     <module>meta</module>
     <module>ui-react</module>
+    <module>test</module>
```
zregvart: Not sure about this change, I think `test` and `../doc` module are included in most cases (i.e. when `non-staging` profile is active), and we don't want those to be included when staging (I guess, no idea what sets `stagingDescription`, could be Sonatype release plugin).

.github/workflows/sonarcloud.io.yml
Delawen: The problem is that this project is inside the folder structure but it is not included inside the maven structure. And that's a huge bug for sonar (you can't even workaround it with "-fn")

app/pom.xml
Delawen: Maybe not, I tried a lot of combinations with basepom and sonar properties until I made it work. I will try without.

app/pom.xml
Delawen: 120 is defined in our main pom.xml: https://github.com/syndesisio/syndesis/blob/1.12.x/app/pom.xml#L999-L1006
Maybe it is not needed anymore?

I added here because I needed to add the following surefireArgLine.

app/pom.xml
Delawen: Well... it wasn't. I'll check how to merge.

app/pom.xml
Delawen: Makes sense.

app/pom.xml
Delawen: This is very very specific to the environment. For example on my local the path is different. I'm not sure if this would be useful outside the github action context.

.github/workflows/sonarcloud.io.yml
Delawen: Can we deactivate a profile and a child from the sonarcloud profile? I'm not sure how to do that.

app/pom.xml
Delawen: Removing it then.

app/pom.xml
Delawen: Right, this is a mistake, I cut/paste from the wrong profile

app/pom.xml
zregvart: Yeah, probably a mistake we made in not configuring via basepom. For simplicity and consistency we should tweak basepom properties rather than changing the plugin configuration directly.

.github/workflows/sonarcloud.io.yml
zregvart: From what I understand, we need to add the `tools.jar` only when running on Java 11 and targeting Java 8. So we can activate a profile that adds that dependency to the plugin dependencies based on the value of `project.build.targetJdk`.

The trick is to think of the opposite, and use profile activation, similar to how `non-staging` profile works, if a property is not defined include the modules if it's defined include it; perhaps based on the `sonar.login` property.

app/pom.xml
Delawen: Fixed and moved to pluginManagement

app/pom.xml
Delawen: Moved to the properties on the top of pom.xml

.github/workflows/sonarcloud.io.yml
Delawen: `tools.jar` is only being used when analyzing the code. We are not really going to use it outside the sonarcloud use case. Is it good to take it out of that profile then?

zregvart: To get the `./mvnw -Dsonar.login=${{secrets.SONAR_TOKEN}} sonar:sonar` to work without specifying profiles, yeah I think it's worth it. Better have things auto-configured than rely on remembering how to invoke Maven, i.e. what profiles are needed and such.

zregvart:
Another thought, I think this should be based on #9328.

app/pom.xml
zregvart: Yup, I've just seen this:
```
[INFO] --- jacoco-maven-plugin:0.8.5:prepare-agent (basepom.default) @ server-endpoint ---
[INFO] argLine set to -javaagent:/home/zregvart/.m2/repository/org/jacoco/org.jacoco.agent/0.8.5/org.jacoco.agent-0.8.5-runtime.jar=destfile=/home/zregvart/tmp/syndesis-1.12.x/app/server/endpoint/target/jacoco.exec
```

Delawen:
Do we push this for pushes only until we find the way to do this per PR?

zregvart:
> Do we push this for pushes only until we find the way to do this per PR?

Yeah, that makes sense, small steps we know will work. Best integrate that in the existing workflow, so we don't build twice.

Delawen:
> > Do we push this for pushes only until we find the way to do this per PR?
> 
> Yeah, that makes sense, small steps we know will work. Best integrate that in the existing workflow, so we don't build twice.

I don't see any building job on push right now. We only have on PR.

Delawen:
Let's see if it works on my fork: https://github.com/Delawen/syndesis/runs/1887069441?check_suite_focus=true

zregvart:
> I don't see any building job on push right now. We only have on PR.

Yeah, my mistake, I thought we had `on: push` here also:

https://github.com/syndesisio/syndesis/blob/54d702655f5d70c27858ef880116ba192e21a274/.github/workflows/on_pr.yml#L3-L4

We could rename that to `build.yml` and ron sonar only if `SONAR_TOKEN` is present? That would be for forks that have the secret defined and PRs originating from this repository.

Delawen:
So we add jacoco and all the verify stuff on a generic build but only send if sonarcloud token is present?
Or do the full build only if the token is present?

I'm worried that jacoco sometimes throws errors unrelated to the code quality and that may block the rest of the build. Unless we build and then we verify on a separated step with the "-fn".

zregvart:
I'd add the sonar step as the last step in the build and make it run only if `SONAR_TOKEN` env variable is present. If the build fails before that I think it's okay not to send the data to sonar.

Seems that the build went [okay](https://github.com/Delawen/syndesis/runs/1887069441?check_suite_focus=true#step:9:142) on your fork.

zregvart:
Hmm, [code coverage is at 0%](https://sonarcloud.io/component_measures?id=syndesis-parent&metric=coverage&view=list), that doesn't seem right.

Delawen:
> I'd add the sonar step as the last step in the build and make it run only if `SONAR_TOKEN` env variable is present. If the build fails before that I think it's okay not to send the data to sonar.
> 
> Seems that the build went [okay](https://github.com/Delawen/syndesis/runs/1887069441?check_suite_focus=true#step:9:142) on your fork.

It finishes right because I do the first `mvn verify ` with `-fn` so it doesn't fail. But we get one error:

And if we build with jacoco included, then we will get this error on a normal build:

```
Error:  Failed to execute goal org.codehaus.mojo:build-helper-maven-plugin:3.0.0:bsh-property (generate-merged-settings) on project s2i: error during Beanshell script execution: Sourced file: inline evaluation of: ``import java.io.*;                 import org.apache.maven.settings.*;            . . . '' : Method Invocation SettingsUtils.merge -> [Help 1]```
```

Only if jacoco is added to the build. I'm going to check why (on this first iteration I just worried it worked with ''-fn'')

**Forget about this, it was because I was doing `mvn verify install` instead of `mvn install` which means `verify` was executed twice.**

Delawen:
> Hmm, [code coverage is at 0%](https://sonarcloud.io/component_measures?id=syndesis-parent&metric=coverage&view=list), that doesn't seem right.

It seems we have to add the argLine of surefire plugin.

zregvart:
> > Hmm, [code coverage is at 0%](https://sonarcloud.io/component_measures?id=syndesis-parent&metric=coverage&view=list), that doesn't seem right.
> 
> It seems we have to add the argLine of surefire plugin.

There is this warning in the logs:

```
2021-02-12T13:01:52.1992334Z [WARNING] No coverage report can be found with sonar.coverage.jacoco.xmlReportPaths='**/target/**/jacoco.xml'. Using default locations: target/site/jacoco/jacoco.xml,target/site/jacoco-it/jacoco.xml,build/reports/jacoco/test/jacocoTestReport.xml
```

We have `jacoco.exec` but no `jacoco.xml` files, which would mean that the `jacoco:report` wasn't run. I think adding `./mvnw -pl '!:extension-bom,!:integration-bom' jacoco:report` would help with that. Let me try that. (also see [MMF-1651](https://jira.sonarsource.com/browse/MMF-1651))

Delawen:
Superseeded by a smaller approach.

PR #9328, opened 2021-02-09
ENTESB-15633 - Main build on Github Actions

KurtStam:
https://issues.redhat.com/browse/ENTESB-15633


zregvart (approved):
LGTM

.github/workflows/on_pr.yml
```diff
@@ -0,0 +1,46 @@
+name: On PR
+
+on: 
+  pull_request
```
zregvart, line 4: Actually, this can be
```suggestion
  pull_request_target
```

It's a bit safer that way.

Delawen:
```There are test failures.```

Did we... mess up tests somewhere?

KurtStam:
@Delawen I've disabled the test that is failing: https://issues.redhat.com/browse/ENTESB-15760 it failed both on GH and CircleCI

.github/workflows/on_pr.yml
pakhuis2e: Sounds good. I will do it in a follwup PR

KurtStam: Sounds good - will do that in a followup PR or else it does not run with this PR.

PR #9337, opened 2021-02-10
backport ENTESB-15633 - Main build on Github Actions

KurtStam:


Delawen:
`Build Error` failing... is that on purpose?

zregvart:
> `Build Error` failing... is that on purpose?

I think we need to remove the CircleCI webhook integration

PR #9351, opened 2021-02-12
feature(analysis): Add SonarCloud at the end of the main building job

Delawen:
Do not merge yet, I'm testing if this works.


Delawen:
The definition of the workflow was bad and github doesn't know how to recover from that, even after another push.

PR #9354, opened 2021-02-12
fix: remove references to `master` branch

zregvart:
We had several references to `master` branch, now that we no longer have
a `master` branch these are no longer valid. The incremental build
support also now uses `GITHUB_*` environment variables as we no longer
build on CircleCI but on GitHub actions.

I've also taken the liberty of removing some of the unused scripts from the tools directory.


zregvart:
retest this please

PR #9362, opened 2021-02-15
ENTESB-15773: Removes legacy operator deployment config

phantomjinx:
* Due to changing the operator from a deployemnt-config to a deployment, it
  is possible for a 1.11 and 1.12 to be installed at the same time. In
  order for 1.12 to replace 1.11, the former removes the deployment-config
  of 1.11 when it is first run-up. This uninstalls 1.11 operator leaving
  the operands intact and allows the 1.12 operator to be deployed and
  upgrade the operands accordingly.

* run.go
 * On the first run-up of the operator, checked for an operator dc and
   removes it. Otherwise, it does nothing.

* syndesis-controller.go
 * Removes unused function

### DIFF
--- a/install/operator/pkg/cmd/internal/run/run.go
+++ b/install/operator/pkg/cmd/internal/run/run.go
@@ -17,17 +17,21 @@
 package run
 
 import (
+	"context"
 	"fmt"
 	"runtime"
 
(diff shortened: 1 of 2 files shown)


phantomjinx:
Incorrect branch and jira.

PR #9367, opened 2021-02-16
feature(sonarcloud.io): Explicitly set sources and binaries for coverage

Delawen:


Delawen:
There are more things we can add, but let's start small: https://docs.sonarqube.org/latest/analysis/languages/java/

PR #9368, opened 2021-02-17
Minor bugfixing

Delawen:
Cleaning up the code a bit

### DIFF
--- a/app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
+++ b/app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
@@ -237,7 +237,8 @@ public class GenerateConnectorInspectionsMojo extends AbstractMojo {
             return shape;
         }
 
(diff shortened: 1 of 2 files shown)


app/connector/support/maven-plugin/src/main/java/io/syndesis/connector/support/maven/GenerateConnectorInspectionsMojo.java
```diff
@@ -237,7 +237,8 @@ private DataShape generateInspections(Connector connector, DataShape shape) thro
             return shape;
         }
 
-        getLog().info("Generating inspections for connector: " + connector.getId().get() + ", and type: " + shape.getType());
+        getLog().info("Generating inspections for connector: " +
+            connector.getId().orElse("NULL") + ", and type: " + shape.getType());
```
zregvart, line 241: Would be better to use placeholders here, string concatenation from logging uses cpu/memory (see http://www.slf4j.org/faq.html#logging_performance). I'm fairly sure that error prone should have caught this, not sure why this wasn't the case. PMD for sure should have [caught this](https://pmd.github.io/latest/pmd_rules_java_errorprone.html#invalidlogmessageformat).

```suggestion
        getLog().info("Generating inspections for connector: {} , and type: {}", connector.getId().orElse("NULL"), shape.getType());
```

PR #9369, opened 2021-02-17
Small refactors and removing code smells. No bugs here, just refactor.

Delawen:
More small refactors suggested by sonarcloud.

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/TagFinder.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/TagFinder.java
@@ -25,7 +25,7 @@ import java.util.TreeSet;
  */
 public class TagFinder {
 
-    private final SortedSet<String> tags = new TreeSet<String>();
+    private final SortedSet<String> tags = new TreeSet<>();
 
     public TagFinder add(ListResult<? extends WithTags> items) {
         if (items.getItems()!=null) {
--- a/app/common/model/src/main/java/io/syndesis/common/model/action/ActionDescriptor.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/action/ActionDescriptor.java
@@ -55,6 +55,7 @@ public interface ActionDescriptor {
 
     @JsonIgnore
     default Map<String, ActionDescriptorStep> getPropertyDefinitionStepsAsMap() {
-        return getPropertyDefinitionSteps().stream().collect(Collectors.toMap(p -> p.getName(), p -> p));
+        return getPropertyDefinitionSteps().stream().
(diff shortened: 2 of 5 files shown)


app/integration/runtime/src/main/java/io/syndesis/integration/runtime/logging/IntegrationLoggingListener.java
```diff
@@ -31,27 +31,25 @@ public IntegrationLoggingListener(ActivityTracker tracker) {
...
-        final String activityId = ActivityTracker.getActivityId(exchange);
-        if (activityId != null) {
-            final Marker marker = camelLogger.getMarker();
-            final String step = marker != null ? marker.getName() : "null";
-
-            final Message in = exchange.getIn();
-            String stepTrackerId = in.getHeader(IntegrationLoggingConstants.STEP_TRACKER_ID, String.class);
-            if( stepTrackerId == null ) {
-                stepTrackerId = KeyGenerator.createKey();
+        if (tracker != null) {
```
zregvart: Prefer early exit
```suggestion
        if (tracker == null) {
        	return message;
        }
```

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/logging/IntegrationLoggingListener.java
```diff
@@ -31,27 +31,25 @@ public IntegrationLoggingListener(ActivityTracker tracker) {
...
+            final String activityId = ActivityTracker.getActivityId(exchange);
+            if (activityId != null) {
```
zregvart:
```suggestion
            if (activityId == null) {
	            return message;
	        }
```

zregvart (commented):
Nitpicks

app/integration/runtime/src/main/java/io/syndesis/integration/runtime/logging/IntegrationLoggingListener.java
Delawen: That was one of the things detected as code smell because you had two different returns for the same value.

zregvart: Well, that smell was wrong. You could do:
```
        if (tracker == null || ActivityTracker.getActivityId(exchange) == null) {
            return message;
        }
```

to make it happy, but that leads to calling `ActivityTracker.getActivityId(exchange)` twice in happy day scenario.

Delawen: I will just remove the change then :)

PR #9370, opened 2021-02-18
fix: Missing patch body in FHIR Patch operation

rkorytkowski:
ENTESB-14237


Delawen:
Re-running integration tests, but looks good.

PR #9379, opened 2021-02-22
General bugfixing 2

Delawen:
Removing some code smells, no bugs.


app/server/logging/jaeger/src/main/java/io/syndesis/server/logging/jaeger/service/JaegerActivityTrackingService.java
```diff
@@ -66,7 +66,7 @@ public JaegerActivityTrackingService(JaegerQueryAPI jaegerQueryApi) {
             Activity activity = null;
             ArrayList<ActivityStep> steps = new ArrayList<>();
 
-            if (trace.spans != null && trace.spans.size() >= 1) {
+            if (trace.spans != null) {
```
zregvart, line 69:
```suggestion
            if (trace.spans == null) {
                continue;
            }
```

Not really important, but since you're refactoring, perhaps on the next PR if you get around to it...

zregvart (approved):

PR #9380, opened 2021-02-22
chore: endpoint parameters in component proxy

zregvart:
When using Pool Enrich EIP, Consumer that's being pooled operates on a
Exchange created from scratch, so we can't pass data from the Exchange
used in the Route flow preceding Poll Enrich. The only option is to
pass data via endpoint parameters. The endpoint that's accessible from
the route is Component Proxy endpoint, and as implemented did not
support any endpoint parameters.

So to pass data from the running exchange into the Poll Enriched
Consumer via endpoint parameters, Component Proxy needed to support
endpoint parameters, i.e. when generating Component Proxy Endpoint we
need to separate parameters that will be passed on the Component Proxy
Endpoint vs parameters that will be passed to the delegate Endpoint
directly. For this `kind` property of the Action now supports value
`"proxyParameter"` in addition to `"path"` and `"parameter"` used to
pass parameters to the delegated Endpoint. When `kind` is set to
`"proxyParameter"` that parameter is passed via the Component Proxy
endpoint, e.g. `component-F-S:?param=value`, instead of setting
parameter on the delegated Endpoint.

With this in Poll Enrich EIP we now can use dynamic endpoints, with the
first use case in the FTP connector when the `fileName` property is
passed via `CamelFileName` header to the FTP Polling Consumer within the
Poll Enrichment EIP.

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/WithConfigurationProperties.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/WithConfigurationProperties.java
@@ -63,6 +63,12 @@ public interface WithConfigurationProperties {
     }
 
     @JsonIgnore
+    default boolean isProxyEndpointProperty(final String propertyName) {
+        final ConfigurationProperty property = getProperties().get(propertyName);
+        return property != null && "proxyParameter".equals(property.getKind());
+    }
+
+    @JsonIgnore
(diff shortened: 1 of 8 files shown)


KurtStam (approved):
Nice :)

KurtStam:
retest this please

PR #9385, opened 2021-02-23
fix: make sure messages are captured in pollenrich

zregvart:
The default aggregation strategy in `PollEnricher` does not persist the
headers needed by Syndesis to perform output message capture. Seems that
aggregation tries to copy data from the `Exchange` created in the
`PollEnrich` but it ends up just copying everything from that `Exchange`
while discarding any values from the previous `Exchange`. We need the
actions that utilize `PollEnrich` to maintain message capture so that
the data is available for mapping. This should also maintain the
activity tracking.

### DIFF
--- a/app/integration/runtime/src/test/java/io/syndesis/integration/runtime/handlers/ConnectorStepHandlerPollEnrichIT.java
+++ b/app/integration/runtime/src/test/java/io/syndesis/integration/runtime/handlers/ConnectorStepHandlerPollEnrichIT.java
@@ -30,6 +30,7 @@ import io.syndesis.common.model.integration.Step;
 import io.syndesis.common.model.integration.StepKind;
 import io.syndesis.integration.component.proxy.ComponentProxyComponent;
 import io.syndesis.integration.component.proxy.ComponentProxyCustomizer;
+import io.syndesis.integration.runtime.capture.OutMessageCaptureProcessor;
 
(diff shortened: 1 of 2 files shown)


app/integration/runtime/src/test/java/io/syndesis/integration/runtime/handlers/ConnectorStepHandlerPollEnrichIT.java
```diff
@@ -174,6 +175,7 @@ public void shouldSupportPollEnriching() throws Exception {
                     .putConfiguredProperty("period", "1000")
                     .build(),
                 new Step.Builder()
+                    .id("expected")
```
KurtStam, line 178: why 'expected' ? I mean shouldn't it be something more dynamic incase we have more then one pollEnrichers?

zregvart: To [assert it easier](https://github.com/syndesisio/syndesis/pull/9385/files#diff-fda11763c721dc439cde8d9d6ff6e8f294d8f4b75e24f7e9975e100f9efb32f4R210), otherwise it gets a randomly generated id, so we can't assert that it's this step that we keep track of. The test now covers a case where captured messages map was missing output from poll enrich.

PR #9390, opened 2021-02-25
chore: Update to AtlasMap 2.1.6

igarashitm:
Release Note: https://github.com/atlasmap/atlasmap/releases/tag/atlasmap-2.1.6


igarashitm:
@christophd do you see what's wrong with `DBToHttp#testDBToHttp()`? Does the order of the records in DB matter here?

igarashitm:
I'm trying to run these IT locally by following what is done in GH checks, but it always fails with `java.lang.NoClassDefFoundError: Could not initialize class io.syndesis.test.itest.SyndesisIntegrationTestSupport`

PR #9403, opened 2021-03-01
refactor: use JUnit Jupiter

zregvart:
Thoroughly refactors tests to depend only on JUnit 5 (Jupiter). To make
this transition a stub with empty JUnit classes was needed
(`junit4-stub` module) to help with compiling against Testcontainers and
Citrus framework classes (amongst others).

Crux of the refactor was changing package names for `Test` annotations
and using builtin JUnit 5 parameterized test support.

Some tests that were disabled and unmaintained were outright removed.
This is the case with the system-test module which we do not run and
haven't maintained in years now.

Also:  fix: make template processing thread safe (1cf4bb7)

Issue popped up when switching to JUnit Jupiter, which runs tests in
random order caused issues around the shared `StringBuilder`. This makes
the `StringBuilder` a local variable and introduces a context object
that holds any values needed by the template processor. Now there is no
shared state to cause thread safety issues.

And:  refactor: sql connector tests (0a216e6)

These tests were refactored a bit more in the JUnit 5 migration, so this
commit contains changes just to the SQL connector.


christophd:
@zregvart Citrus should also provide JUnit5 support. Did you have a look into this already?

```
@ExtendWith(CitrusSpringSupport.class)
```

zregvart:
> @zregvart Citrus should also provide JUnit5 support. Did you have a look into this already?

The version 2.8.0 of Citrus in citrus-core has a dependency both on JUnit 4 and JUnit 5. With JUnit 4 completely removed from classpath this doesn't compile. I've found no reason to use the extension, the integration tests were migrated over without it as the dependency on JUnit 4 in the integration tests wasn't as large.

christophd:
@zregvart ah I see, the dependency issues are solved in Citrus 3.x but we are stuck with 2.8.0 because of Camel 2.x usage

zregvart:
Actually, @christophd I do need to use the `CitrusExtension` here, I can't derive from JUnit 4 based base class.

zregvart:
@christophd you might be interested in changes made in cb9bef80004834e5819cd418299665425678f96e, its a bit difficult to mix Spring `@Configuration` and Citrus application contexts as they're separate. I've tried to bridge them by setting Spring application context's parent to the Citrus application context, but this caused issues when a `@Bean` declared in `@Configuration` was required in Citrus's application context. Not sure what would be idiomatic Citrus solution for this, seems that setting `citrus.spring.application.context` might help but that restricts the use to a single `@Configuration` class (we had many).

zregvart:
retest this please

zregvart:
retest this please

zregvart:
This is too big of an effort to rebase, and the end to end tests seem to fail due to unrelated issues.

PR #9387, opened 2021-02-24
ENTESB-15225: Adds affinity and toleration values on install

phantomjinx:
* During install phase, adds the affinity and toleration values to the
  infrastructure deployment configs
 * Not been possible previously due to not having the deployment configs
   readily to hand as typed objects

* install.go
 * Processes each resource on install and determines if they're a DC
 * If so, adds in the content from the Syndesis resource as blocks
 * Should the syntax in the Syndesis resource be incorrect then there will
   be errors in the operator log

* install_test.go
 * Tests for the PreProcessDeploymentConfig function

PR #9404, opened 2021-03-02
Set affinity/toleration only for integrations

claudio4j:
As #9387 PR sets the affinity/tolerations in install action
it makes no sense to patch the infrastructure components in post
install.
This PR is to remove the patch for infrastructure, the pod_scheduling
action patches only the integration components.

https://github.com/syndesisio/syndesis/pull/9387


claudio4j:
retest this please

phantomjinx (approved):
LGTM

PR #9416, opened 2021-03-05
chore: initial support for mapping CSV data format

zregvart:
This adds initial support for mapping CSV data format, helpful in cases
where data shape can be defined by the user, as we (currently) don't
have any connectors that provide CSV data shape metadata.

For any connector that specifies `ANY` datashape, in the UI user can
provide a sample CSV file, for which the shape of the document will be
generated and available for mapping.

The drawback of the this implementation is that no parameters are passed
to the CSV inspection service, so for example, a way of specifying
`firstRecordAsHeader` parameter to enable column names to appear as
field names in the mapping is not available.

To do:
 - [x] UI ConfigMap missing CSV configuration (operator needs update)
 - [x] Tests

Stretch:
 - [x] Collection support
 - [x] CSV service parameter support

Ref. https://issues.redhat.com/browse/ENTESB-15928


avano:
retest this please

PR #9424, opened 2021-03-09
chore: add 2.x daily release

zregvart:
Ref. https://issues.redhat.com/browse/ENTESB-15938


zregvart:
So my plan is to try this out on 2.x and then prepare a PR for 1.12.x. I've disabled the 2.x daily build on ci.fabric8.io.

PR #9426, opened 2021-03-10
chore: daily release using GitHub Actions

zregvart:
This replaces the daily release jobs running on ci.fabric8.io. Scheduled
GitHub Actions are only picked up from the default branch so the
done on #9424 won't help, so this should build both 1.12.x and 2.x
branches.


Delawen (approved):

zregvart:
I'll disable the 1.12.x daily build on ci.fabric8.io, we'll see about the results tomorrow.

PR #9448, opened 2021-03-18
Proposed fix for release build failure due to not including junit4-stub dependency

phantomjinx:


phantomjinx:
retest this please

app/pom.xml
```diff
@@ -322,6 +322,7 @@
         <basepom.check.skip-all>true</basepom.check.skip-all>
         <basepom.it.skip>true</basepom.it.skip>
         <skipTests>true</skipTests>
+        <maven.test.skip>true</maven.test.skip>
```
zregvart: Not sure we need this, compiler, resource, surefire and failsafe are removed from running below, so this doesn't make a difference.

app/pom.xml
```diff
@@ -371,22 +372,6 @@
...
-            <executions>
-              <execution>
-                <id>attach-sources</id>
-                <goals>
-                  <goal>jar</goal>
-                </goals>
-              </execution>
-            </executions>
-          </plugin>
-
```
zregvart: Don't we want to release sources?

app/pom.xml
```diff
@@ -603,17 +588,6 @@
...
-            <groupId>com.mycila</groupId>
-            <artifactId>license-maven-plugin</artifactId>
-            <executions>
-              <execution>
-                <id>basepom.default</id>
-                <phase />
-              </execution>
-            </executions>
-          </plugin>
-
```
zregvart: This disables license header check, why run that if we're not running any other checks?

app/s2i/pom.xml
```diff
@@ -187,7 +187,36 @@
...
+                </licenseUrlReplacement>
+              </licenseUrlReplacements>
+              <skipAggregateDownloadLicenses>false</skipAggregateDownloadLicenses>
+              <executeOnlyOnRootModule>false</executeOnlyOnRootModule>
+              <licensesOutputDirectory>${image.build.directory}/licenses</licensesOutputDirectory>
+              <licensesOutputFile>${image.build.directory}/licenses.xml</licensesOutputFile>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
```
zregvart: This restores the `license-maven-plugin` which will try to resolve test dependencies, and thus junit4-stub, which hasn't been built (not part of the reactor).

zregvart (changes requested):
Yeah, I don't think this will help, most likely you had the `junit4-stub` installed in your local Maven repository.

app/s2i/pom.xml
phantomjinx: Back to drawing board.

app/pom.xml
phantomjinx: Restoring....

app/pom.xml
phantomjinx: Obviously removed the wrong plugin. These is a duplicate error  as there is 2 maven-source-plugin in the release profile.

app/pom.xml
phantomjinx: Thought this might solve the problem but sadly not.

phantomjinx:
retest this please

PR #9457, opened 2021-03-24
Debugging full-build pr action

phantomjinx:
* du whole / filesystem to understand what files fill up device

* Flash build syndesis to quicken up build time

* Re-instate docker save action to determine if upload-artifact still works
  when provided with just one image (server)

### DIFF
--- a/.github/workflows/on_pr.yml
+++ b/.github/workflows/on_pr.yml
@@ -56,11 +56,9 @@ jobs:
           df -h
           echo Space in Docker ...
           docker system df
-          echo Filesystem Details ...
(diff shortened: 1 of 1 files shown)


KurtStam (commented):
So if we do a flash build here, what is doing running the regular unittests?

phantomjinx:
> So if we do a flash build here, what is doing running the regular unittests?

Is this not taking care of that -> Integration Tests 1.x branch / integration-tests (pull_request)

If it isn't then sure we should restore the 'full' build but it seemed like duplication??

PR #9462, opened 2021-03-26
Use latest release to download syndesis-operator, excluding 2.x

claudio4j:
When instaling syndesis with syndesis script, prints the syndesis
version, so the user can check if the correct version is used.

Excludes releases from 2.x branch when installing with syndesis script.

### DIFF
--- a/tools/bin/commands/util/operator_funcs
+++ b/tools/bin/commands/util/operator_funcs
@@ -89,18 +89,17 @@ download_operator_binary() {
     esac
 
     if [[ "${pattern}" != "unknown" ]]; then
-        release_uri="https://api.github.com/repos/syndesisio/syndesis/releases?per_page=1"
+        release_uri="https://api.github.com/repos/syndesisio/syndesis/releases?per_page=4"
         release_tag="$(readopt --tag)"
 
         if [[ -n "$release_tag" ]]; then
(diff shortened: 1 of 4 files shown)


tools/bin/commands/util/operator_funcs
```diff
@@ -100,7 +100,9 @@ download_operator_binary() {
             | grep browser_download_url \
             | grep ${OPERATOR} \
             | grep ${pattern} \
-            | cut -d '"' -f 4)
+            | grep -v 'download/2' \
+            | cut -d '"' -f 4 \
+            | head -1)
```
zregvart: Would it make a more sense to use `jq` here? For example based on this:

```shell
curl -s https://api.github.com/repos/syndesisio/syndesis/releases\?per_page\=4 \
| jq -r '.[]| select(.tag_name == "1.12.0-20210325")| .assets[] | select(.name |startswith("syndesis-operator-linux-amd64"))| .browser_download_url'
```

zregvart (approved):
Looks straightforward enough, I wonder if we should be using `jq` instead of `grep` to parse the JSON...

tools/bin/commands/util/operator_funcs
claudio4j: Using jq now.

PR #9473, opened 2021-04-06
feat(server): integration user labels

squakez:
Adding the possibility to include integration labels that can eventually been used for UI filtering.

Ref ENTESB-14559


squakez:
cc @kahboom

PR #9475, opened 2021-04-08
feat(server): auditing

zregvart:
Implements auditing support

Ref. https://issues.redhat.com/browse/ENTESB-12232


zregvart:
@phantomjinx I would appreciate if you can have a look at 97d4029, I'm not sure if that's the correct Yaml file to change (doesn't seem up to date), and I'm not sure if some version somewhere needs to be bumped when adding a field in the CRD.

Delawen (approved):
Wow, this grow in complexity since last time I checked!

Delawen:
I think QE is failing due to infrastructure issues. And this change is not something they are testing anyway. So merging? @phantomjinx

zregvart:
Fixed the operator changes, so now the auditing feature seems to work, still not sure about the Syndesis CRD version (if it needs to be bumped or not).
Fixed per-object audit filter, so now only marked objects are audited, bug was that only object creation was checking for the mark.
Added dynamic configuration reload support, so now auditing can be toggled on/off without server restart via the Syndesis CR.

zregvart:
retest this please

zregvart:
retest this please

zregvart:
This has been on hold for a while in review and in flaky test. I'm going to merge this and deal with the fallout as it happens.

PR #9478, opened 2021-04-09
fix(ui) entesb-16038 - unable to delete middle integration step

kahboom:
fixes [entesb-16038](https://issues.redhat.com/browse/ENTESB-16038)

Immer [automatically freezes](https://immerjs.github.io/immer/freezing) any state trees modified using `produce`, which is typically what we'd want. However, "If auto freezing is enabled, recipes are not entirely side-effect free: Any plain object or array that ends up in the produced result, will be frozen, even when these objects were not frozen before the start of the producer!" Also: " Immer freezes everything recursively, for large data objects that won't be changed in the future this might be over-kill, in that case it can be more efficient to shallowly pre-freeze data using the freeze utility."

Specific to this issue, we are modifying the `steps` array prior to the integration being saved for the first time, and disabling auto freezing resolves this issue. To avoid causing future bugs, I've also pre-frozen the newly created `integration` object, which does a shallow freeze and allows `steps` to behave as a draft. Additional unit tests have been added, though as usual can be improved upon.

Note: Auto-freezing is normally disabled in production by Immer, and disabling it would only disable it in development mode.

### DIFF
--- a/app/ui-react/packages/api/src/helpers/__test__/integrationFunctions.spec.ts
+++ b/app/ui-react/packages/api/src/helpers/__test__/integrationFunctions.spec.ts
@@ -1,6 +1,19 @@
-import { DataShape, Step } from '@syndesis/models';
+import {
+  Connection,
+  DataShape,
+  Flow,
+  Integration,
+  Step,
+} from '@syndesis/models';
+import produce from 'immer';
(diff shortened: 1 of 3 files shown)


gashcrumb (approved):
:+1:

PR #9489, opened 2021-04-13
refactor: Remove duplicated code from Camel ServiceNow

Delawen:
Fixing the "TODO" that says to replace the MetadataManager once we get to Camel 2.22. We are on Camel 2.23+

Now, less code to maintain.


app/connector/servicenow/src/main/java/io/syndesis/connector/servicenow/ServiceNowMetadataRetrieval.java
```diff
@@ -34,16 +34,14 @@
 import org.apache.camel.CamelContext;
 import org.apache.camel.component.extension.MetaDataExtension;
 import org.apache.camel.component.servicenow.ServiceNowConstants;
+import org.apache.camel.component.servicenow.ServiceNowMetaDataExtensionWrapper;
 import org.apache.camel.util.ObjectHelper;
 
 public final class ServiceNowMetadataRetrieval extends ComponentMetadataRetrieval {
```
lburgazzoli, line 40: Is this class still needed ?

app/connector/servicenow/src/test/java/io/syndesis/connector/servicenow/ServiceNowMetaDataExtensionTest.java
```diff
@@ -24,18 +24,21 @@
...
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class ServiceNowMetaDataExtensionTest {
 
     @Test
+    @Disabled("This is not a unit test but an integration test")
     public void retrieveTableListTes() {
         CamelContext context = new DefaultCamelContext();
-        ServiceNowMetaDataExtension extension = new ServiceNowMetaDataExtension(context);
+        MetaDataExtension extension = ServiceNowMetaDataExtensionWrapper.getServiceNowMetaDataExtension(context);
```
lburgazzoli: I think you can get the meta extension from the ServicewNow component now

app/connector/servicenow/src/main/java/io/syndesis/connector/servicenow/ServiceNowMetadataRetrieval.java
Delawen: The constructors for the `ServiceNowMetaDataExtension` are protected so I need to create them from somewhere inside the same package.

This was the quickest/cleanest solution I could think of for that. If you know a better approach, I'm all ears.

app/connector/servicenow/src/test/java/io/syndesis/connector/servicenow/ServiceNowMetaDataExtensionTest.java
Delawen: Ah, then that's why the constructors are protected?

lburgazzoli: yes, it is not supposed to be used outside the component

Delawen: But the class is also protected:
https://github.com/apache/camel/blob/master/components/camel-servicenow/camel-servicenow-component/src/main/java/org/apache/camel/component/servicenow/ServiceNowMetaDataExtension.java#L59

That's why I am using MetadataExtension as the return value here: I cannot even import it.

lburgazzoli: Something like that should work:

```java
SalesforceComponent component = context.getComponent("salesforce", SalesforceComponent.class);
MetaDataExtension meta = component.getExtension(MetaDataExtension.class);
```

Delawen: Changed according to this.

PR #9490, opened 2021-04-13
operator-sdk and removal of api deprecations

phantomjinx:


claudio4j:
Just run `GOOSLIST=linux ./build.sh --operator-build go --image-tag 1.12`

it changed three files
```
        modified:   config/crd/bases/syndesis.io_syndesises.yaml
        modified:   pkg/generator/assets/install/cluster/syndesis.yml
        modified:   pkg/generator/assets_vfsdata.go
```
Changing files may be confusing for each build, and the developer may commit them if that should not be the case.
```
-    controller-gen.kubebuilder.io/version: (devel)
+    controller-gen.kubebuilder.io/version: v0.3.0
```

install/operator/config/Makefile
```diff
@@ -0,0 +1,287 @@
...
+ROLE := rbac-role
+
+#
+# Allows for resources to be loaded from outside the root location of
+# the kustomize config file. Ensures that resource don't need to be
+# copied around the file system.
+#
+# See https://kubectl.docs.kubernetes.io/faq/kustomize
+#
+KOPTIONS := --load_restrictor LoadRestrictionsNone
```
claudio4j: This KOPTIONS is already declared above.

phantomjinx:
> Just run `GOOSLIST=linux ./build.sh --operator-build go --image-tag 1.12`
> 
> it changed three files
> 
> ```
>         modified:   config/crd/bases/syndesis.io_syndesises.yaml
>         modified:   pkg/generator/assets/install/cluster/syndesis.yml
>         modified:   pkg/generator/assets_vfsdata.go
> ```
> 
> Changing files may be confusing for each build, and the developer may commit them if that should not be the case.
> 
> ```
> -    controller-gen.kubebuilder.io/version: (devel)
> +    controller-gen.kubebuilder.io/version: v0.3.0
> ```

This is a factor but not unprecedented in that other repositories are quite used to doing it this way, eg. fuse-console, camel-k. It comes down to whether we even commit these files since they are auto-generated (yes for visibility / no for visible false-positive changes).
At the moment, the status-quo is to commit what is generated in the config/manifests directory but not commit the config/bundle directory.

We can mitigate this to an extent, eg. I have compiled my own version of controller-gen and its unlikely many others will do that so the devel/0.3.0 change could probably be committed and most of the time will never occur.

install/operator/config/Makefile
claudio4j: It outputs error 
```
Error: unknown flag: --load_restrictor
```
From kustomize help, the correct parameter is `--load-restrictor`

claudio4j:
> We can mitigate this to an extent, eg. I have compiled my own version of controller-gen and its unlikely many others will do that so the devel/0.3.0 change could probably be committed and most of the time will never occur.

Commiting those files sounds good, if that causes confusion later, we can think of .gitignore them. Thanks for the overview.

install/operator/config/Makefile
phantomjinx: Version 3 of kustomize uses underscores and this is what was being downloaded by the `kustomize` rule in the Makefile. I've upped the version to 4.0.5 (latest) which uses hyphens instead.

claudio4j (approved):

PR #9491, opened 2021-04-13
[maintenance] Small code smells

Delawen:
This doesn't fix any bug, just simplifies maintainability.

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/integration/step/template/AbstractTemplatePreProcessor.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/integration/step/template/AbstractTemplatePreProcessor.java
@@ -59,9 +59,6 @@ abstract class AbstractTemplatePreProcessor<C extends ProcessingContext> impleme
         }
     }
 
-    @Override
-    public abstract boolean isMySymbol(String literal);
-
     private boolean isSymbol(String symbol) throws TemplateProcessingException {
         if (isMySymbol(symbol)) {
             return true;
--- a/app/common/model/src/main/java/io/syndesis/common/model/support/Equivalencer.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/support/Equivalencer.java
@@ -53,7 +53,7 @@ public class Equivalencer implements StringConstants {
 
     private boolean push(String id, Class<?> klazz) {
         if (failureContext == null) {
-            failureContext = new ArrayDeque<EquivContext>();
+            failureContext = new ArrayDeque<>();
         }
 
         EquivContext ctx = new EquivContext(id, klazz);
@@ -63,7 +63,7 @@ public class Equivalencer implements StringConstants {
 
     private boolean push(String id, Class<?> klazz, String failingProperty, Object a, Object b) {
         if (failureContext == null) {
-            failureContext = new ArrayDeque<EquivContext>();
+            failureContext = new ArrayDeque<>();
         }
 
         EquivContext ctx = new EquivContext(id, klazz);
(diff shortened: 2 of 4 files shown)


app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/auth/oauth/OAuthState.java
```diff
@@ -26,16 +26,30 @@
     private volatile State state;
 
     private static final class State {
-        private final String accessToken;
+        private final String _accessToken;
 
-        private final long accessTokenExpiresAt;
+        private final long _accessTokenExpiresAt;
```
zregvart: Not sure about this, why is this a smell?

app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/auth/oauth/OAuthState.java
```diff
@@ -26,16 +26,30 @@
...
+
+        public synchronized String getRefreshToken() {
+            return _refreshToken;
+        }
 
         private State(final String accessToken, final long accessTokenExpiresAt, final String refreshToken) {
-            this.accessToken = accessToken;
-            this.refreshToken = refreshToken;
-            this.accessTokenExpiresAt = accessTokenExpiresAt;
+            synchronized (this){
```
zregvart: This should not be required, why lock here?

app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/auth/oauth/OAuthState.java
```diff
@@ -26,16 +26,30 @@
...
+            return _accessToken;
+        }
+
+        public synchronized long getAccessTokenExpiresAt() {
+            return _accessTokenExpiresAt;
+        }
+
+        public synchronized String getRefreshToken() {
+            return _refreshToken;
+        }
```
zregvart: These are accessors for `final` fields of an `final` immutable class, why the need for locking?

app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/auth/oauth/OAuthState.java
Delawen: I was trying to make the "volatile" useful. When applied to objects, volatile does not warrantee multithreading readings.

app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/auth/oauth/OAuthState.java
Delawen: Same as above.

app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/auth/oauth/OAuthState.java
Delawen: If the "volatile" label is not important in this context, then we can revert this commit.

zregvart: I think in this case `volatile` is is used correctly, it can be replaced with `AtomicReference` as an alternative that might be less confusing. Adding synchronization, especially in the constructor doesn't make much sense. Going through setters for `private final` field also doesn't make much sense, in fact in this case it might lead to `StackOverflowErrors`.

Delawen: Ok, then I will remove this commit from the list.

PR #9495, opened 2021-04-14
chore: upgrade operator-sdk to 0.19.4

zregvart:
I was getting:

```
verifying github.com/operator-framework/operator-sdk@v0.19.3: checksum mismatch
	downloaded: h1:MHp1iMPppIm+ujTwnnm8GJbztW4CXW+EqM9ysiXRf4E=
	go.sum:     h1:gWSpeKFERW42wCgl35yC2oRqel2o6oXy/0FMwE0RiFE=

SECURITY ERROR
This download does NOT match an earlier download recorded in go.sum.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
```

So I removed `go.sum` and run:

```
$ go mod vendor
$ go mod tidy
```

This removed quite a bit of lines from `go.sum` and also changed the
aforementioned line:

```diff
-github.com/operator-framework/operator-sdk v0.19.3 h1:gWSpeKFERW42wCgl35yC2oRqel2o6oXy/0FMwE0RiFE=
+github.com/operator-framework/operator-sdk v0.19.3 h1:MHp1iMPppIm+ujTwnnm8GJbztW4CXW+EqM9ysiXRf4E=
```

I've searched to see what the correct hash is but could not find any
usage of that particular version on GitHub.

So next I've upgraded the operator-sdk to the currently last version of
the 0.19.x line, which is 0.19.4. For that version I can attest that at
some point in time the valid has was indeed the one we now have in
`go.sum`, as it looks exactly like the one in:

https://github.com/openshift/elasticsearch-operator/blob/6aeb7c0412bf6ec6f9f1c0b74369b742bdccdab0/go.sum#L789-L790


phantomjinx (commented):
LGTM

PR #9499, opened 2021-04-15
chore: Updates k8s dependencies as far as possible

phantomjinx:
* This brings forward deps for k8s.io/api and client-go as far as currently
  possible. The dependency holding them back is controller-runtime which
  will require updating to (as yet unreleased) 0.9.0.

* Fixes failing test in subscription_test that occurs as a result of the
  upgrade. The dynamic client has an issue but thankfully can now switch to
  using the runtime client instead so works around it.


claudio4j:
There is a warning in syndesis-operator log. Would it be feasible to upgrade this dependency ? I am using openshift 4.7.3
```
extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```

phantomjinx:
> There is a warning in syndesis-operator log. Would it be feasible to upgrade this dependency ? I am using openshift 4.7.3
> 
> ```
> extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
> ```

I spent a couple of days trying to pin this one down. It seems to be a warning tirggered by client-go which contains both ingress versions and simply selects the v1beta1 version as its first in the list. Syndesis at no point uses ingress at all. I even updated permissions.

claudio4j (approved):

PR #9508, opened 2021-04-16
fix(c/activemq): RMI/JMX deadlock in tests

zregvart:
Fixes deadlock by not starting the RMI used by JMX in in-vm ActiveMQ
brokers and fixes the test not to verify using equals on a Camel
component. These are integration tests disguised as unit tests, so not
much can be done to make them reliable, this is but a small step to try
to make them a bit more reliable.


zregvart:
Needs to be rebased on #9509

PR #9509, opened 2021-04-16
fix(c/http): fix timeout related to network/DNS

zregvart:
It's much better to listen on the loopback interface than on the public
in tests. Also fixes any DNS/network issues when running that lead to
slow execution and test timeouts.


zregvart:
Needs rebase on #9510

Issue #9504, opened 2021-04-16
FHIR connector test failures


zregvart:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[x] Bug report  
[ ] Documentation issue or request
</code></pre>

## Description

```
[INFO] Running io.syndesis.connector.fhir.FhirPatchWithBodyTest
Error:  Tests run: 2, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.388 s <<< FAILURE! - in io.syndesis.connector.fhir.FhirPatchWithBodyTest
Error:  io.syndesis.connector.fhir.FhirPatchWithBodyTest.shouldPatchWithBodyTest  Time elapsed: 0.163 s  <<< ERROR!
org.apache.camel.CamelExecutionException: Exception occurred during execution on the exchange: Exchange[ID-fv-az209-794-1618574321447-18-1]
	at io.syndesis.connector.fhir.FhirPatchWithBodyTest.shouldPatchWithBodyTest(FhirPatchWithBodyTest.java:46)
Caused by: org.apache.camel.RuntimeCamelException: 
Error invoking patchById with {stringId=Patient/1, patchBody=[{"op":"replace","path":"active","value":true}], extraParameters={}, preferReturn=null}: HTTP 404 Not Found: 
                                               Request was not matched
                                               =======================

-----------------------------------------------------------------------------------------------------------------------
| Closest stub                                             | Request                                                  |
-----------------------------------------------------------------------------------------------------------------------
                                                           |
PATCH                                                      | PATCH
/Patient/1?_format=xml                                     | /Patient/1                                          <<<<< URL does not match
                                                           |
                                                           |
-----------------------------------------------------------------------------------------------------------------------

Caused by: ca.uhn.fhir.rest.server.exceptions.ResourceNotFoundException: 
HTTP 404 Not Found: 
                                               Request was not matched
                                               =======================

-----------------------------------------------------------------------------------------------------------------------
| Closest stub                                             | Request                                                  |
-----------------------------------------------------------------------------------------------------------------------
                                                           |
PATCH                                                      | PATCH
/Patient/1?_format=xml                                     | /Patient/1                                          <<<<< URL does not match
                                                           |
                                                           |
-----------------------------------------------------------------------------------------------------------------------


[...]

PR #9510, opened 2021-04-16
fix(c/fhir): disable two failing tests

zregvart:
Ref. #9504


zregvart:
Needs rebase on #9511

PR #9511, opened 2021-04-16
fix(c/odata): listen on localhost in the tests

zregvart:
`ODataTestServer` wasn't configured to listen on `localhost`, which
leads to timeouts in the test. This is getting a bit tiresome.


zregvart:
Needs rebase on #9513

PR #9513, opened 2021-04-16
fix(c/odata2): listen on localhost in the tests

zregvart:
`ODataTestServer` wasn't configured to listen on `localhost`, which
leads to timeouts in the test. This is a bit tiresome.


zregvart:
Needs rebase on #9514

PR #9522, opened 2021-04-19
ENTESB-16083: CRD fields need to be specified as optional

phantomjinx:
* If fields are not specified as omitempty then they are determined as
  required in the CRD. This can cause extraneous validation issues when
  empty parent structures happen to be specified in the CR (as occurs in
  the product version).


zregvart (approved):
I was wondering why an empty (`maven: {}`) was required.

PR #9528, opened 2021-04-20
refactor(github): redo GitHub workflow definitions

zregvart:
Seems that we have accumulated a lot of complexity, unneeded duplication
and straight up unnecessary configuration in the GitHub workflow
definitions. This simplifies, optimizes and reduces the complexity of
those.

We have a single workflow `checks.yml` that applies to both pull
requests and to release branches (`N.M.x`).

Work done is minimized by checking what is changed, and build/checks are
only run for the subset of changes. For any change in the `docs/**` we
don't run any check; backend, operator and ui-react are built only if
changes have been made within their specific directories; images are
built for any change in the backend, operator or ui-react to keep the
image set consistent; and lastly integration tests are run for any
change within the backend.

fix(tools): don't build operator images by default (d925acd)

The convention within the `syndesis` tool is that if no argument is
given for an image to be built the build of the image should not be
performed. This extends this convention to the operator build as well.


zregvart:
I've tested this on my fork with:
 - [docs change](https://github.com/zregvart/syndesis/pull/9)
 - [ui change](https://github.com/zregvart/syndesis/actions/runs/767049502)
 - [backend change](https://github.com/zregvart/syndesis/actions/runs/767059654)
 - [operator change](https://github.com/zregvart/syndesis/actions/runs/767071222)

We can wait and see how those checks fare.

.github/workflows/checks.yml
```diff
@@ -0,0 +1,191 @@
...
+          path: |
+            ~/.m2/repository
+            !~/.m2/repository/io/syndesis
+          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-maven-
+      - name: Build and test
+        run: tools/bin/syndesis build --batch-mode --backend
+      - name: Publish Test Reports
+        if: always()
+        uses: scacap/action-surefire-report@v1
```
Delawen, line 71: Good idea!

.github/workflows/checks.yml
```diff
@@ -0,0 +1,191 @@
...
+    needs: setup
+    runs-on: ubuntu-latest
+    outputs:
+      backend: ${{ steps.filter.outputs.backend }}
+      images: ${{ steps.filter.outputs.images }}
+      operator: ${{ steps.filter.outputs.operator }}
+      ui: ${{ steps.filter.outputs.ui }}
+    steps:
+      - uses: actions/checkout@v2
+        if: needs.setup.outputs.isDefaultBranch == 'true'
```
Delawen, line 33: So if it is not the default branch... it does not run anything at all? Because the changes.outpus.* will be false if this is not run, right?

zregvart: From what I understood in the https://github.com/dorny/paths-filter action needs a git repository only for repository branches, not for pull requests. So this skips the https://github.com/actions/checkout for non-default branches (pull requests and every other branch). This could be expanded to include other branches from the repository. This was a first pass at this, and it'll work for `1.12.x`.

zregvart:
I'm trying to fix the issue that running `syndesis build -m ui-react` is not possible without previously running `syndesis build -m common` (as seen in the https://github.com/zregvart/syndesis/runs/2390379579?check_suite_focus=true). For this I need to de-activate the `non-staging` profile based on existance of `stagingDescription` **AND** `build-selection` properties. This is not possible in Maven out of the box and I'm looking into creating an extension to help with that. There are other extensions available, I've tried 3 or 4 of them, none of them work.

zregvart:
All workflows succeeded on my fork:
 - [1.12.x](https://github.com/zregvart/syndesis/actions/runs/775214831)
 - [UI change](https://github.com/zregvart/syndesis/actions/runs/775393965)
 - [backend change](https://github.com/zregvart/syndesis/actions/runs/775397678)
 - [operator change](https://github.com/zregvart/syndesis/actions/runs/775398981)

I think we can merge this now. WDYT @phantomjinx @Delawen cc @avano

PR #9530, opened 2021-04-21
ENTESB-16258: Ensure that CRD is installed even if all versions present

phantomjinx:
* CRD is not installed if version is already present. This stops the syntax
  of the CRD from being replaced.

* Ensure the versions are updated even if present to ensure any corrections
  to version format and syntax are added.

* Will only return message "cluster resources already installed" if the
  exact format of the CRD resource is already present.

===

 chore(operator): Tighten up removal of legacy operator
    
 * The product renames the operator deployment-config so use label selector
    to correctly identify the legacy operator instead.


mkralik3:
@phantomjinx  I have tried the generated images on cluster where the old CRD ( from 1.12.0-20210414) was installed and the installation succeeded (CRD was updated). :+1:

phantomjinx:
retest please

PR #9540, opened 2021-04-24
ENTESB-16286 Fix load procedures and functions in connector-sql

claudio4j:
PostgreSQL 11 started to support procedures, before 11, the jdbc metadata getProcedures returned a list of functions.
Then after PostgreSQL 11, the getProcedures() method, only returns procedures and getFunctions() returns only functions.
To populate the procedures dropdown in UI, syndesis used to call only getProcedures(), which will not return the functions.
The fix is to use both getProcedures/getFunctions to maintain the same
behavior as before.

Syndesis doesn't show PostgreSQL demo data procedures
https://issues.redhat.com/browse/ENTESB-16286

### DIFF
--- a/app/connector/sql/src/main/java/io/syndesis/connector/sql/SqlSupport.java
+++ b/app/connector/sql/src/main/java/io/syndesis/connector/sql/SqlSupport.java
@@ -34,6 +34,9 @@ import io.syndesis.connector.sql.common.stored.StoredProcedureMetadata;
 import io.syndesis.connector.support.util.ConnectorOptions;
 
 public final class SqlSupport {
+
+    public enum TYPE { STORED_PROCEDURE, FUNCTION }
+
     private SqlSupport() {
     }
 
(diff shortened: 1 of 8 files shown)


app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaMetaDataRetrieval.java
```diff
@@ -47,15 +47,15 @@
...
+        .withApiVersion("apiextensions.k8s.io/v1beta2")
         .withKind("CustomResourceDefinition")
         .withNewMetadata()
             .withName("kafkas.kafka.strimzi.io")
         .endMetadata()
         .withNewSpec()
             .withGroup("kafka.strimzi.io")
             .withScope("Namespaced")
-            .withVersion("v1beta1")
+            .withVersion("v1beta2")
```
zregvart: Should this be part of this PR?

app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java
```diff
@@ -82,15 +82,15 @@
...
+        .withApiVersion("apiextensions.k8s.io/v1beta2")
         .withKind("CustomResourceDefinition")
         .withNewMetadata()
             .withName("syndesises.syndesis.io")
         .endMetadata()
         .withNewSpec()
             .withGroup("syndesis.io")
             .withScope("Namespaced")
-            .withVersion("v1beta1")
+            .withVersion("v1beta2")
```
zregvart: Should this be part of this PR?

zregvart:
Needs to be rebased to pickup the workflow fix in #9541 and the Guava revert from #9542

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaMetaDataRetrieval.java
claudio4j: I will submit another PR with this change.

claudio4j:
@zregvart Is this ok to merge ?

PR #9552, opened 2021-04-29
fix(operator): correct nesting for auditing

zregvart:
The auditing feature toggle was placed after the Jaeger feature toggle,
which did not only change values within `features` key, but terminated
the `features` key and started the `jaeger` top level key. This way when
Jaeger feature was enabled `auditing.enabled` was placed within `jaeger`
and not within `features`. This places the `auditing.enabled` before the
Jaeger feature toggle in the server config template, so now the Jager
feature toggle should not affect placement of the auditing feature
toggle. Tests were added to make sure valid (parsable) YAML files are
generated from the generator and that the auditing feature toggle state
is correctly passed to the template.

Separate commit adds gopter dependency (3278206). Split to ease the review.

Ref. https://issues.redhat.com/browse/ENTESB-16295


install/operator/pkg/generator/assets/infrastructure/03-syndesis-server-config.yml.tmpl
```diff
@@ -78,6 +78,8 @@
           {{- else}}
           enabled: false
           {{- end}}
+        auditing:
+          enabled: {{.Syndesis.Components.Server.Features.Auditing}}
 {{- if .Syndesis.Addons.Jaeger.Enabled}}
```
phantomjinx, line 83: This line (L83 - the if statement) seems to be in the wrong place ??
Does `dblogging` have anything to do with jaeger?

@claudio4j

phantomjinx (approved):

install/operator/pkg/generator/assets/infrastructure/03-syndesis-server-config.yml.tmpl
zregvart: Yeah, we have two implementations JsonDB and Jaeger (https://github.com/syndesisio/syndesis/tree/1.12.x/app/server/logging), so when Jaeger is turned on the JsonDB one is turned off. (but @claudio4j would know better)

claudio4j: +1 @zregvart.

PR #9568, opened 2021-04-29
doc-update-fhir-jason-patch

MelissaFlinn:


zregvart:
@MelissaFlinn I had to remove branch protection that forced checks to succeed to get this merged. I've tried re-triggering the checks by pushing an empty change on top of yours, that didn't help. We have a [paths filter](https://github.com/syndesisio/syndesis/blob/76ef7e121cdd95e2d3f0461047cdc62ef8482b8d/.github/workflows/checks.yml#L9-L10) for anything within the `docs` directory that should skip checks. It seems that there is now an inconsistency between that and required checks by branch protection setup. From what I remember that wasn't the case before, skipped checks reported as skipped and we could have merged pull requests. If this remains to be an issue, we can change the checks workflow definition. I'd wait to see if this repeats.

PR #9573, opened 2021-04-30
fix(ui) [ENTESB-16069] conditional step gets duplicated

kahboom:
fixes [ENTESB-16069](https://issues.redhat.com/browse/ENTESB-16069), where conditional flow steps get duplicated on pressing the 'Back' button in the editor.

Kapture 2021-04-30 at 17 41 21


gashcrumb (approved):
:+1:

PR #9584, opened 2021-05-03
fix(c/sql): jsqlparser for parsing SQL statements

zregvart:
The homegrown SQL parser made a number of assumptions and was very
brittle which made it really difficult to maintain.

This replaces the homegrown SQL parser with jsqlparser, which is more
robust, correct and feature full which should ease maintenance.

Ref. issues.redhat.com/browse/ENTESB-13689

fix(c/sql): make sure test tear down is done (6f85713)

We need to make sure that any object we create is dropped in the test.
Otherwise we might get flaky tests that depend on the order of
execution.

### DIFF
--- a/app/connector/sql/pom.xml
+++ b/app/connector/sql/pom.xml
@@ -43,6 +43,11 @@
       <artifactId>spring-tx</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.github.jsqlparser</groupId>
+      <artifactId>jsqlparser</artifactId>
+      <version>4.0</version>
+    </dependency>
+    <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>connector-support-verifier</artifactId>
       <optional>true</optional>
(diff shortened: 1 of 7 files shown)


KurtStam (commented):
I think it makes sense to open a separate issue for this as it is just the introduction of the jsqlparser. and do the UUID support in a different issue

zregvart:
> I think it makes sense to open a separate issue for this as it is just the introduction of the jsqlparser. and do the UUID support in a different issue

I've created the issue #9587, which is implemented by this pull request. The UUID support already went in #9248 and #9467. This pull request replaces the SQL parser with JSqlParser to enable the `CAST(x AS y)` syntax that would allow UUID column types to be used in expressions with PostgreSQL.

KurtStam (approved):
Thanks for doing this. Looks really good!

zregvart:
retest this please

mkralik3:
retest this please

zregvart:
retest this please

avano:
retest this please

zregvart:
What would you know, it's Tuesday already.

PR #9588, opened 2021-05-04
chore: Update to AtlasMap 2.2.1

igarashitm:
Fixes: https://issues.redhat.com/browse/ENTESB-16353
Contains: atlasmap/atlasmap#2729 - Regression: UI: Mapping for a collection reference expression missing reference layer input field group (non-preview).

It's not a blocker ATM, so we can skip if it bothers productization.


phantomjinx (approved):
LGTM

PR #9592, opened 2021-05-06
fix(c/sql): stored procedure/function metadata

zregvart:
When fetching metadata for functions in PostgreSQL we need to use stored
procedure metadata, not function metadata as it returns nothing.

Ref. https://issues.redhat.com/browse/ENTESB-16358

fix(c/sql): make sure test tear down is done (766e285)

We need to make sure that any object we create is dropped in the test.
Otherwise we might get flaky tests that depend on the order of
execution.

(cherry picked from commit fbcbfeac2a2111e7f43f60458b88a7472bff5778)


zregvart:
retest this please

avano:
retest this please

app/connector/sql/src/main/java/io/syndesis/connector/sql/SqlSupport.java
```diff
@@ -49,15 +47,13 @@ public static Connection createConnection(final Map<String, Object> properties)
     }
 
     public static StoredProcedureMetadata getFunctionOrProcedureMetadata(final Connection connection, final String catalog,
```
claudio4j: As there is no fetchFunctionColumns call, renaming the method back to getProcedureMetadata and adding a comment about why not using fetchFunctionColumns.

claudio4j (changes requested):

zregvart:
Thanks for the review, @claudio4j can you have another look?

claudio4j (approved):

PR #9595, opened 2021-05-07
ENTESB-16374 Fail to install jaeger-operator on openshift3

claudio4j:
Re-generated the assets_vfsdata.go related to a previous change in 03-syndesis-server-config.yml.tmpl
https://issues.redhat.com/browse/ENTESB-16374


avano:
retest this please

install/operator/pkg/syndesis/capabilities/capabilities.go
```diff
@@ -37,15 +37,15 @@ type RequiredApiSpec struct {
 	routes                    string
 	imagestreams              string
 	oauthclientauthorizations string
-	catalogsources            string
+	packagemanifests          string
```
zregvart, line 40: I'm assuming this is part of this PR?

zregvart (approved):
Thanks!

install/operator/pkg/syndesis/capabilities/capabilities.go
zregvart: Looks like it needs to be.

claudio4j: Yes, I replaced `catalogsources` for `packagemanifests` to test if openshift can install from OLM. At least in my openshift 3.11, `catalogsources` does exist but lacks OLM operator and other OLM objetcs. 
Therefore, I found that `packagemanifests` better represents a CR to enable operators to install from operatorhub.

PR #9596, opened 2021-05-07
chore(tools): remove unused images

zregvart:
I don't think these are used for anything, so let's remove them.


avano:
retest this please

PR #9616, opened 2021-05-25
Simplify code: the optional was never going to be null anyway

Delawen:
So enforce it with a NotNull annotation and skip checks

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/connection/ConnectionBase.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/connection/ConnectionBase.java
@@ -27,6 +27,9 @@ import io.syndesis.common.model.WithResourceId;
 import io.syndesis.common.model.WithTags;
 import io.syndesis.common.model.WithUsage;
 import io.syndesis.common.model.environment.Organization;
+import io.syndesis.common.model.validation.AllValidations;
+
+import javax.validation.constraints.NotNull;
 
 public interface ConnectionBase
     extends WithResourceId, WithTags, WithName, WithConfiguredProperties, WithMetadata, ToJson, WithUsage {
@@ -35,6 +38,7 @@ public interface ConnectionBase
 
     Optional<String> getOrganizationId();
 
+    @NotNull(groups = AllValidations.class)
     Optional<Connector> getConnector();
 
     String getConnectorId();
--- a/app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/IntegrationUpdateHandler.java
+++ b/app/server/update-controller/src/main/java/io/syndesis/server/update/controller/bulletin/IntegrationUpdateHandler.java
@@ -21,7 +21,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -197,11 +196,10 @@ public class IntegrationUpdateHandler extends AbstractResourceUpdateHandler<Inte
 
     private Connection includeConnector(final Connection connection) {
         if (connection == null) {
-            return connection;
+            return null;
         }
 
-        final Optional<Connector> connectorOptional = connection.getConnector();
-        if (connectorOptional != null && connectorOptional.isPresent()) {
+        if (connection.getConnector().isPresent()) {
             return connection;
         }
 


Delawen:
All the implementations of getConnector returns a not null value.

app/common/model/src/main/java/io/syndesis/common/model/connection/ConnectionBase.java
```diff
@@ -35,6 +38,7 @@
 
     Optional<String> getOrganizationId();
 
+    @NotNull(groups = AllValidations.class)
     Optional<Connector> getConnector();
```
zregvart, line 42: This might break the UI if the validation is performed and no Connector is provided. Could be a legacy issue. Should be reviewed by someone with the knowledge of UI.

Delawen: @kahboom ?

kahboom: should be ok, the UI will throw an exception if there is no connector ID, but I don't see how we would call that API if we don't already have a connector ID available.

zregvart: This holds the full Connector object, `connectorId` is a different property, or did I miss something?

kahboom: Is the concern that the UI might not supply the Connector when validation is performed? If so, the validation function requires a `connectorId` or will throw an error, see [here](https://github.com/syndesisio/syndesis/blob/33ced86ea33cef4500d72c620f04083eaac80a50/app/ui-react/packages/api/src/useConnectionHelpers.tsx#L77) and [here](https://github.com/syndesisio/syndesis/blob/1.12.x/app/ui-react/syndesis/src/modules/connections/components/WithConnectorForm.tsx#L150), so I can't see how we would call it at all if we don't have a Connector to begin with. If you can't find a Connector with the given ID then the UI should be able to handle that exception. though cc'ing @gashcrumb for additional eyeballs

zregvart: Validation would require the `connector`, `connectorId` is mandatory anyhow. Re-reading this, I don't think the `@NotNull` annotation makes much sense on `Optional` in a Immutables managed class, that's a given, the `Optional` will never be `null` here, the value might be `Optional.empty()` but it'll not be `null`.

Delawen: Does that mean that we can assume that all Optionals that comes from the frontend are never going to be null?

Because we have a lot of shitty code checking that kind of stuff. If we make sure backend is not creating a null there (as on this case) and we know for sure that the parsing from frontend to Java is also not creating nulls if the type is an Optional, then there's a lot of cleanup we can do directly.

zregvart: Yeah, I think Immutables would throw an exception from the builder method for `Optional` type when trying to set `null` value to it. On the backend we should assume that no `Optional` can be `null`, that's kinda the purpose of using `Optional`.

I think for UI empty `Optional` is `undefined`, and the generated API model should generate optional typing (i.e. `val?`).

zregvart: So bottom line, I don't think we need the `@NotNull(groups = AllValidations.class)` bit on the `connector` property as that property can't be `null`.

I've looked at the code Immutables generated, and that confirms it:

```java
  public final ImmutableConnection withConnector(Connector value) {
    @Nullable Connector newValue = Objects.requireNonNull(value, "connector");
    // ...
  }

  @SuppressWarnings("unchecked") // safe covariant cast
  public final ImmutableConnection withConnector(Optional<? extends Connector> optional) {
    @Nullable Connector value = optional.orElse(null);
    // ...
  }
```

Both of these throw an exception if `Connector` is given as `null` or `Optional<Connector>` is given as `null`.

Delawen: So, if I remove the `@NotNull`, is it fine to merge? And also similar code checking if optionals are null if I can't find any explicit code making an optional null because the framework auto-generating this objects from the frontend will never make an optional null?

PR #9620, opened 2021-05-25
Clean and remove outdated and unused code and docs.

claudio4j:


claudio4j:
@gashcrumb is `ux` directory outdated and a candidate to be removed ?

Delawen (approved):
About time!

zregvart (changes requested):
I would not remove `project/proposals`, they're the only technical documentation we have, and it's still relevant as far as I know.

zregvart (changes requested):
Perhaps we need to leave the `project/designs/error-handling.md` as well.

claudio4j:
> I would not remove `project/proposals`, they're the only technical documentation we have, and it's still relevant as far as I know.

I just put it back. Thanks for reviewing.

gashcrumb:
> @gashcrumb is `ux` directory outdated and a candidate to be removed ?

Just to follow up, we could, it wouldn't be hard to bring it back anyways.  @kahboom when's the last time you've needed to look here in [ux](https://github.com/syndesisio/syndesis/tree/1.12.x/ux)?

kahboom:
@gashcrumb I forgot it existed. Used to be useful but now is quite outdated. I don't see any problem with it being removed.

PR #9629, opened 2021-06-04
ENTESB-16464: NullpointerException when creating SOAP Client Connector

zregvart:
Split the work to address the issue in several commits to ease reviewing:

chore(deps): upgrade xmlschema-core 2.2.5 (b7cfa9c)

Makes sense to use a newer, currently latest, version.

chore(c/soap): test data shape with AtlasMap (10dbcec)

This makes sure that any data shape produced by the SOAP API generator
will be accepted by AtlasMap inspection by running the inspection as a
part of the example tests.

fix(c/soap): change xsd:timeInstant to xsd:dateTime (0cac7b4)

SOAP 1.2 renamed `xsd:timeInstant` to `xsd:dateTime` see[1]. Using
`xsd:timeInstant` was breaking the tests with AtlasMap as the XML Schema
data types no longer define `xsd:timeInstant`.

Might be a slight concern with regards to legacy SOAP 1.1 or earlier,
this was an easier fix and it's fairly safe that most deployments are
using SOAP 1.2 these days.

[1] https://w3.org/2000/xp/Group/1/10/11/soap12-part1.html#schemachanges

fix(c/soap): NPE in XmlSchemaSerializer (b402730)

This fixes the `NullPointerException` in `XmlSchemaSerializer` reported
in ENTESB-16464[1]. The namespace context in the target schema was not
initialized and defaulted to `null`, leading to the
`NullPointerException`.

In addition to setting the namespace context to an instance of
`NamespaceMap`, this instance was populated with prefixes for XML Schema
and XML namespaces, as these were not set by default.

[1] https://issues.redhat.com/browse/ENTESB-16464

fix(c/soap): use name instead of concrete name (db434b4)

For an unknown reason the code was using concrete name instead of name
to extract a schema for a SOAP message part, this lead to
`NullPointerException`s when concrete name was not set.

fix(c/soap): add definition of XML schema namespace (0107b48)

XML namespace is an odd one, it's implicitly defined in a XML schema
document, but not in non-XML schema documents. So in the schema sets
created by the SOAP API generator and expected by AtlasMap in the data
shape it needs to be declared and included. Otherwise attributes with
type of `xml:*` can't be unreferenced and the inspection of the data
shape fails.

To circumvent this, XML namespace schema is added automatically to all
schema sets. This might seem unwarranted, but considering that most SOAP
APIs will declare a SOAP fault, and that the `reasonText` is defined
with `xml:lang` attribute inclusion seems warranted.

fix(c/soap): test with SugarCRM WSDL file (fb9e399)

This adds an example test with the WSDL from SugarCRM, being that it's
reported in ENTESB-16464[1], now we can make sure that this WSDL passes
generation and AtlasMap inspection.

[1] https://issues.redhat.com/browse/ENTESB-16464

### DIFF
--- a/app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/BindingHelper.java
+++ b/app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/BindingHelper.java
@@ -74,8 +74,32 @@ import static javax.jws.soap.SOAPBinding.Use.ENCODED;
 public class BindingHelper {
 
     private static final String SCHEMA_SET_XML =
-        "<d:SchemaSet xmlns:d=\"http://atlasmap.io/xml/schemaset/v2\">" +
-            "<d:AdditionalSchemas/>" +
+        "<d:SchemaSet xmlns:d=\"http://atlasmap.io/xml/schemaset/v2\" xmlns:xml=\"http://www.w3.org/XML/1998/namespace\">" +
+            "<d:AdditionalSchemas>" +
+                "<xsd:schema targetNamespace=\"http://www.w3.org/XML/1998/namespace\" xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\">" +
+                    "<xsd:attribute name=\"lang\">" +
+                        "<xsd:simpleType>" +
+                            "<xsd:union memberTypes=\"xsd:language\">" +
+                                "<xsd:simpleType>" +
+                                    "<xsd:restriction base=\"xsd:string\">" +
+                                        "<xsd:enumeration value=\"\"/>" +
+                                    "</xsd:restriction>" +
+                                "</xsd:simpleType>" +
+                            "</xsd:union>" +
+                        "</xsd:simpleType>" +
+                    "</xsd:attribute>" +
+                    "<xsd:attribute name=\"space\">" +
+                        "<xsd:simpleType>" +
+                            "<xsd:restriction base=\"xsd:NCName\">" +
+                                "<xsd:enumeration value=\"default\"/>" +
+                                "<xsd:enumeration value=\"preserve\"/>" +
+                            "</xsd:restriction>" +
+                        "</xsd:simpleType>" +
+                    "</xsd:attribute>" +
+                    "<xsd:attribute name=\"base\" type=\"xsd:anyURI\"/>" +
+                    "<xsd:attribute name=\"id\" type=\"xsd:ID\"/>" +
+                "</xsd:schema>" +
+            "</d:AdditionalSchemas>" +
         "</d:SchemaSet>";
 
     private static final SchemaCollection SOAP_SCHEMAS;
--- a/app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/BindingHelper.java
+++ b/app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/BindingHelper.java
@@ -426,7 +426,7 @@ public class BindingHelper {
             final XmlSchemaAnnotated annotated = part.getXmlSchema() != null ? part.getXmlSchema() :
                 schemaCollection.getTypeByQName(part.getTypeQName());
 
-            final QName name = part.getConcreteName();
+            final QName name = part.getName();
             final XmlSchemaElement element;
             if (annotated instanceof XmlSchemaElement) {
                 // extract element
--- a/app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/XmlSchemaExtractor.java
+++ b/app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/XmlSchemaExtractor.java
@@ -30,6 +30,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
+
+import javax.xml.XMLConstants;
 import javax.xml.namespace.QName;
 
 import org.apache.commons.beanutils.BeanUtils;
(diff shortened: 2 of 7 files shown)


zregvart:
retest this please

mkralik3:
retest this please

dhirajsb:
@zregvart Great work on these issues, nicely done!

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/BindingHelper.java
```diff
@@ -426,7 +450,7 @@ protected XmlSchema createSoapSchema(SchemaCollection targetSchemas) {
             final XmlSchemaAnnotated annotated = part.getXmlSchema() != null ? part.getXmlSchema() :
                 schemaCollection.getTypeByQName(part.getTypeQName());
 
-            final QName name = part.getConcreteName();
```
dhirajsb, line 429: Are you sure we don't need a check to decide which one to use? Maybe if one is not set, use the other? I can't remember why I chose concrete name back then.

igarashitm (approved):
LGTM - nothing jumps out from AtlasMap perspective. Any specific concern I have to look deeper, e.g. possibility to change at AtlasMap side?

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/XmlSchemaExtractor.java
```diff
@@ -224,6 +227,14 @@ public XmlSchema getOrCreateTargetSchema(String namespaceURI) {
...
         if (targetSchema == null) {
             targetSchema = targetSchemas.newXmlSchemaInCollection(namespaceURI);
+            final NamespaceMap namespaceContext = new NamespaceMap();
+
+            // add the usual suspects, for some reason these seem to be missing from the
+            // generated SchemaSet
+            namespaceContext.add("xsd", XMLConstants.W3C_XML_SCHEMA_NS_URI);
+            namespaceContext.add(XMLConstants.XML_NS_PREFIX, XMLConstants.XML_NS_URI);
+
+            targetSchema.setNamespaceContext(namespaceContext);
```
dhirajsb, line 237: Could the ns context from source schema be copied, in case there are other properties defined there?

app/server/api-generator/src/test/resources/soap/StockQuote.wsdl
```diff
@@ -26,7 +26,7 @@
 
     <message name="GetTradePriceInput">
         <part name="tickerSymbol" type="xsd:string"/>
-        <part name="time" type="xsd:timeInstant"/>
```
dhirajsb, line 29: Does the WS-I Basic Profile 2.0 say anything about this ambiguity? Would be nice to point to some standards doc to say why we support it this way.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/BindingHelper.java
zregvart: > LGTM - nothing jumps out from AtlasMap perspective. Any specific concern I have to look deeper, e.g. possibility to change at AtlasMap side?

Not sure if the XML namespace needs to be included by default in AtlasMap, that way it can be omitted from here. It's added for all cases now.

zregvart:
> LGTM - nothing jumps out from AtlasMap perspective. Any specific concern I have to look deeper, e.g. possibility to change at AtlasMap side?

@igarashitm, not sure if the XML namespace needs to be/can be included by default in AtlasMap, that way it can be omitted from here. It's added for all cases now.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/BindingHelper.java
zregvart: There is virtually no documentation on the difference between these, tests seem to confirm that `getName` is good enough.

app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/XmlSchemaExtractor.java
zregvart: This is the default context, added to each target schema set, additional namespaces are added to the context as needed. These were missing, especially difficult one is the XML namespace which seems implicitly defined in some cases only.

app/server/api-generator/src/test/resources/soap/StockQuote.wsdl
zregvart: It's a change introduced in SOAP 1.2: https://www.w3.org/2000/xp/Group/1/10/11/soap12-part1.html#schemachanges, BP 2.0 is based on SOAP 1.2: http://ws-i.org/profiles/BasicProfile-2.0-WGD.html#bp11compatibility

dhirajsb (approved):
LGTM!

igarashitm:
Ah it might be better to have it in AtlasMap, having it has a reserved alias, i.e. it's expected to be standard built-in I guess.
https://github.com/atlasmap/atlasmap/issues/2916

zregvart:
> Ah it might be better to have it in AtlasMap, having it has a reserved alias, i.e. it's expected to be standard built-in I guess.
> [atlasmap/atlasmap#2916](https://github.com/atlasmap/atlasmap/issues/2916)

I'll merge this as a fix that we have, time allowing we can remove it when there is support in AtlasMap.

zregvart:
Thanks for the review @igarashitm & @dhirajsb

PR #9630, opened 2021-06-07
fix(c/sql): use the legacy parser as fall back

zregvart:
The JSQLParser we use is currently not aware of some of the proprietary
SQL syntax PostgreSQL supports and it is strict (as parsers are) as to
the correct syntax of SQL it can parse. This came up in issues like
ENTESB-16608 and ENTESB-16609, so as a strategy to help when SQL
statement parsing fails this brings the legacy parser, the parser used
before we switched to JSQLParser as a fall back. Now when a statement
can't be parsed using JSQLParser an attempt is made to parse the
statement using the legacy parser. If the legacy parser fails, the
original error is reported, i.e. the one from JSQLParser.

Ref. https://issues.redhat.com/browse/ENTESB-16608
Ref. https://issues.redhat.com/browse/ENTESB-16609

### DIFF
--- a/app/connector/sql/src/main/java/io/syndesis/connector/sql/common/SqlStatementLegacyParser.java
+++ b/app/connector/sql/src/main/java/io/syndesis/connector/sql/common/SqlStatementLegacyParser.java
@@ -1,0 +1,322 @@
+/*
+ * Copyright (C) 2016 Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


mkralik3:
Verified that the integration from ENTESB-16609 is able to be edited and query from ENTESB-16608 is able to be used.

zregvart:
> Verified that the integration from ENTESB-16609 is able to be edited and query from ENTESB-16608 is able to be used.

Thanks!

PR #9633, opened 2021-06-09
fix(c/soap): custom SOAP client error handling

zregvart:
This adds error handling to custom SOAP client creation. On the server
side this also cleans up the error message when parsing WSDL so that the
message doesn't show Java internals.

Ref. https://issues.redhat.com/browse/ENTESB-14078


kahboom (approved):
👍

PR #9642, opened 2021-06-16
fix(c/sql): pass schema to legacy parser

zregvart:
Legacy parser could fail if the database object is in a different schema
than the default schema provided by the connection.

Ref. https://issues.redhat.com/browse/ENTESB-16608


mkralik3:
retest this please

PR #9654, opened 2021-06-25
feat: integration labels

squakez:
With this PR we include the possibility to set `Integration` labels on server.

Ref [ENTESB-14559](https://issues.redhat.com/browse/ENTESB-14559)

### DIFF
--- a/app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
+++ b/app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
@@ -15,6 +15,17 @@
  */
 package io.syndesis.server.controller.integration.online;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.StringWriter;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.Set;
+
 import io.syndesis.common.model.integration.Integration;
 import io.syndesis.common.model.integration.IntegrationDeployment;
 import io.syndesis.common.model.integration.IntegrationDeploymentError;
@@ -34,11 +45,6 @@ import io.syndesis.server.openshift.OpenShiftService;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
-
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.StringWriter;
-import java.util.*;
 
 @Qualifier("s2i")
 @Component()
--- a/app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
+++ b/app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
@@ -15,17 +15,6 @@
  */
 package io.syndesis.server.controller.integration.online;
 
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.StringWriter;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Properties;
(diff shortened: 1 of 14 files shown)


squakez:
cc @kahboom

app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/PublishHandler.java
```diff
@@ -149,6 +149,7 @@ private DeploymentData createDeploymentData(Integration integration, Integration
             .addSecretEntry("application.properties", propsToString(applicationProperties));
 
         integration.getConfiguredProperties().forEach(deploymentDataBuilder::addProperty);
+        integration.getLabels().forEach((k, v) -> deploymentDataBuilder.addLabel(Labels.validateKey(k), Labels.validate(v)));
```
zregvart: Here we should expect labels to be valid, I don't think there is a need to re-validate here.

app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/integration/IntegrationHandler.java
```diff
@@ -98,10 +99,19 @@ public Integration create(@Context final SecurityContext sec, final Integration
         final Integration updatedIntegration = new Integration.Builder().createFrom(encryptedIntegration)
             .createdAt(System.currentTimeMillis()).build();
 
+        validateLabels(updatedIntegration);
```
zregvart: Could we use JSR 303 validation instead here? I don't think throwing `IllegalArgumentException` is what we do right now, we have a protocol in place for validation errors and how they're signaled to the UI via `ConstraintViolationExceptionMapper` et al. Look at `@UniqueProperty` and [configuration](https://github.com/syndesisio/syndesis/tree/1.13.x/app/server/dao/src/main/resources/META-INF/validation).

zregvart (commented):
Would be good to rebase this to pickup #9653 to update the OpenAPI documents needed to generate model in the UI.

squakez:
Thanks for the review @zregvart, I've applied your suggestion, please have a further look to confirm it's good to go. I've performed some test, and the API is returning a validation error that can be used from UI now:
### Integration with good labels:
```
$ cat integration.json | http POST https://syndesis-syndesis.192.168.42.139.nip.io/api/v1/integrations \
>   SYNDESIS-XSRF-TOKEN:... \
>   cookie:"..." \
>   --verify=no

{
    "createdAt": 1625219948726,
...
    "id": "i-Mdae51sB5zmxphzQTctz",
    "labels": {
        "key1": "value1",
        "key2": "value2"
    },
    "name": "test-labels",
    "tags": [
        "timer"
    ],
    "updatedAt": 0,
    "version": 1
}
```
### Integration with labels errors:
```
$ cat integration.json | http POST https://syndesis-syndesis.192.168.42.139.nip.io/api/v1/integrations ...

[
    {
        "error": "ValidLabels",
        "message": "Invalid label key 'syndesis.io/test' or value 'value2'",
        "property": "create.obj.labels"
    },
    {
        "error": "NoDuplicateIntegration",
        "message": "Integration name 'test-labels' is not unique",
        "property": "create.obj.name"
    }
]
```

zregvart (approved):
LGTM. I _think_ we might need a XML file here: https://github.com/syndesisio/syndesis/tree/1.13.x/app/server/dao/src/main/resources/META-INF/validation to make this work in the tests, can't remember it was a while ago that we last added a constraint.

squakez:
> LGTM. I _think_ we might need a XML file here: https://github.com/syndesisio/syndesis/tree/1.13.x/app/server/dao/src/main/resources/META-INF/validation to make this work in the tests, can't remember it was a while ago that we last added a constraint.

Thanks Zoran. I wasn't sure if it was good to include a second validation like I'm doing here: https://github.com/syndesisio/syndesis/pull/9654/files#diff-8f22ea2b354414f071792042fd4f9f254e5ec5e46494a3c23554150528a8cf3dR26-R31 - I am reviewing the test and include a separate one if it's needed.

PR #9658, opened 2021-06-28
fix(ui): additional step for SOAP service and port

zregvart:
We used to have two screens on the "Imported Operations" step, akka
review step of the custom API client wizard. That caused issues with
error handling: the screen to specify service and port would throw an
`Error` if the provided WSDL file cannot be parsed. So in #9633 instead
of reaching this screen a notification would appear indicating of this
error situation. Unfortunately, this also changed the flow for the
OpenAPI. So when an error was generated editing the successfully parsed
OpenAPI document we would render the first screen (upload) of the wizard
and show the error notification. And in process loose any changes user
has made to the document.

This instead adds an additional step in the wizard "Additional
Configuration" to specify service/port for WSDL documents, and is
skipped for OpenAPI documents, also changing the order of the
"Imported Operations" (review) screen, so that it precedes this new
screen. This way we won't `Error`-out in trying to access nonexistent
data, as it could not be parsed.

Ref. https://issues.redhat.com/browse/ENTESB-16902

### DIFF
--- a/app/ui-react/syndesis/src/modules/apiClientConnectors/locales/api-client-connectors-translations.en.json
+++ b/app/ui-react/syndesis/src/modules/apiClientConnectors/locales/api-client-connectors-translations.en.json
@@ -9,6 +9,9 @@
   "apiConnectorsPageTitle": "API Client Connectors",
   "basePath": "Base URL",
   "create": {
+    "configuration": {
+      "title": "Additional Configuration"
+    },
     "details": {
       "saveButton": "Create API Connector",
       "successNotification": "API Connector successfully created",
--- a/app/common/model/src/main/java/io/syndesis/common/model/action/ActionsSummary.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/action/ActionsSummary.java
@@ -28,6 +28,10 @@ public interface ActionsSummary {
 
     final class Builder extends ImmutableActionsSummary.Builder {
         // make ImmutableActionsSummary.Builder accessible
+
+        public static ActionsSummary empty() {
+            return new Builder().build();
+        }
     }
 
     Map<String, Integer> getActionCountByTags();
--- a/app/common/model/src/main/java/io/syndesis/common/model/api/APISummary.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/api/APISummary.java
@@ -40,8 +40,8 @@ public interface APISummary extends WithName, WithConfigurationProperties, WithC
         // make ImmutableAPISummary.Builder accessible
 
         public static Builder createFrom(final Connector connector) {
-            final ActionsSummary actionsSummary = new ActionsSummary.Builder()//
-                .totalActions(connector.getActions().size())//
+            final ActionsSummary actionsSummary = new ActionsSummary.Builder()
+                .totalActions(connector.getActions().size())
                 .actionCountByTags(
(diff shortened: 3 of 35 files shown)


zregvart:
cc @gashcrumb, @mkralik3 this is what I think a _proper_ fix would do. It's quite big as you can see. It's also more work to adapt QE tests and probably needs adjusting the docs.
For `1.12.x` I can try to create a quick fix, one that would not be that _proper_. Or we can bite the bullet, up for discussion.

app/ui-react/syndesis/src/modules/apiClientConnectors/locales/api-client-connectors-translations.en.json
```diff
@@ -56,17 +24,14 @@
...
-      "sectionErrors": "ERRORS",
       "title": "Imported Operations"
     },
-    "specification": {
-      "title": "Edit API Definition"
-    },
     "security": {
-      "address": "Address",
       "accessTokenUrl": "Access Token URL",
+      "address": "Address",
```
kahboom: Happy to see it's not just me that appreciates alphabetical order 😆

kahboom:
@zregvart - I mean, it's a lot of changes, but I do like the idea of moving the Service/Port config onto a separate page. I've gone through the changes and tested the branch locally, and it looks good, with a few small requests:

1. Please update the story filenames to match the step number; it just makes it _much_ easier to find them when you need to reference them. This probably also means adding another story for the new step/page.
2. Remove the footer in the `ApiConnectorCreatorService` (found in `packages/ui/src/Customization/apiClientConnectors/create`) since it's no longer needed, otherwise produces a duplicate Next button:

Screen Shot 2021-07-02 at 11 56 08

kahboom (changes requested):
@zregvart - I mean, it's a lot of changes, but I do like the idea of moving the Service/Port config onto a separate page. I've gone through the changes and tested the branch locally, and it looks good, with a few small requests:

1. Please update the story filenames to match the step number; it just makes it _much_ easier to find them when you need to reference them. This probably also means adding another story for the new step/page.
2. Remove the footer in the `ApiConnectorCreatorService` (found in `packages/ui/src/Customization/apiClientConnectors/create`) since it's no longer needed, otherwise produces a duplicate Next button:

Screen Shot 2021-07-02 at 11 56 08

app/ui-react/syndesis/src/modules/apiClientConnectors/locales/api-client-connectors-translations.en.json
zregvart: Yeah I did that in #9652 to ease with OpenAPI JSON files copied between server and UI, but in general that (with time) will help us with merges/backports...

zregvart:
Fat fingered that. Thanks @kahboom for the review, I'll get to fixing this.

zregvart:
Many new fixes were added for this, more details in the commit messages.

zregvart:
This also should fix https://issues.redhat.com/browse/ENTESB-17281 via a671e49

Delawen:
Rerunning jobs because one failed. Just in case.

zregvart:
> Rerunning jobs because one failed. Just in case.

There was a TSLint issue in the UI, should be okay now.

kahboom (approved):

PR #9659, opened 2021-06-29
fix(ui): don't revert to upload on OpenAPI error

zregvart:
When an error is introduced in editing an OpenAPI document we would
revert to the upload screen, this is because in #9633 we introduced
better error handling for the custom SOAP client connectors, where we
needed to prevent reaching the review screen for unparsable WSDL files.

The error handling for both OpenAPI and SOAP custom client connectors
is shared and the intent in #9633 was not to allow reaching the review
screen unless the given document (WSDL, OpenAPI) can be parsed.

Recoverable parse errors can be introduced by editing the OpenAPI
document on the review screen, and in that case the common error handler
would revert to the upload screen, loosing any edits made by the user.

Since the review screen will not error-out for OpenAPI documents, only
for WSDL documents. So this changes the point of raising the `Error` so
errors are raised only for SOAP custom connectors. Thus preventing
reaching the review screen for SOAP connectors.

Ref. https://issues.redhat.com/browse/ENTESB-16902

### DIFF
--- a/app/ui-react/packages/api/src/useApiConnectorSummary.tsx
+++ b/app/ui-react/packages/api/src/useApiConnectorSummary.tsx
@@ -56,7 +56,7 @@ export function useApiConnectorSummary(
         if (summary.errorCode) {
           throw new Error(summary.userMsg);
         }
-        if (Array.isArray(summary.errors) && summary.errors.length > 0) {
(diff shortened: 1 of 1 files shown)


zregvart:
cc @gashcrumb, @mkralik3 this would be a quick-patchy-fix I mentioned in #9658

mkralik3:
retest this please

PR #9673, opened 2021-07-07
fix: soap connector actions missing in info

zregvart:
If a WSDL file contains more than one service the amount of imported
actions is 0. This changes the condition so that the action info is
generated for the first service if at least one service exists. This is
not 100% accurate but it's better than not displaying 0 actions
imported.

### DIFF
--- a/app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/SoapApiModelParser.java
+++ b/app/server/api-generator/src/main/java/io/syndesis/server/api/generator/soap/parser/SoapApiModelParser.java
@@ -138,9 +138,10 @@ public final class SoapApiModelParser {
             builder.model(definition);
             validateModel(definition, builder);
 
-            // set default service, port and address if present
+            // set default service, port and address to the first service, port and address in the WSDL
+            // otherwise we don't generate any action summary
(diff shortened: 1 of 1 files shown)


dhirajsb:
Is the action summary needed at the point before user selects a service? From what I can recall, the UI just shows ports and services to user first, then once a service is selected, action summary is displayed for that particular service. Did the UI flow change around these steps? 

If it did, the changes lgtm, otherwise I'd question what's happening in the UI to show action summary before service selection.

zregvart:
> Is the action summary needed at the point before user selects a service? From what I can recall, the UI just shows ports and services to user first, then once a service is selected, action summary is displayed for that particular service. Did the UI flow change around these steps?
> 
> If it did, the changes lgtm, otherwise I'd question what's happening in the UI to show action summary before service selection.

Yeah, I'm trying to change this flow a bit (see https://github.com/syndesisio/syndesis/pull/9658). We need the info screen before the service selection as the info screen contains any errors that there might be, so when uploading an invalid WSDL XML that screen would break.

I think I need to rethink this a bit, perhaps we could return an array of summaries instead.

zregvart:
I have a number of fixes in this area I'll continue on #9658

PR #9694, opened 2021-07-23
ENTESB-16809 Camel Clients Getting Started with Managed Kafka

claudio4j:
* Updated camel to 2.23.2.fuse-790053
* Camel update also updated cxf and was necessary to add one more cxf artifact to the dependencyManagement
* Add cluster permission to verify kafka strimzi crd is installed
* Changed KafkaCustomizer to KafkaConnector to retrieve the clean text
  password
* Add dependency to strimzi oauth-client, to use the JAAS Login Callback
  Handles class
* Enhanced doc about kafka connector


app/connector/kafka/src/main/resources/META-INF/syndesis/connector/kafka.json
```diff
@@ -140,11 +111,11 @@
       "secret": false,
       "type": "typeahead"
     },
-    "transportProtocol": {
```
zregvart: We need a migration script for this.

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaMetaDataExtension.java
```diff
@@ -40,11 +40,17 @@
...
-        final String brokers = ConnectorOptions.extractOption(parameters, "brokers");
-        final String certificate = ConnectorOptions.extractOption(parameters, "brokerCertificate");
-        final String transportProtocol = ConnectorOptions.extractOption(parameters, "transportProtocol");
-        LOG.debug("Getting metadata from Kafka connection to {} with protocol {}", brokers, transportProtocol);
+        final String topic = ConnectorOptions.extractOption(parameters, KafkaBrokerService.TOPIC);
+        final String brokers = ConnectorOptions.extractOption(parameters, KafkaBrokerService.BROKERS);
+        final String certificate = ConnectorOptions.extractOption(parameters, KafkaBrokerService.BROKER_CERTIFICATE);
+        final String securityProtocol = ConnectorOptions.extractOption(parameters, KafkaBrokerService.SECURITY_PROTOCOL);
+        final String saslMechanism = ConnectorOptions.extractOption(parameters, KafkaBrokerService.SASL_MECHANISM);
+        final String saslLoginCallbackHandlerClass = ConnectorOptions.extractOption(parameters, KafkaBrokerService.SASL_LOGIN_CALLBACK_HANDLER_CLASS);
```
zregvart, line 48: Could we make it simpler on the user and pick one (say OAuth bearer) so they don't need figure out the right combination of parameters?

app/connector/kafka/src/main/resources/META-INF/syndesis/connector/kafka.json
```diff
@@ -153,18 +124,102 @@
...
+      "javaType": "java.lang.String",
+      "kind": "parameter",
+      "label": "security",
+      "labelHint": "OAuth Client Secret or the password",
+      "order": "5",
+      "required": false,
+      "secret": true,
+      "type": "string"
+    },
+    "saslLoginCallbackHandlerClass": {
```
zregvart, line 196: Can we make this the default and not configurable by the user.

app/connector/kafka/src/main/resources/META-INF/syndesis/connector/kafka.json
```diff
@@ -153,18 +124,102 @@
...
+      "javaType": "java.lang.String",
+      "kind": "parameter",
+      "label": "common",
+      "order": "6",
+      "required": false,
+      "secret": false,
+      "placeholder": "io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler",
+      "type": "string"
+    },
+    "oauthTokenEndpointURI": {
```
zregvart, line 211: Is there a way to discover this? Perhaps via [well-known](https://datatracker.ietf.org/doc/html/rfc5785)? So the users don't need to provide the value?

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaVerifierExtension.java
```diff
@@ -65,27 +62,68 @@ protected Result verifyConnectivity(Map<String, Object> parameters) {
...
+        final String password = ConnectorOptions.extractOption(parameters, KafkaBrokerService.PASSWORD);
+        final String oauthTokenEndpointURI = ConnectorOptions.extractOption(parameters, KafkaBrokerService.OAUTH_TOKEN_ENDPOINT_URI);
         if (ObjectHelper.isNotEmpty(brokers)) {
-            KafkaBrokerService kafkaBrokerService = new KafkaBrokerServiceImpl(brokers, transportProtocol, certificate);
-            try {
-                kafkaBrokerService.ping();
-            } catch (KafkaBrokerServiceException e) {
+            boolean validParameters = true;
+            if (KafkaBrokerService.SASL_SSL.equals(securityProtocol)) {
+                if (StringUtils.isBlank(username)) {
```
zregvart: We can avoid depending on Commons Lang by doing this

```suggestion
                if (ObjectHelper.isEmpty(username)) {
```

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/service/KafkaBrokerServiceImpl.java
```diff
@@ -83,22 +87,66 @@ public void ping() throws KafkaBrokerServiceException {
...
+        properties.put(CommonClientConfigs.DEFAULT_API_TIMEOUT_MS_CONFIG, 5000);
+        // use certificate to authenticate to kafka server
+        if (brokerKeyStoreHelper != null) {
+            properties.put(SslConfigs.SSL_ENDPOINT_IDENTIFICATION_ALGORITHM_CONFIG, "");
+            properties.put(SslConfigs.SSL_KEYSTORE_LOCATION_CONFIG, brokerKeyStoreHelper.getKeyStorePath());
+            properties.put(SslConfigs.SSL_KEYSTORE_PASSWORD_CONFIG, brokerKeyStoreHelper.getPassword());
+            properties.put(SslConfigs.SSL_KEY_PASSWORD_CONFIG, brokerKeyStoreHelper.getPassword());
+            properties.put(SslConfigs.SSL_TRUSTSTORE_LOCATION_CONFIG, brokerKeyStoreHelper.getKeyStorePath());
+            properties.put(SslConfigs.SSL_TRUSTSTORE_PASSWORD_CONFIG, brokerKeyStoreHelper.getPassword());
+        } else if (StringUtils.isNotBlank(username) && StringUtils.isNotBlank(password)) {
```
zregvart: Let's not add a dependency we might not need

```suggestion
        } else if (ObjectHelper.isNotEmpty(username) && ObjectHelper.isNotEmpty(password)) {
```

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaConnector.java
```diff
@@ -15,40 +15,85 @@
...
-
 import org.apache.commons.beanutils.PropertyUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.kafka.common.config.SaslConfigs;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import static io.syndesis.connector.kafka.service.KafkaBrokerService.OAUTHBEARER;
+import static io.syndesis.connector.kafka.service.KafkaBrokerService.PLAIN;
 
-public class KafkaConnectionCustomizer implements ComponentProxyCustomizer {
```
zregvart, line 39: I'm still not sure why the customizer approach doesn't work here, is how secrets are fetched in the REST Swagger connector, for example fetching password here:

https://github.com/syndesisio/syndesis/blob/78d0ffb36cd15810b01aab864f6878aecb8e2f5f/app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/auth/basic/Basic.java#L33

Leads to:

https://github.com/syndesisio/syndesis/blob/78d0ffb36cd15810b01aab864f6878aecb8e2f5f/app/connector/rest-swagger/src/main/java/io/syndesis/connector/rest/swagger/Configuration.java#L77

Password being secret:

https://github.com/syndesisio/syndesis/blob/78d0ffb36cd15810b01aab864f6878aecb8e2f5f/app/connector/rest-swagger/src/main/resources/META-INF/syndesis/connector/rest-swagger.json#L171

I wonder if this is getting more complex than it needs to be.

zregvart (commented):
Overall looks good, some nitpickery added. @MelissaFlinn you might want to have a look at the documentation changes.

app/connector/kafka/src/main/resources/META-INF/syndesis/connector/kafka.json
MelissaFlinn: Thank you, Zoran. I created FUSEDOC-4596 to address related documentation updates...

MelissaFlinn (commented):
I created FUSEDOC-4596 to address doc updates related to this feature.

app/connector/kafka/src/main/resources/META-INF/syndesis/connector/kafka.json
claudio4j: I will revert this change, it was mostly to reflect the parameter name in camel-kafka component, but not worth to spend too much on this renaming.

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaMetaDataExtension.java
claudio4j: I thought about having it as default (handler class and token uri) when user selects oauthbearer, but as I am not familiar with java kafka clients, I am not sure if those values could change a lot. I see kafka clients may have different sets of combinations for example scram, jwt, etc. But as you did notice this, I think it would be a good idea to have them as default and documenting it.

app/connector/kafka/src/main/resources/META-INF/syndesis/connector/kafka.json
claudio4j: Going to have a default one (as defined in the placeholder) if user selects oauthbearer, and having the field available if the user wants to set a different class from kafka clients.jar.

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/KafkaVerifierExtension.java
claudio4j: Thanks for the tip, I will change it.

app/connector/kafka/src/main/java/io/syndesis/connector/kafka/service/KafkaBrokerServiceImpl.java
claudio4j: Thanks for the tip, I will change it.

zregvart (approved):
Let's merge this and polish as needed in subsequent PRs

claudio4j:
@zregvart I changed the code to accomodate some valid points you made. Thanks for reviewing.

zregvart (approved):

zregvart:
Things I'd like us to follow up on:

- [ ] docs (tracked by [FUSEDOC-4596](https://issues.redhat.com/browse/FUSEDOC-4596))
- [ ] add an integration test
- [ ] investigate the need for `ComponentProxyComponent` vs `ComponentProxyCustomizer`

PR #9705, opened 2021-08-02
fix: Add a test ID to the form group containing a select typeahead for ENTESB-13112

gashcrumb:
This is to work around https://github.com/patternfly/patternfly-react/issues/6107 by providing a test ID on the element that's as close as we can control, the enclosing form group.  The ID would be in the format `form-group-` and then the object property name.

Hopefully addresses [ENTESB-13112](https://issues.redhat.com/browse/ENTESB-13112)


kahboom (approved):
👍

PR #9707, opened 2021-08-03
Fixes for labelling error and removal of service account yaml

phantomjinx:
These fixes brought over from fuse-online-operator-prod-operator-metadata to sync up bundle generation.


phantomjinx:
Invalid PR.

install/operator/config/manifests/kustomization.yaml
```diff
@@ -12,8 +12,6 @@ kind: Kustomization
 #namePrefix: memcached-operator-
 
 # Labels to add to all resources and selectors.
-commonLabels:
-  app: syndesis
```
claudio4j, line 16: Shouldn't this be removed from upstream too ?

Issue #8707, opened 2020-06-16
Upgrade PatternFly to latest v4 and React 17


riccardo-forina:
## This is a...

<pre><code>
[ ] Feature request
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Documentation issue or request
[X] Chore
</code></pre>

## Description

https://medium.com/patternfly/patternflys-major-release-everything-you-need-to-know-1f9e50be81e8
https://github.com/patternfly/pf-codemods

This needs to be done in sync with AtlasMap: https://github.com/atlasmap/atlasmap/issues/2186

igarashitm:
I'd like to have this reconsidered, it's almost done in AtlasMap upstream
https://github.com/atlasmap/atlasmap/pull/2752
https://github.com/atlasmap/atlasmap/pull/2779

PR #9708, opened 2021-08-04
chore(deps): more UI pkg upgrades

seanforyou23:
This PR aligns dependency usage among several submodules. We previously had a somewhat [hacky](https://github.com/syndesisio/syndesis/blob/1.13.x/app/ui-react/packages/ui/src/index.css#L4-L12) way of importing patternfly styles, and to be honest I'm not sure I ever fully understood why that was necessary. My guess is that there were stylesheet load order issues coming from us having several submodules which all had different versions of patternfly being used. Anyway, I've managed to make all the versions aligned now and was able to get rid of all those individual stylesheet layers for a single import of `@patternfly/react-core/dist/styles/base.css` which is how it should be.

`@rh-uxd/integration-*` pkgs are no longer under active development, so we've ripped those bits out of @syndesis/syndesis and @syndesis/ui - we're no longer getting duplicate styles from this module.

One main thing to note is that I don't think we can merge this PR as is, I think we might need to bump PF versions yet again in the [atlasmap](https://github.com/atlasmap/atlasmap/blob/master/ui/packages/atlasmap/package.json#L74-L78) repo one more time and get that pulled in before everything will work out of the box. I was seeing `Cannot find module: '@patternfly/react-core/dist/js/experimental'` - which is old PF stuff that we're not using anymore, so I had to update the PF versions in atlasmap locally to get everything to build right. I can work on a PR to make these updates in atlasmap also, but it will be a couple days before I can work on that.

Overall things are looking pretty darn good! Please let me know if you see any issues or have any questions about these changes.

Screen Shot 2021-08-04 at 3 41 20 PM

steps toward resolving [ENTESB-14379](https://issues.redhat.com/browse/ENTESB-14379) / https://github.com/syndesisio/syndesis/issues/8707
remove `@rh-uxd/*` pkg (it is archived)
remove cross console navigation
further alignment of patternfly dependency versions
update usage of labels for integration status
shuffle yarn workspace hoist configurations


app/ui-react/syndesis/src/index.tsx
```diff
@@ -1,13 +1,14 @@
+import './index.css';
+// tslint:disable-next-line:ordered-imports
+// Don't move this line, otherwise we anger the pf4
```
gashcrumb: you moved that line, is that ok?  :smile: 

Wonder do we need this comment still?

app/ui-react/typings/rh-uxd/index.d.ts
```diff
@@ -0,0 +1,9 @@
+declare module "patternfly-react";
+declare module '@rh-uxd/integration-core/styles/*.css' {
+  const styles: any;
+  export default styles;
+}
+declare module '@rh-uxd/integration-core/styles/*.png' {
+  const images: any;
+  export default images;
+}
```
gashcrumb: Did you pull this package out?  Wonder do we still need these definitions and the package.json below?

app/ui-react/packages/ui/src/Layout/PfVerticalNavItem.tsx
```diff
@@ -53,17 +53,20 @@ function PfVerticalNavItem({
...
-        <Link
+      <NavItem>
+        <NavLink
           to={to}
-          className={classNames('pf-c-nav__link', {
-            ['pf-m-current']: isActive,
-          })}
+          className={
+            classNames(
+              // 'pf-c-nav__link',
```
gashcrumb: be good to delete this line?

app/ui-react/packages/api/package.json
```diff
@@ -60,8 +59,8 @@
     "immer": "^8.0.1",
     "monaco-editor-core": "^0.19.3",
```
gashcrumb: I didn't realize that we still had monaco references in here, these should be leftovers from removing the datavirt components.  I wonder if cleaning up these along with any relevant craco config (the `"vscode": require.resolve('monaco-languageclient/lib/vscode-compatibility'),` line) could help at all with the build error?  I can look at removing these in a separate PR tomorrow.

app/ui-react/syndesis/src/index.tsx
seanforyou23: 👍 yea I think it can be removed now

app/ui-react/typings/rh-uxd/index.d.ts
seanforyou23: Yep, and nope! This was leftover from before I realized it could just go away completely.. I'll remove

app/ui-react/packages/api/package.json
seanforyou23: Good to know, I'll see if I can remove them while I'm addressing the rest of the cleanup.

kahboom:
@seanforyou23 - Thanks for getting this sorted, it's looking pretty good! I don't have much to add with @gashcrumb 's review other than maybe we should close this until we get the PR for AtlasMap if it will take a few days? Just because we wouldn't be able to get this approved & merged without the build passing. What do you think? Also, let me know if you want any help with the upgrade for AtlasMap.

seanforyou23:
No problem @kahboom - I can close this PR for now and can reopen a new one once we have a newer version for these deps in atlasmap. I don't have time to send a PR for atlasmap at the moment, I will try to make some time to do it soon as I bring a couple other work items to a close.

PR #9722, opened 2021-09-08
fix(ci): align maven-deploy-plugin

squakez:
The plugin version diverged after the upgrade of base pom. Both extensions an integration boms are not depending from that pom and we must align versions in order to let the dependencies been deployed correctly. Deploy was failing because of the following error:
```
[ERROR] Found multiple major versions of maven-deploy-plugin; this is a malformed project.
	[org.apache.maven.plugins:maven-deploy-plugin:maven-plugin:2.7:runtime, org.apache.maven.plugins:maven-deploy-plugin:maven-plugin:3.0.0-M1:runtime]
```


squakez:
@zregvart do you think we can apply this?

zregvart (approved):
LGTM

PR #9728, opened 2021-09-16
Pf upgrade followup

seanforyou23:
This PR aligns dependency usage among several submodules. We previously had a somewhat [hacky](https://github.com/syndesisio/syndesis/blob/1.13.x/app/ui-react/packages/ui/src/index.css#L4-L12) way of importing patternfly styles, and to be honest I'm not sure I ever fully understood why that was necessary. My guess is that there were stylesheet load order issues coming from us having several submodules which all had different versions of patternfly being used. Anyway, I've managed to make all the versions aligned now and was able to get rid of all those individual stylesheet layers for a single import of `@patternfly/react-core/dist/styles/base.css` which is how it should be.

`@rh-uxd/integration-*` pkgs are no longer under active development, so we've ripped those bits out of @syndesis/syndesis and @syndesis/ui - we're no longer getting duplicate styles from this module.

Overall things are looking pretty darn good! Please let me know if you see any issues or have any questions about these changes.

Screen Shot 2021-08-04 at 3 41 20 PM

steps toward resolving [ENTESB-14379](https://issues.redhat.com/browse/ENTESB-14379) / https://github.com/syndesisio/syndesis/issues/8707
remove `@rh-uxd/*` pkg (it is archived)
remove cross console navigation
further alignment of patternfly dependency versions
update usage of labels for integration status
shuffle yarn workspace hoist configurations


app/ui-react/packages/atlasmap-adapter/src/DataMapperAdapter.tsx
```diff
@@ -76,6 +76,7 @@ export const DataMapperAdapter: React.FunctionComponent<IDataMapperAdapterProps>
     console.log('Atlasmap document', JSON.stringify(externalDocument));
     return (
       <AtlasmapProvider
+        logLevel={'info'}
```
gashcrumb: Should probably take this out, might be too noisy.

app/ui-react/packages/ui/src/Integration/IntegrationStatus.tsx
```diff
@@ -30,24 +29,24 @@ export class IntegrationStatus extends React.Component<
...
-        style = { background: global_active_color_100.value };
+        color = 'blue';
         break;
       case UNPUBLISHED:
         label = this.props.i18nUnpublished;
         break;
       case ERROR:
         label = this.props.i18nError;
-        style = { background: global_danger_color_100 };
+        color = 'red';
```
zregvart, line 43: Are `blue` and `red` show up as in the palette we're using, or are these `#f00` and `#00f`?

gashcrumb:
I suspect the QE checks fail because this PR only bumps the UI.  @igarashitm wonder, would it be easier to try and apply the updates on the Java side on this PR?  Or is there an easier way to tackle this?

app/ui-react/packages/ui/src/Layout/AppLayout.tsx
```diff
@@ -73,10 +67,6 @@ export const AppLayout: React.FunctionComponent<ILayoutBase> = ({
     : onNavigationExpand;
 
   const [breadcrumb, setHasBreadcrumb] = React.useState(null);
-  const [availableApps, setHasAvailableApps] = React.useState<
```
zregvart, line 76: I'm guessing we no longer need this? (ref #8561)

app/ui-react/packages/atlasmap-adapter/src/DataMapperAdapter.tsx
seanforyou23: Ah ok, yea I just added it because it throws a TS error that I otherwise couldn't seem to get around. Maybe we could make it an optional property or have a 'silent' type option?

zregvart (commented):
I guess we no longer need the bits implemented in #8561?

app/ui-react/packages/atlasmap-adapter/src/DataMapperAdapter.tsx
igarashitm: `warn` or `silent` should work.

igarashitm:
> apply the updates on the Java side on this PR

I sent a PR to @seanforyou23 's branch
https://github.com/seanforyou23/syndesis/pull/1

gashcrumb:
Thanks @igarashitm!

I built the branch here locally and had a look at the "import-edit-integration" integration there on staging and indeed the AtlasMap UI doesn't load up properly there, let's see how the QE checks do once the AtlasMap backend is in sync.

app/ui-react/packages/ui/src/Integration/IntegrationStatus.tsx
seanforyou23: Yep they correspond the hex values defined in the PF color palette, they don't export that type that lists the colors so I just redefined it there in the component - not sure the specific hex values off the top of my head but I can take a look soon - totally customizable if we need btw, I just used these because it provided better contrast than the previous dark text against the darker background color.

app/ui-react/packages/ui/src/Layout/AppLayout.tsx
seanforyou23: Good catch! I'll remove it.

app/ui-react/packages/ui/src/Integration/IntegrationStatus.tsx
gashcrumb: Ah, did those constants get removed then?

seanforyou23: No actually they are new! We didn't used to have all the fancy color support for the Label component in PF, and so lots of consumers were manually managing the background and text color which can be hit and miss - so I figured we could probably just use the built in color options for a more consistent look since there was concern about these labels and the text color being hard to read.

For example, 'blue' applies the css variable `--pf-c-label--m-blue--BackgroundColor` to the container - here's the [mapping](https://github.com/patternfly/patternfly-react/blob/main/packages/react-core/src/components/Label/Label.tsx#L57-L65) for reference

app/ui-react/packages/ui/src/Layout/AppLayout.tsx
seanforyou23: Oops, just realized I read your comment/question wrong - you are correct, the cross console navigation effort was sunset so basically just needed to rip that stuff out and put it back like it was. Most of the discussion about it was offline, but here's an [issue](https://github.com/rh-uxd/appservices-patternfly/issues/53) that confirms it in case it's helpful .

gashcrumb:
Looks like we had an API change @igarashitm 

```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project server-endpoint: Compilation failure
Error:  /home/runner/work/syndesis/syndesis/app/server/endpoint/src/main/java/io/syndesis/server/endpoint/v1/handler/atlas/SyndesisJavaService.java:[19,26] error: cannot find symbol
Error:    symbol:   class MavenClasspathResponse
Error:    location: package io.atlasmap.java.v2
Error:  -> [Help 1]
```

Also @seanforyou23 is having a look to be sure the CSS is being applied correctly on AtlasMap, he may have some questions there.

seanforyou23:
Ah, the base styles for atlasmap were not being imported. After adding it everything seems to have snapped right back in place.. win! If we can get the atlasmap stylesheet published to dist like the [other packages](https://github.com/syndesisio/syndesis/pull/9728/files#diff-15ebdf20b76e95ab793492940c6b2fd716353ac3869d0392199a7c98992d23caR8) it would make for a cleaner import path. In the mean time I'm just importing it from node_modules which I suppose it fine for now. Let me know if you find anything else!

Screen Shot 2021-09-17 at 5 04 08 PM

igarashitm:
ugh right, maven classpath request has been removed - removed it and trying to build but then I hit "unused dependency" error in verifier, trying to see what's wrong.

igarashitm:
@seanforyou23 OK this should fix the backend build error - https://github.com/seanforyou23/syndesis/pull/2

igarashitm:
@seanforyou23 one more chore, unused dependency in POM I couldn't find in local build - https://github.com/seanforyou23/syndesis/pull/3

gashcrumb:
@mkralik3 wonder is there an outage?  Seems the PR checks are stuck.

mkralik3:
@gashcrumb I will try to rerun it.

mkralik3:
retest this please

gashcrumb:
retest this please

mkralik3:
We can ignore that 1 failed test because the latest Syndesis upstream contains updated AtlasMap and probably some selector was changed so UI test suite needs an update too. I will do it after verified the last prod build which contains the older AtlasMap.

gashcrumb:
Ok thanks @mkralik3!

gashcrumb (approved):
:+1:

zregvart:
Merged, let's get this in and polish if/as needed.

PR #9734, opened 2021-09-23
chore(deps): remove node-sass download

zregvart:
Seems we are no longer using node-sass, so this removes the build
configuration used to download the prebuilt binary.

Ref. https://issues.redhat.com/browse/ENTESB-17290


zregvart:
cc @cunningt @squakez

squakez (approved):

PR #9740, opened 2021-09-29
ENTESB-17341 nodeAffinity and toleration are not set in an integration pod

claudio4j:
https://issues.redhat.com/browse/ENTESB-17341

OpenShiftServiceImpl was using an outdated v1beta2 syndesis CRD version
updated to v1beta3

The syndesis CRD version check is in operator build, because any syndesis CRD upgrade is driven by the operator code, so it is my understanding that if there is a change in the syndesis api version, that is going to occur in operator code.


phantomjinx (commented):
I hadn't seen this but have gone another way by fetching the version of the CRD at runtime so removing the need for the constant. See #9741.

claudio4j:
Closing this in favor of https://github.com/syndesisio/syndesis/pull/9741

PR #9747, opened 2021-10-05
feat: set integration pod environment variables

zregvart:
This adds the ability to add custom environment variables set on the
DeploymentConfig and available in the container of the running
integration Pod on the Save integration screen of the integration edit
wizard.

Ref. https://issues.redhat.com/browse/ENTESB-16398


zregvart:
For reviewers, this _is_ a quick & dirty way of implementing this functionality, stuff that I know can be improved:
 - there are no tests
 - there is no (new) inline documentation
 - ~~I did not localize the UI~~
 - binding is iffy at the least
 - UI looks awkward

Screenshot:
Screenshot 2021-10-05 at 23-11-38 Save the integration - Syndesis

zregvart:
Converted to draft, most likely won't make it to 1.13.x, so let's target for 1.14.x

zregvart:
I'm happier with this version, the environment section is now incorporated into the auto-form of the save page instead of being a separate component, so no more bind-in-render. Still would benefit from some UX help.

localhost_3000_integrations_

phantomjinx (approved):
LGTM - but another set of eyes would be advantagous

app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java
```diff
@@ -272,12 +274,14 @@ protected void ensureDeploymentConfig(String name, DeploymentData deploymentData
...
-                    new EnvVar("JAEGER_SAMPLER_PARAM", "1", null) };
+            final List<EnvVar> vars = new ArrayList<>(Arrays.asList(
+                new EnvVar("LOADER_HOME", config.getIntegrationDataPath(), null),
+                new EnvVar("AB_JMX_EXPORTER_CONFIG", "/tmp/src/prometheus-config.yml", null),
+                new EnvVar("JAEGER_ENDPOINT", jaegerCollectorUri, null),
+                new EnvVar("JAEGER_TAGS", "integration.version="+deploymentData.getVersion(), null),
+                new EnvVar("JAEGER_SAMPLER_TYPE", "const", null),
+                new EnvVar("JAEGER_SAMPLER_PARAM", "1", null)));
+            vars.addAll(deploymentData.getEnvironment());
             final Map<String, EnvVar> envVarMap = new HashMap<>();
```
claudio4j, line 285: The environment variables are added only for existing integrations, this logic to add the environment variables is missing when there is a new integration, see line 417.

claudio4j (changes requested):

app/server/openshift/src/main/java/io/syndesis/server/openshift/OpenShiftServiceImpl.java
zregvart: Good catch, I think 6e249d9b944910a5e31c908abfb338cfdddf51ae should fix that, can you have another look?

claudio4j: The `deploymentData.getEnvironment()` is empty in `ensureDeploymentConfig` method, I didn't investigate why.

zregvart: Was at least one environment variable set for the Integration you were testing with?

Should be set from here:

https://github.com/syndesisio/syndesis/pull/9747/files#diff-792449018802d5dc5b366781e0a37a4f0d7b3d033400b95260e7953d8e9ffa1eR153

claudio4j: The `integration.getEnvironment()` is empty in `PublishHandler.java createDeploymentData()`. I built syndesis with `syndesis build -b -m ui-react,s2i -f -i`

claudio4j: common/model/target/generated-sources/annotations/io/syndesis/common/model/integration/ImmutableIntegration.java was missing the WithEnvironment attributes, just did `mvn clean` in common dir, the ImmutableIntegration.java was created correctly. I just forgot to wipe old stuff build earlier. Sorry.

claudio4j (approved):

kahboom:
@zregvart nice improvement in the UI from the original PR :)

zregvart:
> @zregvart nice improvement in the UI from the original PR :)

Still a bit awkward UX...

PR #9750, opened 2021-10-08
chore: upgrade org.glassfish:jakarta.el to 3.0.4

zregvart:
Ref. https://issues.redhat.com/browse/ENTESB-16630


zregvart:
cc @claudio4j

claudio4j:
The issue was changed to https://issues.redhat.com/browse/ENTESB-17580
Thanks for taking this one.

PR #9754, opened 2021-10-12
fix: maintain data shape parameters

zregvart:
fix: maintain data shape parameters (afc06bb)

Even though an issue in AtlasMap was resolved[1] it doesn't seem to help
in the Syndesis use case. This refactors the support for providing data
shape parameters and uses one dirty trick to avoid resetting the
internal state of AtlasMap's `ParametersDialog`. The key is in setting
the `enabled=true` parameter which should leave the value present in the
dialog. The refactoring focused on removing the complexity of AtlasMap
to the AtlasMap adapter so that pages that use the
`DataShapeParametersDialog` can have a simpler interface and need not
take any concern of the AtlasMap internal state.

Ref. https://issues.redhat.com/browse/ENTESB-17416

[1] https://github.com/atlasmap/atlasmap/issues/2990

chore: don't hardcode AtlasMap CSV parameters (05c8328)

We now can access AtlasMap supported CSV parameters directly from
AtlasMap[1], so we don't need to hardcode them on our end.

### DIFF
--- a/app/ui-react/packages/atlasmap-adapter/package.json
+++ b/app/ui-react/packages/atlasmap-adapter/package.json
@@ -71,6 +71,7 @@
   },
   "dependencies": {
     "@atlasmap/atlasmap": "^2.3.1",
+    "@atlasmap/core": "^2.3.1",
     "@types/patternfly-react": "*",
     "loglevel": "^1.7.1",
     "react-dnd-html5-backend": "^14.0.0",
--- a/app/ui-react/packages/atlasmap-adapter/src/DataMapperAdapter.tsx
+++ b/app/ui-react/packages/atlasmap-adapter/src/DataMapperAdapter.tsx
@@ -102,7 +102,7 @@ export interface IParameterOption {
   value: string;
 }
 
-export interface IParameter {
+export interface IParameterDefinition {
   name: string;
   label: string;
   value: string;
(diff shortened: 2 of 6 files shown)


app/ui-react/packages/atlasmap-adapter/package.json
```diff
@@ -71,6 +71,7 @@
   },
   "dependencies": {
     "@atlasmap/atlasmap": "^2.3.0-M.3",
+    "@atlasmap/core": "^2.3.0-M.3",
```
apupier: The [2.3.0 has been released](https://github.com/atlasmap/atlasmap/releases/tag/atlasmap-2.3.0) few hours ago with the fix of the [pointed issue](https://github.com/atlasmap/atlasmap/issues/2990). Would it be a possibility to upgrade to 2.3.0 and avoid any other changes?

apupier: hum, I guess I misunderstood, the 2.0.0-M3 seemed to already contains the fix for pointed issue. So would mean that both the fix upstream and the workaround in this PR are needed.

zregvart: I've looked at the pull request for that issue and found nothing that would help in this use case. I don't think that issue was fixed for us.

mkralik3:
retest this please

kahboom:
Added some comments, but not about anything major. I don't see issues with these changes, however, it's difficult to tell the impact on other parts of the app. Would be nice to get someone who has specifically worked on the AtlasMap adapter before to take a look as well.

zregvart:
Thanks for having a look @kahboom, I don't see the comments did you submit your review?

PR #9755, opened 2021-10-12
chore: pull images as possible from quay.io

### LINKED ISSUE #9751: Upgrade 1.12 to 1.13 doesn't work
## This is a...
<pre><code>
[ ] Feature request
[x] Bug report  
</code></pre>

## Description
Upgrade Syndesis 1.12.1 to 1.13.0-20211011 doesn't work. The pod `db-backup-` is not able to be initialized successfully due to this error:
```
Events:
  Type     Reason          Age   From               Message
  ----     ------          ----  ----               -------
  Normal   Scheduled       27s   default-scheduler  Successfully assigned mkralik/db-backup-1633952344-w6z94 to crc-scmfr-master-0
  Normal   AddedInterface  24s   multus             Add eth0 [10.217.0.85/23] from openshift-sdn
  Normal   Pulling         24s   kubelet            Pulling image "crunchydata/crunchy-pgdump:centos7-12.5-4.5.1"
  Normal   Pulled          5s    kubelet            Successfully pulled image "crunchydata/crunchy-pgdump:centos7-12.5-4.5.1" in 18.667838578s
  Normal   Created         5s    kubelet            Created container backup-db-compiler
  Normal   Started         5s    kubelet            Started container backup-db-compiler
  Normal   Pulling         5s    kubelet            Pulling image "centos:7"
  Warning  Failed          4s    kubelet            Failed to pull image "centos:7": rpc error: code = Unknown desc = unable to retrieve auth token: invalid username/password: unauthorized: Invalid Username or Password
  Warning  Failed          4s    kubelet            Error: ErrImagePull
  Normal   BackOff         3s    kubelet            Back-off pulling image "centos:7"
  Warning  Failed          3s    kubelet            Error: ImagePullBackOff

```

Tried on OCP 4.6, OCP 4.8 and CRC 1.33.1 (OCP 4.8.12).

zregvart:
chore: pull images as possible from quay.io (bcce7e6d64f7035ed223862a21640b0a3004a983)

All but the crunchy images are available on quay.io, so we can pull from
there and try to avoid Docker Hub rate limits. This also upgrades some
of the releases, mostly as older releases are not available on quay.io.

Fixes #9751

chore: tag default branch releases with latest (2d1ade6)

On Docker Hub we have a super old release tagged with `latest`, on
quay.io we don't have any image tagged with `latest`. This tags a
release from the GitHub default branch with `latest` on (pre-)release.

### DIFF
--- a/install/operator/build/conf/config-test.yaml
+++ b/install/operator/build/conf/config-test.yaml
@@ -14,14 +14,14 @@ Syndesis:
                 Channel: "stable"
             SamplerType: "const"
             SamplerParam: "0"
-            ImageAgent: "jaegertracing/jaeger-agent:1.13"
-            ImageAllInOne: "jaegertracing/all-in-one:1.13"
-            ImageOperator: "jaegertracing/jaeger-operator:1.13"
+            ImageAgent: "quay.io/jaegertracing/jaeger-agent:1.27"
+            ImageAllInOne: "quay.io/jaegertracing/all-in-one:1.27"
+            ImageOperator: "quay.io/jaegertracing/jaeger-operator:1.27"
         Ops:
             Enabled: false
         Todo:
             Enabled: false
-            Image: "docker.io/centos/php-71-centos7"
+            Image: "quay.io/centos7/php-73-centos7:7.3"
         Knative:
             Enabled: false
         PublicApi:
@@ -29,14 +29,14 @@ Syndesis:
             RouteHostname: "mypublichost.com"
     Components:
         Oauth:
-            Image: "quay.io/openshift/origin-oauth-proxy:v4.0.0"
+            Image: "quay.io/openshift/origin-oauth-proxy:4.9"
         UI:
-            Image: "docker.io/syndesis/syndesis-ui:latest"
+            Image: "quay.io/syndesis/syndesis-ui:latest"
         S2I:
-            Image: "docker.io/syndesis/syndesis-s2i:latest"
+            Image: "quay.io/syndesis/syndesis-s2i:latest"
         Prometheus:
             Rules: ""
-            Image: "docker.io/prom/prometheus:v2.1.0"
+            Image: "quay.io/prometheus/prometheus:v2.30.3"
             Resources:
                 Limit:
                     Memory: "512Mi"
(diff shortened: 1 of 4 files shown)


mkralik3:
If I recall correctly, the Jaeger cannot be update (>1.17.1) due to this limitation on OCP 3.11 https://issues.redhat.com/browse/ENTESB-16313 ( workaround can be using the old logging on OCP 3.11 )

mkralik3:
retest this please 
(wrong qe jenkins was triggered)

mkralik3:
I have tried the upgrade manually on OCP4 and it works so it looks that the other images work ok. 
But I noticed that the syndesis images (s2i, meta, server, ui, upgrade) in quay don't have `latest` tag. 
e.g. https://quay.io/repository/syndesis/syndesis-meta?tag=latest&tab=tags

zregvart:
> But I noticed that the syndesis images (s2i, meta, server, ui, upgrade) in quay don't have `latest` tag.
> e.g. [quay.io/repository/syndesis/syndesis-meta?tag=latest&tab=tags](https://quay.io/repository/syndesis/syndesis-meta?tag=latest&tab=tags)

That should be addressed by 2d1ade6, we haven't pushed the `latest` tag to Docker Hub either, and the `latest` there is 8 months old by now.

zregvart:
> If I recall correctly, the Jaeger cannot be update (>1.17.1) due to this limitation on OCP 3.11 [issues.redhat.com/browse/ENTESB-16313](https://issues.redhat.com/browse/ENTESB-16313) ( workaround can be using the old logging on OCP 3.11 )

I would be okay if we dropped support for OpenShift 3 in the upstream releases. That would mean dropping support for minishift in favor of CRC. What do others think?

claudio4j:
There is an error running syndesis-oauthproxy on OCP 3.11 
```
github.com/openshift/oauth-proxy/providers/openshift/provider.go:347: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps "oauth-serving-cert" is forbidden: User "system:serviceaccount:syndesis:syndesis-oauth-client" cannot list configmaps in the namespace "openshift-config-managed": no RBAC policy matched
```
Perhaps because the origin-oauth-proxy was updated to 4.9 ?

claudio4j:
> I would be okay if we dropped support for OpenShift 3 in the upstream releases. That would mean dropping support for minishift in favor of CRC. What do others think?

That worries me because we may find problems very much later in development cycle when the productized builds are tested.

mkralik3:
I think we can leave previous old Jaeger images there because on OCP4, Jaeger is installed via OLM so the Jaeger images from config are used only in OCP3 (correct me if i am wrong). Without support OCP3, I don't see any benefit to update them ( maybe if OCP4 can be without OLM (?) )

zregvart:
Seems like this needs to mature a bit. I'll convert this to draft and we can discuss what needs to be done for 1.14, for 1.13 I'll create a change just for the `centos:7` image.

zregvart:
Okay the release tagging of images with `latest` for the default branch is merged via #9757 (picked from 2d1ade6, and a the change of centos pull specification is in #9758. I'll rebase this and create a new issue to drop support for OpenShift 3, use more images from quay.io and see about not using crunchy images.

zregvart:
This is now ready for review, I guess some of the issues around not supporting OpenShift 3 should be done in #9760, for example removing the need to install Jaeger.

PR #9709, opened 2021-08-04
ENTESB-12106 Remove the deployIntegrations flag from the CR

claudio4j:
https://issues.redhat.com/browse/ENTESB-12106

* Increase apiVersion to v1beta3 due to removing a field from CRD
* Remove unused `install/operator/version/version.go`
* Added a `synapi` prefix to syndesis api imports to make it easier to change api version in the future.

PR #9766, opened 2021-10-18
chore: remove `demoData` flag from Syndesis CR

zregvart:
This removes the `demoData` flag from Syndesis Custom resource. The same
effect can now be achieved by enabling the `todo` addon.

Since the operator merge code doesn't support map entry removal, only
value replacement and loosely entry addition. Properties relating to the
sampledb database on resources, namely `DeploymentConfig`s for
`syndesis-server` and `syndesis-db` need to remain. Their values are
changed so that the functionality of enabling/disabling the `todo` addon
is performed.

To apply the change to the PostgresSQL service, post-start lifecycle
hook in `syndesis-db` `DeploymentConfig` is changed to trigger the
redeployment. Otherwise changing the `syndesis-sampledb-config`
`ConfigMap` would not take effect. This causes, hopefully a brief,
downtime of the PostgreSQL pod, which could be avoided if the operator
did not apply changes via the templating system, and it would connect to
the database to apply the changes. This is left as an exercise for
later to make this change as simple as possible.

Ref. https://issues.redhat.com/browse/ENTESB-15863


zregvart:
retest this please

claudio4j:
Doesn't this api change, requires a new v1beta4 api and api migration ? In case there is users with demodate=true in v1beta3

zregvart:
> Doesn't this api change, requires a new v1beta4 api and api migration ? In case there is users with demodate=true in v1beta3

We introduced v1beta3 in 1.13.x via #9709, so I don't think we need to migrate or increase custom resource the version until we release 1.13.0.

I hypothesized that most folk that have `demoData` turned on also turn on the `todo` addon -- that's how the quickstart examples do it. So don't think a migration from `v1beta2` or earlier is necessary, especially as this is a demo feature not a production feature, so performing a manual step of enabling the `todo` addon, in what I expect to be an almost non-existent minority of cases, should not be an issue.

claudio4j (approved):

PR #9767, opened 2021-10-20
Large memory usage in custom SOAP API client connectors

zregvart:
When dealing with large WSDL files (>1MB) containing a large number (100s) of operations the amount of data used by strings or concatenation of strings in `StringBuffer`s/`StringBuilder`s for specifications stored in connector's configured properties or in the data shapes of actions memory limits, either on the server or in the browser, can be easily breached.

This solves that issue by not storing the WSDL specification in connector's configured properties but in the file store (blob) and pointing to it from the `specification` configured property (e.g. has the value `db:<connectorId>`); also the data shapes are not generated when the custom connector is created but declared dynamic, by adding the `dynamic` tag to their actions, and computed on demand when needed by the UI, i.e. when adding the connection's action utilizing the custom SOAP client connector to a flow of an integration.

For a ~1MB WSDL file with 115 operations this results in an connector JSON of ~60KB rather than ~23MB.

Ref. https://issues.redhat.com/browse/ENTESB-15863


zregvart:
Some issues remaining to finish:

- [x] restore fetching the WSDL specification via URL (currently removed)
- [x] go through the changes once more and do more thorough review
- [x] test integrations using OpenAPI and SOAP client connectors, IIRC we have an OpenAPI end-to-end integration test (that includes the generator), not sure about SOAP

app/ui-react/packages/api/src/useApiConnectorSummary.tsx
```diff
@@ -29,26 +31,41 @@ export function useApiConnectorSummary(
...
     const fetchSummary = async () => {
       setLoading(true);
 
+      const body = new FormData();
+      const connectorSettings = {
+        configuredProperties: {
+          ...configured,
+        } as IApiConnectorSummaryOptions,
+        connectorTemplateId: connectorTemplateId
+          ? connectorTemplateId
```
kahboom: Probably could just do `connectTemplateId ?? 'swagger-connector-template'`

app/ui-react/packages/api/src/useApiConnectorSummary.tsx
```diff
@@ -29,26 +31,41 @@ export function useApiConnectorSummary(
   >(undefined);
 
   React.useEffect(() => {
-    if (!specification) {
+    if (!specification && !url) {
```
kahboom, line 34: Why are we only returning if there is _both_ no specification and url? Just curious why it's not `||` instead, I may have missed something.

app/ui-react/syndesis/src/modules/apiClientConnectors/pages/create/DetailsPage.tsx
```diff
@@ -25,17 +25,24 @@ import routes from '../../routes';
...
 export const DetailsPage: React.FunctionComponent = () => {
   const { t } = useTranslation(['apiClientConnectors', 'shared']);
   const { pushNotification } = React.useContext(UIContext);
   const { state, history } = useRouteData<null, IDetailsPageRouteState>();
-  const createApiConnector = useApiConnectorCreator();
+  // if there is a specification configured, that means that it was provided by the server in the `.../info` response
+  // which means that the server prefers receiving the specification via configuredProperties, not as a separate mime part
+  // and passing a defined value for specification to useApiConnectorCreator results in a separate mime part
+  const createApiConnector = useApiConnectorCreator(
+    state.configured?.specification ? undefined : state.specification
```
kahboom: same here. for `const createApiConnector = useApiConnectorCreator(state.configured?.specification ? undefined : state.specification` we can now use the nullish coallescing operator, like this: `const createApiConnector = useApiConnectorCreator(state.configured?.specification ?? state.specification`

app/ui-react/syndesis/src/modules/apiClientConnectors/pages/create/DetailsPage.tsx
```diff
@@ -60,14 +67,25 @@ export const DetailsPage: React.FunctionComponent = () => {
       {({ allowNavigation }) => {
         const onSubmit = async (values: IConnectorValues, actions: any) => {
           actions.setSubmitting(true);
+          const { name, description, icon, ...configured } = values;
+          if (state.configured) {
+            Object.entries(state.configured).forEach(([k, v]) => {
```
kahboom, line 72: can we map here?:
```
state.configured.map((k, v) => { if(v) { configured[k] = v as string; } });
```
or maybe even `v.toString();` (I forget what type `v` is?)

kahboom (approved):
@zregvart - In terms of the UI parts, overall, this looks good to me. I added some comments, but most were just syntax related so I think it's really up to you whether you think it'd be worth making those changes.

app/ui-react/packages/api/src/useApiConnectorSummary.tsx
zregvart: Cool, didn't know about that syntax

app/ui-react/packages/api/src/useApiConnectorSummary.tsx
zregvart: Without a `specification` _and_ without `url` to the specification it doesn't make sense to invoke the server API. I think this needs a comment to explain it, thanks for the feedback!

app/ui-react/syndesis/src/modules/apiClientConnectors/pages/create/DetailsPage.tsx
zregvart: `state.configured` is a [`ICreateConnectorProps`](https://github.com/syndesisio/syndesis/blob/7f5f86f1e6802caa062b99654f036fbe443927af/app/ui-react/syndesis/src/modules/apiClientConnectors/models.ts#L4-L28), so values are either a `string` or a `boolean`, and I need to convert it to `{ [key: string]: string }` for `configuredProperties` [here](https://github.com/syndesisio/syndesis/pull/9767/files#diff-0e78c5f9c21b28707cf7299c13680bedeccd11bf8a4a9ccc9207d5002850ccf4R13-R15) as in the server model all `configuredProperties` are `Map<String, String>`. I tried to not bundle concerns of client OpenAPI clients vs SOAP clients, i.e. not have a type that contains properties of both, but did that only partially, a bigger refactor would make the wizard oblivious to the type of the connector and rely on the response from the custom connector summary API, but I didn't want to make this bigger than it already is.

zregvart: Anyhow, changed to `v.toString()` not to get a runtime error, good catch!

PR #9777, opened 2021-10-27
Upgrade Fabric8 Kubernetes Client to 4.13.3

### LINKED ISSUE #9773: Upgrade Fabric8 Kubernetes client
For OpenShift 4.9 compatibility the API version of CRDs needs to be `v1`, see #9772. I think we need to support both `apiextensions.k8s.io/v1beta2` and `apiextensions.k8s.io/v1` at the same time.

zregvart:
chore: upgrade Fabric8 Kubernetes Client to 4.13.3 (e25788a)

Also addresses some of the quirks of the upgrade and in the Kafka broker
/topic auto discovery allows for `apiextensions.k8s.io/v1beta1` and
`apiextensions.k8s.io/v1` versions.

refactor: integration scheduling from ConfigMap (79af4ed)
Seems odd to tie the server back to the Syndesis CR, it adds an
additional dependency that we can do without and increases the
complexity. This removes the Syndesis CRD definition from the server and
uses the values from the ConfigMap instead. The net effect of this
should not be visible in any behavioral change, e.g. ConfigMap is
watched and the configuration will reload if Syndesis CR is changed and
the operator updates it.

TODO:

- [x] test with several versions of Strimzi
- [x] test on OpenShift 4.9

Fixes #9773


phantomjinx (approved):
LGTM

PR #9786, opened 2021-11-04
chore: upgrade operator-sdk to v1.14.0

zregvart:
This seems to trickle down and upgrade quite a bit of dependencies as
well. We can't upgrade logr to v1.2.0 (from 0.4.0) as it seems to have a
breaking change in it as `sigs.k8s.io/controller-runtime/pkg/log` and
`k8s.io/klog/v2` would need to upgrade to it as well.

It builds and tests pass, not sure if an additional step is needed to
regenerate the scaffolded code.


phantomjinx:
This dependency is separate from the scaffolding version which is specified in [config/var/Makefile](https://github.com/syndesisio/syndesis/blob/1.13.x/install/operator/config/vars/Makefile#L2), which is currently set to 1.5.0 but in actual fact has a bundle config generated with 1.9.2 (locally installed here).

Migrating the scaffolding requires going through the migration docs in https://sdk.operatorframework.io/docs/upgrading-sdk-version/

zregvart:
Closing old pull requests.

PR #9791, opened 2021-11-08
fix: manage environment variables on publish

zregvart:
Instead of managing the environment variables, i.e. in this case removal
of environment variables, at the point the integration is updated this
manages the environment variables at the point of publishing by
examining environment variables of previous deployments and adding any
environment variables that are not present on the latest deployment to
the list of environment variables to remove.

Ref. https://issues.redhat.com/browse/ENTESB-17748

### DIFF
--- a/app/common/model/src/main/java/io/syndesis/common/model/WithEnvironment.java
+++ b/app/common/model/src/main/java/io/syndesis/common/model/WithEnvironment.java
@@ -17,7 +17,6 @@ package io.syndesis.common.model;
 
 import java.util.Collections;
 import java.util.Map;
-import java.util.Set;
 
 import org.immutables.value.Value;
 
@@ -28,8 +27,4 @@ public interface WithEnvironment {
         return Collections.emptyMap();
     }
 
-    @Value.Default
-    default Set<String> getRemovedEnvironment() {
-        return Collections.emptySet();
-    }
 }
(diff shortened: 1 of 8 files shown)


zregvart:
@mkralik3 looking at the smoke test failure, and it's a bit of a low-resolution image in the report, the environment variable has name set to `ENV4ENV4NAMEUPDATED`, which would mean that this `input.clear()` did not clear the `input`, does that ring any bells to you?

https://github.com/syndesisio/syndesis-qe/blob/8cf8a2fb3076b9bca78f638b4d47042111703a31/ui-common/src/main/java/io/syndesis/qe/pages/integrations/editor/CreateIntegration.java#L118

mkralik3:
@zregvart Probably. I will try it manually.

mkralik3:
@zregvart  It works on Firefox. It looks that it doesn't work only on Chrome. I have found in the different part of the test suite that there are some other steps ( i recall some problems was also on macOS )
```
        input.sendKeys(Keys.chord(Keys.SHIFT, Keys.HOME));
        input.sendKeys(Keys.BACK_SPACE);
        input.clear();
```

anyway, the test passed so we can merge this PR.

zregvart:
> @zregvart It works on Firefox. It looks that it doesn't work only on Chrome. I have found in the different part of the test suite that there are some other steps ( i recall some problems was also on macOS )

Could it be that the UI binds the value after the `.clear()` was invoked? So the order of events is:

1. `input.clear()`
2. React binds attribute to `value="ENV4"`
3. Additional keys are sent

Anyhow thank you for double checking this!

PR #9800, opened 2021-11-09
ENTESB-17767 Integration with MongoDB connector is not able to start

claudio4j:
https://issues.redhat.com/browse/ENTESB-17767

* Updated mongodb java driver as camel mongodb component requires it
* Updated to camel 2.23.2 fuse 7.10
* For some unknown reason there should be a wait time before connecting to mongodb embedded server

### DIFF
--- a/app/extension/bom/pom.xml
+++ b/app/extension/bom/pom.xml
@@ -28,7 +28,7 @@
 
   <properties>
     <spring-boot.version>2.3.12.RELEASE</spring-boot.version>
-    <camel.version>2.23.2.fuse-790053</camel.version>
+    <camel.version>2.23.2.fuse-7_10_0-00016</camel.version>
     <!-- we must ensure to have this version aligned with extension-bom and integration-bom -->
     <dep.plugin.deploy.version>2.8.2</dep.plugin.deploy.version>
   </properties>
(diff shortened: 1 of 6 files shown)


app/extension/bom/pom.xml
```diff
@@ -28,7 +28,7 @@
 
   <properties>
     <spring-boot.version>2.3.12.RELEASE</spring-boot.version>
-    <camel.version>2.23.2.fuse-790053</camel.version>
+    <camel.version>2.23.2.fuse-7_10_0-00016</camel.version>
```
zregvart, line 31: Let's do this in a separate PR, seems that we're hitting ENTESB-16583

claudio4j: Unfortunately these two should be together, if syndesis uses 2.23.2.fuse-790053, then we should use `com.mongodb.MongoClient` in mongodb connector. Also, there are tests failures, because camel component wants to use the `com.mongodb.MongoClient`.

zregvart: Chicken or the egg, well, we can't do this upstream it seams. Not unless the PR for ENTESB-16583 gets merged and all dependencies get pushed to early access.

claudio4j:
Pushed a change to override xstream to use a version from a public repo.

claudio4j:
Putting in draft mode to wonder how to use a fuse camel with non public artifacts.

PR #9802, opened 2021-11-10
ENTESB-17480 Fix to align jetty version into syndesis-meta

claudio4j:


zregvart:
Hmm, from this commit onward (3x) the integration tests stopped working. Might be worth having a look.

PR #9809, opened 2021-11-12
chore: add dependency report to the S2I image

zregvart:
When trying to track down why a dependency is present in the S2I image
we often need to drill down Maven POM files in search of the origin of a
dependency, it is also difficult to understand what brought in a
transitive dependency.

This adds the dependency report in a file called `dependencies.txt` to
the S2I image and the repository ZIP to help with that.


zregvart:
retest this please

claudio4j (approved):

PR #9819, opened 2021-11-15
chore: upgrade jaegger and opentracing

zregvart:
Ref. https://issues.redhat.com/browse/ENTESB-17769

PR #9822, opened 2021-11-16
fix: accidental Tomcat inclusion

zregvart:
In #9819 embedded Tomcat dependency sneaked in, which caused it to be
used instead of Undertow, but more importantly the integration failed to
start as the version that was brought in by libthrift was not compatible
with the version that the spring-boot-tomcat-starter expected.


oscerd (approved):

zregvart:
retest this please

PR #9827, opened 2021-11-24
fix: declare spring-boot-starter version

zregvart:
When we removed import of `spring-boot-dependencies` BOM we lost the
declaration, among others, of `spring-boot-starter` version. This is not
an issue for integrations built with provided set of connectors/steps,
but it could be an issue with extensions that could declare it as a
dependency.

Ref. https://issues.redhat.com/browse/ENTESB-17890


mkralik3:
retest this please

PR #9833, opened 2021-12-10
chore: configure CodeQL security analysis

zregvart:
This adds CodeQL security analysis scanning. We might need to add the configuration in `.github/codeql/` as well. We'll see how this goes it might increase the build time as well.


zregvart:
One immediate drawback is that we should skip CodeQL checks for doc changes, perhaps a better approach would be to integrate the pull request/branch merge scanning into the existing checks workflow and keep only the cron invocation here. Not sure if CodeQL instruments the compiled code in some way, so it needs to be kept separate.

zregvart:
retest this please

Delawen (approved):

oscerd (approved):

zregvart:
retest this please

PR #9881, opened 2022-01-27
fix(server): correct totalCount in ListResult

zregvart:
When filters (e.g. pagination) are used the `totalCount` gets reset to
whatever the filter outcome size is (e.g. page size), and the initial
`totalCount` that reflects the number of available items is lost.

This refactors the `ListResult` to clearly support two cases: partial or
complete item set was provided; also for the existing filtering
operations partial case is used with the `totalCount` of the original
item set.


zregvart:
retest this please

I think the local Maven repository might have had an older version of common-model cached, let's see if re-running the test helps.

PR #9890, opened 2022-02-03
refactor: use embedded resources

zregvart:
The way we bundle resources within the executable is problematic.
Resources need to be included by running code generation, so if skipped
older version of the resource will be included; and the generated .go
files contain the resources as byte arrays in hex, nullifying the
utility of diff tools and making resolving conflicts difficult.

This instead uses the `//go:embed` directive to embed the resources at
compilation time, making sure that the most recent version of the
resource is included and making potential merge conflicts explicit and
simpler to resolve.


mkralik3:
retest this please
(QE tests fixed)

install/operator/pkg/generator/generator.go
```diff
@@ -186,7 +171,7 @@ func Render(filePath string, context interface{}) ([]unstructured.Unstructured,
 		// It's ok if a template chooses not to generate any resources..
 
 	default:
-		return nil, fmt.Errorf("Unexptected yaml unmarshal type: %v", obj)
+		return nil, fmt.Errorf("unexptected yaml unmarshal type: %v", obj)
```
phantomjinx: erroneous "t" in the middle of "unexpected"

phantomjinx: erroneous "t" in the middle of "unexpected"

phantomjinx (approved):

install/operator/pkg/generator/generator.go
zregvart: Thanks!

PR #9892, opened 2022-02-04
refactor(ui): replace Learna with Lage

zregvart:
Lage[1] makes a good replacement for Lerna because of strict topological
workflow -- so tasks in workspaces are run in accordance to the how they
depend on each other; and because of the incremental cache -- so tasks
are skipped if inputs have not changed.

This dramatically decreases the build time by allowing the `build` task
to be run in parallel, skipping over tasks for which the cache is hot.

[1] https://microsoft.github.io/lage/


mkralik3:
retest this please
(QE tests fixed)

kahboom (approved):

zregvart:
I've converted this to draft as I'm unsure about the changes within the GitHub Actions workflows I made. The Lage cache looks very small and it seems to hold only the cache keys. I was hoping that the output of the build would be held in cache but it doesn't seem to be. Without the output in the cache I'm not sure that the entire build is correct when parts of the build are skipped. Works okay as far as I can tell locally (outside of CI), as the outputs are already in `dist` and don't need to be in the cache.

zregvart:
retest this please

zregvart:
I've reverted the GitHub Actions workflow configuration, so this will end up only helping in local/development workflows.

Issue #9893, opened 2022-02-08
Using embed.Open to read a directory causes error


claudio4j:
For some reason the `embed.Open` throws an error when trying to read a directory ending with a `/`, as in this [install.go](https://github.com/syndesisio/syndesis/blob/1.14.x/install/operator/pkg/syndesis/action/install.go#L236) reading the addons directory.
The error is `open assets/addons/jaeger/: file does not exist` when installing the syndesis CR with jaeger addon enabled.

PR #9894, opened 2022-02-08
Fix operator install with jaeger enabled, fail to install

claudio4j:
Using embed.Open to read a directory causes error
https://github.com/syndesisio/syndesis/issues/9893


zregvart (approved):
Thanks!

zregvart:
retest this please

Issue #9898, opened 2022-02-09
origin-oauth-proxy:4.9 requires permission to read openshift-config-managed


claudio4j:
There is an error when starting syndesis-oauthproxy pod
```
github.com/openshift/oauth-proxy/providers/openshift/provider.go:347: Failed to watch *v1.ConfigMap: failed to list *v1.ConfigMap: configmaps "oauth-serving-cert" is forbidden: User "system:serviceaccount:syndesis:syndesis-oauth-client" cannot list resource "configmaps" in API group "" in the namespace "openshift-config-managed"
```
openshift/origin-oauth-proxy:4.9 requires permission to read configmaps from `openshift-config-managed` namespace.

https://github.com/openshift/oauth-proxy/issues/229

PR #9899, opened 2022-02-10
Fix rbac permission for syndesis-oauthproxy read configmaps

claudio4j:
origin-oauth-proxy:4.9 requires permission to read openshift-config-managed #9898
https://github.com/syndesisio/syndesis/issues/9898


zregvart (approved):
LGTM

PR #9911, opened 2022-02-22
Cleanup OpenShiftServiceImpl

zregvart:
Contains the following fixes and refactorings:

[fix(server): syndesis-s2i not s2i-java as default](https://github.com/syndesisio/syndesis/commit/8d4bd6057fe336f93ab1af1453442b635044a698)

This is not an issue at runtime as by default the operator sets this
to whatever image reference is configured in the Syndesis CR. It is
problematic however in local development when running without the
operator, say running the Syndesis server from the IDE and connecting to
a OpenShift cluster.

[refactor](https://github.com/syndesisio/syndesis/commit/5c4b26280a1b5f5ddfdfefab5eb86eb93927223d)[(server): null-safe by default](https://github.com/syndesisio/syndesis/commit/5c4b26280a1b5f5ddfdfefab5eb86eb93927223d)

Makes sure that there is no need to check if `scheduling` is `null` or
not, as it cannot be `null`.

[fix(server): wrong variable reference](https://github.com/syndesisio/syndesis/commit/f06987906ef9a72d647785ddfdb5a85c4f83fcc3)

The setter `setOpenShiftHost` is setting the existing value of
`masterUrlHost` and ignoring the provided new value, thus effectively it
is implemented as a no-op. This makes sure that the given value replaces
the existing value.

[refactor(server): simplify unpublishing](https://github.com/syndesisio/syndesis/commit/964133684ac2c48d031395ffa2a113a8b2189df8)

This refactors the unpublishing step to perform a much simpler operation
of scaling to zero the existing deployment found via deployment name.

Also refactors the placement of the scale annotation to the
`DeploymentConfig`'s metadata rather than on container's metadata. The
whole deployment is scaled not an individual container. This makes for a
simpler update that doesn't require a new revision of the deployment and
prevents further re-deployments.

Subsequent commit will also include this change at deployment time.

[chore(deps): align version of f8:mockwebserver](https://github.com/syndesisio/syndesis/commit/8829764984dd37749182ef35e6a70c20c6286e90)

[refactor(server): openshift service refactor](https://github.com/syndesisio/syndesis/commit/3565f12e6449fd90df5b4b5e8dbc54b845956736)

The `OpenShiftServiceImpl` had major issues with code duplication and
unnecessary complexity. For example code paths for creating a new
`DeploymentConfig` and editing the existing `DeploymentConfig` were
nearly but not exactly identical, making maintenance difficult as one
would need to make changes to two code paths simultaneously.

One long standing issue with `OpenShiftServiceImpl` is also the
superfluous deployments that would generate new revisions, and as such
redeploy the integration multiple times during publishing. This
refactoring makes sure that for each publish there is only a single
deployment revision -- i.e. single restart of the integration pod.

Several code smells were also refactored.

### DIFF
--- a/app/pom.xml
+++ b/app/pom.xml
@@ -3531,7 +3531,7 @@
       <dependency>
         <groupId>io.fabric8</groupId>
         <artifactId>mockwebserver</artifactId>
-        <version>0.1.3</version>
+        <version>0.1.8</version>
         <scope>test</scope>
         <exclusions>
           <exclusion>
--- a/app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/UnpublishHandler.java
+++ b/app/server/controller/src/main/java/io/syndesis/server/controller/integration/online/UnpublishHandler.java
@@ -17,16 +17,11 @@ package io.syndesis.server.controller.integration.online;
 
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
-import java.util.stream.Collectors;
 
-import io.fabric8.openshift.api.model.DeploymentConfig;
 import io.syndesis.common.model.integration.IntegrationDeployment;
 import io.syndesis.common.model.integration.IntegrationDeploymentState;
-import io.syndesis.common.util.Labels;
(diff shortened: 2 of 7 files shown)


oscerd (approved):

zregvart:
retest this please

PR #9915, opened 2022-03-01
Update camel version to 2.23.2.fuse-7_11_0-00010

claudio4j:
Also, changed the url repo to verify the early access releases of camel.
This camel release will allow [a camel-slack fix](https://issues.redhat.com/browse/ENTESB-17387) to come in syndesis

### DIFF
--- a/app/extension/bom/pom.xml
+++ b/app/extension/bom/pom.xml
@@ -29,7 +29,7 @@
   <properties>
     <!-- we must ensure to have this version aligned with extension-bom and integration-bom -->
     <spring-boot.version>2.3.12.RELEASE</spring-boot.version>
-    <camel.version>2.23.2.fuse-7_10_0-00020-redhat-00001</camel.version>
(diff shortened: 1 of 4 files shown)


zregvart:
I thought we agreed not to use downstream development dependencies and have removed the references to those (see #9814). The argument was that we cannot reliably build against JBoss repositories because the repository is not stable enough for upstream use and because the development artifacts contain pointers to private Maven repositories (see [ENTESB-16583](https://issues.redhat.com/browse/ENTESB-16583)).
This is unfortunate but I don't see another way to have reliable upstream builds.
What we have done in the past is to make changes downstream against the development builds, test them and forward port them upstream (see https://github.com/syndesisio/syndesis/commit/05f8e864ee260d2d0c4582bfc611a0ad5f6f5658). This is, again, unfortunate, but unless the issues outlined above are corrected, again, I don't see any other way of handling this.

claudio4j:
Yeah, that's ok

PR #9927, opened 2022-04-01
Upgrades Spring to 5.3.18 and Spring Boot to 2.5.12

zregvart:
Upgrading Spring alone would not have been an issue, upgrading Spring Boot seems to have been a bigger issue. One issue unrelated to Spring upgrades is that the WSDL stack we use seems to fetch and parse remote XML schemas without following HTTP redirects, so I've added local cache for those.

Here are the results of that:

[fix(extension): don't use removed method](https://github.com/syndesisio/syndesis/commit/c51c578ff24f173d4ada318becb54450e5fb491d)

The method `Layout::getLibraryDestination` was renamed to `Layout::getLibraryLocation`.

[fix(api-generator): don't load remote schemas](https://github.com/syndesisio/syndesis/commit/800357b3a2dd7148acf92f3429e23bc029cbc18c)

We should not reach out and load remote XML schemas from
schemas.xmlsoap.org.

Background: it seems that recently schemas.xmlsoap.org website
configured HTTP to HTTPS redirect and in doing so broke our WSDL parser
as it will try to fetch without following HTTP redirects. This results
in empty HTTP body that the parser tries to parse and fails with:

```
org.xml.sax.SAXParseException: Premature end of file.
```

To resolve this this stores the two identified XSD files locally and
sets the custom EntityResolver via a custom WSDLLocator to load those
instead. The same is done for the CXF schema parsing classes which can
load the catalog files from `META-INF/jax-ws-catalog.xml`.

[fix(deps): missing javax.annotation.Nullable](https://github.com/syndesisio/syndesis/commit/e3180d7084e6e694f0274308236e464b56d45622)

Seems that some dependency was pulling this in, but it is not pulling it in any more after Spring/Spring Boot upgrade. This adds it to prevent a compile error.

[fix(extension): compile issues Spring Boot upgrade](https://github.com/syndesisio/syndesis/commit/e836abe3ba3893ffb376caffab118a06b771e9d9)

The `AbstractDependencyFilterMojo:filterDependencies` method that we override has become `final`, so this copies and repackages `AbstractDependencyFilterMojo` and all of the code that is referenced from it in a package private/protected way so we can compile the plugin. 

The Spring Boot code included here is ASF licensed, so it should be okay.

Some code quality/style checks changes were required.

[fix(deps): upgrade to Spring Cloud 2020.0.4](https://github.com/syndesisio/syndesis/commit/94722040e53326c22b09fff835af2f95bf61b84b)

Upgrades to the next version after Hoxton.SR12, as Hoxton.SR12 was not compatible with Spring Boot 2.5.x.

[fix(deps): declare hystrix-core version](https://github.com/syndesisio/syndesis/commit/b208a2f26ecfcf4341f35cba386f2d25cb7a5910)

I think Spring Cloud was declaring it and then dropped it from the release we need to upgrade to, so we need to declare it here instead.

[fix(deps): add missing dependency verisions](https://github.com/syndesisio/syndesis/commit/00175b065bd84e73ef2b779414c93e7277b763fd)

Upgrading to Spring Cloud we lost some of the versions declared here, I picked to the best of my ability the versions, there could be a better selection.

[fix(server): enable prometheus export](https://github.com/syndesisio/syndesis/commit/21dba005288794b7d48ee88d7aaaa2bff5ba8b7a)

It is no longer by default on if Prometheus exporter classes are detected on the classpath, we need to enable it explicitly.

Also adds the `debug=true` to the to the server integration tests so we can tell why a bean wasn't introduced into the context.

[fix(server): adapt KubernetesClient test](https://github.com/syndesisio/syndesis/commit/cf843b2a147650d682de55e509e3cefc238dbba9)

Seems that we need to explicitly load the `Application` configuration, otherwise Spring Boot auto-configuration doesn't kick in.

### DIFF
--- a/app/extension/bom/pom.xml
+++ b/app/extension/bom/pom.xml
@@ -28,6 +28,7 @@
 
   <properties>
     <!-- we must ensure to have this version aligned with extension-bom and integration-bom -->
+    <spring.version>5.3.18</spring.version>
     <spring-boot.version>2.5.12.RELEASE</spring-boot.version>
     <camel.version>2.23.2.fuse-7_10_0-00020-redhat-00001</camel.version>
     <atlasmap.version>2.3.14</atlasmap.version>
@@ -170,6 +171,14 @@
         <version>${org.apache.log4j.version}</version>
         <scope>import</scope>
         <type>pom</type>
+      </dependency>
+
+      <dependency>
+        <groupId>org.springframework</groupId>
+        <artifactId>spring-framework-bom</artifactId>
+        <version>${spring.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
       </dependency>
 
       <!-- Camel BOM for Spring-boot -->
--- a/app/extension/bom/pom.xml
+++ b/app/extension/bom/pom.xml
@@ -30,7 +30,7 @@
     <!-- we must ensure to have this version aligned with extension-bom and integration-bom -->
     <spring-boot.version>2.5.12.RELEASE</spring-boot.version>
     <camel.version>2.23.2.fuse-7_10_0-00020-redhat-00001</camel.version>
-    <atlasmap.version>2.3.13</atlasmap.version>
+    <atlasmap.version>2.3.14</atlasmap.version>
     <jackson.version>2.11.2</jackson.version>
     <mongodb.version>3.9.0</mongodb.version>
     <jetty.version>9.4.43.v20210629</jetty.version>
--- a/app/extension/bom/pom.xml
+++ b/app/extension/bom/pom.xml
@@ -29,7 +29,7 @@
   <properties>
     <!-- we must ensure to have this version aligned with extension-bom and integration-bom -->
     <spring.version>5.3.18</spring.version>
-    <spring-boot.version>2.5.12.RELEASE</spring-boot.version>
+    <spring-boot.version>2.5.12</spring-boot.version>
     <camel.version>2.23.2.fuse-7_10_0-00020-redhat-00001</camel.version>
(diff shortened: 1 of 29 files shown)


phantomjinx (approved):
LGTM

zregvart:
retest this please

mkralik3:
The wrong qe jenkins was triggered, the running job for the latest changes is here: https://fuse-7x-jenkins-csb-fuse-qe.apps.ocp4.prod.psi.redhat.com/view/Fuse%20Online%207.11/job/fuse-7.11-syndesis-tests-PR-ENG/154/console

claudio4j:
Add `org.springframework.boot.actuate.autoconfigure.metrics.mongo.MongoMetricsAutoConfiguration` to `spring.autoconfigure.exclude` in `app/meta/src/main/resources/application.yml`, otherwise it throws IllegalStateException: Failed to introspect Class [org.springframework.boot.actuate.autoconfigure.metrics.mongo.MongoMetricsAutoConfiguration$ and a ClassNotFoundException

zregvart:
@claudio4j I've added those exclusions but also a bit more so we're not bitten with this in the future. I've added tests to check if Meta starts up and one test for the `/connectors` endpoint. I've found the same issue as in 21dba005 in Meta as well.

claudio4j (approved):

mkralik3:
retest this please

zregvart:
retest this please

mkralik3:
The QE job doesn't work properly. We need to fix checking images on Quay. I am working on it.

mkralik3:
retest this please

PR #9938, opened 2022-04-25
ENTESB-18466 - Reworking of operator conditions ensuring step upgrades

phantomjinx:
While performing a double upgrade, ie. 1.13.0 -> 1.14.0 -> 1.14.1, the reconcile function was never being called and hence the operatorcondition for delaying the operator upgrade was never applied. To workaround this a new init-container is executed in front of the operator container that calls a partner binary (operator-init). This binary sits in the same operator image so no need to build another image (just override the command).
operator-init detects existence of a Syndesis resource, and if present, applies the operatorcondition. Since this is an init-container, the operator has not even started so no race conditions and guranteed that the next upgrade just cannot start ahead of the operatorcondition being set.
Only when the reconciled has correctly processed the operands, done the upgrade etc., is the operatorcondition released, allowing the next operator to kick in.

The visual manifestation of this is that the 1.14.1 operator appears in the list of operators but remains pending.

### DIFF
--- a/install/operator/.lib.sh
+++ b/install/operator/.lib.sh
@@ -136,6 +136,11 @@ build_operator()
             go build -o ./dist/${GOOS}-${GOARCH}/platform-detect \
                 -gcflags all=-trimpath=${GOPATH} -asmflags all=-trimpath=${GOPATH} -mod=vendor \
                 ./cmd/detect
+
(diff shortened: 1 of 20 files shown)


claudio4j (approved):

zregvart (approved):

zregvart:
retest this please

PR #9942, opened 2022-05-13
[ENTESB-18911] UnsatisfiedDependencyException in integration with MongoDB connector

rnetuka:
Issue: https://issues.redhat.com/browse/ENTESB-18911

Spring Boot 2.5.12 uses MongoDB 4.2.3
https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-dependencies/2.5.12/spring-boot-dependencies-2.5.12.pom

This needs to be synchronized.

### DIFF
--- a/app/connector/mongodb/pom.xml
+++ b/app/connector/mongodb/pom.xml
@@ -42,7 +42,7 @@
       </dependency>
       <dependency>
         <groupId>org.mongodb</groupId>
-        <artifactId>mongo-java-driver</artifactId>
+        <artifactId>mongodb-driver-legacy</artifactId>
         <version>${mongodb.version}</version>
       </dependency>
     </dependencies>
--- a/app/extension/bom/pom.xml
+++ b/app/extension/bom/pom.xml
@@ -33,7 +33,7 @@
     <camel.version>2.23.2.fuse-7_10_0-00020-redhat-00001</camel.version>
     <atlasmap.version>2.3.14</atlasmap.version>
     <jackson.version>2.11.2</jackson.version>
-    <mongodb.version>3.9.0</mongodb.version>
+    <mongodb.version>4.2.3</mongodb.version>
     <jetty.version>9.4.43.v20210629</jetty.version>
     <netty.version>4.1.68.Final</netty.version>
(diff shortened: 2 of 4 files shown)


claudio4j:
Shouldn't this mongodb driver go to app/connector/mongodb connector and integration/bom/pom.xml ?

rnetuka:
@claudio4j integration/bom/pom.xml updated. Shall app/extension/bom/pom.xml keep the update or discard it?

claudio4j:
Both pom.xml should be updated. Looks like the CI tests are ok. Were you able to run an integration to confirm it's working ?
Thanks for fixing it.

claudio4j:
There is https://github.com/syndesisio/syndesis/blob/1.14.x/app/pom.xml#L150 to change too.

claudio4j:
I was looking at [mvn repo](https://mvnrepository.com/artifact/org.mongodb/mongo-java-driver) and there is no 4.x version.
Also, looking at [camel-mongodb component](https://github.com/jboss-fuse/camel/blob/2.23.x.redhat-7-x/parent/pom.xml#L521) it uses 3.12.7, I think we can continue to use this version.
camel spring boot is one of the runtimes to run camel routes, I think we are safe to use this version, but in syndesis there should be an exclusion of mongodb lib from spring boot dependency.
Were you able to run integrations to test it ?

rnetuka:
MongoDB drivers 4.x are published as different artifacts. https://mvnrepository.com/artifact/org.mongodb/mongodb-driver-legacy/4.2.3

Unfortunately, I wasn't able to test it since my OS refuses to run VirtualBox / Minishift.

I've updated the files with additional changes. It's still only a proposal.

syndesisqe:

rnetuka:
Upgrading to MongoDB 4.2.3 is not an option because Camel still uses 3.12.7

Replaced by PR https://github.com/syndesisio/syndesis/pull/9943

rnetuka:
Upgrading to MongoDB 4.2.3 is not an option because Camel still uses 3.12.7

Replaced by PR https://github.com/syndesisio/syndesis/pull/9943

PR #9964, opened 2022-08-31
fix operator: link service account to secret token

claudio4j:
* Openshift 4.11 is based on kubernetes 1.24, which introduced KEP-2799 which doesn't create the Secret (token) anymore, then the user should create the Secret (token) and link to the SA.
* fix build: copy operator-init binary
* fix ImageStream of operator-init for dev build in operator deployment
  yaml


phantomjinx (approved):
LGTM

PR #9970, opened 2022-10-25
fix: adjust bouncy castle dependencies

zregvart:
Makes sure that the managed version of `wss4j-ws-security-common` is the where the exclusions are handled, and excludes the Bouncy Castle version 1.69 dependency from `kubernetes-client`.


grgrzybek (approved):

oscerd (approved):

zregvart:
I guess we'll just merge this...

